PR #130, opened 2014-10-14
Ceph config rework

shinespb:
Changed config generation.
Now you can add necessary parameters in playbook, and default vars will be rewrited. No more hardcoded parameters. All new params you can add in your playbook.

It's will be something like this:

ceph_config:
  global:
    disable_in_memory_logs: true
    osd pool default pg num: 256
    osd pool default pgp num: 256
  osd:
    some variable here: true
  mds:
    new mds variable in release: 123


leseb:
Resurrected at #215

Issue #162, opened 2014-12-23
Support for a CRUSH location string in config file


mwheckmann:
It would be great if we could support setting a custom CRUSH location string in the config file for larger non-standard setups.

Just a placeholder for future work.

leseb:
What do you mean? Something like:

[osd.X]
osd crush location = "root=location"

mwheckmann:
Yes, exactly. An example would be: 

osd crush location = "root=default pod=b1 host={{ ansible_hostname }}"

Ideally, the right hand side of the '=' would be in a configurable variable.

Default would be "root=default host={{ ansible_hostname }}" (this is the Ceph default).

leseb:
Ok I see, since it's an OSD section, it might be really tricky.
We don't support OSD section at the moment, and will be difficult to support it.
I've been thinking of using the assemble module for that. I'll have a look again.

leseb:
Ok I have a really tricky prototype for you. I'm gonna push a branch soon.

PR #177, opened 2015-01-08
WIP: Implement OSD sections

leseb:
Still WIP, @mwheckmann free to test
As requested by #162

Current known issue, since ceph.conf gets modified during every single
run (at the end during the merge) so this will restart ceph daemons.

### DIFF
--- a/group_vars/osds
+++ b/group_vars/osds
@@ -5,8 +5,27 @@
 # Dummy variable to avoid error because ansible does not recognize the file as a good configuration file when no variable in it.
 dummy:
 
-## Ceph options
+####################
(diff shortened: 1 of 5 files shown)


leseb:
If we want to somehow give more variable we could use the Ansible inventory for that something like:

[osds]
osd0 ceph_root= default ceph_rack=foo

leseb:
Ok everything works now, the only thing that I'll like to avoid is to restart Ceph during each run.

leseb:
Merging this.
Simply note that everytime you replay the playbook it will restart all the OSD daemons :-(

PR #197, opened 2015-01-24
fsid & ceph_stable_redhat_distro fixs up

flisky:


group_vars/all
```diff
@@ -19,7 +19,7 @@ dummy:
 #ceph_stable_release: giant # ceph stable release
 
 # This option is needed for _both_ stable and dev version, so please always fill the right version
-# # supported distros are el6, rhel6, f18, f19, opensuse12.2, sles11, centos7 (see http://ceph.com/rpm-firefly/)
+# # supported distros are el6, rhel6, f20, el7, rhel7 (see http://ceph.com/rpm-giant/)
```
leseb: Since this changes every release we probably should say something like:

"for supported distros see: http://ceph.com/rpm-$ceph-release"

leseb:
Can you also update roles/ceph-common/default/?

PR #204, opened 2015-02-12
Update ceph.conf.j2

marmot21:
This is just a simple change that will allow multiple gateways


leseb:
AFAIR we don't support multi RGW yet, I still need to remember the reason why...

marmot21:
If there's a reason why I'd love to know, this seemed to work in our situation

leseb:
Ok maybe this is just me and without remembering I introduced a change to allow that :).
Do you also use HAproxy?

marmot21:
No :)

leseb:
Actually I had to reverse this since it breaks RGW startup.

leseb:
However I reworked the multi-rgw, tested and validated. See #210

PR #215, opened 2015-02-23
[WIP] DO NOT MERGE Shinespb master

leseb:


leseb:
Note for myself, see if this doesn't break the CRUSH OSD section support.

leseb:
I don't think there is any way to fix the ceph config layout.
Will do a final attempt and maybe close it eventually.

shinespb:
Sorry for my long absence. What I need to fix now? I saw that you've made some changes. What now? Now I have some time and could finally do it.

leseb:
Well the big issue at the moment is the layout of the ceph.conf, I can get something decent however, I haven't found anything to get it structured like this:

- global
- mon
- osd
- mds
- rgw

Basically the order of the section should be the same as the declaration. It's mainly a python issue that can't really by solved. If you can come up with a good strategy to tackle then I'll be happy :)

shinespb:
Ok, I'll test it.

leseb:
Thanks!

shinespb:
Ok. Now I've done some changes and got:
[client]
[global]
[mds]
[mon]
  [mon.ceph-mon0]
  [mon.ceph-mon1]
  [mon.ceph-mon2]
[osd]
[rgw]

Are you sure, that order of this sections really does matter?

leseb:
It doesn't really matter, it's just not elegant at all...
I just don't want to confuse users.

shinespb:
I think, users must change configuration through ansible and not directly in ceph.conf.

leseb:
I think so too, but it'd be nice if we could get a standard layout :)

shinespb:
I could rewrite it, but ceph.conf.j2 would be more complicated and ugly.

leseb:
Hum then I think we should forget this.
Can't we keep static sections and then loop within each section instead if creating everything?

Something like:

[global]
{% if option == "global": %}
then we fill the section global section and nothing more

[mon]
{% if option == "mon": %}
then we fill the section global section and nothing more

So basically we just catch the section we want  instead of letting python creating the section for us.

Hope it's clear :)

shinespb:
Yes, I already testing changes with hardcoded sections. But when I wrote it first time, I was thinking about flexibility. You could add new sections and new parameters without changing any templates.

leseb:
I don't think we will have more sections in the future.
Ceph is not going to have another component any time soon.
Even if it does, it's just a simple PR, and way better than one PR = one new option...

shinespb:
Send patch to you by mail.

leseb:
@shinespb just pushed some changes, I'm still testing this, thanks again!

shinespb:
I think we need to resolve all conflicts and merge it.
If you doesn't have time for that I could do it myself and then make a pull request for current version of code.

leseb:
Sorry, I've been really busy.
I tried the patch and changed a couple of things.

The layout is not quite right.
Some sections get an indentation and they should not.

shinespb:
So, it will not be merged?

leseb:
@shinespb sorry I didn't mean to close it, I did some (wrong) git commands and just realized that this closed some PR.
I'm still considering your work as being really valuable, I just need to get another round on this one.

leseb:
@shinespb unfortunately this will never get merged.
Sorry for wasting your time on this. There is a new ansible module that aims to do exactly this.

Hope your understand.

PR #218, opened 2015-03-02
Add support of civetweb for Rados Gateway

leseb:
The default is now to use civet web since it has proven to be a bit more reliable and easy to setup than Apache.
We don't support multi backend yet.
Radow GW supports both CGI and civet web on the same machine.


leseb:
Handlers will have to be reworked a bit too, especially for RGW.
Note for myself Ubuntu upstart init for RGW don't seem to work...

PR #230, opened 2015-03-19
Faster Travis build with Docker

leseb:
http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/


leseb:
Will see how to fix this later.

PR #248, opened 2015-04-14
Allow restapi=false

aisrael:
Just a quick PR to allow specifying `restapi=false` in Ansible to skip the tasks in `ceph-restapi` role.

### DIFF
--- a/roles/ceph-restapi/tasks/main.yml
+++ b/roles/ceph-restapi/tasks/main.yml
@@ -1,12 +1,14 @@
 ---
 - include: pre_requisite.yml
+  when: restapi
 
 - name: Check if Ceph REST API is already started
   shell: "ps aux|grep [c]eph-rest-api"
   register: restapi_status
+  when: restapi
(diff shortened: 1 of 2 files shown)


leseb:
Why don't you simply remove this line?
https://github.com/ceph/ceph-ansible/blob/master/site.yml#L17

aisrael:
@leseb I could (and I did, at first), but this way I figured one can choose to enable/disable ceph-restapi at deploy-time, without having to modify the ceph-ansible sources.

leseb:
Hum I don't like adding a lot of when statements, can you do something like this instead?

- { role: ceph-restapi, when: not restapi }

The restapi var exists here: https://github.com/ceph/ceph-ansible/blob/master/group_vars/all#L116
This will also disable the section in the ceph.conf

aisrael:
Sorry, you're right. That's what I should've done in the first place. Fixed.

leseb:
Thanks!

PR #250, opened 2015-04-14
rolling_update cleanup

flisky:


flisky:
By the way, the radosgw service is named "ceph-radosgw" for el6/el7.

@leseb, have you notice that? Should I take another PR?

leseb:
Yes I noticed.
Now that I think about it, this playbook doesn't handle RedHat distros.

You can fix in this PR if you want :)

rolling_update.yml
```diff
@@ -59,7 +59,7 @@
       until: result.rc == 0
       retries: 5
       delay: 10
-      delegate_to: 127.0.0.1
+      delegate_to: "{{ groups.mons[0] }}"
```
leseb: did you actually test this statement? can you confirm that it works?
I had so much trouble getting this work properly, that's why I ended up with this stupid with_items :)

roles/ceph-common/tasks/os_tuning.yml
```diff
@@ -22,4 +22,5 @@
     value={{ item.value }}
     state=present
     sysctl_file=/etc/sysctl.conf
+    ignoreerrors=yes
```
leseb: I guess you meant: ignore_errors: True
This should appear below the with_items statement.

leseb:
up?

rolling_update.yml
```diff
@@ -59,7 +63,7 @@
...
```
leseb: Not sure about this one, using the first monitor is not that good if it's the one that has been or is been updated

roles/ceph-common/tasks/os_tuning.yml
flisky: FYI, `ignoreerrors` is a option for `sysctl`. Here is the [doc](http://docs.ansible.com/sysctl_module.html#options).

rolling_update.yml
flisky: localhost is not working for me, since ansible mgt machine is not a part of the ceph cluster.

Oh, I see your concern. How about this?

```
- name: select a running monitor
  set_fact: mon_host={{ item }}
  with_items: groups.mons
  when: item != inventory_hostname

- name: Waiting for clean PGs...
  ...
  delegate_to: {{ mon_host }}
```

roles/ceph-common/tasks/os_tuning.yml
leseb: Ok my bad.

rolling_update.yml
leseb: This solution looks good to me, we must make sure to take a monitor that is being updated :)

flisky: My approach takes the last non-self monitor. It seems I should take the first non-self one. Does it?

leseb: No I think it's good this way too.
Only the non-self matters here.

flisky: Okay, committed.

rolling_update.yml
```diff
@@ -59,7 +68,7 @@
       until: result.rc == 0
       retries: 5
       delay: 10
-      delegate_to: 127.0.0.1
+      delegate_to: "{{ mon_host }}"
```
leseb: Sorry to be picky but assuming I have 3 monitors, this variable will contain 2 monitors right?
So we actually perform this action on 2 mons, this is not really an issue, but ideally we would pick the first one of this list as well.

leseb:
Thanks for the last commit, do you still plan on proposing something for the radosgw service name on RHEL vs Debian systems?

rolling_update.yml
flisky: No, the mon_host is the last one set by `set_fact`.

flisky:
Not in this PR. Maybe I'll take a try in next few days.

leseb:
I think we are good to go, one more thing, did you actually test all your changes?

flisky:
Yes, and I test them again. I think it's safe to merge them in :)

leseb:
Thanks!

PR #260, opened 2015-05-14
make sure journal_size configured

flisky:


leseb:
I'm fine with this change, however it breaks the vagrant deployment.
Can you insert a new var after this line: https://github.com/ceph/ceph-ansible/blob/master/Vagrantfile#L55

With journal_size: 100

Thanks!

flisky:
Didn't notice that, and thanks for the hint. Committed.

PR #268, opened 2015-06-08
Directory should be executable

bjne:
This looks like a bug, if not, it should have a comment that tells why

### DIFF
--- a/roles/ceph-mon/tasks/deploy_monitors.yml
+++ b/roles/ceph-mon/tasks/deploy_monitors.yml
@@ -17,7 +17,7 @@
     state=directory
     owner=root
     group=root
-    mode=0644
+    mode=0755
 
 - name: ceph monitor mkfs
   command: >


leseb:
I've never had any issue with those permissions.
Why should this directory be executable?

Looks like ceph-deploy uses 755 as well...

bjne:
If you are supposed to be able to enter a directory, it needs to be
executable..

2015-06-08 17:56 GMT+02:00 Leseb <notifications@github.com>:

> I've never had any issue with those permissions.
> Why should this directory be executable?
>
> Looks like ceph-deploy uses 755 as well...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ceph/ceph-ansible/pull/268#issuecomment-110052224>.
>



-- 
Bj(/)rnar

leseb:
For a common change please keep the same PR next time.
Thanks!

PR #272, opened 2015-06-08
Add access to "rbd concurrent_management_ops"

bjne:


leseb:
Can you also edit `group_vars/all` and `roles/ceph-common/defaults/mail.yml` to reflect this change?

PR #287, opened 2015-06-25
Correct the logic for running 'ceph-disk prepare'

bstillwell:
The logic was broken here for repeated runs.  We only want to run
'ceph-disk prepare' when the disk does not contain a ceph partition, is
not a partition, and raw_multi_journal is set.  Previously it would
attempt to run 'ceph-disk prepare' when there was a ceph partition
because the second half of the 'or' was still true since it isn't a
partition.


leseb:
Thanks @bstillwell !

PR #317, opened 2015-07-06
Add more client options

leseb:
Add `rbd_default_features` and `rbd_default_map_options`.

Signed-off-by: leseb <seb@redhat.com>


elsonrodriguez:
@leseb Looks like this got reverted somehow.

Was it intentional?

leseb:
@elsonrodriguez this got reverted because we now use overrides, see https://github.com/ceph/ceph-ansible/blob/master/group_vars/all.sample#L312-L327

So we don't need to add every single Ceph options.

PR #321, opened 2015-07-07
Do not use latest package

leseb:
While re-running the playbook we do not want to check for new packages.
We shouldn't perform upgrades, we leave this to the operators.

Signed-off-by: leseb <seb@redhat.com>

### DIFF
--- a/roles/ceph-common/tasks/install_on_debian.yml
+++ b/roles/ceph-common/tasks/install_on_debian.yml
@@ -52,7 +52,7 @@
 - name: install Ceph
   apt: >
     pkg={{ item }}
-    state=latest
+    state=present
   with_items:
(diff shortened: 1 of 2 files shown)


mwheckmann:
This change breaks the rolling_update.yml Playbook. Intentional?

leseb:
@mwheckmann you're right, my mistake going to revert this.

leseb:
@mwheckmann I'm currently on PTO, would you mind doing a PR for this? Otherwise this will have to wait next week...

mwheckmann:
Hi there, I was actually on PTO too :). Thanks for fixing this.

leseb:
@mwheckmann np :)

PR #352, opened 2015-07-28
Failed_when and stat module

msambol:
Changed ignore_errors to failed_when so the output doesn't show in
red. Also changed a few instances of the shell module to stat.

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -26,15 +26,15 @@
   when: ansible_os_family == 'Debian'
 
 - name: check for a ceph socket
-  shell: "stat /var/run/ceph/*.asok > /dev/null 2>&1"
+  stat: path=/var/run/ceph/*.asok
   changed_when: false
-  ignore_errors: true
+  failed_when: false
   register: socket
 
 - name: check for a rados gateway socket
-  shell: "stat {{ rbd_client_admin_socket_path }}*.asok > /dev/null 2>&1"
+  stat: path={{ rbd_client_admin_socket_path }}*.asok
   changed_when: false
-  ignore_errors: true
+  failed_when: false
   register: socketrgw
 
 - name: create a local fetch directory if it doesn't exist
--- a/cluster-os-migration.yml
+++ b/cluster-os-migration.yml
@@ -21,7 +21,7 @@
       stat: >
         path=/var/lib/ceph/mon/ceph-{{ ansible_hostname }}/migration_completed
       register: migration_completed
-      ignore_errors: True
+      failed_when: false
 
     - name: Check for failed run
       stat: >
@@ -248,7 +248,7 @@
       stat: >
         path=/var/lib/ceph/migration_completed
       register: migration_completed
-      ignore_errors: True
+      failed_when: false
 
     - name: Check for failed run
       stat: >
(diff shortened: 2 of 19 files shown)


leseb:
@msambol could you please rebase your work?

roles/ceph-osd/tasks/scenarios/journal_collocation.yml
```diff
@@ -12,7 +12,7 @@
 # failed, this is why we check if the device is a partition too.
 - name: automatic prepare osd disk(s) without partitions
   command: ceph-disk prepare "/dev/{{ item.key }}"
-  ignore_errors: true
+  failed_when: false
```
leseb: I'd like to keep this one with `ignore_errors: true`. I don't want to have hidden.
It can be useful to see which disk didn't succeed to get formatted or something else.

roles/ceph-osd/tasks/scenarios/journal_collocation.yml
```diff
@@ -24,7 +24,7 @@
 
 - name: manually Prepare osd disk(s)
   command: "ceph-disk prepare {{ item.2 }}"
-  ignore_errors: true
+  failed_when: false
```
leseb: Same like to keep this one as well.

roles/ceph-osd/tasks/scenarios/raw_multi_journal.yml
```diff
@@ -18,7 +18,7 @@
     - devices
     - raw_journal_devices
   changed_when: false
-  ignore_errors: true
+  failed_when: false
```
leseb: Same like to keep this one as well.

msambol:
Hmm, showing up a little weird. I think because I used `--amend`.
But it's rebased.

roles/ceph-common/tasks/main.yml
```diff
@@ -26,15 +26,15 @@
   when: ansible_os_family == 'Debian'
 
 - name: check for a ceph socket
-  shell: "stat /var/run/ceph/*.asok > /dev/null 2>&1"
+  stat: path=/var/run/ceph/*.asok
```
leseb: I'm not sure about this one, AFAIR stat doesn't support * 
Did you get a chance to validate this?

roles/ceph-common/tasks/main.yml
```diff
@@ -26,15 +26,15 @@
...
   changed_when: false
-  ignore_errors: true
+  failed_when: false
   register: socket
 
 - name: check for a rados gateway socket
-  shell: "stat {{ rbd_client_admin_socket_path }}*.asok > /dev/null 2>&1"
+  stat: path={{ rbd_client_admin_socket_path }}*.asok
```
leseb: Same as above.

msambol: I tested and it passed these tasks.

leseb: Ok thanks, I think we are good to go then.

roles/ceph-common/tasks/main.yml
leseb: Looks like this breaks the deployment in the end... :/ socket.rc can't be evaluated.
Same for rgw socket.

leseb:
Since probably most of the 'stat' break the deployment or upgrade, I'll revert this.
Then I might re-merge it after more intensive testings.

PR #357, opened 2015-07-29
Failed_when instead of ignore_errors

msambol:
Changed ignore_errors to failed_when so the output doesn't show in
red.


msambol:
I'll test the stat module more intensively later. Sorry about that.
It passed on our initial tests.

msambol:
Never mind, looks like you already took care of this.

leseb:
Thanks!

PR #361, opened 2015-08-01
DO NOT MERGE - Add keys from url

msambol:
I want to get your opinion on this before I test. What do you
think about getting rid of the keys in the repo and just
grabbing it from the url? That way you don't have to update the
keys in your repo if/when they change.


jimmymccrory:
It looks like something similar was done before, not sure why it was seemingly backed out.
https://github.com/ceph/ceph-ansible/commit/b6fcd37bc01fb55adbcd07f14bd7471fcb19675c

msambol:
I guess if the machine doesn't have public Internet access?

leseb:
As @jimmymccrory mentioned this was originally the case.
The main reason was to speed up the deployment because doing a lookup on a local file is faster than retrieving it from the internet.
Another reason was to ease the configuration as well by not overloading users with tons of configuration options.
ceph-ansible already has a lot of configuration option and didn't find these being really useful.

Since the keys are not likely to change, I'd like to keep it that way. (unless you can come up with a good argument). We could potentially try to profile this to verify the benefit from local files to fetch over internet but to me the local lookup is the clear winner here.
Even if they do it's not much work to fix it.

Issue #42, opened 2014-03-13
Enable cluster name support


leseb:
By default we consider that the cluster is named 'ceph', it would be nice to allow to define another name for the cluster.

leseb:
Not 100% sure of the real value of this, I'm closing this from now.
Feel free to re-open...

alfredodeza:
This is pretty important for us. The ability to handle multiple Ceph instances is a nice-to-have feature.

PR #365, opened 2015-08-03
Add support for cluster name

leseb:
Thanks to the 'cluster' variable you can use a different clsuter name
than the default one being 'ceph'.

This also cleans up the old docker.yml file from the OSD.
And fixes the radosgw sudoers template as well...

fixes: #42


leseb:
This probably breaks more things than adding real value to the playbook...
Radosgw init doens't really support cluster with names different than 'ceph'. I might re-open this one in the future when **all** the distros moved to systemd...

PR #367, opened 2015-08-06
Use stat module instead of shell

msambol:


leseb:
Looks good thanks @msambol

PR #371, opened 2015-08-07
Revert "Use stat module instead of shell"

msambol:
Reverts ceph/ceph-ansible#367

Shoot. Real sorry about this, but I think it's best if we revert this for now. We were using a custom Ansible and it was working fine but upon testing with an old version of Ansible it appears one of the checks isn't working properly.


leseb:
Indeed services don't seem to get restarted anymore. The handler doesn't work.

PR #374, opened 2015-08-14
Fix devices check for raw_multi_journal

andymcc:
Fix the logic for the mandatory devices check so that it applies to
raw_multi_journal and journal_collocation scenarios.


andymcc:
https://github.com/ceph/ceph-ansible/issues/375
Happy to change the "when" format for all, but it only seems to not work on the "or" statement.

andymcc:
Should be fixed better now so there is one check when using raw_multi_journal and one when using journal_collocation.

leseb:
Thanks @andymcc

PR #379, opened 2015-08-17
Check to ensure device checks only happen on osds

andymcc:
Add bool for osd_group_name in group_names for device checks.


leseb:
Thanks @andymcc, can you also update the two previous osd checks?

leseb:
Thanks!

PR #380, opened 2015-08-17
enable ceph-create-keys in Debian

Abhishekvrshny:


leseb:
Hum since we support both Debian and Red Hat based systems there is no point of using a condition. Can you remove the `when`? Thanks!

Abhishekvrshny:
Hi @leseb , I have removed the `when` condition. Please review this PR again

leseb:
Thanks @Abhishekvrshny!

PR #381, opened 2015-08-17
Allow configurable pg_num for OpenStack pools

git-harry:
Currently the OpenStack pools that get created use the default pg_num.
This commit updates the ceph-mon role to allow the pg_num for each pool
to be customised.


leseb:
Thanks for this PR @git-harry, could you please also reflect the pgp_num please?

leseb:
Arf never mind the pgp inherits from the pg value which is good enough :)

PR #382, opened 2015-08-17
Remove parition check from ceph-osd role

msambol:
I'm removing the ceph paritition check from `activate osd(s) when device is a disk` 
because the ceph parition does not exist when parted was registered (on a fresh install). 
This was causing the activate step to be skipped.


leseb:
Minor not: item.2 becomes item.1

msambol:
Fixed. Thanks..

PR #389, opened 2015-08-26
Fix for error when the nodes don't have the same interface name.

darkcrux:
This is a rare case but it happens. Since we're just calling
`monitor_interface` and not `hostvars[host]['monitor_interface'],
an error may occur when the current host's interface does not
exist on the other hosts. (eg. eth0 exists for node0, but it does
not exist on node1 and node2)

Fix for this is to use hostvars[host]['monitor_interface']

### DIFF
--- a/roles/ceph-common/templates/ceph.conf.j2
+++ b/roles/ceph-common/templates/ceph.conf.j2
@@ -91,7 +91,7 @@
   {% if hostvars[host]['ansible_hostname'] is defined %}
   [mon.{{ hostvars[host]['ansible_hostname'] }}]
     host = {{ hostvars[host]['ansible_hostname'] }}
-    mon addr = {{ hostvars[host]['ansible_' + monitor_interface]['ipv4']['address'] }}
+    mon addr = {{ hostvars[host]['ansible_' + hostvars[host][monitor_interface]]['ipv4']['address'] }}
   {% endif %}
 {% endfor %}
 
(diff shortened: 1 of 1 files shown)


leseb:
It seems to be failing for me with:

```fatal: [mon0] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'monitor_interface'", 'failed': True}```

**Comment by darkcrux** [comment]
Hmm. strange. I'm testing this on vagrant now to see what went wrong. Was running the installation and tests on Baremetal yesterday with ceph-ansible added as a submodule to another ansible project and it seemed to work. Will update soon.

**Comment by darkcrux** [comment]
I see the difference now. Since I got different network interface names on the my nodes, I had to define the variables as hostvars on the inventory file:

```
node1 monitor_interface=eth0
node2 monitor_interface=eth0
node3 monitor_interface=p1p4
```

The fix worked in this instance since `monitor_interface` is a host variable.

This fails on Vagrant since `monitor_interface` is defined in  `extra_vars` so I don't think it gets added to the group or host vars. 

Maybe a solution is to use the hostvars if they are defined and fallback to just `plain `monitor_interface` if not. Something like this:

```
mon addr = {{ hostvars[host]['ansible_' + (hostvars[host]['monitor_interface'] if hostvars[host]['monitor_interface'] is defined else monitor_interface)]['ipv4']['address'] }}
```

code now looks longer though. :(

wdyt @leseb?

leseb:
@darkcrux testing this :)

leseb:
Looks good to me thanks @darkcrux

PR #391, opened 2015-08-26
Prevent failure from race creating fetch directory

git-harry:
When multiple monitor hosts attempt to create the fetch directory there
is the potential for the task to fail with:

  "OSError: [Errno 17] File exists: 'fetch'"

This appear to be an issue with the file module trying to create the
same directory at the same time when the tasks has been delegated to a
single host.

This commit enables run_once on the affected task which should address
the issue.


leseb:
Thanks @git-harry

PR #393, opened 2015-08-27
Vagrant storagectl as option

HanXHX:
Hi,

On many Vagrant boxes, default controller is not "SATAController". This PR provides a new option. You can use now every boxes you want without editing Vagrantfile.

Example:  "deb/wheezy-amd64" uses "SATA Controller".

Cheers!

Emilien


leseb:
Thanks!

Issue #384, opened 2015-08-24
ceph-common dependencies in RHEL7.1


ti-mo:
Hey @leseb !

Some minor issues with an Enterprise 1.3 install on RHEL7.1 (fresh from the RH site):

ceph-common/tasks/install_on_redhat.yml:
```
- name: install dependencies
...
    - yum-plugin-priorities.noarch
    - epel-release
```

* `yum-plugin-priorities.noarch` is in `rhel-7-server-optional-rpms`, the user needs to manually intervene with `subscription-manager repos --enable rhel-7-server-optional-rpms`
* The `epel-release` RPM is absent from RH mirrors

`yum-plugin-priorities` is useful to have, but the code doesn't enable `optional-rpms` by default.
Could we conditionally install epel? Perhaps only on CentOS/Fedora/SLES; epel-release-*.rpm seems to be included out-of-the-box there according a quick pbone search.

Cheers, let me know if I can help!

Timo

leseb:
Hey @ti-mo I understand the issue, the reasoning looks good to me. Do you want me to fix this or do you you to submit a PR?

ti-mo:
I'll try and get a PR out later this week, thanks for the reply.

leseb:
@ti-mo thanks! Looking forward to seeing your PR :)

PR #394, opened 2015-08-27
Enable optional-rpms on official RHEL for yum-plugin-priorities

ti-mo:
Hey @leseb 

An extra to #384 

Not sure if this is the solution we want or if we want to force optional-rpms onto the user at all. This is the only option if we want to install yum-plugin-priorities without adding non-RH repo's. (official RH support compliance, etc.)

Let me know what you think!

Timo

### DIFF
--- a/roles/ceph-common/tasks/pre_requisites/prerequisite_rh_storage_cdn_install.yml
+++ b/roles/ceph-common/tasks/pre_requisites/prerequisite_rh_storage_cdn_install.yml
@@ -4,6 +4,17 @@
   register: subscription
   changed_when: false
 
+- name: check if the red hat optional repo is present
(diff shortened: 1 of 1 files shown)


leseb:
I've just made everything configurable in order to avoid a specific condition for yum-plugin-priority since this would have called the yum module.
I'd like to move this in `pre_requisites/prerequisite_rh_storage_cdn_install.yml` since this is called before the install tasks. You can also remove the when condition since this will always be RHEL.

leseb:
@ti-mo ping?

ti-mo:
@leseb Sorry for the downtime! I agree this patch would really only be useful in the CDN scenario, but enabling the ISO scenario doesn't automatically remove epel and yum-plugin-prio from the list.

Regardless, I don't see a use case for optional-rpms outside of the RH CDN scenario.

leseb:
@ti-mo thanks!

Issue #383, opened 2015-08-24
Allow fetch directory location to be specified


mattt416:
Hi @leseb,

I'd like to store the fetch directory outside of my current working directory.  I'm happy to create a PR for this adding the necessary variable to each role's defaults file, and then referencing that instead of "fetch/".  However, before I do this I wanted to confirm there is no specific reason for using the working directory.

Thanks!

--Matt

leseb:
Hey @mattt416, AFAIR there is no specific reasons for storing files under the `fetch` directory unless that this is in the git repo. I don't think this is a limitation of the module either so you can go ahead with this :).

leseb:
Fixes by #395

PR #395, opened 2015-08-27
Make fetch directory configurable

mattt416:
Currently, the fetch directory is created in your working directory
(where ansible is run from).  We prefer to not keep any state in this
directory and would prefer to have the fetch directory configurable so
we can store it outside of our code checkout.

This commit creates a new variable in each role called
`fetch_directory` (defaulting to the previous value of 'fetch/'), and
then updates each reference to 'fetch' to use the new variable instead.

Closes issue #383

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -4,6 +4,12 @@
 # Dummy variable to avoid error because ansible does not recognize the file as a good configuration file when no variable in it.
 dummy:
 
+###########
+# GENERAL #
+###########
+
+#fetch_directory: fetch/
+
 #########
 # INSTALL
 #########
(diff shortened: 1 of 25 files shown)


leseb:
Hey @mattt416k, thanks for submitting this PR.
One tiny thing, could you please reflect all the changes you've made in `defaults/main.yml` on `group_vars/*.sample` as well?
Thanks!

mattt416:
Hi @leseb, I've set (commented) defaults in group_vars/*.sample!  Shout if there are any issues or those entries should go in uncommented.

leseb:
Looks good, I'm just testing this with Vagrant, shouldn't take that long.

leseb:
Thanks @mattt416!

Issue #397, opened 2015-09-02
sysctl fails with whitespace separated multi values for a key


Abhishekvrshny:
When the `os_tuning_params` in `group_vars/all` contain a multi valued field, like

```
os_tuning_params:
  - { name: net.ipv4.ip_local_port_range, value: 2048 65535 }
```

the following error is seen:

```
failed: [10.32.165.19] => (item={'name': 'net.ipv4.ip_local_port_range', 'value': '2048 65535'}) => {"failed": true, "item": {"name": "net.ipv4.ip_local_port_range", "value": "2048 65535"}}
msg: this module requires key=value arguments (['name=net.ipv4.ip_local_port_range', 'value=2048', '65535', 'state=present', 'sysctl_file=/etc/sysctl.d/d42.conf', 'ignoreerrors=yes'])
```
I have sent a PR to fix this. Please review.

https://github.com/ceph/ceph-ansible/pull/398

PR #398, opened 2015-09-02
fixing sysctl command to take whitespace-separated multi values for a key

Abhishekvrshny:
https://github.com/ceph/ceph-ansible/issues/397


leseb:
Thanks @Abhishekvrshny

PR #401, opened 2015-09-03
Support setting the "mon pg warn max object skew" in ceph.conf

mwheckmann:
It should be used to disable health warnings about number of PGs
being too low if some pools have very few objects bringing down
the average number of objects per pool. This happens when running RadosGW.

The default is 10 and since the warnings only occur with some use cases,
the default here is 10 as well. Set to 20 or more to silence the warnings.


leseb:
Thanks!

PR #409, opened 2015-09-21
Fix and improve checks when ceph_origin == 'distro'

HanXHX:
Hi,

If you use `ceph_origin` == 'distro' and you don't set to true `ceph_stable`... `ceph_common` fails.

This PR fix this issue and manages well ceph_origin / "ceph sources" (when `ceph_origin` == 'upstream').

Cheers!

Emilien


leseb:
Thanks!

PR #412, opened 2015-10-07
fix regex for partition names, enhance for NVM

bengland2:
Improve regular expression for recognizing partition name.  Regex now uses "or" syntax to represent different families of devices - SCSI, IDE, CCISS, NVM, and it's easy to add others.  The partition number is factored out at the end.  Learned from previous version that we could control how many letters or digits to allow - I allow for 2 digits in partition number so poor folk like me can have >= 10 disks/SSD in their journal device.  

In the future, should we just ask Linux if it's a block device, and then look for a partition number on the end?  That would be a much simpler regex to maintain.

### DIFF
--- a/roles/ceph-osd/tasks/check_devices.yml
+++ b/roles/ceph-osd/tasks/check_devices.yml
@@ -4,9 +4,8 @@
 # it should exist we rc=0 and don't do anything unless we do something like --force
 # As as a final word, I prefer to keep the partition check instead of running ceph-disk prepare with "failed_when: false"
 # I believe it's safer
-
 - name: check if the device is a partition or a disk
-  shell: "echo '{{ item }}' | egrep '/dev/[sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p][0-9]{1,2}'"
+  shell: "echo '{{ item }}' | egrep '/dev/[sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p][0-9]{1,2}$'"
   with_items: devices
   changed_when: false
   failed_when: false
--- a/roles/ceph-osd/tasks/check_devices.yml
+++ b/roles/ceph-osd/tasks/check_devices.yml
@@ -4,8 +4,14 @@
 # it should exist we rc=0 and don't do anything unless we do something like --force
 # As as a final word, I prefer to keep the partition check instead of running ceph-disk prepare with "failed_when: false"
 # I believe it's safer
-- name: check if the device is a partition or a disk
-  shell: "echo '{{ item }}' | egrep '/dev/[sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p][0-9]{1,2}$'"
+#
+# regex syntax uses (pat1|pat2|...|patN) for different families of device
+# names, but has a common expression for partition number at the end.
+# allow 2-digit partition numbers so fast SSDs can be shared by > 9 disks
+# for SSD journals.
+
+- name: check if the device is a partition
+  shell: "echo '{{ item }}' | egrep '/dev/(sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'"
   with_items: devices
   changed_when: false
   failed_when: false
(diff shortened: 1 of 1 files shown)


roles/ceph-osd/tasks/check_devices.yml
```diff
@@ -4,8 +4,9 @@
 # it should exist we rc=0 and don't do anything unless we do something like --force
 # As as a final word, I prefer to keep the partition check instead of running ceph-disk prepare with "failed_when: false"
 # I believe it's safer
+
 - name: check if the device is a partition or a disk
-  shell: "echo '{{ item }}' | egrep '/dev/(([a-z]{3,4}[0-9]$)|(cciss/c[0-9]{1}d[0-9]{1}p[0-9]$))'"
+  shell: "echo '{{ item }}' | egrep '/dev/[sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p][0-9]{1,2}'"
```
leseb: Hum shouldn't it be something more like:

`egrep '/dev/sd[a-z]{1,2}[0-9]$|hd ... ...`

Otherwise it looks like the condition is always true.

bengland2: https://github.com/ceph/ceph-ansible/pull/412 updated with $ on end as you suggested.  So there are different families of device names, but the partition number in the regex is shared across all of them.

'/dev/[sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p][0-9]{1,2}$'"

ok?

----- Original Message -----

bengland2: Has anyone seen this?  When I run

ansible -m setup hp60ds3 says that device /dev/nvme0n1 (Intel P3700 SSD) has no partitions, but it has 58 of them.  Is this a bug in ansible module or task, or is this a problem with ceph-ansible?  I tend to think it's not anything to do with ceph-ansible.  I'm using RHEL7.2 RC4 and ansible-1.9.2-1.el7.noarch.  Also the model and vendor look wrong. 

-ben

hp60ds3 | success >> {
    "ansible_facts": {
...
        "ansible_devices": {
            "nvme0n1": {
...
                "model": null,
                "partitions": {},
                "removable": "0",
                "rotational": "0",
                "scheduler_mode": "",
                "sectors": "1562824368",
                "sectorsize": "512",
                "size": "745.21 GB",
                "support_discard": "512",
                "vendor": null
            },
...

leseb:
The thing is we have to check if the device passed is a partition or not. We can't run `ceph-disk prepare` if we have a partition.

roles/ceph-osd/tasks/check_devices.yml
```diff
@@ -5,7 +5,7 @@
 # As as a final word, I prefer to keep the partition check instead of running ceph-disk prepare with "failed_when: false"
 # I believe it's safer
 - name: check if the device is a partition or a disk
-  shell: "echo '{{ item }}' | egrep '/dev/(([a-z]{3,4}[0-9]$)|(cciss/c[0-9]{1}d[0-9]{1}p[0-9]$))'"
+  shell: "echo '{{ item }}' | egrep '/dev/[sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p][0-9]{1,2}$'"
```
leseb: I think the first bracket is not needed here.
For me, this will do:

`/dev/sd[a-z]{1,2}$|hd[a-z]{1,2}$|cciss/c[0-9]{1}d[0-9]{1}p[0-9]$|nvme[0-9]n[0-9]p][0-9]{1,2}$`

What do you think?

bengland2: Sebastien, you were right, the regex I specified is wrong, thanks for pointing it out.  Here's a correction to the regex syntax along with a quick method of testing it.

# echo /dev/hdab99 | egrep '/dev/(sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]{1,2}p|nvme[0-9]n[0-9]p)[0-9]{1,2}$' ; echo $?
/dev/hdab99
0

So I think this works, try running different device names through it and see if it works for you.  Again, the first part of the regex represents the different flavors of device names, and the last part is the partition number syntax.  Does this work for you?  I tried to limit to 99 disks/server, since Ceph is designed for scale-out, not to cram as many disks as possible into 1 server.    But I wanted to have option to have >= 10 partitions per device, since in the case of NVM SSDs we may need to get away with that sometimes.

If you agree with it I'll submit a new pull request with just this change.

-ben

leseb: Ben, thanks this. This is exactly the method I used to test all the previous regex :).
The last one looks good to me, so feel free to amend your commit and push force the last changes. No need for an extra PR.

Thanks @bengland2 !

leseb:
@bengland2 oops I missed your last patchset, thanks!

Issue #413, opened 2015-10-08
Still failing to import the ceph stable repo key


bsanders:
I don't see a way to re-open issue #403, which I ran into yesterday (and added comments on).  Apologies if I'm breaking github-etiquette by creating a new issue...

Wget'ing the file with an escaped string and running the import command by hand allows the import to succeed (though it still won't pass this step in the playbook).  Likewise, if you change the rpm_key path to something local and have ansible copy the file over, the import succeeds and (in this case) ansible can move on.  I also noted at the end of that issue [https://github.com/ceph/ceph-ansible/issues/403#issuecomment-146341201] that there's a 301 HTTP code when you do the wget, which if Ansible is correctly escaping the URL, it may still be choking on.

Didn't Ceph's keys change recently?  Is that where the 301 came from?

Issue #403, opened 2015-09-04
Failure importing the ceph stable repo key


matthewrees:
This issue seems to be intermittent, but it has happened twice in the last 2 weeks and the first time it was no longer an issue when I ran the playbook less than an hour later.

Workstation: Linux Mint 17.1 (based on Ubuntu 14.04) with Ansible 1.9.3
Servers: CentOS 7.1 with latest updates

Trace of the playbook failing:

TASK: [ceph-common | install the ceph stable repository key] ****************** 
failed: [XXX] => {"failed": true, "parsed": false}
BECOME-SUCCESS-vpcuvjvkadqieptjwkavpyewfqoisxnh
Traceback (most recent call last):
  File "/root/.ansible/tmp/ansible-tmp-1441356162.03-168602942242488/rpm_key", line 2587, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1441356162.03-168602942242488/rpm_key", line 199, in main
    RpmKey(module)
  File "/root/.ansible/tmp/ansible-tmp-1441356162.03-168602942242488/rpm_key", line 87, in __init__
    keyfile = self.fetch_key(key)
  File "/root/.ansible/tmp/ansible-tmp-1441356162.03-168602942242488/rpm_key", line 121, in fetch_key
    key = rsp.read()
AttributeError: 'NoneType' object has no attribute 'read'
OpenSSH_6.6.1, OpenSSL 1.0.1f 6 Jan 2014
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 19: Applying options for *
debug1: auto-mux: Trying existing master
debug1: mux_client_request_session: master session id: 2
Shared connection to XXX closed.

Testing importing the key manually on the ceph host:

[root@XXX ~]# rpm --import https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/release.asc
error: https://ceph.com/git/?p=ceph.git: key 1 not an armored public key.

For what it's worth, when running this playbook, this test fails on both hosts being provisioned for the first time as well as hosts that have already been provisioned with ceph-ansible and are actively running ceph.

leseb:
Hum can't really reproduce this, I suspect this is a system issue not related to ansible or anything.

matthewrees:
[...]

PR #414, opened 2015-10-16
Changed URL for Ceph release and developer GPG keys, because previous…

bsanders:
… key URL suffers sporadic timeouts.

This addresses the issue I've been having in #403 and #413. It's also being discussed on the Ceph-users mailing list that there are ipv6 issues with the other URL.


leseb:
Thanks @bsanders

PR #433, opened 2015-11-06
This change address radosgw not getting started on CentOS

dvusboy:
* The explicit check of 'ansible_distribution' being 'RedHat'
  will not work on CentOS hosts.
* Check the value of 'ansible_os_family' instead.


dvusboy:
I was following the bare-metal demo and tried the playbook on CentOS 7.1 hosts. I found out ```radosgw``` didn't get started and discovered the problem with this check.

PR #434, opened 2015-11-06
Defaults for the ceph-mon role should be setting the 'mon_group_name'

dvusboy:
Without this, and if all these group name variables are commented out in ```group_vars/all```, then the handler in ```ceph-common``` that depends on the value of ```mon_group_name``` would fail:

```
NOTIFIED: [ceph-common | restart ceph mons] ***********************************
fatal: [ctrl-gprt] => error while evaluating conditional: socket.rc == 0 and ansible_distribution != 'Ubuntu' and mon_group_name in group_names
```

which does look like some sort of unexpected behaviour on Ansible because those ```*_group_name``` defaults are set in the ```ceph-common``` role. I'm using Ansible 1.9.4.

### DIFF
--- a/roles/ceph-mon/defaults/main.yml
+++ b/roles/ceph-mon/defaults/main.yml
@@ -7,7 +7,7 @@
 
 fetch_directory: fetch/
 
-rgw_group_name: rgws
+mon_group_name: mons
 
 # ACTIVATE BOTH FSID AND MONITOR_SECRET VARIABLES FOR NON-VAGRANT DEPLOYMENT
 fsid: "{{ cluster_uuid.stdout }}"


dvusboy:
Thanks for merging this pull request.

I wanted to point out something regarding ```defaults``` on dependent roles: They don't really trickle through to the parent role as assumed by how these default group names are set in the ```ceph-common``` role. I looked into [the code](https://github.com/ansible/ansible/blob/stable-1.9/lib/ansible/playbook/play.py#L246) for the ```stable-1.9``` branch of Ansible and these defaults are only applicable to the role in play; they are not transitive at all. Ansible 2 is a completely rewrite, especially in this area, so I haven't checked the behaviour yet.

Issue #437, opened 2015-11-19
ceph-osd | prepare osd disk(s) ignored when GPT labels missing


BjoernT:
The current ceph-osd tasks are ignoring disks with missing GPT labels leading to issues during later tasks to actually prepare/activate the OSDs so you end up with missing OSDs in your tree.
I personally would prefer to bail out the tasks so it can be fixed. Or is this intentionally ?

TASK: [ceph-osd | prepare osd disk(s)] ****************************************
skipping: [617569-cinder01] => (item=[{'cmd': "parted --script /dev/sdc print | egrep -sq '^ 1.*ceph'", 'end': '2015-11-19 19:45:10.241914', 'stderr': u'', 'stdout': u'', 'item': u'/dev/sdc', 'changed': False, 'rc': 0, 'failed': False, 'warnings': [], 'delta': '0:00:00.004694', 'invocation': {'module_name': u'shell', 'module_complex_args': {}, 'module_args': u"parted --script /dev/sdc print | egrep -sq '^ 1.*ceph'"}, 'stdout_lines': [], 'failed_when_result': False, 'start': '2015-11-19 19:45:10.237220'}, {'cmd': "echo '/dev/sdc' | egrep '/dev/(([a-z]{3,4}[0-9]$)|(cciss/c[0-9]{1}d[0-9]{1}p[0-9]$))'", 'end': '2015-11-19 19:45:10.069739', 'stderr': u'', 'stdout': u'', 'item': u'/dev/sdc', 'changed': False, 'rc': 1, 'failed': False, 'warnings': [], 'delta': '0:00:00.002533', 'invocation': {'module_name': u'shell', 'module_complex_args': {}, 'module_args': u"echo '/dev/sdc' | egrep '/dev/(([a-z]{3,4}[0-9]$)|(cciss/c[0-9]{1}d[0-9]{1}p[0-9]$))'"}, 'stdout_lines': [], 'failed_when_result': False, 'start': '2015-11-19 19:45:10.067206'}, u'/dev/sdc', u'/dev/sdc'])
[...]

PR #439, opened 2015-11-20
Check partition status

leseb:
Verify that partitions (for both osd disks and journal disks) are sane
before attempting to prepare the device. Fail if parted fails for
whatever reason.

Closes: #437


leseb:
this breaks vagrant... ugly fix coming :(

PR #440, opened 2015-11-20
Add systemd-specific start/restart tasks/handlers for el7/infernalis

eikef:
This makes the roles work on el7; the old sysv-initscript invocations no longer work. Unfortunately the "restart osd with systemd" handler is now a "restart everything on the node" handler at the moment (to fix this a list of OSD-ids would need to be generated somehow, there no longer is a "restart all osds" option like in SysV and Upstart).

### DIFF
--- a/roles/ceph-mds/tasks/pre_requisite.yml
+++ b/roles/ceph-mds/tasks/pre_requisite.yml
@@ -208,11 +208,30 @@
   changed_when: false
   when: ansible_distribution == "Ubuntu"
 
-- name: start and add that the metadata service to the init sequence
+- name: start and add that the metadata service to the init sequence (before infernalis)
   service:
     name: ceph
     state: started
     enabled: yes
     args: mds
   changed_when: false
-  when: ansible_distribution != "Ubuntu"
+  when: ansible_distribution != "Ubuntu" and
+    (ceph_stable_release == 'dumpling' or
+    ceph_stable_release == 'emperor' or
+    ceph_stable_release == 'firefly' or
+    ceph_stable_release == 'giant' or
+    ceph_stable_release == 'hammer')
+
+
+- name: start and add that the metadata service to the init sequence (for or after infernalis)
+  service:
+    name: ceph-mds@{{ ansible_hostname }}
+    state: started
+    enabled: yes
+  changed_when: false
+  when: ansible_distribution != "Ubuntu" and not
+    (ceph_stable_release == 'dumpling' or
+    ceph_stable_release == 'emperor' or
+    ceph_stable_release == 'firefly' or
+    ceph_stable_release == 'giant' or
+    ceph_stable_release == 'hammer')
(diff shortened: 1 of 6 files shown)


roles/ceph-rgw/tasks/start_radosgw.yml
```diff
@@ -24,9 +24,26 @@
...
+    ceph_stable_release == 'giant' or
+    ceph_stable_release == 'hammer')
+
+- name: start rgw on red hat (after infernalis)
+  service:
+    name: ceph-radosgw@{{ ansible_hostname }}
+    state: started
+    enabled: yes
+  when: ansible_os_family == 'RedHat' and not
+    (ceph_stable_release == 'dumpling' or
```
leseb: this should be `!=`

roles/ceph-osd/tasks/activate_osds.yml
```diff
@@ -23,8 +23,29 @@
...
+     ceph_stable_release == 'hammer')
+
+- name: start and add that the osd service(s) to the init sequence (on or after infernalis)
+  service:
+    name: ceph.target
+    state: started
+    enabled: yes
+  when:
+    ansible_distribution != "Ubuntu" and not
+    (ceph_stable_release == 'dumpling' or
```
leseb: this should be `!=`

roles/ceph-mds/tasks/pre_requisite.yml
```diff
@@ -208,11 +208,30 @@
...
+
+
+- name: start and add that the metadata service to the init sequence (for or after infernalis)
+  service:
+    name: ceph-mds@{{ ansible_hostname }}
+    state: started
+    enabled: yes
+  changed_when: false
+  when: ansible_distribution != "Ubuntu" and not
+    (ceph_stable_release == 'dumpling' or
```
leseb: this should be `!=`

roles/ceph-osd/tasks/activate_osds.yml
```diff
@@ -23,8 +23,29 @@
...
   when: crush_location
 
-- name: start and add that the osd service(s) to the init sequence
+- name: start and add that the osd service(s) to the init sequence (before infernalis)
   service:
     name: ceph
     state: started
     enabled: yes
+  when:
+    ansible_distribution == "Ubuntu" or
```
leseb: use 'and' instead of 'or'?

eikef: I wanted the behavior to stay the same when the role runs with either Ubuntu or an older supported version of ceph/el. Since I assume that the role works on Ubuntu with Infernalis right now (maybe that is a wrong assumption) that's the first check, and the second check should cancel out any ceph that run on el6 (as infernalis is not supported on el6).

Maybe there is a better way, but the or was intentional :)

leseb: Correct, sorry it was just me reading too fast :).

rokka-n:
If I need to restart osd or mon or cento7 with playbook, what would be the commands for this?

leseb:
@rokka-n 

For a monitor: `systemctl restart ceph-mon@<hostname>.service`
For an OSD: `systemctl restart ceph-osd@<id>.service`

PR #441, opened 2015-11-23
Updated yum to dnf for installation.

safiyat:
The package manager has been changed from yum to dnf since Fedora 22.
Ref: https://fedoraproject.org/wiki/Changes/ReplaceYumWithDNF

### DIFF
--- a/install-ansible.sh
+++ b/install-ansible.sh
@@ -26,6 +26,6 @@ if [[ -x $(which lsb_release 2>/dev/null) ]]; then
       apt-get install -y ansible
   fi
 elif [[ -r /etc/redhat-release ]]; then
-  yum install -y epel-release
-  yum install -y ansible
+  dnf install -y epel-release
+  dnf install -y ansible
 fi


leseb:
@safiyat what about people not using f22?

safiyat:
@leseb Oh, my bad! Redhat has replaced yum with dnf in all its releases but it seems Cent OS still uses yum as the package manager.
This change to `dnf` can be done once Cent OS switches loyalty from `yum` too.

leseb:
@safiyat or you can make an exception for centos only? :)

safiyat:
@leseb I don't understand you, but can be done. :)

leseb:
@safiyat what I meant is that you can use `dnf` for every redhat based systems except for centos and rhel.

safiyat:
@leseb Alright.
I think we can defer this transition to `dnf` for now. Though most of the Redhat based systems internally interpret `yum` calls as `dnf`.
Let's close this issue I think.

PR #443, opened 2015-11-24
Tweak the ansible configuration file

leseb:
This change reduces the time to run ansible.
Note that when pipelining is enabled root should be used or
disable the requiretty option for the remote_user


leseb:
Looks like this breaks some pieces of the playbook.
Feel free to re-open if someone has the time to investigate.

Issue #410, opened 2015-09-23
ceph-mon fail at "copy keys to the ansible server"


xahare:
ceph-ansible version:
commit 2c01567a4f0016aa5a0762c871f479ff40689fcf
Date:   Mon Sep 21 18:29:20 2015 +0200

when trying this on 3 dells with a stock, updated ubuntu 14.04 lts. they used to run ceph giant. 

heres a typescript of the ansible run and the vars used. https://gist.github.com/xahare/d67078edd88cd6eb6c60

the problem seems to start at line 3002:

[0;31mfailed: [ceph1] => (item=/var/lib/ceph/bootstrap-rgw/ceph.keyring) => {"checksum": "da39a3ee5e6b4b0d3255bfef95601890afd80709", "dest": "/void/pixel/ansible/westcoastdata/fetch/6AAABE05-8E3A-476F-B55C-6F5AACC342AC/var/lib/ceph/bootstrap-rgw/ceph.keyring", "failed": true, "file": "/var/lib/ceph/bootstrap-rgw/ceph.keyring", "item": "/var/lib/ceph/bootstrap-rgw/ceph.keyring", "md5sum": "d41d8cd98f00b204e9800998ecf8427e", "remote_checksum": "074703eeaa1b4288e19ed432437d211683620140", "remote_md5sum": null}[0m
[0;31mmsg: checksum mismatch[0m

cant reproduce this on vmware or virtualbox.

leseb:
Hum could it be an underlying problem? Like a fs issue or something.
If you calculate the md5 manually what do you get?

xahare:
wrote a script to generate preseeded ubuntu isos to make sure all the nodes were the same and reinstalled on all the nodes. maybe a hardware problem? any ideas on how to look? threw a couple ceph command after the 2nd run on the end. this time it thinks it worked, but im still skeptical.

as before, everything was fine on vmware. heres the same section this time. had to take the /usr/sbin/ off sgdisk in the purge playbook for that work. manually running after the playbook, they all had the same md5sum.

[...]

PR #448, opened 2015-12-04
Add run_once to 'copy keys to the ansible server'

mattt416:
I have seen a number of failures on this task due to mismatch of
checksum of source file and destination.  I suspect this is due to a
race condition caused by several hosts simultaneously copying the same
file to single location on the deployment server.

This change simply updates the 'copy keys to the ansible server' task
by adding 'run_once', which limits the task to being run on a single
MON host.

Closes issue #410

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -39,6 +39,7 @@
     src: "{{ item }}"
     dest: "{{ fetch_directory }}/{{ fsid }}/{{ item }}"
     flat: yes
+  run_once: true
   with_items:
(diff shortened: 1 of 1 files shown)


mattt416:
I have just instantiated 10 deployments with this change in place and have yet to receive a failure on this task.  Yesterday I was hitting this issue around 2/3 times.  It would be great to get some feedback from those who were hitting this issue to confirm it resolves it before we go ahead and merge.

leseb:
@mattt416 thanks for the PR, this change is harmless but let's wait for more inputs. According to your results I feel pretty confident that this fixes the issue though...

d34dh0r53:
:+1: I too have seen this, and I think this is an appropriate fix

Issue #446, opened 2015-12-02
ceph_stable_release logic is incorrect


michaelforman:
Multiple locations in  roles/ceph-mon/tasks/deploy_monitors.yml
The inverse of 
```
  when:
    ceph_stable_release == 'dumpling' or
    ceph_stable_release == 'emperor' or
    ceph_stable_release == 'firefly' or
    ceph_stable_release == 'giant' or
    ceph_stable_release == 'hammer'
```
Is not
```
    ceph_stable_release != 'dumpling' or
    ceph_stable_release != 'emperor' or
    ceph_stable_release != 'firefly' or
    ceph_stable_release != 'giant' or
    ceph_stable_release != 'hammer'
```

Either swap the the ORs with ANDs  or negate the original check

PR #449, opened 2015-12-04
fixes ceph_stable_release logic mentioned in issue #446

nexecook:
fixes ceph_stable_release logic mentioned in issue #446

I changed the argument used for starting the mds server. (pre
infernalis)
```
service ceph start mds
```
errors, while
```
service ceph start mds.$hostname
```
correctly starts the service.

I also changed the mds directory ownership from ceph:cephh to root:root
again, for pre-infernalis.

And finally, add the ceph_stable_releases checks for the upstart
activation task `for or after infernalis release'.

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -47,6 +47,16 @@ dummy:
 #ceph_stable_key: https://download.ceph.com/keys/release.asc
 #ceph_stable_release: infernalis # ceph stable release
 
+####################
+## Stable Releases #
+####################
+#ceph_stable_releases:
+#  - dumpling
+#  - emperor
+#  - firefly
+#  - giant
+#  - hammer
+#          
 # Use the option below to specify your applicable package tree, eg. when using non-LTS Ubuntu versions
 # # for a list of available Debian distributions, visit http://ceph.com/debian-{{ ceph_stable_release }}/dists/
 # for more info read: https://github.com/ceph/ceph-ansible/issues/305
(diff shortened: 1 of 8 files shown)


leseb:
@nexecook thanks for the fix, I'd like to enhance the logic a bit more could you please do the following as well?

In `roles/ceph-mon/defaults/main.yml`:
```
###################
# Stable Releases #
###################
ceph_stable_releases:
  - dumpling
  - emperor
  - firefly
  - giant
  - hammer
```

And then do: `ceph_stable_release not in ceph_stable_releases` or `ceph_stable_release not in ceph_stable_releases` when appropriate?

nexecook:
What should be added to group_vars, that array? (I am not very familiar with ansible)

leseb:
@nexecook `group_vars/all` must have the exact same content as `roles/ceph-common/defaults/main.yml`.

group_vars/all.sample
```diff
@@ -47,6 +47,16 @@ dummy:
 #ceph_stable_key: https://download.ceph.com/keys/release.asc
 #ceph_stable_release: infernalis # ceph stable release
 
+####################
+## Stable Releases #
```
leseb: Can you just remove one '#' so we only have `# Stable Releases`. :)

roles/ceph-common/defaults/main.yml
```diff
@@ -45,6 +45,16 @@ ceph_stable: false # use ceph stable branch
 ceph_stable_key: https://download.ceph.com/keys/release.asc
 ceph_stable_release: infernalis # ceph stable release
 
+####################
+## Stable Releases #
```
leseb: Can you just remove one '#' so we only have `# Stable Releases`. :)

leseb:
I don't have anything to test at the moment, will have a look soon.

leseb:
@nexecook tested and validated on both Hammer and Infernalis, thanks!

Issue #451, opened 2015-12-07
Add configuration options for osd scrub begin/end hour


BjoernT:
The begin and end of the scrubbing process should be configurable, please add overrides for 

osd scrub begin hour
osd scrub end hour

PR #452, opened 2015-12-07
Adding osd_scrub_begin_hour and osd_scrub_end_hour

BjoernT:
Adding osd_scrub_begin_hour and osd_scrub_end_hour to limit scrub window of 24h if needed Fixes-Bug: #451

### DIFF
--- a/roles/ceph-common/templates/ceph.conf.j2
+++ b/roles/ceph-common/templates/ceph.conf.j2
@@ -115,6 +115,10 @@
   filestore op threads = {{ filestore_op_threads }}
   filestore max sync interval = {{ filestore_max_sync_interval }}
   osd max scrubs = {{ osd_max_scrubs }}
+{% if ceph_stable_release not in ['argonaut','bobtail','cuttlefish','dumpling','emperor','firefly','giant'] %}
+  osd scrub begin hour = {{ osd_scrub_begin_hour }}
+  osd scrub end hour = {{ osd_scrub_end_hour }}
+{% endif %}
   # Recovery tuning
   osd recovery max active = {{ osd_recovery_max_active }}
   osd max backfills = {{ osd_max_backfills }}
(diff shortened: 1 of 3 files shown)


leseb:
@BjoernT Thanks for submitting a PR!
:-1: Unfortunately, this will only work for Hammer and above. We need a warning in the vars file and a condition in the template.

BjoernT:
Oh, didn't know that older versions are still supported. I can work on adding some logic into it

leseb:
Well officially I'm trying to do my best to at least support the last 2 LTSs. Firefly is still heavily used I believe, so my wait next year to drop its support :). After this, we will probably stick with a single LTS.
Thanks!

BjoernT:
@leseb  I added a condition that these options are only configured from hammer and later. It's a bit cheesy but we can drop it once firefly got removed

leseb:
@BjoernT thanks! one last thing can you had a warning in the var file to explain that this will only work for hammer and above? And please edit the group_vars/all.sample as well :)

BjoernT:
@leseb  Sure just updated.

leseb:
@BjoernT can we have the same warning message in `group_vars/all.sample` as well? :)

BjoernT:
@leseb No problem, updated

roles/ceph-common/templates/ceph.conf.j2
```diff
@@ -115,6 +115,10 @@
   filestore op threads = {{ filestore_op_threads }}
   filestore max sync interval = {{ filestore_max_sync_interval }}
   osd max scrubs = {{ osd_max_scrubs }}
+{% if ceph_stable_release not in ['argonaut','bobtail','cuttlefish','dumpling','emperor','firefly'] %}
```
leseb: did you miss giant as well?

BjoernT: oops, the giant was apparently to small, adding it :-) Thanks

leseb:
Thanks!

BjoernT:
Is there any back port needed for any other branch ?

leseb:
Unfortunately, I need to clean up the branches and probably start creating some stable branches as well... So the answer is no. Thanks for pointing that out :)

PR #442, opened 2015-11-23
Support infernalis for rh storage

leseb:
WIP

Signed-off-by: Sébastien Han <seb@redhat.com>

PR #453, opened 2015-12-11
corrections for infernalis on EL7

crcceph:
CentOS7 needs python-setuptools for /usr/bin/ceph-detect-init so I added this to the redhat dependencies.

The playbook hangs at gathering the mon keys, which is a result of ceph-mon not being able to create the sock in /var/run/ceph because of permissions.

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -100,10 +100,30 @@
     - restart ceph rgws on red hat
     - restart ceph rgws with systemd
 
-- name: create rbd client directory
+- name: create rbd client directory on other
   file:
     path: "{{ rbd_client_admin_socket_path }}"
     state: directory
     owner: root
     group: root
     mode: 0644
+  when: ansible_os_family != 'RedHat' or
+    (ceph_stable_release == 'dumpling' or
+    ceph_stable_release == 'emperor' or
+    ceph_stable_release == 'firefly' or
+    ceph_stable_release == 'giant' or
+    ceph_stable_release == 'hammer')
+
+- name: create rbd client directory on red hat (for infernalis or after and systemd and ceph user)
+  file:
+    path: "{{ rbd_client_admin_socket_path }}"
+    state: directory
+    owner: ceph
+    group: ceph
+    mode: 0744
+  when: ansible_os_family == 'RedHat' and not
+    (ceph_stable_release == 'dumpling' or
+    ceph_stable_release == 'emperor' or
+    ceph_stable_release == 'firefly' or
+    ceph_stable_release == 'giant' or
+    ceph_stable_release == 'hammer')
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -107,7 +107,7 @@
     owner: root
     group: root
     mode: 0644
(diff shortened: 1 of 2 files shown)


leseb:
Thanks @crcceph . LGTM, let's just wait a little bit for #449 so you can re-use the logic.

crcceph:
While doing some testing on another issue today I noticed that before the permissions are set by the playbook the mode was 0770 (set during the package install), so that might be a better choice.
I looked at #449 and agree that test should be used once it is merged.  I am using only C7, so I am not sure what other logic options might be required (Ubuntu?).

roles/ceph-common/tasks/main.yml
```diff
@@ -100,10 +100,30 @@
...
 
-- name: create rbd client directory
+- name: create rbd client directory on other
   file:
     path: "{{ rbd_client_admin_socket_path }}"
     state: directory
     owner: root
     group: root
     mode: 0644
+  when: ansible_os_family != 'RedHat' or
```
leseb: There is nothing distro specific here, so you can just use the logic implemented in #449

roles/ceph-common/tasks/main.yml
```diff
@@ -100,10 +100,30 @@
...
+    ceph_stable_release == 'hammer')
+
+- name: create rbd client directory on red hat (for infernalis or after and systemd and ceph user)
+  file:
+    path: "{{ rbd_client_admin_socket_path }}"
+    state: directory
+    owner: ceph
+    group: ceph
+    mode: 0744
+  when: ansible_os_family == 'RedHat' and not
```
leseb: There is nothing distro specific here, so you can just use the logic implemented in #449

leseb:
@crcceph could you re-use the logic of #442? Thanks!

roles/ceph-common/tasks/main.yml
```diff
@@ -100,10 +100,20 @@
...
 
-- name: create rbd client directory
+- name: create rbd client directory (before infernalis)
   file:
     path: "{{ rbd_client_admin_socket_path }}"
     state: directory
     owner: root
     group: root
     mode: 0644
+  when: ceph_stable_release in ceph_stable_releases
```
leseb: `when: not is_ceph_infernalis`

roles/ceph-common/tasks/main.yml
```diff
@@ -100,10 +100,20 @@
...
+
+- name: create rbd client directory on red hat (for infernalis or after)
+  file:
+    path: "{{ rbd_client_admin_socket_path }}"
+    state: directory
+    owner: ceph
+    group: ceph
+    mode: 0770
+  when: ceph_stable_release not in ceph_stable_releases
```
leseb: `when: is_ceph_infernalis`

roles/ceph-common/tasks/main.yml
```diff
@@ -100,10 +100,20 @@
...
     - restart ceph rgws with systemd
 
-- name: create rbd client directory
+- name: create rbd client directory (before infernalis)
   file:
     path: "{{ rbd_client_admin_socket_path }}"
     state: directory
     owner: root
     group: root
     mode: 0644
```
leseb: `mode: 0770`

Issue #455, opened 2015-12-15
Issues with rgw print continue = false


Abhishekvrshny:
Hi @leseb 

I ran into an issue where the PUTs where latent on a cluster deployed through ceph-ansible, wherein, there was a delay of ~3s in every PUT. The cluster was running rgw with civetweb and the client was aws-sdk-java version 1.9.4. I was able to pin-point it to `rgw print continue = false` in `ceph.conf.j2`. The default value for this parameter is true as specified here: http://docs.ceph.com/docs/master/radosgw/config-ref/. Is there any specific reason for setting it to `false`?

Related Ceph tracker bug : http://tracker.ceph.com/issues/12640

leseb:
@Abhishekvrshny I don't really remember why we set that value to `false` however this doesn't seem to make more sense to keep it disabled. After looking at the tracker bug, I don't see any reason why we should keep that value as this seems to bring problems.

PR #456, opened 2015-12-15
removed rgw print continue = false, to set it to its default value of…

Abhishekvrshny:
… true

Fixes : https://github.com/ceph/ceph-ansible/issues/455


leseb:
Thanks @Abhishekvrshny

PR #460, opened 2015-12-16
fixed for infernalis on EL7

crcceph:
EL7 needs python-setuptools as a dependency
when running under ceph user the /var/run/ceph directory needs to be owned by ceph

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -30,6 +30,7 @@ redhat_package_dependencies:
   - yum-plugin-priorities.noarch
   - epel-release
   - ntp
+  - python-setuptools
 
 ## Configure package origin
 #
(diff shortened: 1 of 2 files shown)


leseb:
Argh this actually won't work since the fact is being created after...
Can you also move the set_fact here: https://github.com/crcceph/ceph-ansible/blob/master/roles/ceph-common/tasks/main.yml#L72

There is no need to create a new PR, just do a git push force :).
Thanks!

crcceph:
I tried getting the previous PR to rebase but wasn't having much luck.  I just moved the set above the test.

leseb:
Please put the fact here: https://github.com/crcceph/ceph-ansible/blob/master/roles/ceph-common/tasks/main.yml#L39 

Thanks!

crcceph:
Moved as requested
Still an ansilble newb, so didn't think about putting near the top where it belongs.

leseb:
Thanks @crcceph!

PR #461, opened 2015-12-16
Using config_template

leseb:
Thanks to @cloudnull great patch at
https://github.com/ansible/ansible/pull/12555
we now have the ability to add more configuration options instead of
having to push a PR to add a new option to the template. So you can
dynamically add and remove flags.

To use it, edit `ceph_conf_overrides` in `group_vars/all` like so:

```
ceph_conf_overrides
 global:
     foo: 12345
     bar: 6789
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ group_vars/rgws
 group_vars/restapis
 *.DS_Store
 site.yml
+*.pyc
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,2 +1,3 @@
 [defaults]
 ansible_managed = Please do not change this file directly since it is managed by Ansible and will be overwritten
+action_plugins = plugins/actions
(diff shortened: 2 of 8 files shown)


leseb:
@cloudnull while trying to add new values, I'm getting the following error: http://pastebin.com/bx248qVv
Any inputs would be appreciated, thanks :)

leseb:
hum looks like 'indent' are not supported/compliant with the INI style format...

leseb:
Sorry for everyone using the module already. This will likely perform a restart of the cluster while running the playbook again since the structure of the `ceph.conf` will change. One thing to do is to remove all the indent from the `ceph.conf`.

cloudnull:
@leseb Just following up, I've been on holiday for the past little bit. While it seems like all is well I wanted to follow up to make. Was everything figured out regarding the `config_template`?

leseb:
@cloudnull Hey! No worries, thanks for catching up :). Yes once I figured the indent issue everything went smooth. Thanks again for this module. Hopefully it'll get merged next early next year.

PR #463, opened 2015-12-17
create new ceph-install role.

erikh:
This pulls out most of ceph-common's install process and makes ceph-common
depend on it. This is useful for incorporating ceph installation into VM
images, but for allowing post-start configuration via ceph-common and the other
roles.

We have already done this in volplugin's ansible and it works really, really well. I know this is large, but please consider this patch. 

Thank you for your time.

Signed-off-by: Erik Hollensbe <github@hollensbe.org>


leseb:
Hum not sure to really understand the use case. Can't you simply add an 'install' tag to ceph-common to isolate the installation instead of creating a new role? And then use another tag for post-config?
Thanks!

erikh:
It partitions the notion of installing the product vs. configuring it. I'm happy to do it the other way, however. I'll need a few days.

leseb:
As you know, this is a huge change that will impact a lot of users. We don't use tags at all anyway. I think we should, this is a good exercice and a good start I guess.
I can wait a few days :) Thanks!

PR #464, opened 2015-12-18
Re organise roles structure

leseb:
We now use numbers as prefix for tasks files. It is easier and more
visual to determine the execution order.


leseb:
To much work, few benefit, only cosmetic...

Issue #466, opened 2016-01-08
RFE firewall support? iptables/firewalld


pschiffe:
Hello,

is there, or are you planning to add support for firewall configuration? It would be really handy for new users.

leseb:
Hey,

I've been wondering about this one for a while.
I usually believe that firewall rules should sit outside of this playbook and should be managed by an infrastructure playbook.
I'm just not sure how to address that yet, the only thing I can do for now is to provide the list of ports that must be opened...

wdyt?

pschiffe:
That would by actually great start. Document the needed ports on the readme. Because I didn't find it anywhere here, even demo video didn't mention this, so when I started the playbook, it failed on timeout, and I had to investigate what's going on. And since on clean OS firewall is enabled by default, I was hit by that.

Another step could by just to add configuration option to open ports in firewall, commented out by default. I mean in all.sample in group_vars. I think there are valid use cases for this, when deploying on internal infra, or when networking isn't very complicated..

leseb:
Hum actually one thing I can potentially do is to perform some networking test in the playbook and fail if we detect that some ports are not reachable. So if it does, you know that's a firewalling issue and can act accordingly.
wdyt?

pschiffe:
Sounds good to me, definitely better than current state.

leseb:
Ok tested and valided on #469. Feel free to test, closing this :)

PR #469, opened 2016-01-08
Check for blocked ports

leseb:
As raised in #466 it is important in order to avoid unnecessary
troubleshooting to check that ceph ports are allowed on the platform.
The check runs a nmap command from the host running Ansible
to all the ceph nodes with their respective ports.


leseb:
Still testing this.

PR #475, opened 2016-01-11
Stable rh storage repo

bengland2:
For ceph_stable_rh_storage: true option, the repo file wasn't getting created for some reason.  Does this make sense to you?  Specifically from group_vars/all:

ceph_origin: 'distro'
ceph_stable_rh_storage: true
ceph_stable_rh_storage_iso_install: true # usually used when nodes don't have access to cdn.redhat.com


leseb:
Hey @bengland2, I think the error might be here: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/installs/install_on_redhat.yml#L10

You need to set it to upstream however this is confusing and this should be distro (just like you did).

leseb:
@bengland2 any update on this?

bengland2:
You are right, the code was sort of there already, I'll generate a different pull request to fix what already exists.

PR #476, opened 2016-01-11
allow customization of debian ceph stable repo url

Logan2211:
My ceph cluster is not connected to the internet and must use local repo sources. Unfortunately at this time that is not possible with this role. My change moves the repo url to a defaults variable to allow overriding it.


leseb:
@Logan2211 Can you update `group_vars/all.sample`as well?

Logan2211:
Done

PR #481, opened 2016-01-15
install Red Hat Ceph Storage repo when needed

bengland2:
The code to install RHCS repo template was there but was not getting enabled.  Also, because we have tasks/installs/ subdirectory, ansible was not finding the template file.  These 2 lines of code change seem to resolve the problem.  Do they break anything else?   Would it be possible to have regression tests check RHCS case also?

### DIFF
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -7,7 +7,7 @@
 
 - name: configure ceph yum repository
   include: redhat_ceph_repository.yml
-  when: ceph_origin == 'upstream'
+  when: ceph_origin == 'upstream' or ceph_stable_rh_storage
 
 - name: install ceph
   yum:
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -7,7 +7,8 @@
 
 - name: configure ceph yum repository
   include: redhat_ceph_repository.yml
-  when: ceph_origin == 'upstream' or ceph_stable_rh_storage
+  when: ceph_origin == 'upstream' or 
+        ceph_stable_rh_storage
 
 - name: install ceph
   yum:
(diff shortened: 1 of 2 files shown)


roles/ceph-common/tasks/installs/install_on_redhat.yml
```diff
@@ -7,7 +7,7 @@
 
 - name: configure ceph yum repository
   include: redhat_ceph_repository.yml
-  when: ceph_origin == 'upstream'
+  when: ceph_origin == 'upstream' or ceph_stable_rh_storage
```
leseb: Can you please use the following syntax style? (readability purpose)

```
when:
  ceph_origin == 'upstream' or
  ceph_stable_rh_storage
```

Thanks!

bengland2:
added and tested "readability" commit

leseb:
Why don't you simply turn `ceph_origin` to upstream instead of forcing `ceph_stable_rh_storage` in the condition?

bengland2:
so when I try that, it still gives me this error:

Perform task: ceph-common | add red hat storage repository (y/n/c):  ********** 
fatal: [rhs-cli-01] => input file not found at /mnt/oldroot/home/ben/ceph/dm-cache/ceph-ansible/roles/ceph-common/tasks/templates/redhat_storage_repo.j2 or /mnt/oldroot/home/ben/ceph/dm-cache/ceph-ansible/redhat_storage_repo.j2

That's what the 2nd commit is for.  So some change is needed.  

Also, this ISN'T upstream.   What is the purpose of the ceph_origin variable then?

leseb:
Yes I'm in favour of the template path fix.

One way to describe `ceph_origin` is when you use an external repo source from your distribution.
All the distributions come with default repository, if you want to use the ceph version of that repo then change `ceph_origin` to origin, if you plan on adding an extra repo file then just stick with `ceph_origin` to upstream.

I can understand that somehow this might be confusing, perhaps I should add a comment to better explain when the variable should be enabled or not.

wdyt?

bengland2:
So ceph_origin = 'upstream' really means distro-external repo will be added, which is what RHCS wants. So I'll create a new PR with just the 2nd commit.  

It would be useful to explain this a little more in the comments near the variable in the all.sample file, something like "ceph_origin = 'distro' means that no separate repo file will be added, and you will get whatever version of Ceph is included in the Linux distro.".   But I'll leave that up to you.

leseb:
Will do that @bengland2 , sorry for the confusion :)

PR #485, opened 2016-01-17
Configurable Ceph Conf Directory Permissions

byronmccollum:
This change allows for configurable Ceph Conf Directory permissions. This
is required for integrators of Ceph, like OpenStack Cinder, which needs to
read from /etc/ceph for operation.

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -7,6 +7,20 @@
 
 fetch_directory: fetch/
 
+###############
+# PERMISSIONS #
+###############
+
+# Permissions for /etc/ceph configuration directory
+conf_directory_owner: root
+conf_directory_group: root
+conf_directory_mode: 644
+
+# Permissions for /etc/ceph/ceph.conf configuration file
+conf_file_owner: root
+conf_file_group: root
+conf_file_mode: 644
+
 ###########
 # INSTALL #
 ###########
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -10,6 +10,20 @@ dummy:
 
 #fetch_directory: fetch/
 
+###############
+# PERMISSIONS #
+###############
+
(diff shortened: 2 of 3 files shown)


leseb:
Thanks for this PR. Just 2 things:

* could you please add a comment to explain which file will get the configured permissions?
* update group_vars/all.sample to reflect this change 

Thanks!

leseb:
Oh and the ceph.conf file should probably follow the same thing don't you think?

byronmccollum:
Done and done.

byronmccollum:
@leseb Will this eventually end up in `ansible-ceph-common` repo?

leseb:
Yes it will. 
I just need to push it :)

I'm currently traveling, will do this asap. 
I also noticed that my push script is broken. 

Thanks for your patience.

roles/ceph-common/defaults/main.yml
```diff
@@ -7,6 +7,20 @@
...
 fetch_directory: fetch/
 
+###############
+# PERMISSIONS #
+###############
+
+# Permissions for /etc/ceph configuration directory
+conf_directory_owner: root
+conf_directory_group: root
+conf_directory_mode: 644
```
prometheanfire: directories need to be executable

leseb: why?

prometheanfire: In order to be traversed / listed they need to be executable.

https://unix.stackexchange.com/a/21252

leseb: Correct but since it's owned by `root` we don't really care.
Things might change with infernalis though.

prometheanfire: As long as you know the path to the file you are going to you are fine, if you need to run search on the directory then it won't work so well.

PR #486, opened 2016-01-19
ceph-osd: Change ownership of osd directories.

guits:
When using the 4th scenario in ceph-osd role (directory instead of a disk for osd), the directories given in the `osd_directories` variable in `group_vars/osds` are created with `root:root` ownership.
This cause the command `ceph-disk activate` to fail in the task 'activate OSD(s)' in `roles/ceph-osd/tasks/scenarios/osd_directory.yml` because of
the arguments `--setuser ceph --setgroup ceph` that are passed to the command `/usr/bin/ceph-osd` launched by `ceph-disk activate`.

### DIFF
--- a/roles/ceph-osd/tasks/pre_requisite.yml
+++ b/roles/ceph-osd/tasks/pre_requisite.yml
@@ -54,3 +54,15 @@
   when:
     cephx and
     not is_ceph_infernalis
+
+- name: set ownership for osd directories
+  set_fact:
+    osd_dir_owner: "root"
+    osd_dir_group: "root"
+  when: not is_ceph_infernalis
+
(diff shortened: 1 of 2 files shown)


leseb:
You need to check wether we are running on infernalis or not. Please see https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mds/tasks/pre_requisite.yml as a reference. 

Use "when: is_ceph_infernalis" to perform your check

rootfs:
lgtm

leseb:
I like the logic, but if you use it you would have to refactor the entire OSD role so we keep things consistent :).

Do you mind @guits? :D

alfredodeza:
test this please

PR #488, opened 2016-01-22
Adding python-setuptools dependency for CentOS 7.

PiotrProkop:
Ceph-detect-init requires python-setuptools, when using upstream repositories.

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -47,7 +47,7 @@ dummy:
 #  - yum-plugin-priorities.noarch
 #  - epel-release
 #  - ntp
-
+#  - python-setuptools
 ## Configure package origin
 #
 #ceph_origin: 'upstream' # or 'distro'
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -48,6 +48,7 @@ dummy:
 #  - epel-release
 #  - ntp
 #  - python-setuptools
+
 ## Configure package origin
 #
 #ceph_origin: 'upstream' # or 'distro'


rootfs:
that's odd, I don't hit this issue. Which centos vagrant box you are using?

PiotrProkop:
I use bare-metal CentOS 7 installed with cobbler.
```
$ cat /etc/*release
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

```

rootfs:
Ok, thanks. I might running a few rev behind yours but it looks to me that's bug in ceph package: it should list python-setuptools as a dependency so yum install will pick that up.

PiotrProkop:
Anyhow I think  it can save someone time to add this dependency for now then to debug why ansible-playbook is crashing.

rootfs:
yes, i agree. this is absolutely helpful.

PR #489, opened 2016-01-22
updates related to using vagrant and virtualbox for testing

andrewschoen:
I was going through the readme and trying to get things going with vagrant and virtualbox and having some difficulties. Specifically, the vm.* settings in ``os_tuning_params`` were locking up my OSDs and ansible was losing the ssh connection to the nodes.  With these updates I'm able to get everything up and running. I was using vagrant==1.8.1 and virtualbox==5.0.4

I have not tested yet if anything here causes trouble for libvirt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,6 @@ Clone me:
 
 ```bash
 git clone https://github.com/ceph/ceph-ansible.git
-cp vagrant_variables.yml.sample vagrant_variables.yml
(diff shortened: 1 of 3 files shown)


leseb:
Just came across this issue yesterday, I think the problem is with the 4GB being set in `vm.min_free_kbytes`.
Thanks!

PR #492, opened 2016-01-25
WIP: initial steps to provision CentOS Atomic host

rootfs:
- Add centos/atomic-host
- Tagging yum tasks and skip when provisioning CentOS Atomic host.

### DIFF
--- a/vagrant_variables.yml.atomic
+++ b/vagrant_variables.yml.atomic
@@ -1,0 +1,34 @@
+---
+
+# DEFINE THE NUMBER OF VMS TO RUN
+mon_vms: 1
+osd_vms: 1
+mds_vms: 0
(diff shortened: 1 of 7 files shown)


vagrant_variables.yml.atomic
```diff
@@ -0,0 +1,34 @@
...
+# SUBNET TO USE FOR THE VMS
+subnet: 192.168.0
+
+# MEMORY
+memory: 1024
+
+disks: "[ '/dev/sdb', '/dev/sdc' ]"
+
+eth: 'enp0s3'
+# VAGRANT BOX
```
leseb: Since the file is atomic oriented you can remove the link of the other boxes I guess ;)

PR #493, opened 2016-01-25
CI does not use root, but sudo will do fine

alfredodeza:

### DIFF
--- a/test.yml
+++ b/test.yml
@@ -1,6 +1,7 @@
 ---
 - hosts: localhost
   remote_user: root
+  sudo: true
   roles:
     - ceph-common
     - ceph-mon
--- a/test.yml
+++ b/test.yml
@@ -1,6 +1,5 @@
 ---
 - hosts: localhost
-  remote_user: root
   sudo: true
   roles:
     - ceph-common
--- a/tests/inventories/single-machine.yml
+++ b/tests/inventories/single-machine.yml
@@ -1,0 +1,8 @@
+[mons]
+localhost
+
+[osds]
+localhost
+
+[rgws]
+localhost


alfredodeza:
@leseb so this means we are very close to proper CI testing :) would you mind providing some feedback into what is it that we are missing here?

https://jenkins.ceph.com/job/ceph-ansible-pull-requests/11/console

leseb:
@alfredodeza thanks a lot for this :)

I see a couple of issues:

* the playbook is not run using vagrant, shouldn't be a real issue but we need to know the amount of disks available. It looks like it's a single machine which is not ideal but better than the current state :)
* `monitor_interface` is not properly set

As for the error it seems to be related to your inventory file. Can you share it?
It should look like the following:

```
[mons]
localhost
[osds]
localhost
[rgws]
localhost
```

Thanks again!

alfredodeza:
We don't have an actual inventory file, we are passing it on the fly:

https://github.com/ceph/ceph-build/blob/master/ceph-ansible-pull-requests/build/build#L12

    $VENV/ansible-playbook -i "localhost," -c local test.yml

We aren't using vagrant because it would be a pain to make it work with these one-off machines we are spinning from OpenStack. Maybe we need a specific inventory file for these that can specify localhost for all? I will try with a new testing directory that has the right inventory file

alfredodeza:
hrmn, tried with an actual inventory file but got the same error:

    TASK: [ceph-common | generate ceph configuration file] ************************ 
    fatal: [localhost] => One or more undefined variables: 'dict object' has no attribute u'mons'

alfredodeza:
please test this

alfredodeza:
test this please

alfredodeza:
test this please

leseb:
can you try with `hosts: all` in `test.yml`, using `site.yml` should be fine as well.

alfredodeza:
test this please

leseb:
test this please

andrewschoen:
test this please

leseb:
@alfredodeza @andrewschoen we need to put raw devices so we can have OSDs. We need to add something like: `devices=[ '/dev/sdb', '/dev/sdc' ]`.
Looking at the Ansible command, it seems that `monitor_interface` is still not set properly too :(

alfredodeza:
@leseb yes, we need to add those, yesterday we were implementing provisioning of nodes with multiple (available) devices. Thanks for keeping an eye on it!

alfredodeza:
test this please

andrewschoen:
test this please

leseb:
Looks like your sudo user doesn't have enough permission to set sysctl commands

andrewschoen:
@leseb yeah, that's the confusing part as I thought we were giving that user sudo.

This is the ansible for doing that on a jenkins slave: https://github.com/ceph/ceph-build/blob/master/ansible/slave.yml.j2#L25

Also, I see this in /etc/sudoers ``jenkins-build   ALL=(ALL:ALL) NOPASSWD:ALL``

Are we missing something else here?

test.yml
```diff
@@ -1,6 +1,7 @@
 ---
 - hosts: localhost
   remote_user: root
```
leseb: `{{ jenkins_user }}`?

leseb:
@andrewschoen please look at my comment :)

test.yml
andrewschoen: @leseb thanks, I'll try that. I think I'd prefer to just remove it though so it uses the current user.

leseb:
test this please

leseb:
Still using `root`. :(

andrewschoen:
@leseb I thought I got around that issue with the user. I see the last build you triggered failed because ``fsid`` wasn't set, or at least that's what it looks like to me.

https://jenkins.ceph.com/job/ceph-ansible-pull-requests/32/console

```
TASK [ceph-common : generate cluster uuid] *************************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "ERROR! The conditional check 'fsid != '4a158d27-f750-41d5-9e7f-26ce4c9d2d45'' failed. The error was: ERROR! error while evaluating conditional (fsid != '4a158d27-f750-41d5-9e7f-26ce4c9d2d45'): ERROR! ERROR! 'cluster_uuid' is undefined\n\nThe error appears to have been in '/home/jenkins-build/build/workspace/ceph-ansible-pull-requests/ceph-ansible/roles/ceph-common/tasks/main.yml': line 66, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: generate cluster uuid\n  ^ here\n"}
```

leseb:
@andrewschoen argh this error started to show up on ansible v2 recently.
This error makes sense. The condition can't be evaluated since the variable doesn't exist yet...

Working on a fix.

In the meantime, you can force both `fsid` and `monitor_key` to overcome this problem.
Something like:

```
fsid: 4a158d27-f750-41d5-9e7f-26ce4c9d2d45
monitor_secret: AQAWqilTCDh7CBAAawXt6kyTgLFCxSvJhTEmuw==
```

Thanks!

andrewschoen:
test this please

leseb:
looks like one variable used in the template doesn't exist... investigating

andrewschoen:
test this please

andrewschoen:
@leseb that issue with the template was because of an incorrectly set ``monitor_interface``. That's fixed here https://github.com/ceph/ceph-build/commit/0238df1f27f00a6d5739667114c24ce259142f5a

I'm seeing a different failure on centos now. https://jenkins.ceph.com/job/ceph-ansible-pull-requests/39/console

leseb:
@andrewschoen ok great.
I saw this error but it does not make much sense to me at the moment. Since `ceph_stable_ice` is not enabled the task should be skipped...

andrewschoen:
test this please

leseb:
I think we should try to use https://docs.ansible.com/ansible/package_module.html at some point, however it's hard to tell/force users to use Ansible v2... It is to recent I believe.

andrewschoen:
@leseb yeah, the new package module looks very cool but I agree that it's a bit too soon to require Ansible v2.

Alright, we have successful ansible runs now on both centos7 and trusty. The last thing we're trying to resolve is a way to cleanup the jenkins slave used for these runs. We're doing that now but the node is being removed too soon and makes the job always report failure.

leseb:
I'm working on a fix for dnf and yum to solve the issue `importError: No module named yum` here 
https://jenkins.ceph.com/job/ceph-ansible-pull-requests/45/console

leseb:
test this please

leseb:
test this please

alfredodeza:
test this please

alfredodeza:
it looks like this is good to go now @leseb  :) I only need to fix the bit where we teardown the machine, but the testing is now working and reporting correctly! woot woot

thanks to @andrewschoen too!

leseb:
@alfredodeza @andrewschoen thank you both for your perseverance :).
I'm really glad to have this finally ready :D

Should I go ahead and merge this?

@alfredodeza another thing, will it be possible to trigger an action after a PR is merged?
The idea is to run:

`ansible-playbook -i dummy-ansible-hosts contrib/splitup.yml --tags update`

So we also update the separate ansible galaxy roles.

andrewschoen:
@leseb I'm +1 on merging this now.

We're testing either trusty or centos7, depending on which slave jenkins picks. I've been thinking we might want to make this a matrix job so that for each PR we test both of those distros. Thoughts?

Yeah, we should be able to create a job that updates the ansible galaxy roles on merge to master without too much trouble I'd think.

leseb:
@andrewschoen ok let me open a separate issue for this then :)

PR #495, opened 2016-01-27
WIP: use systemd to manage ceph daemons

rootfs:
mon is now started by systemd on Atomic.

### DIFF
--- a/roles/ceph-mon/meta/main.yml
+++ b/roles/ceph-mon/meta/main.yml
@@ -11,4 +11,4 @@ galaxy_info:
   categories:
     - system
 dependencies:
-  - { role: ceph-common, when: not mon_containerized_deployment }
+  - { role: ceph-common }
--- a/roles/ceph-mon/tasks/docker/main.yml
+++ b/roles/ceph-mon/tasks/docker/main.yml
@@ -5,6 +5,14 @@
   changed_when: false
   failed_when: false
 
+- name: check if it is Atomic host
+  stat: path=/run/ostree-booted
+  register: stat_ostree
+
+- name: set fact for using Atomic host
+  set_fact:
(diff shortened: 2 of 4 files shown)


roles/ceph-mon/tasks/docker/main.yml
```diff
@@ -5,6 +5,14 @@
   changed_when: false
   failed_when: false
 
+- name: check if it is Atomic host
+  stat: path=/run/ostree-booted
```
leseb: Please use this syntax:
```
stat:
  path: /run/ostree-booted
```

roles/ceph-mon/tasks/docker/start_docker_monitor.yml
```diff
@@ -10,6 +10,43 @@
 - name: pull ceph daemon image
   shell: "docker pull {{ ceph_mon_docker_username }}/{{ ceph_mon_docker_imagename }}"
 
+# Use systemd to manage container on Atomic host
+- name: generate systemd unit file
+  sudo: true
+  config_template:
```
leseb: I don't think `config_template` is needed here (unless you want users to be able to add more variables to the unit file).
Simply use the `template` module.

roles/ceph-mon/meta/main.yml
```diff
@@ -11,4 +11,4 @@ galaxy_info:
   categories:
     - system
 dependencies:
-  - { role: ceph-common, when: not mon_containerized_deployment }
```
leseb: maybe it's part of your previous patches but why do we need to include `ceph-common` role for docker?

rootfs:
we should label containers and use `atomic` command to run and install containers. Will open another PR for Atomic.

PR #501, opened 2016-01-28
Add defaults for ceph_stable_ice_tmp_path and ceph_stable_ice_kmod

andrewschoen:
It seems that in ansible 2.0 even if a task is skipped by it's `when`
clause not evaluating to true the variables in the play are still
rendered. Because these were not defined in defaults/main.yml ansible
was failing in installs/install_on_redhat where those variables are
being used in a `with_items` stanza.

An example of this failure can be seen here: https://jenkins.ceph.com/job/ceph-ansible-pull-requests/39/console


leseb:
Thanks @andrewschoen

PR #502, opened 2016-01-28
use dnf when yum is not available

leseb:
fixes the `importError: No module named yum` see here:
https://jenkins.ceph.com/job/ceph-ansible-pull-requests/45/console

### DIFF
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -4,6 +4,14 @@
     name: "{{ item }}"
     state: present
   with_items: redhat_package_dependencies
+  when: ansible_pkg_mgr == "yum"
+
+- name: install dependencies
+  dnf:
+    name: "{{ item }}"
+    state: present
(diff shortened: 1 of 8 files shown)


leseb:
weird this doesn't seem to solve the problem :/ still hitting `ImportError: No module named yum`

leseb:
@andrewschoen isn't dnf the package manager used on this VM?

rootfs:
@leseb or maybe it is an Atomic host?

leseb:
@rootfs I doubt that but I'll @andrewschoen confirm or not :)

andrewschoen:
@leseb I'm wondering if this PR just needs the code from https://github.com/ceph/ceph-ansible/pull/493 before it'll give a valid test.

This node used for this PR testing should be exactly the same as the one from this run, which passed with the code from #493 https://jenkins.ceph.com/job/ceph-ansible-pull-requests/46/consoleFull

They are both centos7 nodes created from the same image and configured by the same ansible playbook.

leseb:
@andrewschoen well let' merge this and rebase the other patch if needed.

Issue #506, opened 2016-01-29
Avoiding unnecessary package updates


mattt416:
Hi all,

What are everyone's thoughts on the change in https://github.com/ceph/ceph-ansible/commit/476c5df38feb3525b7c7c32e79b1498f5f2df7d2?  Is this likely to cause issues if you re-run playbooks resulting in package updates?  You won't get a major version update and this won't necessarily result in service restarts (unless there are config changes that happen also), but with something as critical as a storage system I do wonder if package updates need to be more controlled.

As this was done specifically for the rolling upgrade play, an immediate solution that jumps to mind is to require a variable passed in on the command line when running the rolling upgrade play which then decides if packages are installed to present or latest.  I've seen this pattern used elsewhere and it seems to work OK.

I'd love to hear what others think before I put a PR through unnecessarily.

Thanks!

--Matt

leseb:
Yes the commit partially adresses the concern where we don't really want to upgrade the storage system. I think we should find a way to control the behaviour, perhaps we could use a flag for rolling upgrades.
So just exactly as you suggested the variable thing should do the trick.

I'd be more than happy to see a PR for this.
Thanks for highlighting this @mattt416

leseb:
Addressed by #508

PR #508, opened 2016-01-29
Control when ceph packages get updated

mattt416:
Currently, all the ceph package installation resources use
"state=latest", which means subsequent runs of the ceph playbooks
could result in ceph being upgraded if there are package updates
available in the selected repo.

This commit adds a new variable to ceph-common called
'upgrade_ceph_packages' which defaults to False.  This variable is used
in the package installation resources for ceph packages to determine if
the resource should use "state=present" or "state=latest".  If the
variable gets set to True, "state=latest" will be used.

Additionally, we update rolling_update.yml to override
upgrade_ceph_packages to true to permit package upgrades in this
context specifically.

Closes issue #506

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -34,6 +34,11 @@ dummy:
 #mds_group_name: mdss
 #restapi_group_name: restapis
 
+# This variable determines if ceph packages can be updated.  If False, the
+# package resources will use "state=present".  If True, they will use
+# "state=latest".
+#upgrade_ceph_packages: False
+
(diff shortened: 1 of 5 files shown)


mattt416:
Leaving this in WIP (work in progress) until I've done some further testing.  I'd also love to hear from others if this is the right pattern to take or if we should be duplicating these tasks out in separate playbooks which get included depending on the value of `upgrade_ceph_packages`.

leseb:
@mattt416 I kinda like this approach. It's flexible and without too much changes.

leseb:
@alfredodeza thoughts?

alfredodeza:
This looks reasonable to me! 

+1

PR #511, opened 2016-02-02
ceph-common: use a generic way to create uuid

### LINKED ISSUE #510: UUID generation does not fail if uuid-runtime is not installed
Just ran into this today deploying from an LXC container which does not have uuid-runtime installed.

TASK: [ceph-common | generate cluster uuid] ***********************************
changed: [lsn-d5481_ceph-mon_container-bb84e883 -> 127.0.0.1]
changed: [lsn-d4057_ceph-mon_container-e078483d -> 127.0.0.1]
changed: [lsn-d3544_ceph-mon_container-900f1327 -> 127.0.0.1]

This writes an empty ceph_cluster_uuid file to the fetch directory and the play does not fail until later when it is trying to deploy the monitors and tries to pass an empty uuid.

using > redirection instead of piping to tee would cause the task to fail due to return code, but it would still cause an empty ceph_cluster_uuid file to be written...

leseb:
as reported in #510 some systems don't have uuidgen installed so we
better use a more global way to generate it. It sounds like python
should be available in case uuidgen is not.
Otherwise we will have to find another way :)

closes #510

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -64,7 +64,7 @@
   run_once: true
 
 - name: generate cluster uuid
-  local_action: shell uuidgen | tee {{ fetch_directory }}/ceph_cluster_uuid.conf
+  local_action: shell python -c 'import uuid; print str(uuid.uuid4())' | tee {{ fetch_directory }}/ceph_cluster_uuid.conf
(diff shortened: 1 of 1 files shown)


leseb:
@Logan2211 will this work in your container?

PR #514, opened 2016-02-03
Adds a package-install tag to all installation tasks

andrewschoen:
We have a requirement to install the packages first without
configuration. These tags should allow us to target the tasks need to do
that.


leseb:
@andrewschoen looks good thanks! just waiting for the CI :)

PR #530, opened 2016-02-10
config_template: Add interpreter line

zmc:
Fixes the error: "module is missing interpreter line"


leseb:
Not sure about this, it should work without that...

zmc:
Yeah, now I think this error message was only appearing because I was incorrectly using the ``--module-path`` flag.

PR #535, opened 2016-02-11
Allow RGW and client VMs together

dang:
RGW and Client VMs were sharing a network range, causing vagrant
failure.  Bump the RGW range to the next value.

Note that this needs to be applied to the docker branch as well.


leseb:
Nice catch :), thanks @dang

rootfs:
thanks @dang

PR #536, opened 2016-02-11
ceph-osd: fix the auto discovery scenario

leseb:
While this is not widly used (AFAIK :p) the feature was broken. Thanks
to @zmc for reporting it. You can now set `osd_auto_discovery` to
true in your group_vars/osd and it will go through all the devices
available and will make them OSDs.

### DIFF
--- a/roles/ceph-osd/tasks/check_devices.yml
+++ b/roles/ceph-osd/tasks/check_devices.yml
@@ -11,7 +11,7 @@
 # for SSD journals.
 
 - name: check if the device is a partition
-  shell: "echo '{{ item }}' | egrep '/dev/(sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'"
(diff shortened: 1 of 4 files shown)


roles/ceph-osd/tasks/check_devices.yml
```diff
@@ -16,6 +16,18 @@
   changed_when: false
   failed_when: false
   register: ispartition
+  when: not osd_auto_discovery
+
+- name: check if the device is a partition (autodiscover disks)
+  shell: "echo '/dev/{{ item.key }}' | egrep '/dev/(sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'"
```
zmc: Could the regexp be simplified slightly from:
``'/dev/(sd[a-z]{1,2}|hd[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'`` to:
``'/dev/({h,s,v}d[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'`` ?

The 'v' would cause it to also look for virtualized disks.

leseb: Definitely :)

leseb:
We are being bitten by this stupid behaviour: https://github.com/ansible/ansible/issues/4297
Working a fix...

PR #539, opened 2016-02-12
rhscon (storage agent)

alfredodeza:
This implements the `ceph-agent` role which allows remote nodes to register with a Calamari master server by configuring salt to talk to a master server.

It tries to follow the same conventions as the other roles, hopefully I didn't miss anything!

### DIFF
--- a/roles/ceph-agent/defaults/main.yml
+++ b/roles/ceph-agent/defaults/main.yml
@@ -5,6 +5,6 @@
 # GENERAL #
 ###########
 
-# The agent needs a master_host defined so that it can connect
+# The agent needs and agent_master_host variable defined so that it can connect
 # and push information back to it
-master_host: "localhost"
+agent_master_host: "localhost"
--- a/roles/ceph-agent/defaults/main.yml
+++ b/roles/ceph-agent/defaults/main.yml
@@ -1,0 +1,10 @@
+---
+# You can override vars by using host or group vars
+
+###########
+# GENERAL #
+###########
+
+# The agent needs a master_host defined so that it can connect
+# and push information back to it
+master_host: "localhost"
--- a/roles/ceph-agent/defaults/main.yml
+++ b/roles/ceph-agent/defaults/main.yml
@@ -5,6 +5,6 @@
 # GENERAL #
 ###########
 
-# The agent needs and agent_master_host variable defined so that it can connect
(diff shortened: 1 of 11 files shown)


roles/ceph-agent/tasks/pre_requisite.yml
```diff
@@ -0,0 +1,33 @@
+---
+- name: determine if node is registered with subscription-manager.
```
leseb: shouldn't we only execute this on RH?

roles/ceph-agent/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+# You can override vars by using host or group vars
+
+###########
+# GENERAL #
+###########
+
+# The agent needs and agent_master_host variable defined so that it can connect
```
leseb: missing word?

"The agent needs ??? and agent_master_host..."

leseb:
@alfredodeza thanks for this PR :)
Can you also:

* add a `group_vars/agent` file?
* edit `site.yml.sample` with an example

Thanks!

alfredodeza:
@leseb I think the latest changes address all your feedback. Thank you for pointing out the missing pieces!

leseb:
One last thing, can you update the `.gitignore` with `group_vars/agent`? Thanks!

site.yml.sample
```diff
@@ -5,6 +5,7 @@
   become: True
   roles:
   - ceph-mon
+  - ceph-agent
```
leseb: I'd rather have a dedicated host for that, wdyt?

GregMeno:
@alfredodeza I am really confused by this pull request. Would you please walk me through the different workflows you see this supporting?

alfredodeza:
@GregMeno this PR provides a 1:1 functionality for that of `ceph-deploy calamari connect` with the one requirement of defining where the `master` lives.  I see two distinct cases:

* user will deploy this on some node
* user will deploy this alongside a monitor

site.yml.sample
alfredodeza: Someone trying to get this installed will usually want to deploy it on a monitor node as it needs to talk to the rest of the cluster. It can certainly be deployed to other nodes though.

leseb: Ok seems legit.

PR #541, opened 2016-02-14
ceph-common: use explicit action syntax for config_template

haasn:
This fails for me with "ERROR! no action detected in task" otherwise.

```
ERROR! no action detected in task. This often indicates a misspelled module name, or incorrect module path.

The error appears to have been in '/etc/ansible/include/ceph-ansible/roles/ceph-common/tasks/main.yml': line 149, column 3, but may

- name: generate ceph configuration file
  ^ here
```

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -147,7 +147,8 @@
     mode: "{{ conf_directory_mode }}"
 
 - name: generate ceph configuration file
-  config_template:
+  action: config_template
+  args:
     src: ceph.conf.j2
     dest: /etc/ceph/ceph.conf
     owner: "{{ conf_file_owner }}"


leseb:
Never saw this before, weird but seems to be working on both `2.0.0.2` and `1.9.4` so I'm fine with it.
Thanks @haasn!

haasn:
It just might be that something is screwy in my environment because I'm not running these playbooks from ceph-ansible directly but instead symlinking stuff from my own site-wide ansible directory layout.

I just noticed that there's an extra library/ directory that I didn't symlink anywhere, maybe that has something to do with it?

leseb:
@haasn did you follow these instructions? https://github.com/ceph/ceph-ansible/tree/master/roles/ceph-common#requirements

haasn:
@leseb I did update my action_plugins path to point to the directory in which action_plugins.py ultimately resides.

Those instructions are not quite clear - it's ambiguous between moving the entire plugins/actions/config_template.py structure and just moving the file itself. But given the context, I assume the former is what was meant

leseb:
@haasn well if you already have a `plugins/actions` directory at the top level of your playbooks, just symlinking `config_template.py` should be sufficient.
Now I wonder if we shouldn't reverse the change or not...

haasn:
Let me clarify: I have the symlink in place and *still* experience this error. (I've had the symlink there all along)

It might simply be due to using a newer version of ansible (2.1.0).

leseb:
Oh I understand now, ok well as I said this doesn't break anything on the other ansible versions so I'm fine with this :)

PR #542, opened 2016-02-15
Rewrite purge-cluster playbook

KGoreczny:
Rewritten purge-cluster playbook. 
Tested with Ceph Hammer and Infernalis on CentOs 7 and Ubuntu 14.04 with Ansible 2.0.1.0

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -225,11 +225,11 @@
       ansible_distribution == 'Ubuntu'
 
   - name: remove Upstart nad SysV files
-    shell: 'find /etc -name "*ceph*" -delete'
+    shell: "find /etc -name '*ceph*' -delete"
     when:
       ansible_distribution == 'Ubuntu'
 
   - name: remove Upstart and apt logs and cache
-    shell: 'find /var -name "*ceph*" -delete'
+    shell: "find /var -name '*ceph*' -delete"
     when:
       ansible_distribution == 'Ubuntu'
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -2,42 +2,193 @@
 # This playbook purges Ceph
 # It removes: packages, configuration files and ALL THE DATA
 
-- hosts:
-  - mons
-  - osds
-
-  tasks:
-
-  - name: purge ceph
-    command: ceph-deploy purge {{ ansible_fqdn }}
-    delegate_to: 127.0.0.1
-
-  - name: remove osd data
-    shell: rm -rf /var/lib/ceph/osd/*/*
-    failed_when: false
-
-  - name: purge remaining data
-    command: ceph-deploy purgedata {{ ansible_fqdn }}
-    delegate_to: 127.0.0.1
-
-- hosts:
-  - osds
+- name: stop ceph cluster
+  hosts:
+    - mons
+    - osds
+    - mdss
+  become: yes
 
   vars:
-    devices: [ '/dev/sdb', '/dev/sdc', '/dev/sdd', '/dev/sde', '/dev/sdf' ]
-    partitions: [ '1', '2', '3' ]
+# When set to true both groups of packages are purged.
+# This can couse problem with qemu-kvm
+    purge_all_packages: true
+
+    ceph_packages:
+      - ceph
+      - ceph-common
+      - ceph-fs-common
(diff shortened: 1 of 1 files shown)


purge-cluster.yml
```diff
@@ -2,42 +2,193 @@
...
+- name: stop ceph cluster
+  hosts:
+    - mons
+    - osds
+    - mdss
+  become: yes
+
+  vars:
+# When set to true both groups of packages are purged.
+# This can couse problem with qemu-kvm
```
leseb: s/couse/cause

leseb:
@KGoreczny thanks for this :) just a few comments :)

purge-cluster.yml
```diff
@@ -2,42 +2,193 @@
...
+  - name: get osd data mount points
+    shell: ls /var/lib/ceph/osd
+    register: mounted_osd
+    changed_when: false
+    failed_when: mounted_osd.rc != 0 and mounted_osd.rc != 2
+    when:
+     osd_group_name in group_names
+
+  - name: umount osd data partition
+    shell: umount /var/lib/ceph/osd/{{ item }}
```
leseb: mount module instead?

leseb:
I'm a little bit hesitant since this only supports ansible v2. (because of the package module).

purge-cluster.yml
KGoreczny: Problem with mount module is that it operates on /etc/fstab, not current system state. So for me it is not useful in this case as all osds are mounted dynamically, without any change is fstab.

leseb: Correct we don't need to touch `/etc/fstab`.

purge-cluster.yml
```diff
@@ -2,42 +2,215 @@
...
+      ansible_distribution != 'Ubuntu'
+
+  - name: purge ceph packages with apt
+    apt:
+      name: "{{ item }}"
+      state: absent
+    with_items:
+      - "{{ ceph_packages }}"
+    when:
+      ansible_distribution == 'Ubuntu'
```
leseb: `when: ansible_pkg_mgr == "apt"`?

purge-cluster.yml
```diff
@@ -2,42 +2,215 @@
...
+     osd_group_name in group_names
+
+  - name: purge ceph packages with yum
+    yum:
+      name: "{{ item }}"
+      state: absent
+    with_items:
+      - "{{ ceph_packages }}"
+    when:
+      ansible_distribution != 'Ubuntu'
```
leseb: `when: ansible_pkg_mgr == "yum"`?

purge-cluster.yml
```diff
@@ -2,42 +2,215 @@
...
+      ansible_distribution == 'Ubuntu'
+
+  - name: purge remaining ceph packages with yum
+    yum:
+      name: "{{ item }}"
+      state: absent
+    with_items:
+      - "{{ ceph_remaining_packages }}"
+    when:
+      ansible_distribution != 'Ubuntu' and
```
leseb: when: ansible_pkg_mgr == "yum"?

purge-cluster.yml
```diff
@@ -2,42 +2,215 @@
...
+      purge_all_packages == true
+
+  - name: purge remaining ceph packages with apt
+    apt:
+      name: "{{ item }}"
+      state: absent
+    with_items:
+      - "{{ ceph_remaining_packages }}"
+    when:
+      ansible_distribution == 'Ubuntu' and
```
leseb: when: ansible_pkg_mgr == "apt"?

purge-cluster.yml
```diff
@@ -2,42 +2,234 @@
...
+     path: /var/log/ceph
+     state: absent
+
+  - name: remove form SysV
+    shell: "update-rc.d -f ceph remove"
+    when:
+      ansible_distribution == 'Ubuntu'
+
+  - name: remove Upstart nad SysV files
+    shell: 'find /etc -name "*ceph*" -delete'
```
leseb: Why do you sometimes use simple or double quotes for the shell module?
Double quotes all the time are probably the way to go.

leseb:
Thanks @KGoreczny for addressing the comments.

PR #543, opened 2016-02-15
ceph-common: calculate vm.min_free_kbytes

leseb:
based on the os memory we calculate the value for vm.min_free_kbytes
we don't go over 4GB of memory, even for systems with more than 48GB of RAM.

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -283,4 +283,4 @@ dummy:
 #  - { name: fs.file-max, value: 26234859 }
 #  - { name: vm.zone_reclaim_mode, value: 0 }
 #  - { name: vm.vfs_cache_pressure, value: 50 }
-#  - { name: vm.min_free_kbytes, value: 4194303 }
+#  - { name: vm.min_free_kbytes, value: "{{ vm_min_free_kbytes }}" }
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -295,7 +295,7 @@ os_tuning_params:
   - { name: fs.file-max, value: 26234859 }
   - { name: vm.zone_reclaim_mode, value: 0 }
   - { name: vm.vfs_cache_pressure, value: 50 }
-  - { name: vm.min_free_kbytes, value: 4194303 }
+  - { name: vm.min_free_kbytes, value: "{{ vm_min_free_kbytes }}" }
 
 
 ##########
--- a/roles/ceph-common/tasks/misc/system_tuning.yml
+++ b/roles/ceph-common/tasks/misc/system_tuning.yml
@@ -16,9 +16,13 @@
   failed_when: false
   when: disable_swap
 
+- name: get default vm.min_free_kbytes
+  command: sysctl -b vm.min_free_kbytes
+  register: default_vm_min_free_kbytes
+
 - name: calculate vm.min_free_kbytes
   set_fact:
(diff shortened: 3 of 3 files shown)


leseb:
@mcsage @mattt416 @git-harry @byronmccollum does this work for you guys?

byronmccollum:
@leseb I will test this out today...thanks.

leseb:
test this please

mattt416:
I don't do much kernel tweaking, so googled vm.min_free_kbytes and found [1] which indicates that setting this to 5-10% of total system memory is too high.  Are we sure this needs to be so high?  Is this the recommended percentage when running ceph?

[1] https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Performance_Tuning_Guide/s-memory-tunables.html

mattt416:
Anyway, this actual change worked for me for me on my AIO (ansible_memtotal_mb == 7980, vm.min_free_kbytes got set to 81715, which is just slightly higher than the default of 67584).  I guess my concern here is that this number jumps drastically if you go below 48 GB to over.

leseb:
@mattt416 perhaps we can forget the else statement then?

leseb:
@mattt416 this is probably clearer to preserve kernel's default value, added a new commit for that

leseb:
Correct.

leseb:
test this please

mattt416:
Hi @leseb and @byronmccollum ,

This PR merged before I could offer a bit more input.  I don't think we would hit any issues here since our production OSD hosts use > 48 GB of RAM (I _think_), and @byronmccollum's issue should now be resolved, however I think the following points do warrant consideration:

1. There is no way as a user to override `vm_min_free_kbytes`, which means you either use what is enforced by the role or override the entire `os_tuning_params` list
2. The ratio of `vm_min_free_kbytes` to total RAM changes drastically when ansible_memtotal_mb >= 49152 and ansible_memtotal_mb < 49152
3. The following [1] for example suggests a blanket 256M value, would we be better off applying a single value regardless of ansible_memtotal_mb and allow power users to override if necessary?

I'm certainly no authority here and cannot suggest what value is best, however I do think we need to make it easy to override and have a more consistent experience irrespective of how much RAM your host has.

Thoughts?  :)

[1] http://lists.ceph.com/pipermail/ceph-users-ceph.com/2015-July/002770.html

leseb:
Hi @mattt416 

Well if you put a value and remove `vm_min_free_kbytes` here you should be good, right?
https://github.com/ceph/ceph-ansible/blob/master/group_vars/all.sample#L286

byronmccollum:
I think it will need a `when: vm_min_free_kbytes is not defined` here...

https://github.com/ceph/ceph-ansible/blob/f86da22ce62f38e9bbfbd3a5c72b6c2b738b6c40/roles/ceph-common/tasks/misc/system_tuning.yml#L23-L25

leseb:
@byronmccollum not sure, by definition `vm_min_free_kbytes` is never defined. Am I missing something?

mattt416:
@leseb, if you add the when to [1] then it will only set the fact if the user hasn't overridden the value of `vm_min_free_kbytes`.  That means you can override `vm_min_free_kbytes` without having to override the entire `os_tuning_params` list.  The downside to doing the latter is that if other defaults in that list change and you override it then you'll never pick up those changes.

[1] https://github.com/ceph/ceph-ansible/blob/f86da22ce62f38e9bbfbd3a5c72b6c2b738b6c40/roles/ceph-common/tasks/misc/system_tuning.yml#L23-L25

Issue #545, opened 2016-02-16
No package matching 'calamari-server' is available.


armaan:
TASK [ceph-agent : install dependencies] ***************************************
fatal: [192.168.122.115]: FAILED! => {"changed": false, "failed": true, "msg": "No package matching 'calamari-server' is available"}
fatal: [192.168.122.116]: FAILED! => {"changed": false, "failed": true, "msg": "No package matching 'calamari-server' is available"}
fatal: [192.168.122.114]: FAILED! => {"changed": false, "failed": true, "msg": "No package matching 'calamari-server' is available"}

leseb:
@armaan a bit more information about your setup will be useful. On which distro are you running on?
@alfredodeza :)

armaan:
@leseb: A 3 node cluster running Ubuntu 14.04.3 LTS. I moved the HEAD to 20e6f932dd50a87bc2db868098d7a61263572562 and successfully installed Ceph.

leseb:
Closing this because of #546 , feel free to re-open.

PR #546, opened 2016-02-16
separate the agent so it doesn't break standard installs

alfredodeza:
Mainly because some packages are still not there for upstream which will cause
it to break installs that would otherwise would succeed

Should fix issue #545

### DIFF
--- a/site.yml.sample
+++ b/site.yml.sample
@@ -7,6 +7,11 @@
   - ceph-mon
   - ceph-agent
 
+- hosts: agents
+  become: True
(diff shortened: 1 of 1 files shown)


leseb:
@alfredodeza should fix #545 but #545 could also change its `site.yml` file :)

alfredodeza:
@leseb true, but since by default we are failing, I think we could behave a bit better in the meantime

leseb:
@alfredodeza correct :) not against the patch ;)

PR #548, opened 2016-02-16
adds some more --package-install tags; make check_mandatory_vars more specific

andrewschoen:
We have a use case where we won't be configuring mons and osds in the same call, so we want to allow the playbook to run and only check for variables when absolutely necessary. For example, there is no need to check ``journal_size`` when only configuring a mon.

### DIFF
--- a/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
+++ b/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
@@ -5,6 +5,8 @@
   when:
     ceph_origin != 'upstream' and
     ceph_origin != 'distro'
+  tags:
+    - package-install
 
 - name: make sure an installation source was chosen
   fail:
@@ -15,6 +17,8 @@
     not ceph_dev and
     not ceph_stable_ice and
     not ceph_stable_rh_storage
+  tags:
+    - package-install
 
 - name: verify that a method was chosen for red hat storage
   fail:
@@ -23,6 +27,8 @@
     ceph_stable_rh_storage and
     not ceph_stable_rh_storage_cdn_install and
     not ceph_stable_rh_storage_iso_install
+  tags:
+    - package-install
 
 - name: make sure journal_size configured
   fail:
--- a/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
+++ b/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
@@ -33,22 +33,30 @@
 - name: make sure journal_size configured
   fail:
     msg: "journal_size must be configured. See http://ceph.com/docs/master/rados/configuration/osd-config-ref/"
-  when: journal_size|int == 0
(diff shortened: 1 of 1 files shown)


roles/ceph-common/tasks/checks/check_mandatory_vars.yml
```diff
@@ -23,26 +27,36 @@
...
+  tags:
+    - package-install
 
 - name: make sure journal_size configured
   fail:
     msg: "journal_size must be configured. See http://ceph.com/docs/master/rados/configuration/osd-config-ref/"
-  when: journal_size|int == 0
+  when:
+    journal_size|int == 0 and
+    osd_group_name in group_names
```
leseb: Unfortunately I'm afraid this will break the `ceph.conf` build. Assuming we leave it like and that we first deploy monitors and then later osd, the `ceph.conf` will change which will result in monitors being restarted.

Obviously, this won't affect your deployment when you only deploy monitors. However how realistic is it? :)
Same issue for `cluster_network` and `public_network`.

andrewschoen: @leseb The requirements I have are: 1) install packages on mon nodes, 2) install packages on osd nodes, 3) configure mon nodes and then finally 4) configure osds. 

Are you saying that the mon still needs the configuration for OSDs? What's the best approach here if these all need to happen in separate steps?

leseb: Monitors do not need to have the osd conf to work but it is better to have the osd settings from the get go.
The only thing that all the daemons share in common is the `ceph.conf`, so if you discover some setting after the monitors have been bootstrapped, the file will change and then handlers will get triggered which will issue an unnecessary restart of the monitors. Perhaps if you use `--limit` while running the playbook you can avoid that but you will get inconsistent `ceph.conf` between monitors and osds.

How difficult is it to add these options even on the mons bootstrap?

andrewschoen: It's not difficult to add those options, it'd just mean requiring post data be sent to an API endpoint that doesn't have anything to do with what the endpoint is actually doing. It feels strange to require ``journal_size`` on an api call that's suppose to only be configuring mons. For reference: https://github.com/ceph/ceph-installer#ceph-installer

When you talk about the handlers triggering restarts, are you referring to the ansible handlers? That shouldn't be an issue for us here because we only operate against one node at a time at these endpoints.

Are there any other issues with having an inconsistent ``ceph.conf`` between monitors and osds?

leseb: Ok I see.
Yes I'm referring to the ansible handler.

No there is no real issue with having an inconsistent `ceph.conf` expect the inconsistency ;).
Let me run this one time locally but this should be fine in the end.

PR #550, opened 2016-02-18
Adds a new role, ceph-fetch-keys

andrewschoen:
This is a simple role that connects to monitors and downloads all the ceph keys into the local ``fetch_directory``. We use this so that we can add a new OSD to an existing cluster without the need to run ceph-common or ceph-mon against all of our monitors again. The existing monitors must exist in the ``mons`` group in your inventory so that the ceph.conf is created correctly for the new OSD.

Also adds a simple playbook for adding an OSD to an existing cluster.

### DIFF
--- a/roles/ceph-fetch-keys/tasks/main.yml
+++ b/roles/ceph-fetch-keys/tasks/main.yml
@@ -1,0 +1,26 @@
+---
+- name: find ceph keys
+  shell: ls -1 /etc/ceph/*.keyring
+  changed_when: false
+  register: ceph_keys
+
+- name: set keys permissions
+  file:
+    path: "{{ item }}"
+    mode: 0600
+    owner: root
+    group: root
+  with_items:
+    - "{{ ceph_keys.stdout_lines }}"
+
+- name: copy keys to the ansible server
+  fetch:
+    src: "{{ item }}"
+    dest: "{{ fetch_directory }}/{{ fsid }}/{{ item }}"
+    flat: yes
+  run_once: true
+  with_items:
+    - "{{ ceph_keys.stdout_lines }}"
+    - /var/lib/ceph/bootstrap-osd/ceph.keyring
+    - /var/lib/ceph/bootstrap-rgw/ceph.keyring
+    - /var/lib/ceph/bootstrap-mds/ceph.keyring
--- a/osd-configure.yml
+++ b/osd-configure.yml
@@ -1,0 +1,18 @@
+---
+# This playbook is used to add a new OSD to
+# an existing cluster without the need for running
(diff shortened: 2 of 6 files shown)


roles/ceph-fetch-keys/tasks/main.yml
```diff
@@ -0,0 +1,26 @@
...
+- name: find ceph keys
+  shell: ls -1 /etc/ceph/*.keyring
+  changed_when: false
+  register: ceph_keys
+
+- name: set keys permissions
+  file:
+    path: "{{ item }}"
+    mode: 0600
+    owner: root
```
leseb: shouldn't we set `ceph` user depending on the ceph version?

leseb:
The rest looks good :)

roles/ceph-fetch-keys/tasks/main.yml
andrewschoen: @leseb yeah, maybe so. Do we need to fix that here as well? https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/ceph_keys.yml#L28

Running a ``ceph osd tree`` on the new OSD does give me an error about how it can't find the keyrings. Maybe this is indicating a permissions problem? They look ok to me though.

```
vagrant@vagrant:~$ sudo ceph osd tree
2016-02-18 16:37:02.236868 7f27b5194700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin: (2) No such file or directory
2016-02-18 16:37:02.237400 7f27b5194700 -1 monclient(hunting): ERROR: missing keyring, cannot use cephx for authentication
2016-02-18 16:37:02.237960 7f27b5194700  0 librados: client.admin initialization error (2) No such file or directory
Error connecting to cluster: ObjectNotFound
vagrant@vagrant:~$ sudo ls -l /var/lib/ceph/bootstrap-osd
total 4
-rw------- 1 ceph ceph 71 Feb 17 23:18 ceph.keyring
```

leseb: Getting this error is normal while running your command like (without other argument), it implies you use `--name client.admin` and then it looks for `/etc/ceph/client.admin.keyring`.

OSD nodes shouldn't have the admin key so you're all good :)

I think it's fine keeping root as user for the keys in `/etc/ceph` as it only contains the admin one. So nop we are good :)

leseb:
gonna create the ansible galaxy setup now

andrewschoen:
@leseb thanks!

PR #553, opened 2016-02-19
Use object map for xattrs with ext4

stpierre:
These three commits add a number of useful tunables for the ceph.conf template.

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -186,6 +186,13 @@ dummy:
 # Object backend
 #osd_objectstore: filestore
 
+# xattrs. by default, 'filestore xattr use omap' is set to 'true' if
+# 'osd_mkfs_type' is set to 'ext4'; otherwise it isn't set. This can
+# be set to 'true' or 'false' to explicitly override those
+# defaults. Leave it 'null' to use the default for your chosen mkfs
+# type.
+# filestore_xattr_use_omap: null
+
 # Performance tuning
 #filestore_merge_threshold: 40
 #filestore_split_multiple: 8
--- a/roles/ceph-common/templates/ceph.conf.j2
+++ b/roles/ceph-common/templates/ceph.conf.j2
@@ -114,6 +114,12 @@ filestore split multiple = {{ filestore_split_multiple }}
 osd op threads = {{ osd_op_threads }}
 filestore op threads = {{ filestore_op_threads }}
 filestore max sync interval = {{ filestore_max_sync_interval }}
+{% if filestore_xattr_use_omap != None %}
+filestore xattr use omap = {{ filestore_xattr_use_omap }}
+{% elif osd_mkfs_type == "ext4" %}
+filestore xattr use omap = true
+{# else, default is false #}
+{% endif %}
 osd max scrubs = {{ osd_max_scrubs }}
 {% if ceph_stable_release not in ['argonaut','bobtail','cuttlefish','dumpling','emperor','firefly','giant'] %}
 osd scrub begin hour = {{ osd_scrub_begin_hour }}
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -198,6 +198,13 @@ osd_crush_update_on_start: "true"
 # Object backend
 osd_objectstore: filestore
 
+# xattrs. by default, 'filestore xattr use omap' is set to 'true' if
+# 'osd_mkfs_type' is set to 'ext4'; otherwise it isn't set. This can
+# be set to 'true' or 'false' to explicitly override those
+# defaults. Leave it 'null' to use the default for your chosen mkfs
+# type.
+filestore_xattr_use_omap: null
+
 # Performance tuning
 filestore_merge_threshold: 40
 filestore_split_multiple: 8


leseb:
Hey @stpierre, thanks for your PR.  I'm not sure if I should continue to accept config option additions since we have the ability to add new options without impacting the template file.
If you look at https://github.com/ceph/ceph-ansible/blob/master/group_vars/all.sample#L257-L272 you will notice that you can add your own options.

However, I'd like to share interesting config options to tweak as well so people are aware of them. So perhaps we should start modifying this instead https://github.com/ceph/ceph-ansible/blob/master/group_vars/all.sample#L257-L272

@alfredodeza thoughts?

alfredodeza:
@leseb I think the title for that section is a misnomer: I read "overrides" as something that exists already but I will be overwriting it with a different value.

The issue I see is that it is not clear where should a new configuration item is, we should perhaps document it clearly to determine what goes where? E.g. Is all *new* configuration supposed to go into the overrides section? Or just certain ones (which ones?) ?

stpierre:
Cool, I didn't know about that. I think the OSD scrub stuff can clearly go in there, so I can drop that commit from my PR.

The other two I think are valuable enough to include, though:

You really should be using the object map with ext4, so this just makes the module smarter; if you select ext4, you automatically get the right xattr storage.

And ``log_to_syslog`` sets a bunch of options; it's more of a matter of convenience, but setting six options in one fell swoop has its own distinct value.

leseb:
@alfredodeza I agree that the name is a bit misleading, override usually implies overriding an existing value, in our situation this could actually be both. With this module we have the ability to override existing configuration options or add new once that are not part of the template. Basically before the module the only options to get more config flag in the template file was to add one variable along with a change in the template file.
Since we now have the "override" module we do not need to do this anymore, however I still believe that some config options are worth being highlighted. If it was just me we would put everything here: https://github.com/ceph/ceph-ansible/blob/master/group_vars/all.sample#L263-L278 and do a huge cleanup in the template file https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/templates/ceph.conf.j2 However this will likely change people current `ceph.conf` and will trigger an unnecessary restart of the cluster... :/
That's where my main concern is at the moment.

Thoughts?

@stpierre would you mind updating your PR and remove the log options, I think it's worth keeping  `filestore_xattr_use_omap` in the template file since there is a condition going on.

stpierre:
Done.

leseb:
Test this please

stpierre:
As on the other PR, got a pointer to any docs for how to do that? Thanks.

andrewschoen:
@stpierre the "test this please" phrase is a trigger to let our CI system know to run a test on this PR. :)

stpierre:
@andrewschoen Everything is illuminated. Thank you.

I'm not used to people being polite to their CI. :)

leseb:
test this please

leseb:
@stpierre could you please reflect this change in `group_vars/all`  so we can go ahead and merge this?

stpierre:
Done.

leseb:
@stpierre can you just make sure that both content are identical please?

stpierre:
Done.

If they're supposed to be identical, `all.sample` could be trivially generated with one line of `sed`. (There are currently a bunch of differences between them, so that approach would improve the current state of affairs, too.)

leseb:
@stpierre you're right, would you mind re-generating all of them in another PR?
Thanks!

group_vars/all.sample
```diff
@@ -186,6 +186,11 @@ dummy:
 # Object backend
 #osd_objectstore: filestore
 
+# xattrs. by default, 'filestore xattr use omap' is set to 'true' if
+# ext4 if 'osd_mkfs_type' is set to 'ext4'; otherwise it isn't set. This
+# can be set to 'true' or 'false' to explicitly override those defaults.
+# filestore_xattr_use_omap: null
```
leseb: read the first sentence again, I think a word is missing.
moreover shouldn't we set "false" instead of null?

roles/ceph-common/templates/ceph.conf.j2
```diff
@@ -114,6 +114,12 @@ filestore split multiple = {{ filestore_split_multiple }}
 osd op threads = {{ osd_op_threads }}
 filestore op threads = {{ filestore_op_threads }}
 filestore max sync interval = {{ filestore_max_sync_interval }}
+{% if filestore_xattr_use_omap %}
+filestore xattr use omap = {{ filestore_xattr_use_omap }}
+{% elif osd_mkfs_type == "ext4" %}
```
mattt416: What if osd_mkfs_type == "ext4" but I do not want to set filestore xattr use omap?

group_vars/all.sample
stpierre: Docs fixed. I think I got lost halfway through that sentence.

`false`/`null` doesn't really matter; it just needs to be a non-true value. But I think `null` is more intuitive, since the difference between `false` (the boolean) and `"false"` (the string) is super critical here and not super obvious. But if you've got religion about this being `false` I'll change it.

roles/ceph-common/templates/ceph.conf.j2
stpierre: That's why this starts with:

    {% if filestore_xattr_use_omap %}

Just set that variable to override the default.

Can you suggest an improvement to the comments in `main.yml` that would make this clearer? Thanks.

mattt416: Hi @stpierre, I see your logic but according to my local testing (ansible 1.9.4) if I set `filestore_xattr_use_omap: false` then that condition is skipped, which means `filestore xattr use omap` may get set if osd_mkfs_type == "ext4".

stpierre: `false` isn't the same as `"false"`. :)

Notice that the other boolean settings in the variables use string values as well.

mattt416: Yeah, I see now.  :)  I personally find this a bit misleading (if something tells me to use true/false, I'll use the YAML boolean not a string), but you are right that this convention seems to be used elsewhere in defaults files.

stpierre: I've updated it to explicitly check for `None`, so that you can use a boolean value and it'll do its best to work. (Assuming ceph itself parses "False" as "false".)

roles/ceph-common/templates/ceph.conf.j2
```diff
@@ -114,6 +114,12 @@ filestore split multiple = {{ filestore_split_multiple }}
 osd op threads = {{ osd_op_threads }}
 filestore op threads = {{ filestore_op_threads }}
 filestore max sync interval = {{ filestore_max_sync_interval }}
+{% if filestore_xattr_use_omap is not None %}
```
mattt416: Does this need to be `{% if filestore_xattr_use_omap == none %}`?  I'm wondering if this is what caused the run failure.

leseb:
test this please

stpierre:
Yep, looks like `is not None` was an issue. Fixed.

PR #558, opened 2016-02-23
Improve firewall checks

stpierre:
The firewall checks can fail for any number of reasons -- e.g., the
ceph cluster hostnames are unresolvable from the ansible host, or the
ports are filtered by some intermediate hop, etc. Make two changes to
make those checks better:

* Set pipefail when running the checks, so if nmap itself fails the
  command will be marked as 'failed'. Specifically, this fixes the
  case where the hostnames cannot be resolved.
* Add a new variable, check_firewall, which can be used to disable
  checks entirely. Specifically, this fixes the case where some
  intermediate firewall filters the ports, so nmap returns "filtered".

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -31,6 +31,12 @@ rgw_group_name: rgws
 mds_group_name: mdss
 restapi_group_name: restapis
 
+# If check_firewall is true, then ansible will try to determine if the
+# Ceph ports are blocked by a firewall. If the machine running ansible
+# cannot reach the Ceph ports for some other reason, you may need or
+# want to set this to False to skip those checks.
+check_firewall: True
+
 # This variable determines if ceph packages can be updated.  If False, the
 # package resources will use "state=present".  If True, they will use
 # "state=latest".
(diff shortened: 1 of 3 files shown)


roles/ceph-common/tasks/checks/check_firewall.yml
```diff
@@ -4,39 +4,47 @@
...
   changed_when: false
   failed_when: false
   with_items: groups.{{ mon_group_name }}
   register: monportstate
   when:
+    check_firewall and
     mon_group_name in group_names and
     nmapexist.rc == 0
+  tags:
+    - check-firewall
```
leseb: Why is this the only tag?

roles/ceph-common/defaults/main.yml
```diff
@@ -31,11 +31,17 @@ rgw_group_name: rgws
 mds_group_name: mdss
 restapi_group_name: restapis
```
leseb: Can you please update group_vars/all.sample to reflect this change?

roles/ceph-common/tasks/checks/check_firewall.yml
stpierre: Leftover cruft from manual testing. I've removed it.

roles/ceph-common/defaults/main.yml
stpierre: Done.

leseb:
Thanks!

PR #559, opened 2016-02-23
Make RHEL installs honor ceph_origin: distro

stpierre:
Installs on RHEL with ceph_origin set to distro previously would fail
because no packages would get installed, but all of the checks passed
fine. This adds support for ceph_origin: distro, simply installing the
packages using yum/dnf and assuming that the sysadmin has provided a
repository containing them.

This also supports the use case where Satellite or a similar local
mirror is in use, and the admin does not or cannot use the additional
repositories the role would otherwise add.

### DIFF
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -23,7 +23,7 @@
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
   when: not ceph_stable_rh_storage
 
-- name: install red hat storage ceph mon
+- name: install distro or red hat storage ceph mon
   yum:
     name: "{{ item }}"
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
@@ -31,11 +31,11 @@
     - ceph
     - ceph-mon
   when:
-    ceph_stable_rh_storage and
+    (ceph_origin == "distro" or ceph_stable_rh_storage) and
     mon_group_name in group_names and
     ansible_pkg_mgr == "yum"
 
-- name: install red hat storage ceph mon
+- name: install distro or red hat storage ceph mon
   dnf:
     name: "{{ item }}"
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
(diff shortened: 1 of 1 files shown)


leseb:
Why don't you simply set upstream for ceph_distro?
I know it has been a bit of a debate recently. 

From my perspective, everything that is not part of the default distro repository is considered as being upstream. As soon as you need to activate an external repo or add an extra repo source this is upstream. Perhaps the term upstream is confusing and we should change it. 
If you have any idea.

stpierre:
As far as I can tell, every option for RHEL ends up adding some extra repository, which is explicitly what I need to avoid. If you set ``ceph_origin: "upstream"``, then you must also set ``ceph_stable``, ``ceph_dev``, ``ceph_stable_ice``, or ``ceph_stable_rh_storage``, but I don't want to (and cannot) get Ceph from any of those sources because each one of them relies on an external repository. It is possible that "distro" is a little misleading here, but not terribly; it really does just rely on your distro to provide ceph, without any external repo.

leseb:
Ok I get your point now, sorry I've read too fast. This is all valid to me. 
One thing that bothers me is that I don't think packages named ceph-mon and ceph-osd exist. I think it's RHCS specific so in this case the install might fail. 
Not sure of the return code if yum/dnf if the package is not found.

ktdreyer:
Packages named "ceph-osd" and "ceph-mon" will exist in Jewel: https://github.com/ceph/ceph/pull/7746

stpierre:
The task errors out if a package cannot be found.

It's also possible that the sysadmin has built hir own RPMs locally, or has mirrored RH Storage with Satellite.

I'll be happy to update the wiki to add more detail about the various installation source options as part of this PR.

leseb:
@ktdreyer thanks for the update on that.
@stpierre yes a bit of documentation won't hurt, so please go ahead :).

stpierre:
Wiki updated.

leseb:
I can't see you change

stpierre:
https://github.com/ceph/ceph-ansible/wiki/Home/_compare/57b24e212fcd1e3af3c5ca19a8861635a240c2b0...0d3bd0f93755492caaffd54aadeb89c269b4a9dc

leseb:
Oh right, thanks for taking the time for this.

Issue #556, opened 2016-02-22
osd_auto_discovery not working on (our) kvm instances


zmc:
I'm working on getting ceph-ansible testable via [teuthology](https://github.com/ceph/teuthology). The ``osd_auto_discovery`` feature isn't working on our KVM-based virtual machines. It *might* be working in our OpenStack test lab. I'm attempting to figure out what is causing the difference in behavior but I figured I should file an issue while I do that, in case there is a fix obvious to someone more experienced with ceph-ansible.

The error I see is:
``error while evaluating conditional: item.0.rc != 0 and item.1.rc != 0 and journal_collocation and not osd_auto_discovery``

Here is an example of the same job [executed on OpenStack](http://teuthology.ovh.sepia.ceph.com/teuthology/zack-2016-02-22_16:20:20-ceph-ansible-master---basic-openstack/3360/teuthology.log) vs. [executed on KVM](http://qa-proxy.ceph.com/teuthology/zack-2016-02-22_08:26:37-ceph-ansible-master---basic-vps/20458/teuthology.log).

That's a ton of output so I'll attempt to isolate some relevant things. First, I'm using a [separate branch of ceph-ansible with extra debugging](https://github.com/ceph/ceph-ansible/compare/wip-teuthology).

Here is just the [debug from OpenStack](https://gist.github.com/zmc/1ef9daa99c0c768b1da8/raw/4bf7a4107a8f27849523cc71880d010f290df7fb/ovh-debug.txt) vs. the [debug from KVM](https://gist.github.com/zmc/1ef9daa99c0c768b1da8/raw/4bf7a4107a8f27849523cc71880d010f290df7fb/vps-debug.txt)

``combined_parted_results.results`` and ``combined_ispartition_results.results`` are very different between the two instances. What I haven't unraveled yet is *why* - aside from the obvious difference in device counts.

leseb:
@zmc at a first glance, if the output is so different is mainly because the KVM env has more disks than the OpenStack. On OpenStack, you seem to only have a single drive `/dev/vda` where on KVM you have `/dev/vda`, `/dev/vdb`, `/dev/vdc`, `/dev/vdd`.

I just figured out that this https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/check_devices.yml#L21-L30 is useless while using `osd_auto_discovery`, the device list will always be a raw device. So this shouldn't run.
[...]

PR #563, opened 2016-02-24
Do not prepare skipped disks

stpierre:
When autodiscovering disks, disks can be skipped if either they are
removable, or if they have partitions on them. Skipped actions have no
'rc' attribute, though, so the 'ceph prepare' conditional fails unless
we first check to ensure that the results were not skipped before
checking the return value.

### DIFF
--- a/roles/ceph-osd/tasks/scenarios/journal_collocation.yml
+++ b/roles/ceph-osd/tasks/scenarios/journal_collocation.yml
@@ -29,8 +29,10 @@
     - combined_ispartition_results.results
     - devices
   when:
-    item.0.rc != 0 and
-    item.1.rc != 0 and
+    not item.0.get("skipped") and
+    not item.1.get("skipped") and
+    item.0.get("rc", 0) != 0 and
+    item.1.get("rc", 0) != 0 and
     journal_collocation and not
     osd_auto_discovery
 
--- a/roles/ceph-osd/tasks/activate_osds.yml
+++ b/roles/ceph-osd/tasks/activate_osds.yml
@@ -1,6 +1,18 @@
 ---
 # NOTE (leseb) : this task is for disk devices only because of the explicit use of the first
 # partition.
+
+- name: automatically activate osd disk(s) without partitions
+  command: ceph-disk activate "/dev/{{ item.key }}"
+  ignore_errors: true
+  with_dict: ansible_devices
+  when:
+    ansible_devices is defined and
+    item.value.removable == "0" and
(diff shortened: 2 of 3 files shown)


roles/ceph-osd/tasks/scenarios/journal_collocation.yml
```diff
@@ -29,9 +29,11 @@
     - combined_ispartition_results.results
     - devices
   when:
+    not item.0.skipped and
+    not item.1.skipped and
     item.0.rc != 0 and
     item.1.rc != 0 and
     journal_collocation and not
-    osd_auto_discovery
+    osd_auto_discovery and
```
leseb: This will likely fail, can you remove the "and". Thanks!

stpierre: Done, sorry. I rearranged this during testing and didn't do a very good job of un-rearranging it.

leseb:
You also need update https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/scenarios/raw_multi_journal.yml

stpierre:
Done.

leseb:
Test this please

stpierre:
This is broken and needs more work.

stpierre:
Updated. The dicts describing a skipped action and command output are neatly disjoint, which makes this check uglier.

leseb:
Test this please

leseb:
I believe this fixes #556, @zmc can you give it a try?

zmc:
@leseb Great! I'll test it today.

zmc:
I am seeing: http://fpaste.org/328763/35093514/

stpierre:
I'm working on a patch for that, too -- I think 'ceph activate' uses the wrong set of devices when `osd_auto_discovery` is enabled.

stpierre:
Actually there are two issues: one is that the 'ceph activate' stuff also has a bogus conditional, and that's fixed in this PR. The other issue, which you may run into next, is the one I describe; I'll put in another PR for that.

stpierre:
The other changes depended on these changes, so I've just appended another commit to this PR. Should now fix all of the issues mentioned in this thread, and #556.

zmc:
Great! Retesting right now.

leseb:
@zmc do you have a status on this? :)

zmc:
@leseb It appears to fix the issue. I haven't managed to fully verify it yet, but I can open a separate issue if there are further problems.

PR #564, opened 2016-02-24
Make /etc/ceph.conf 755 instead of 644

zmc:
Otherwise, the 644 permissions on ceph.conf don't actually allow
non-root to read the file.

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -14,7 +14,7 @@ fetch_directory: fetch/
 # Permissions for /etc/ceph configuration directory
 conf_directory_owner: root
 conf_directory_group: root
-conf_directory_mode: 644
+conf_directory_mode: 755
 
 # Permissions for /etc/ceph/ceph.conf configuration file
 conf_file_owner: root


stpierre:
655 is still bizarre permissions for a directory. Shouldn't we just make it 755 rather than depending on root's ability to execute directories without +x?

zmc:
You're absolutely right @stpierre - will repush

leseb:
Alright, I think we can do a bit of cleanup here :).
I'd like to use the following logic: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/main.yml#L59

So perhaps we can remove these: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/defaults/main.yml#L14-L22

And make the use of those: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/main.yml#L86-L96

I know that the name of the variable is not that clear, but the permission matches so...
What do you guys think?

zmc:
@leseb wow, I hadn't noticed all that logic there. I don't quite understand it at first glance. It does seem like all those set_fact calls could be reduced to, like two.

leseb:
@zmc well if you think we can factories this again then I'm good with that too :)

zmc:
@leseb is this blocking on the refactor you mentioned then?

leseb:
Yes would you mind giving it a try?
Thanks!

zmc:
Oh, I didn't realize you wanted me to do it. I can try - though I'm not exactly sure when I'll be able to get to it.

leseb:
No worries, I'll take care of that next week then.

leseb:
Closing this in favour of #575

PR #568, opened 2016-02-25
Skip keyring tasks when cephx is disabled

stpierre:
Skip a number of ceph keyring-related tasks (or remove the keyring
portion of some tasks) when cephx is disabled. Specifically, avoid
generating the initial keyring, which only clutters up the ansible
repo if cephx is not in use.

PR #569, opened 2016-02-25
Add option to skip generating fsid

stpierre:
If using another method to generate a consistent fsid, then we can
skip creation of an (unused) cluster UUID file. If cephx is disabled
as well, we can skip creation of the fetch directory entirely.

Skipping fetch directory creation when cephx is disabled requires #568 to be merged as well.

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -144,7 +144,15 @@ dummy:
 
 ## Ceph options
 #
+# Each cluster requires a unique, consistent filesystem ID. By
+# default, the playbook generates one for you and stores it in a file
+# in `fetch_directory`. If you want to customize how the fsid is
+# generated, you may find it useful to disable fsid generation to
+# avoid cluttering up your ansible repo. If you set `generate_fsid` to
+# false, you *must* generate `fsid` in another way.
 #fsid: "{{ cluster_uuid.stdout }}"
+#generate_fsid: true
+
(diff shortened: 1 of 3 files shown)


leseb:
Thanks. 
Can you please add a check to make sure that if generate_fsid is true and fsid empty then we fail early? See the check.yml file in ceph-common for that. I know you've put a note (and thanks for the clarify of it) but I don't trust users :).

stpierre:
A check sounds like a good idea, but I'm not sure what it would actually check. If we just check for:

    not generate_fsid and not fsid

...that will only catch the condition where a user sets something like:

    fsid: null
    generate_fsid: false

That seems unlikely. What we would want to check for is someone who has set `generate_fsid: false` without changing fsid, but a) I'm not sure how to check for that; and b) that will fail quickly anyway, because `cluster_uuid` will not get set.

As much as I hate to, I think this is a place where we kind of have to trust users, since we can't realistically check for the most common ways to screw this up.

If you've got suggestions, I'll be happy to implement them -- you clearly know loads more about ansible than I do, so there's quite possibly something I'm missing.

leseb:
Yeah I've forgot about cluster_uuid, this is tricky to check as the var is not registered yet. I'm away from my laptop for the weekend I'll think about that early next week.

leseb:
test this please

PR #570, opened 2016-02-26
Use saner defaults for config dir/files mode

gpocentek:
The mode argument requires octal (prefixed with 0) values to properly
work.


leseb:
Thanks for your PR. Unfortunately if you look at #564 you will see that we are in the process of removing these options.

gpocentek:
I should have checked the other PRs first! I'm happy to see this fixed in some way in any case.

PR #571, opened 2016-02-26
Fix openstack config skipping

stpierre:
Previously, creating pools was skipped if cephx was disabled; instead,
we should only skip key creation if cephx is disabled, and create
pools any time openstack_config is true.


leseb:
Nice catch thanks!

PR #572, opened 2016-02-28
help RGW install from RHCS .iso

bengland2:
While doing RGW install using ceph-ansible with RHCS .iso image, I saw that ceph-radosgw RPM install step was failing, it is packaged in this Tools subdirectory of the .iso image.   This change adds the Tools subdirectory into /etc/yum.repos.d/ceph.repo.


leseb:
Thanks @bengland2

PR #573, opened 2016-02-28
extend to handle rgws, read in ceph-common defaults

bengland2:
purge-cluster.yml blows up for me on expression "osd_group_name in group_names".  This seems to be because it isn't reading roles/ceph-common/defaults/main.yml which defines osd_group_name.  By explicitly calling out ceph-common role I was able to work around this, but how does site.yml work then?

Also, it ignores rgws role so I tried to add that in, it gets farther but does not yet remove RPMs from rgws hosts, probably doesn't shut down services.

If you need one of these two fixes but not the other, let me know and I'll push separate PR.

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -7,7 +7,12 @@
     - mons
     - osds
     - mdss
+    - rgws
+
   become: yes
+
+  roles:
+    - ceph-common
 
   vars:
 # When set to true both groups of packages are purged.
@@ -21,6 +26,7 @@
       - ceph-fuse
       - ceph-mds
       - ceph-release
+      - ceph-radosgw
 
     ceph_remaining_packages:
       - libcephfs1


leseb:
@bengland2 can't we just add these variables https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/defaults/main.yml#L28-L32 instead of parsing the entire ceph-common role?

bengland2:
I tried your suggestion at one point to copy *group_name into vars section of purge-cluster.yml, but discovered there were more undefined variables.   Copying parts of roles/ceph-common/defaults/main.yml is error-prone, because my purge-cluster.yml tests at least don't cover all cases (such as Ubuntu, upstream, etc).  I agree it is inefficient to bring in the entire ceph-common role that installs packages, etc.  IMHO it's important for purge-cluster.yml to reliably work, because it makes ceph-ansible behavior much more deterministic (not affected by outcome of previous installations).

leseb:
@bengland2 right, I understand the concern.
@KGoreczny would you mind chiming in as you were the author of this recent change? :) Thanks! How did you get this working?

PR #580, opened 2016-02-29
Generate all.sample automagically

stpierre:
This adds a script, generate_group_vars_sample.sh, that generates
group_vars/all.sample from roles/ceph-common/defaults/main.yml to
avoid discrepancies between the two files. It also converts the line
endings in main.yml from DOS to Unix, since generating all.sample was
spreading the line ending plague around to more files.


leseb:
Would you mind doing this for all the roles? Perhaps using `$1` as role input.

stpierre:
I've just added a loop to properly generate all roles every time.

group_vars/all.sample
```diff
@@ -1,18 +1,33 @@
 ---
 # Variables here are applicable to all host groups NOT roles
```
leseb: why did you merge all the files?
IMHO `all.sample` should container `ceph-common`.

leseb:
@stpierre thanks for this! did you generated all the `group_vars` files?

stpierre:
Okay, I think this works like it's supposed to now. Sorry for being dense. :)

leseb:
Because of #553 merged there is a conflict, can you rebase your work please?

stpierre:
Rebased.

leseb:
@stpierre not sure to understand what happened to the `defaults/main.yml` files? Why did they change?

leseb:
Ok got it.

PR #583, opened 2016-02-29
Fix pre-infernalis RBD client directory mode

stpierre:
0644 should never be a directory mode. 1777 makes it so that any user
can create a ceph client, not just root. (This is helpful if, for
instance, nova-compute is running as non-root.)

I'm not entirely sure this is the right fix, but it works for me, and 0644 is plainly wrong.

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -98,7 +98,7 @@
 - set_fact:
     rbd_client_dir_owner: root
     rbd_client_dir_group: root
-    rbd_client_dir_mode: "0644"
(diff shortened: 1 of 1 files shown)


leseb:
test this please

leseb:
shouldn't we do the same `when: is_ceph_infernalis`?

stpierre:
I wasn't sure. When `is_ceph_infernalis` is true, we set it to 0660 ceph:ceph, so any user in the ceph group can create a client socket. That's more secure and more granular, but requires extra setup -- the admin must add other users to that group. Since I don't have an infernalis cluster to test on, I didn't want to change that without knowing for sure how it was intended to work.

PR #594, opened 2016-03-02
rollback previous change for ceph-common change

leseb:
changing the name of the directory causes issues with git subtree which
will create new commits. Creating a symlink for vagrant to be happy.


leseb:
@mattt416 that's the fix, phew... what a relief :D

mattt416:
Thanks @leseb!

PR #603, opened 2016-03-05
fix purge-request.yml so non-OSD hosts will not fail

bengland2:
The problem with only doing "get osd numbers" task for OSD hosts is that subsequent tasks iterating on the "osd_ids" register variable will fail for non-OSD hosts, such as task "stop ceph-osd with systemd".  You would think that the when clause would be sufficient to prevent this problem, but it does not.  By defining "osd_ids" for all hosts (empty for non-OSD hosts), we avoid this problem.

The "ls" command will fail for non-OSD hosts (or for OSD hosts that have already been purged), but it's ok, return status from entire command is zero.

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -43,7 +43,7 @@
 
   tasks:
   - name: get osd numbers
-    shell: ls /var/lib/ceph/osd | cut -d "-" -f 2
+    shell: "if [ -d /var/lib/ceph/osd ] ; then ls /var/lib/ceph/osd | cut -d '-' -f 2 ; fi"
     register: osd_ids
     changed_when: false
 
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -46,8 +46,6 @@
     shell: ls /var/lib/ceph/osd | cut -d "-" -f 2
     register: osd_ids
     changed_when: false
-    when:
-     osd_group_name in group_names
 
 # Infernalis
   - name: stop ceph.target with systemd


bengland2:
oops should have named this pull request "fix purge-cluster.yml"

leseb:
I'm fine with this change, although I don't really understand the issue here.
What's the failure exactly? I only see one block using `osd_ids` and it is the service stop for OSD.

So which other task will fail?

bengland2:
Correct me if I'm wrong, but when a task fails (and no ignore_errors), then subsequent tasks directed to that set of hosts are skipped, right?  This includes purging ceph RPMs and /var/lib/ceph, which is really important for enabling a clean re-install.

leseb:
Yes but why would this task fail if it is been run only on OSDs?

bengland2:
Your question makes perfect sense, but what happens is that on non-OSD hosts, the register variable "osd_ids" is undefined so ansible itself throws an error.  It shouldn't, because it should evaluate the "when" clause before evaluating the items for the task, but instead an error happens and consequently it starts skipping tasks on non-OSD hosts.  This change just prevents that error from happening.  This isn't the first time I've seen a bug in ansible that I've implemented  a workaround for.  I should stop what I'm doing and try to fix ansible, but I need to do something resembling my day job ;-)

leseb:
What I don't understand is that osd_ids is only used by another task, the task can not possibly fail since it only runs on an OSD node. Moreover because of the `| cut` the return code will always be 0 so it should not cause the task to fail.
So I'm understand how your commit is solving your problem :(

bengland2:
Yes, the | cut makes the return code always be zero.  So the "get osd numbers" task never fails.  But before I made this change, this task WAS NEVER EXECUTED for a non-OSD host, because of the when clause!  So "osd_ids" was not defined when subsequent task was run, and so task failed (although ansible shouldn't have failed it). 

Note: you cannot reproduce this problem with a single VM running all roles, which I just verified.  You have to have different roles on different hosts for this problem to show up.   A cluster with a mon VM and an OSD VM should see it, for example.  Does ceph-ansible have a CI setup with 1 VM per role?  Does it test purge-cluster.yml?

bengland2:
The error I get without this change is from the non-OSD hosts here:

TASK: [stop ceph-osd with systemd] ******************************************** 
fatal: [cephmon2-10ge] => with_items expects a list or a set
fatal: [gprfc073-10ge] => with_items expects a list or a set
fatal: [gprfc074-10ge] => with_items expects a list or a set
fatal: [gprfc075-10ge] => with_items expects a list or a set
fatal: [gprfc076-10ge] => with_items expects a list or a set

leseb:
Can you describe the role of each host?

Can you try to do:

```
when: 
  osd_group_name is defined and
  osd_group_name in group_names
```

And tell me if by any chance it fixes your issue.

bengland2:
I will try that, sorry I didn't know about "is defined", that would be better way to solve.

----- Original Message -----

bengland2:
I tried doing something similar to what you suggested, but it didn't seem to work.  Instead of "osd_group_name is defined" (it always is), I tried "osd_ids is defined" in the shutdown-osds task.  Again, I think it SHOULD work, but it appears that ansible is not evaluating the when clause before the rest of the task.  

Instead, I changed "get osd numbers" step so that it would not do "ls" on non-existent directory, even though this appears to be harmless.  So fact "osd_ids" is always defined, even on non-OSD hosts, but it would be empty in non-OSD hosts so nothing happens in task that iterates on it.     Make sense?

leseb:
@bengland2 ok thanks for the perseverance on this, we haven't quite clarified why this is happening but your last proposition looks good.

PR #610, opened 2016-03-08
Install ceph-mds on Debian platforms only if mds_group_name is set

### LINKED ISSUE #547: do not install ceph-mds if not needed
from https://github.com/ceph/ansible-ceph-common/issues/5

fcharlier:
This should allow to not install ceph-mds on Debian platforms when it is not required.

Fixes #547


leseb:
can you do the same for RH please? in `roles/ceph-common/tasks/installs/install_on_redhat.yml`

leseb:
@fcharlier naz?!

fcharlier:
@leseb there's no mds related package for RHEL platforms.

leseb:
@fcharlier my bad! thanks!

PR #611, opened 2016-03-08
support for the jewel version of red hat ceph storage on ubuntu

andrewschoen:
The packages aren't released yet, but this enables us to install them on Ubuntu when they are released.

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -35,7 +35,16 @@
     - package-install
 
 - include: ./installs/install_on_debian.yml
-  when: ansible_os_family == 'Debian'
+  when:
+    ansible_os_family == 'Debian' and
+    not ceph_stable_rh_storage
+  tags:
+    - package-install
+
+- include: ./installs/install_rh_storage_on_debian.yml
+  when:
+    ansible_os_family == 'Debian' and
+    ceph_stable_rh_storage
   tags:
     - package-install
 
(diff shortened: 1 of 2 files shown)


ktdreyer:
:+1:

roles/ceph-common/tasks/main.yml
```diff
@@ -67,6 +78,12 @@
     is_ceph_infernalis={{ (ceph_stable and ceph_stable_release not in ceph_stable_releases) or (ceph_dev) or (ceph_stable_rh_storage and (rh_storage_version.stdout | version_compare('0.94', '>'))) }}
 
 - set_fact:
+    is_ceph_infernalis=True
```
leseb: why another fact?
the previous one should handle that

andrewschoen: @leseb I was trying to avoid running a shell command to get the rhcs version on debian systems. In this case we can just assume that anything debian and rh storage is gonna be jewel or later.

I can make a change to get the rhcs_storage_version on debian systems if you'd like.

ktdreyer: I vote that we keep this PR as-is. We are introducing ceph-ansible support for RHCS with RHCS 2, which is Jewel-based. Simply setting this fact to "True" here keeps things simple.

Honestly I'd rather we ripped out hammer support (including the `is_ceph_infernalis` variable) throughout the tree entirely, to drastically simplify the codebase. But maybe that will need to happen in ceph-ansible 2.x.

leseb: Ok it makes sense. However we can't drop the multi Ceph version support all of the sudden. 
So yes I think the next big change will happen when I'll move the repo to ansible 2 so ceph-ansible 2.0 probably.

PR #614, opened 2016-03-09
Allow to install the calamari-server on monitors

alfredodeza:
This enables monitors to have the "rest api" provided by `calamari-server`. It lives in the `ceph-mon` role because it requires to live where the monitors are.

Reference BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1305259


leseb:
Thanks!

PR #618, opened 2016-03-10
adds the ability to set osd_crush_chooseleaf_type

andrewschoen:

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -220,6 +220,9 @@ dummy:
 # CRUSH
 #pool_default_crush_rule: 0
 #osd_crush_update_on_start: "true"
+# the default of 1 maps to the 'host' type
+# reference: http://docs.ceph.com/docs/master/rados/operations/crush-map/#crush-map-bucket-types
+#osd_crush_chooseleaf_type: 1
 
 # Object backend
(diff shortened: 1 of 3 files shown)


leseb:
We don't need extra parameters anymore. We can use the override variable from group_vars/all and set an extra option. 
It's a bit tricky as some options are interesting to share with users. So not sure how to deal with those PRs....

Any idea?

andrewschoen:
@leseb In my case I wasn't aware that the override variable existed. Maybe we can just promote it's usage more with some documentation? I'm unsure if ceph-ansible is where people should look to learn how to tweak ceph though, thoughts?

I do think that the override variable makes more sense than providing a 1 to 1 mapping of ansible vars in ceph-ansible to config options in ceph. I can see how that could get difficult to maintain over time.

leseb:
Yes, highlighting this in the documentation is probably the best we can do. We should also say that we do not accept anymore pull requests on the template file unless this is helping the deployment, #553 was a good exception for example.
Would you mind doing a bit of docs? :)

Thanks @andrewschoen

andrewschoen:
@leseb yeah, I can write those docs. I'll close this and make a new PR for that.

leseb:
Great thanks!

PR #619, opened 2016-03-11
ceph-osd: Set selinux to permissive

dpkshetty:
Currently we don't yet support runnings OSDs w/ selinux in
enforcing mode. Thus its better to ensure that ceph-ansible
explicitly makes selinux permissive. This should help in
scenarios such as hyperconverged where OSDs are colocated
with VMs on compute nodes which needs selinux enforcing, but
OSDs don't.

Signed-off-by: Deepak C Shetty <deepakcs@redhat.com>

### DIFF
--- a/roles/ceph-osd/tasks/activate_osds.yml
+++ b/roles/ceph-osd/tasks/activate_osds.yml
@@ -41,6 +41,16 @@
 - include: osd_fragment.yml
   when: crush_location
 
+- name: set selinux to permissive and make it persistent
+  selinux:
+    policy: targeted
+    state: permissive
+  when:
+    ansible_selinux != false and
+    ansible_selinux['status'] == 'enabled' and
+    ansible_selinux['config_mode'] != 'disabled' and
+    not is_ceph_infernalis
+
 - name: start and add that the osd service(s) to the init sequence (before infernalis)
   service:
     name: ceph


detiber:
@dpkshetty instead of disabling SELinux, why not run the OSDs using an unconfined or permissive domain?

https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Security-Enhanced_Linux/sect-Security-Enhanced_Linux-Targeted_Policy-Unconfined_Processes.html

http://danwalsh.livejournal.com/42394.html

leseb:
@dpkshetty hum I don't recall having any issues running OSDs with SELinux enabled. Where did you see it wasn't supported?

dpkshetty:
@leseb The BZ [1] was opened for this purpose and the patch [2] fixes it in
triple heat templates (as part  of RDO-Manager / OSP director)
Per the latest releasenotes [3] (grep for selinux), looks like the selinux
policy was added post Hammer release, so we can disable it for Hammer only ?

My inspiration for the patch was from the fact that today higher level
orchestrators like RDO Manager disable selinux for OSD nodes, which should
ideally be contained within ceph-ansible.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1241422
[2]: https://review.openstack.org/#/c/201259/
[3]: http://docs.ceph.com/docs/master/release-notes/

dpkshetty:
@detiber Yes i meant permissive, changed the issue title.

roles/ceph-osd/tasks/activate_osds.yml
```diff
@@ -41,6 +41,23 @@
 - include: osd_fragment.yml
   when: crush_location
 
+- name: check if selinux is enabled
+  command: getenforce
+  register: sestatus
+  changed_when: false
+
+- name: set selinux to permissive
```
leseb: could you please use the selinux module instead? http://docs.ansible.com/ansible/selinux_module.html

leseb:
@dpkshetty yes we should disable it only when release is not infernalis. So you can do something like:

```
when:
  not is_ceph_infernalis
````

File: roles/ceph-osd/tasks/activate_osds.yml [hunk 2]
```diff
@@ -41,6 +41,18 @@
 - include: osd_fragment.yml
   when: crush_location
 
+- name: check if selinux is enabled
+  command: getenforce
+  register: sestatus
+  changed_when: false
```
leseb: add `failed_when: false` and/or a `when: ansible_os_family == 'RedHat'` so if the command does not exist the play doesn't fail.

detiber: You can also use:
```
when: ansible_selinux && ansible_selinux_status == 'enabled'
```

detiber: There should also be an ansible_selinux_config_mode fact available to avoid the need for calling out to getenforce.

leseb: Right so something like:

```
when:
  ansible_selinux != false and
  ansible_selinux['status'] == 'enabled' and
  ansible_selinux['config_mode'] != 'disabled'
```

?

roles/ceph-osd/tasks/activate_osds.yml
```diff
@@ -41,6 +41,18 @@
...
+
+- name: set selinux to permissive and make it persistent
+  selinux: policy=targeted state=permissive
```
leseb: please use the the following syntax style:

```
selinux:
  policy: targeted
  state: permissive
```

roles/ceph-osd/tasks/activate_osds.yml
detiber: Yes, that should do the trick.

dpkshetty:
@leseb @detiber Thanks for your inputs, posted next patchset

leseb:
Thanks for addressing the comments!

PR #630, opened 2016-03-16
Deployment of Ceph Monitors with kv_backend 

PiotrProkop:


roles/ceph-common/templates/ceph.conf.j2
```diff
@@ -15,7 +15,7 @@ auth service required = none
 auth client required = none
 auth supported = none
 {% endif %}
-fsid = {{ fsid }}
+{% if not mon_containerized_deployment_with_kv%}fsid = {{ fsid }}{% endif %}
```
leseb: There is no need to do anything in the ceph-common role since we run containers. Common is not a dependency of monitor role when we run in containers.

PiotrProkop: I am generating a ceph.,conf which i put into etcd.

leseb: what about this then? https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/meta/main.yml#L14

PiotrProkop: I do not use this role just its templatate , already ceph-mon dockerized deployment uses it with or without kv backend https://github.com/ceph/ceph-ansible/blob/docker/roles/ceph-mon/tasks/docker/create_configs.yml#L12

leseb: I've kinda missed the patch that I introduce this change. The ceph-docker project has its own mechanism to build the ceph.conf file. If you use populate_kv and generate the ceph.conf with ansible, populate becomes useless. So we should be able to allow both scenarios, either ceph-ansible generating the ceph.conf either let the container generates its own configuration file.

roles/ceph-mon/templates/ceph-mon.service.j2
```diff
@@ -3,12 +3,17 @@ Description=Ceph Monitor
 After=docker.service
 
 [Service]
-EnvironmentFile=/etc/environment
+EnvironmentFile=-/etc/environment
```
leseb: what does `-` do?

PiotrProkop: If /etc/environment doesn't exist then this unit file won't fail.

leseb: Ok thanks

roles/ceph-mon/defaults/main.yml
```diff
@@ -10,6 +10,7 @@ fetch_directory: fetch/
 mon_group_name: mons
 
 # ACTIVATE BOTH FSID AND MONITOR_SECRET VARIABLES FOR NON-VAGRANT DEPLOYMENT
+# Comment out fsid when using mon_containerized_deployment_with_kv
```
leseb: This one is tricky, how do you make sure the fsid in the ceph.conf generated will match with the one generated by ceph-docker during the init monitor bootstrap?
We can add an option in ceph-docker to specify a FSID and then re-use the same here.

roles/ceph-common/templates/ceph.conf.j2
PiotrProkop: Now i just mount ceph.conf as /etc/ceph/ceph.defaults) into populate docker and let him  push this into etcd. Add a flag for a user to decide if use default ceph-docker/ceph.conf or to push custom one ?

roles/ceph-mon/defaults/main.yml
PiotrProkop: That is why i add {% if not mon_containerized_deployment_with_kv%}fsid = {{ fsid }}{% endif %} in ceph.conf so etcd populate will be the only one which  generate fsid.

roles/ceph-common/templates/ceph.conf.j2
leseb: yes something like this please. let's keep the default with the ceph-docker ceph.conf

PiotrProkop: Added `mon_containerized_default_ceph_conf_with_kv` variable.

leseb: Cool thanks, just one thing, there is a missing space.

PiotrProkop: Fixed

PR #640, opened 2016-03-22
Ensure purge-cluster doesn't call install tasks

dpkshetty:
Currently calling purge-cluster calls ceph-common role
which ends up installing common ceph deps. This isn't
correct as we are in the process of purging the cluster.

Add a new purge_cluster var and ensure all the tasks tagged
with package-install are skipped if purge_cluster is true.
Default value of purge_cluster is false.

A better way to fix this would be to use --skip-tags, but
unfortunately there is no way to embed skip-tags inside a
playbook, thus falling back to this method.

Signed-off-by: Deepak C Shetty <deepakcs@redhat.com>

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -12,7 +12,7 @@
   become: yes
 
   roles:
-    - ceph-common
+    - { role: ceph-common, purge_cluster: true }
 
   vars:
 # When set to true both groups of packages are purged.
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -7,6 +7,12 @@
 
 fetch_directory: fetch/
 
+#############
+# UNINSTALL #
+#############
+
+purge_cluster: false
+
 ###########
 # INSTALL #
 ###########
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -9,7 +9,9 @@
   when: osd_group_name in group_names
 
 - include: ./pre_requisites/prerequisite_ice.yml
-  when: ceph_stable_ice
+  when:
+    ceph_stable_ice and
+    not purge_cluster
   tags:
     - package-install
 
@@ -17,7 +19,8 @@
   when:
     ceph_stable_rh_storage and
     ceph_stable_rh_storage_iso_install and
-    ansible_os_family == "RedHat"
+    ansible_os_family == "RedHat" and
+    not purge_cluster
   tags:
     - package-install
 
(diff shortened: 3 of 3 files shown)


leseb:
Hey @dpkshetty thanks for your PR, however @bengland2 has a slightly better approach in #644 which does not touch the ceph-common role at all. So I think we should stick with ben's proposal here.

dpkshetty:
@leseb @bengland2 PR looks good. I specifically liked the fact that he is fixing the /var/lib/ceph device / resource busy issue there (I haven't tested it, but the reboot in his patch suggests me that), which I was hitting consistently and was hoping to root cause it :)

Havign said that, his patch, instead of calling ceph-common role, puts the vars in purge yml file and removes the need for infernalis check by using systemd check (which i think is a bit hacky). This also means if someone changes {osd|mon|rgw}_group_name in ceph-common's default.yml, he/she needs to change it in purge-cluster too (multipel copies of same var issue)... but calling ceph-common directly with purge_cluster: true avoid the double edit and/or double mgmt of vars. My 2 cents.

dpkshetty:
@leseb also using different logic for infernalis check in ceph-common role and a altoghether diff logic in purge-cluster (as suggested by ben's patch) looks inconsistent, purge-cluster using already existing vars is better IMHO

leseb:
Well the thing is I'd like to have the purge playbook running in standalone so without dependency to any role so we can run it on any installation even clusters that weren't deployed with ceph-ansible.

bengland2:
Deepak, 

- It turns out that I can get rid of the 4 *_group_name variables by just doing 

    include: ./roles/ceph-common/defaults/main.yml

Where they are defined, I'll try that tomorrow, but it does compile if inserted in the vars: section of purge-cluster.yml with proper indentation.  Thanks for complaining about that.

- As for using different logic, I disagree.  is_ceph_infernalis fact computation requires you to install ceph to even get to the point where you can test the RPM version.  And it doesn't take into account what version of Ceph is installed at the present time, which might not be the one you expect.  

My method assumes nothing about what is or isn't already there.  If ceph is utilizing native systemd unit files, it acts accordingly.  The only thing I am unsure of is that Debian and other Linux distros use the same directory, /usr/lib/systemd/system to contain unit files.  Perhaps there is a more portable way to test for presence of native ceph unit files, such as "systemctl | grep ceph"?  

The point of this change is to eliminate having to install ceph in order to de-install it. This speeds up the de-install process as well and reduces the number of things that can go wrong.  de-installation is important to get repeatable behavior for cases where you need to retry the installation with different configurations or versions for example.

-ben

----- Original Message -----

bengland2:
I wrote this example to test my hypothesis about including roles/ceph-common/defaults/main.yml in purge-cluster.yml, and it fails

---
- name: stop ceph cluster
  hosts:
    - mons

  vars:
   include: ./roles/ceph-common/defaults/main.yml

  tasks:
  - name: show vars
    shell: "echo {{ osd_group_name }}"

-------------------------
and the result of running it is:
-------------------------

[root@cephmon ceph-ansible]# ansible-playbook -v t.yml 

PLAY [stop ceph cluster] ****************************************************** 

GATHERING FACTS *************************************************************** 
ok: [cephmon]

TASK: [show vars] ************************************************************* 
fatal: [cephmon] => One or more undefined variables: 'osd_group_name' is undefined

FATAL: all hosts have already failed -- aborting

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/root/t.retry

cephmon                    : ok=1    changed=0    unreachable=1    failed=0   
-------------------------------

However osd_group_name is very clearly defined in the included file.

The root cause is that the "include:" statement looks like a variable assignment, so it tries to create a variable named "include".  I can't do the include statement outside the vars: declaration, because there is no "vars:" line in the included file. 

Drat.



----- Original Message -----

dpkshetty:
@leseb  I didn't realise that your intention was to make purge-cluster a standalone, since you had the ceph-common role being called to begin with :) thus i was more thinking on the lines of re-usability in the right sense

@bengland2 I had tried using include and i had hit some other error, i guess i tried including before the vars... in any case, i had tried couple of things before concluding on this patch, since nothing worked for me except this one... but i must admit I am not a ansible expert, so my conclusion might not be fair :) 

So if the intention is to make purge-cluster standalone then lets work towards it, if not, then i think my patch still makes sense.

leseb:
@dpkshetty thanks for your inputs :), closing this in favour of ben's patch then.

PR #643, opened 2016-03-22
Provide the ability to provide monitor_address instead of an interface

andrewschoen:
This would allow users who don't know what interface to provide to
give an IP address to use for the monitor instead.

Note: the includes are needed in ceph.conf.j2 because without them
jinja2 can not properly evaluate the template and will complain about a
missing 'ansible_interface' variable. The includes allow the template to
be evaluated correctly and then the correct include will be used during
render time.

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -191,6 +191,8 @@ dummy:
 
 ## Monitor options
 #
+# You must define either monitor_interface or monitor_address. Preference
+# will go to monitor_interface if both are defined.
 #monitor_interface: interface
 #monitor_address: 0.0.0.0
 #mon_use_fqdn: false # if set to true, the MON name used will be the fqdn in the ceph.conf
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -192,6 +192,7 @@ dummy:
 ## Monitor options
 #
 #monitor_interface: interface
+#monitor_address: 0.0.0.0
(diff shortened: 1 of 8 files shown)


leseb:
Thanks @andrewschoen.
One question, why dedicated jinja include files?

andrewschoen:
@leseb without the include files jinja fails to evaluate the template because ``hostvars[host]['ansible_interface']`` is undefined. If we use the includes jinja can validate ``ceph.conf.j2`` as a valid template and only render the appropriate include.

I was unable to find another way around this issue.

leseb:
We must make sure that the user doesn't populate both values (eth and ip). So we might need an extra check for this.
Also I think we need to highlight in the var file that the user must choose one or the other but not both.

leseb:
Hum wait, actually the first check in place should be sufficient since jinja is also doing a check. We can't possibly get both configured. Never mind :).

andrewschoen:
@leseb I've added the comments you requested. Defining both is fine, but ``monitor_interface`` will get priority if both are provided.

group_vars/all.sample
```diff
@@ -191,7 +191,10 @@ dummy:
 
 ## Monitor options
 #
+# You must use define either monitor_interface or monitor_address. Preference
```
leseb: s/use define/define/

andrewschoen: fixed

leseb:
Thanks @andrewschoen

PR #644, opened 2016-03-22
make purge-cluster.yml both more reliable and faster

bengland2:
This change has multiple enhancements:
- no longer uses ceph-common role
- works around open RHEL7.2 bz 1312676 that prevents removal of /var/lib/ceph/osd/* mountpoints after all OSDs are unmounted

Sorry this has multiple fixes in it, if necessary I can separate them out into separate PRs.  But I can now run this sequence as many times in a row as I want and it will come up clean every time with Hammer (RHCS 1.3.1) on RHEL7.2.

# ansible-playbook purge-cluster.yml
# ansible-playbook site.yml

To eliminate ceph-common dependency, I had to eliminate the dependency on is_ceph_infernalis fact, which is generated inside roles/ceph-common/tasks/main.yml .  To do this, the "are we using systemd" task counts native systemd unit files and we use this fact instead.

To finish deleting /var/lib/ceph/osd/* mountpoints, unfortunately it is necessary to reboot the OSD host.   This is done by using handlers.  It is only done where absolutely required. This works because we zap the OSD devices and remove ceph s/w before reboot, so that after reboot they wlil not be remounted.

Note that the removal of /var/lib/ceph is done in a handler now, so that it happens immediately for any system that doesn't have the above problem (see "request data removal" task at bottom).

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -11,10 +11,13 @@
 
   become: yes
 
-  roles:
-    - ceph-common
-
   vars:
+    osd_group_name: osds
+    mon_group_name: mons
+    rgw_group_name: rgws
+    mds_group_name: mdss
+    process_success: 0  # process exit status for success
+
 # When set to true both groups of packages are purged.
 # This can cause problem with qemu-kvm
     purge_all_packages: true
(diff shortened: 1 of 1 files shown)


bengland2:
sorry about huge font in .-1, I forgot I was using markdown language.

purge-cluster.yml
```diff
@@ -11,10 +11,13 @@
...
 
-  roles:
-    - ceph-common
-
   vars:
+    osd_group_name: osds
+    mon_group_name: mons
+    rgw_group_name: rgws
+    mds_group_name: mdss
+    process_success: 0  # process exit status for success
```
leseb: Looks like this is never used. What's the purpose of this variable?

bengland2: I updated the pull request to remove it.

----- Original Message -----

PR #650, opened 2016-03-25
Make RBD client directory owner/group configurable

stpierre:
Some versions (?) of libvirt provide a 'libvirt' group instead of
'libvirtd'. (Observed with libvirt-daemon-1.2.17-13.el7_2.2.x86_64.)
This makes the RBD client directory owner and group configurable to
allow for this.


leseb:
test this please

PR #651, opened 2016-03-25
Simplification and correction of when clauses for yum install

bengland2:
This is following up on discussion in issue 641, where you suggested a pull request like this.

The first commit is just trying to simplify and correct when clauses.  There are 4 distro types you have to support, but for these steps I don't think the decision tree is any different.  The existing when clause is wrong because it evaluates "ceph_stable_rh_storage and ceph_stable" but they are never both true, right?  I  I think the logic in the PR is cleaner and easier to understand unless I'm missing something, plz review. 

The 2nd commit is just to fail early where it's easy to understand what went wrong, rather than failing way into some complicated logic path where it's much harder to determine what went wrong.

### DIFF
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -58,7 +58,7 @@
     - ceph-mon
   when:
     mon_group_name in group_names and
-    ansible_pkg_mgr == "dnf" and
+    ansible_pkg_mgr == "dnf"
 
 - name: install distro or red hat storage ceph osd
   yum:
@@ -69,7 +69,7 @@
     - ceph-osd
   when:
     osd_group_name in group_names and
-    ansible_pkg_mgr == "yum" and
+    ansible_pkg_mgr == "yum"
 
 - name: install distro or red hat storage ceph osd
   dnf:
@@ -80,7 +80,7 @@
     - ceph-osd
   when:
     osd_group_name in group_names and
-    ansible_pkg_mgr == "dnf" and
+    ansible_pkg_mgr == "dnf"
 
 - name: install ceph-test
   yum:
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -34,7 +34,9 @@
   yum:
     name: ceph
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
-  when: not ceph_stable_rh_storage
+  when: 
+    not ceph_stable_rh_storage and
+    ansible_pkg_mgr == "yum"
 
 - name: install distro or red hat storage ceph mon
   yum:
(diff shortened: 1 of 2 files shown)


leseb:
Finally got the time to look into this, please see my comments.

roles/ceph-common/tasks/installs/install_on_redhat.yml
```diff
@@ -44,11 +46,8 @@
     - ceph
     - ceph-mon
   when:
-    (ceph_origin == "distro" or ceph_stable_rh_storage) and
     mon_group_name in group_names and
-    ansible_pkg_mgr == "yum" and
-    ceph_stable and
-    ceph_stable_release not in ceph_stable_releases
+    ansible_pkg_mgr == "yum"
```
leseb: we can not remove the check for stable releases because older versions don't have `ceph-mon` or `ceph-osd` packages. Please use the following when statement:

```
  when:
    (ceph_origin == "distro" or
    ceph_stable_rh_storage or
    (ceph_stable and ceph_stable_release not in ceph_stable_releases)) and
    mon_group_name in group_names and
    ansible_pkg_mgr == "yum"
```

Validated this on RHEL 7.2 and RHCS ISO installation.

roles/ceph-common/tasks/installs/install_on_redhat.yml
```diff
@@ -58,11 +57,8 @@
     - ceph
     - ceph-mon
   when:
-    (ceph_origin == "distro" or ceph_stable_rh_storage) and
     mon_group_name in group_names and
-    ansible_pkg_mgr == "dnf" and
-    ceph_stable and
-    ceph_stable_release not in ceph_stable_releases
+    ansible_pkg_mgr == "dnf"
```
leseb: dito

roles/ceph-common/tasks/installs/install_on_redhat.yml
```diff
@@ -72,11 +68,8 @@
     - ceph
     - ceph-osd
   when:
-    (ceph_origin == "distro" or ceph_stable_rh_storage) and
     osd_group_name in group_names and
-    ansible_pkg_mgr == "yum" and
-    ceph_stable and
-    ceph_stable_release not in ceph_stable_releases
+    ansible_pkg_mgr == "yum"
```
leseb: dito

roles/ceph-common/tasks/installs/install_on_redhat.yml
```diff
@@ -86,11 +79,8 @@
     - ceph
     - ceph-osd
   when:
-    (ceph_origin == "distro" or ceph_stable_rh_storage) and
     osd_group_name in group_names and
-    ansible_pkg_mgr == "dnf" and
-    ceph_stable and
-    ceph_stable_release not in ceph_stable_releases
+    ansible_pkg_mgr == "dnf"
```
leseb: dito

leseb:
@bengland2 could you please have a look at my comments? :)
Thanks!

bengland2:
Yes that makes sense, the "and" becomes an "or" so it should work.   Will try to update PR today.

leseb:
Great thanks!

bengland2:
This pull request is too messy at this point, I'm closing it and pushing what I think is a cleaner pull request based on comments above in next 5 min.

PR #653, opened 2016-03-26
ceph-common: Fix for issue no. 652

ksingh7:
ceph.conf file generation task in ceph-common role was getting failed
because it ansible cant find defination of varriable mon_containerized_deployment_with_kv
This fix declare mon_containerized_deployment_with_kv under ceph-common/defaults/main.yml which fixes this issue

Signed-off-by: ksingh7 <karan.singh731987@gmail.com>

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -340,3 +340,7 @@ dummy:
 
 #docker: false
 
+# Do not comment the variable mon_containerized_deployment_with_kv here. This variable is being used
+# by ceph.conf.j2 template. so it should always be defined
+#mon_containerized_deployment_with_kv: false
+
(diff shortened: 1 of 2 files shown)


bengland2:
I saw the same problem, added mon_containerized_deployment_with_kv to defaults and it worked.  Thx for fixing this.

bengland2:
when I actually tried what was committed here, it did not work because fsid value was not separated by a newline from max open files parameter that followed.   The jinja2 format was slightly off, I used this and it worked...

{% if not mon_containerized_deployment_with_kv %}
fsid = {{ fsid }}
{% endif %}

This gives fsid parameter its own separate record.

ksingh7:
@bengland2 it worked for me without newline but anyways nice catch , its in the PR now #657

PR #654, opened 2016-03-27
if user created OSD directories, don't zap disks

bengland2:
If the user is specifying osd_directory: true in group_vars/osds, then this is an indication that the user is taking responsibility for creating, and purging, OSD directories, so we don't want to do something as radical as zapping OSD block devices.    I tried commenting out the devices: variable in group_vars/osds -- when I do this, the task "check for a device list" fails.  If I supply a devices: variable that is empty, then this effectively disables disk zapping as well.   But it's not obvious that this is what you have to do.  It seems safer to do it this way.

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -237,7 +237,8 @@
     shell: ceph-disk zap "{{ item }}"
     with_items: devices
     when:
-     osd_group_name in group_names
+     osd_group_name in group_names and
+     not osd_directory
 
   - name: zap journal devices
     shell: ceph-disk zap "{{ item }}"


purge-cluster.yml
```diff
@@ -237,7 +237,8 @@
     shell: ceph-disk zap "{{ item }}"
     with_items: devices
     when:
-     osd_group_name in group_names
+     osd_group_name in group_names and
+     not osd_directory
```
leseb: won't this fail if `osd_directory` is not defined?

leseb: Since we don't depend on ceph-common we can not evaluate `osd_directory` since it might not be defined...

bengland2:
Sebastien you are correct, specifically osd_directory is defined in 

./roles/ceph-osd/defaults/main.yml:osd_directory: false

which is not read in unless you specify the "osds" role in purge-cluster.yml, which we do not want to do.   It was working for me only because I happened to have defined osd_directory in group_vars/osds.  I still think the problem description is valid but my proposed solution will not work.   I see you defined zap_journal_disks in purge-cluster.yml for this reason, I'm guessing.

How about extending zap_journal_disks to be named zap_block_devices and use this for both zap tasks?  This way, you can at least turn off zapping at the command line with:

# ansible-playbook -e zap_block_devices=false purge-cluster.yml

bengland2:
I'm closing this one.

leseb:
what about trying to detect `osd_directory` with something like:

`if [[ ! "$(find /var/lib/ceph/osd -prune -empty)" && ! "$(grep ceph /proc/partitions)" ]];``

Then we can avoid zap the disks.

bengland2:
I don't understand "find /var/lib/ceph/osd -prune -empty".  ceph-disk prepare creates an OSD directory like "ceph-1" here for each OSD, regardless of whether we provided a directory or a device for the OSD.  

However, I think it would be safe to ok to skip disk zap if the device doesn't contain a partition with a label that contains the string "ceph data" in case of data partition or "ceph journal" in case of journal partition.  We can tell if a device is a partition by just stripping the partition number off (and possibly the letter p) and checking to see if there is a parent block device, and we only want to zap a parent block device, not a partition, LVM device, etc.    I'll work on that as a PR instead.

leseb:
Sounds good thanks!

PR #655, opened 2016-03-28
ceph-common: use variable rgw_group_name instead of hard coded group name 'rgws'

runsisi:
ceph-common: use variable rgw_group_name instead of hard coded group name `rgws` when checking rgw port.

Signed-off-by: runsisi <runsisi@hust.edu.cn>


xiexingguo:
:+1:

PR #664, opened 2016-03-29
Add infernalis to list of stable releases comment

jsaintrocc:
Minor edit to the comment.


leseb:
`group_vars/all.sample` reflects the content of `ceph-common/defaults/main.yml` so next time someone run `generate_group_vars_sample.sh` this will get overwritten.
What's the idea behind your patch?

Thanks!

jsaintrocc:
Hi Sebastien,

Ah, I didn't see that a shell script was generating this.  I'll submit another pull request with a revised `ceph-common/defaults/main.yml`.  To answer your question, this fixes an inconsistency in the comments where infernalis is the default for ceph_stable_releases yet in the list of Stable Releases the last release mentioned is hammer.

Thanks, James

jsaintrocc:
Looks like adding infernalis to `ceph_stable_releases` in `ceph-common/defaults/main.yml`caused issues in how `is_ceph_infernalis` was determined.  My solution is to just have ansible ask ceph what version it's running in all cases.  That seems to fix the `ceph_stable_releases` issue and tidies up the logic for `is_ceph_infernalis` to boot :-).

roles/ceph-common/tasks/main.yml
```diff
@@ -64,24 +64,16 @@
...
-  when: ceph_stable_rh_storage
-
-# NOTE (leseb): be careful with the following
-# somehow the YAML syntax using "is_ceph_infernalis: {{"
-# does NOT work, so we keep this syntax styling...
-- set_fact:
-    is_ceph_infernalis={{ (ceph_stable and ceph_stable_release not in ceph_stable_releases) or (ceph_dev) or (ceph_stable_rh_storage and (rh_storage_version.stdout | version_compare('0.94', '>'))) }}
+  register: ceph_version
 
+# NOTE (leseb/jsaintrocc): Your supposed to qoute variables
```
leseb: s/qoute/quote/

leseb:
@jsaintrocc you should not need to add infernalis in the list of `ceph_stable_releases`, this variable is used in the `is_ceph_infernalis` fact later on and more recently here: https://github.com/ceph/ceph-ansible/pull/684/files. I like the fact simplification, we should keep it.
Since we are on it, do you mind renaming `is_ceph_infernalis` by `is_after_hammer`? And reflect this change in the entire playbook? Please use another commit for this :).
This will tremendously clarify its usage.

Thanks!

jsaintrocc:
Sounds good.  I'll make the changes to revert #684 in this PR, and start a new one to rename `is_ceph_infernalis` to `is_after_hammer`.  Thanks for catching the spelling by the way.

Happy to help :)

leseb:
@jsaintrocc sounds good thanks!

jsaintrocc:
Hum.  Looks like a problem with Jenkins vs the code?

```
Triggering ceph-ansible: Pull Requests » centos7,small,unique
Triggering ceph-ansible: Pull Requests » trusty,small,unique
Configuration ceph-ansible: Pull Requests » centos7,small,unique is still in the queue: There are no nodes with the label ‘small&&unique&&centos7’
Configuration ceph-ansible: Pull Requests » centos7,small,unique is still in the queue: 158.69.91.54+centos7_small_unique__87947213-1a0d-4f43-a8ac-23bbc972663f is offline
Configuration ceph-ansible: Pull Requests » centos7,small,unique is still in the queue: Waiting for next available executor on 158.69.91.54+centos7_small_unique__87947213-1a0d-4f43-a8ac-23bbc972663f
ceph-ansible: Pull Requests » centos7,small,unique completed with result FAILURE
Configuration ceph-ansible: Pull Requests » trusty,small,unique is still in the queue: There are no nodes with the label ‘small&&unique&&trusty’
Configuration ceph-ansible: Pull Requests » trusty,small,unique is still in the queue: 158.69.91.53+trusty_small_unique__d7707082-955a-4c9c-ba23-ced153c55364 is offline
Configuration ceph-ansible: Pull Requests » trusty,small,unique is still in the queue: Waiting for next available executor on 158.69.91.53+trusty_small_unique__d7707082-955a-4c9c-ba23-ced153c55364
ceph-ansible: Pull Requests » trusty,small,unique completed with result FAILURE
```

leseb:
test this please

leseb:
@andrewschoen any ideas why we are playing ceph-agent and ceph-fetch-keys?

roles/ceph-common/tasks/main.yml
```diff
@@ -64,24 +64,16 @@
   tags:
     - package-install
 
-- name: get ceph rhcs version
-  shell: rpm -q --qf "%{version}\n" ceph-common | cut -f1,2 -d '.'
+- name: get ceph version
+  shell: ceph --version | cut -f3 -d ' '
```
leseb: Can we add something like `| tee out.txt ; test ${PIPESTATUS[0]} -eq 0` to that command so we can trap an eventual installation problem?
The issue here is the `| cut` always returning 0 even if the `ceph` command failed.

jsaintrocc: Sounds good.  Or better yet, how about `set -o pipefail; ceph --version | cut -f3 -d ' '`.  This will return a fail if any command in the pipeline fails, and is a little cleaner ;-).  I found it in this [thread](http://stackoverflow.com/questions/1221833/bash-pipe-output-and-capture-exit-status)

jsaintrocc: Turns out, because Ubuntu uses `dash` when Ansbile invokes a shell, neither `PIPESTATUS` nor `-o pipefail` will work.  I could set the shell executable to `/bin/bash` but that means hard coding the executable path.  Another solution is to just get rid of the pipe and split out the version on the ansible side:

```
  - name: get ceph version
    command: ceph --version
    register: ceph_version

  - set_fact:
      ceph_version: "{{ ceph_version.stdout.split(' ')[2] }}"
```

Only downside to that is that `ceph_version.stdin` is used in a couple of other places and will have to be modified.  Does that work?

leseb: The latest approach looks good to me.
Do you mind modifying the other part using ceph_version too?

I think we need to remove this one: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/start_monitor.yml#L57-L61 and update those two https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/create_mds_filesystems.yml#L12 and https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/create_mds_filesystems.yml#L17 with `0.80.0`.
Finally those https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/secure_cluster.yml#L5 and https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/secure_cluster.yml#L12 with `0.94.0`.

Thanks!

alfredodeza:
test this please

jsaintrocc:
I tested the current version on both Ubuntu and Centos without issues.  I'll submit the requested updates later today.

Thanks, James

alfredodeza:
@jsaintrocc sorry, the "test this please" is meant to be a trigger for the Jenkins CI tests :)

jsaintrocc:
Sorry I didn't make it in time for the release tag.  Everything should be good to go on this.  I was silly and based this PR on the master of my fork.  Will recreating my fork cause issues for this PR?  There a couple of other commits I'd like to work on.

Thanks!

roles/ceph-mon/tasks/create_mds_filesystems.yml
```diff
@@ -9,9 +9,9 @@
     - cephfs_data
     - cephfs_metadata
   changed_when: false
-  when: not {{ ceph_version.stdout | version_compare('0.84', '<') }}
+  when: not {{ ceph_version | version_compare('0.80.0', '<') }}
```
leseb: s/80/84/

roles/ceph-mon/tasks/create_mds_filesystems.yml
```diff
@@ -9,9 +9,9 @@
...
 
 - name: create ceph filesystem
   command: ceph --cluster {{ cluster }} fs new {{ cephfs }} {{ cephfs_metadata }} {{ cephfs_data }}
   changed_when: false
-  when: not {{ ceph_version.stdout | version_compare('0.84', '<') }}
+  when: not {{ ceph_version | version_compare('0.80.0', '<') }}
```
leseb: s/80/84/

leseb:
test this please

leseb:
@jsaintrocc what are the other commit you want to integrate? Are they directly related to this PR?

roles/ceph-mon/tasks/create_mds_filesystems.yml
jsaintrocc: Done

roles/ceph-mon/tasks/create_mds_filesystems.yml
jsaintrocc: Done

jsaintrocc:
We discussed renaming `is_ceph_infernalis` to `is_after_hammer` as a separate commit.  Also wanted to update the sample vagrant file with `bento/centos-7.2` vs `bento/centos-7.1`.  So I don't think those are really related to this commit.

leseb:
@jsaintrocc thanks for the work!

Issue #667, opened 2016-03-30
ceph.ceph-common : generate ceph configuration file fails with python TypeError


analbeard:
```
TASK [ceph.ceph-common : generate ceph configuration file] *********************
task path: /root/ceph-02/ceph-ansible/roles/ceph-common/tasks/main.yml:178
An exception occurred during task execution. The full traceback is:
Traceback (most recent call last):
  File "/root/ansible/lib/ansible/executor/task_executor.py", line 124, in run
    res = self._execute()
  File "/root/ansible/lib/ansible/executor/task_executor.py", line 426, in _execute
    result = self._handler.run(task_vars=variables)
  File "/root/ceph-02/ceph-ansible/plugins/actions/config_template.py", line 261, in run
    tmp = self._make_tmp_path()
TypeError: _make_tmp_path() takes exactly 2 arguments (1 given)

fatal: [mon01-c02-rd]: FAILED! => {"failed": true, "msg": "Unexpected failure during module execution.", "stdout": ""}
```

I'm running ansible v2.1.0 and the following modules installed through pip:

Jinja2==2.8
PyYAML==3.11
httplib2==0.9.2
paramiko==1.16.0
six==1.10.0

OS is Trusty and I can provide my relevant group_vars if necessary.

leseb:
@cloudnull would you mind having a look at this?
I noticed you released another version for OpenStack, is this version compatible with 1.9.4 and 2.x.x?
I didn't have the time to update it yet.

Thanks!

cloudnull:
looking now. more soon. :)

cloudnull:
our current module only supports v1 and we have a PR to upstream Ansible to get a V2 module in for broader community consumption which can be found here: [ https://github.com/ansible/ansible/pull/12555 ]. We are working on making the module we're carrying v2 compatible however not completed that work yet; I'll work on that some more today. Once we have a solid v1 / v2 compatible action plugin I'll post back here and or PR the change into the main repo.

cloudnull:
If anyone has time to review the plugin within upstream Ansible it'd be much appreciated. I'm hopeful we'll be able to get it in for 2.1 though I have no confirmation on that at this point.

cloudnull:
@leseb I've posted a review to our main plugins repo -- https://review.openstack.org/#/c/299598 -- Once that change set has a few reviews and assuming it merges I'll pull the change in here.

cloudnull:
[...]

PR #668, opened 2016-03-30
Update the config_template plugin for ansible2

cloudnull:
This change updates the config_template plugin to load
the version of the config_template needed based on the
detected version of ansible running. Note, Once the
upstream change merges for ansible 2 [0] we should
drop the ansible 2  version of this plugin.

This change resolves the Ansible 2.x incompatibility issues as mentioned in this bug: https://github.com/ceph/ceph-ansible/issues/667

[0] - https://github.com/ansible/ansible/pull/12555

Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>


leseb:
Awesome @cloudnull, thanks a lot for the fast response and PR, really appreciate that.

PR #671, opened 2016-03-31
Added when: cephx to prevent waiting on keyring with cephx: false

squizzi:
We unnecessarily wait for `/etc/ceph/{{ cluster }}.client.admin.keyring` to be created when `cephx: false` is set in `group_vars/all ` which causes this task to timeout resulting in an abort. 

Adding a `when: cephx` clause fixes this.

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -4,6 +4,7 @@
 - name: wait for client.admin key exists
   wait_for:
     path: /etc/ceph/{{ cluster }}.client.admin.keyring
+  when: cephx
 
 - name: create ceph rest api keyring when mon is not containerized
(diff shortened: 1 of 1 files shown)


group_vars/common-coreoss.sample
```diff
@@ -14,4 +14,3 @@ dummy:
 #pypy_binary_directory: /opt/bin
 #pip_url: https://bootstrap.pypa.io/get-pip.py
 #local_temp_directory: /tmp
-
```
leseb: can you undo this change? this shouldn't be part of your commit. Thanks!

leseb:
Good catch @squizzi please address my comment and we are good to go.

group_vars/common-coreoss.sample
squizzi: woopsie, no idea how this got in ;) -- I'll fix it

leseb:
test this please

PR #683, opened 2016-04-05
Fix mis-use of config_template in docker tasks

stpierre:
As written, generating the config file for ceph-mon in Docker yielded:

ERROR: config_template is not a legal parameter in an Ansible task or
handler

This fixes that error condition.


stpierre:
Test this please

PR #684, opened 2016-04-06
fix when clauses for installation of ceph RPM

bengland2:
This pull request is a cleaner (I hope!) commit for the issues discussed in PR 651, which should be closed.    The important thing is that the old code was in effect evaluating "ceph_stable_rh_storage and ceph_stable" but these should not both be true.  The new code changes the "and" to an "or".  This code was tested by installing with RHCS 1.3.1 .iso and 1.3.2 .iso.    It would be nice if there was a way to factor out the common when clause expression used here, but other than "set_fact" I can't think of one.

### DIFF
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -36,6 +36,13 @@
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
   when: not ceph_stable_rh_storage
 
+# include Infernalis in the set of releases that do not have
+# a separate ceph-mon package or ceph-osd package
+# Note: Red Hat Ceph Storage is different, DOES separate them even in hammer
+
+- set_fact: > 
+    ceph_stable_rel_pkg="{{ ceph_stable_releases | union([ 'infernalis' ]) }}"
+
 - name: install distro or red hat storage ceph mon
   yum:
     name: "{{ item }}"
@@ -45,7 +52,7 @@
     - ceph-mon
   when:
     (ceph_origin == "distro" or ceph_stable_rh_storage or
-     (ceph_stable and ceph_stable_release not in ceph_stable_releases)) and
+     (ceph_stable and ceph_stable_release not in ceph_stable_rel_pkg)) and
     mon_group_name in group_names and
     ansible_pkg_mgr == "yum"
 
@@ -58,7 +65,7 @@
     - ceph-mon
   when:
     (ceph_origin == "distro" or ceph_stable_rh_storage or
-     (ceph_stable and ceph_stable_release not in ceph_stable_releases)) and
+     (ceph_stable and ceph_stable_release not in ceph_stable_rel_pkg)) and
     mon_group_name in group_names and
     ansible_pkg_mgr == "dnf"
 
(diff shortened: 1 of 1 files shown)


bengland2:
I'm stumped by the log result, which first says no package matching ceph-mon found, then it says ceph is already installed.  Are we properly reinitializing the test host to have no Ceph-related RPMs whatsoever prior to installation?   I'd like to find out what was in group_vars/ .

ailed: [localhost] => (item=[u'ceph', u'ceph-mon']) => {"changed": false, "failed": true, "invocation": {"module_args": {"conf_file": null, "disable_gpg_check": false, "disablerepo": null, "enablerepo": null, "exclude": null, "install_repoquery": true, "list": null, "name": ["ceph", "ceph-mon"], "state": "present", "update_cache": false}, "module_name": "yum"}, "item": ["ceph", "ceph-mon"], "msg": "No Package matching 'ceph-mon' found available, installed or updated", "rc": 0, "results": ["ceph-1:9.2.1-0.el7.x86_64 providing ceph is already installed"]}

leseb:
test this please

bengland2:
I did test in my own configuration, and it did work.    I do not understand the errors being reported by Jenkins, and the console log is not enlightening to me.  Can you explain why it failed?  The error reported the 2nd time is different than the error reported the first time, which makes no sense because the code hasn't changed.  

Does the code change make sense to you at least?

leseb:
test this please

leseb:
@bengland2 I suspect the CI is a bit broken because it looks like the play didn't even start...
And yes the change looks good to me. Let's wait for another round from jenkins.

leseb:
This is the failure: https://jenkins.ceph.com/job/ceph-ansible-pull-requests/DIST=centos7,MACHINE_SIZE=small,UNIQUE=unique/5/console

It can not found any ceph-mon package.

alfredodeza:
This is failing on CentOS7: https://jenkins.ceph.com/job/ceph-ansible-pull-requests/5/DIST=centos7,MACHINE_SIZE=small,UNIQUE=unique/console because it is testing infernalis but there is no ceph-mon in infernalis.

The package split happens in Jewel

leseb:
We need to add 'infernalis' to the list of checks in `ceph_stable_releases` but **only** in these `when` conditions.

bengland2:
There certainly is a ceph-mon RPM in Hammer on RHCS 1.3.1 and 1.3.2.

[root@cephmon ~]# rpm -q ceph-mon
ceph-mon-0.94.3-3.el7cp.x86_64

Is this a difference between upstream and RHCS?

How did you know it was testing with Infernalis?   Where did it say that?  I'm not trying to be difficult -  I need to understand how this is being tested.  I'm not familiar with the CI environment, so if I hit an error reported by it, I need to understand how to reproduce it or at least see what case it was covering.

And why did the build appear to succeed on (Ubuntu) trusty?  If it's the same version of Ceph, then shouldn't it fail equally on both Cenos7 and trusty?

I'll test with upstream Ceph next.

leseb:
@bengland2 you're right RHCS has ceph-mon and ceph-osd, it has always been the case as of Hammer.
Unfortunately upstream is a different story, only Jewel will have ceph-mon and ceph-osd package.
The CI is testing against both CentOS and Ubuntu on Infernalis (as stated here: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/defaults/main.yml#L72

It succeeded on Ubuntu because ceph-mon and ceph-osd don't exist yet (only with Jewel).
So it will just install the 'ceph' package.

leseb:
@bengland2 can you do something like:

```set_fact: ceph_stable_releases_pkg="{{ ceph_stable_releases | union([ 'infernalis' ]) }}"```

Earlier in the file? So we will have the proper list.
Then update the `when` tasks to use `ceph_stable_releases_pkg` instead of `ceph_stable_releases`.

After this Jenkins should be happy.
Thanks!

leseb:
Looks good and Jenkins is happy :)
Thanks @bengland2!

PR #685, opened 2016-04-06
create /etc/sysconfig/ceph if it doesn't already exist

bengland2:
This change deals with case where /etc/sysconfig/ceph doesn't already exist.  I ran into this while installing RHCS 1.3.1 iso image.  Perhaps later RPM versions create this file for you?  But the install errored out because of this.


leseb:
I think I've ran into this as well, however the issue was that the ceph package didn't get installed.
I'm fine with this change anyway, it doesn't hurt.

PR #688, opened 2016-04-06
low-priority - more robust, simpler, idempotent purge-cluster.yml

bengland2:
the first set of changes makes it safe to run purge-cluster twice, or to run it on a partially installed system.    For sysvinit-based service shutdowns, we first do a "service foo status bar" command and only proceed to the "service foo stop bar" command if the return status was 0 (success).  This ensures that we don't try to stop the service if the software isn't installed.   If the software is installed, it's harmless to stop a service that is already stopped.  We  can get rid of the when clause boolean expression:
  systemd_unit_files.stdout == "0"

The 2nd set of changes tests for two things: is ceph-disk utility installed, and are there partitions that were created with ceph-disk?  Only if both conditions are true do we allow zap tasks to proceed.   These tests are implemented in such a way that they should not care what Linux distro they are on.  For example, the ceph-disk test just sees if ceph-disk is in one of the PATH directories.  

Since we're not dealing with just disks I renamed zap_journal_disks to zap_block_devs and we use this var in the when clause for both zap tasks.

I use this in conjunction with osd_directory: true case - since I'm creating my own OSD filesystems for dm-cache testing, I omit "ceph data" or "ceph journal" from the partition labels for these devices, and now purge-cluster.yml won't touch them.


purge-cluster.yml
```diff
@@ -223,6 +223,16 @@
     register: check_for_running_ceph
     failed_when: check_for_running_ceph.rc == 0
 
+  - name: see if ceph-disk-created data partitions are present
+    shell: "ls /dev/disk/by-partlabel | grep -q 'ceph data'"
```
leseb: On two setups (ubuntu and centos) it is named `ceph\x20data` so this will give:

`ls /dev/disk/by-partlabel/ | egrep 'ceph\\x20data'`.

purge-cluster.yml
```diff
@@ -223,6 +223,16 @@
...
+    failed_when: false
+    register: ceph_data_partlabels
+
+  - name: see if ceph-disk-created journal partitions are present
+    shell: "ls /dev/disk/by-partlabel | grep -q 'ceph journal'"
```
leseb: On two setups (ubuntu and centos) it is named `ceph\x20journal` so this will give:

`ls /dev/disk/by-partlabel/ | egrep 'ceph\\x20journal'`.

bengland2:
isn't it interesting that the log of the failed play shows the first failure to be "Problem opening /dev/vdb for reading! Error is 2.\nThe specified file does not exist!\nProblem opening '' for writing! Program will now terminate.\nWarning! MBR not overwritten! Error is 2!" .    So I don't think my script deleted /dev/vdb.  Furthermore, the only file that changes in this commit is purge-cluster.yml and this file is never used in the CI regression test, is it?  Will run the test again and see if it gets the same result.  Am investigating your other suggestion.

leseb:
The issue on CI is not related to your changes.
Actually the CI doesn't use `purge-cluster.yml`.
Let me run it one more time and yes please have a look at my comments ;)

leseb:
test this please

alfredodeza:
This is failing because the `ceph-agent` role has been enabled:

    "msg": "No package matching 'calamari-server' is available"

It should get fixed with PR #705

bengland2:
thx much.  I've had to learn to think upstream as well.

----- Original Message -----

Issue #693, opened 2016-04-07
ceph.conf template has defaults that override better defaults in jewel


jdurgin:
There are a bunch of tunings and settings in ceph.conf that make sense to apply to older version of ceph, but are replaced by better defaults in jewel. As they change, it would be better to let the newer ceph defaults come into effect, instead of always overriding them in ceph.conf.

It would be great to have a more minimal ceph.conf for jewel and later - perhaps just via a 'no tuning' option - that does not put any of 'rbd default', 'rbd cache', 'osd *', 'filestore *', 'debug *' settings into ceph.conf (with the exception of setup config, like journal size and mkfs options if those are still needed).

vasukulkarni:
perhaps its better to have 'no tuning' which in general doesn't set any defaults for any section and let the user do all the setting.

leseb:
@jdurgin thanks for raising this one up.
Actually I've been thinking of removing entirely most of the default since they tend to pollute a bit the `ceph.conf`. I'm saying 'pollute' because we recently release the ability to dynamically overwrite the content of the `ceph.conf` using simple variables, see: https://github.com/ceph/ceph-ansible/blob/master/group_vars/all.sample#L306-L321
The reason why we have so many defaults is that at some point I wanted to highlight some of the best config options that users might want to tweak.
If I still haven't reworked this part is that changing the `ceph.conf`will trigger an unnecessary restart of all the ceph daemons... Moreover, this will reset most the settings that users implemented.

Perhaps I can do this for the next release of ceph-ansible and then add a major warning for all the running users.
Let me give a shot at purging the `ceph.conf` a bit, will put you in cc for review.

PR #694, opened 2016-04-07
ceph-common: purge ceph.conf file

leseb:
Since ##461 we have been having the ability to override ceph default
options. Previously we had to add a new line in the template and then
another variable as well. Doing a PR for one option was such a pain. As
a result, we now have tons of options that we need to maintain across
all the ceph version, yet another painful thing to do.
This commit removes all the ceph options so they are handled by ceph
directly. If you want to add a new option, feel free to to use the
`ceph_conf_overrides` variable of your `group_vars/all`.

Risks, for those who have been managing their ceph using ceph-ansible
this is not a trivial change as it will trigger a change in your
`ceph.conf` and then restart all your ceph services. Moreover if you did
some specific tweaks as well, prior to run ansible you should update the
`ceph_conf_overrides` variable to reflect your previous changes.

To avoid service restart, you need to know a bit of ansible for this,
but generally the idea would be to run ansible on a dummy host to
generate the ceph.conf, then scp this file to all your ceph hosts and
you should be good.

Closes: #693

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -167,21 +167,7 @@ dummy:
 #generate_fsid: true
 
 #cephx: true
-#cephx_require_signatures: true # Kernel RBD does NOT support signatures for Kernels < 3.18!
-#cephx_cluster_require_signatures: true
-#cephx_service_require_signatures: false
 #max_open_files: 131072
-#disable_in_memory_logs: true # set this to false while enabling the options below
-
-# Debug logs
-#enable_debug_global: false
-#debug_global_level: 20
-#enable_debug_mon: false
-#debug_mon_level: 20
-#enable_debug_osd: false
-#debug_osd_level: 20
-#enable_debug_mds: false
-#debug_mds_level: 20
 
 ## Client options
 #
@@ -224,9 +210,6 @@ dummy:
 #rbd_client_log_path: /var/log/ceph
 #rbd_client_log_file: "{{ rbd_client_log_path }}/qemu-guest-$pid.log" # must be writable by QEMU and allowed by SELinux or AppArmor
 #rbd_client_admin_socket_path: /var/run/ceph # must be writable by QEMU and allowed by SELinux or AppArmor
-#rbd_default_features: 3
-#rbd_default_map_options: rw
-#rbd_default_format: 2
 
 ## Monitor options
 #
@@ -235,36 +218,15 @@ dummy:
 #monitor_interface: interface
 #monitor_address: 0.0.0.0
 #mon_use_fqdn: false # if set to true, the MON name used will be the fqdn in the ceph.conf
-#mon_osd_down_out_interval: 600
-#mon_osd_min_down_reporters: 7 # number of OSDs per host + 1
-#mon_clock_drift_allowed: .15
-#mon_clock_drift_warn_backoff: 30
-#mon_osd_full_ratio: .95
-#mon_osd_nearfull_ratio: .85
-#mon_osd_report_timeout: 300
-#mon_pg_warn_max_per_osd: 0 # disable complains about low pgs numbers per osd
-#mon_osd_allow_primary_affinity: "true"
-#mon_pg_warn_max_object_skew: 10 # set to 20 or higher to disable complaints about number of PGs being too low if some pools have very few objects bringing down the average number of objects per pool. This happens when running RadosGW. Ceph default is 10
 
 ## OSD options
 #
 #journal_size: 0
-#pool_default_pg_num: 128
-#pool_default_pgp_num: 128
-#pool_default_size: 3
-#pool_default_min_size: 0 # 0 means no specific default; ceph will use (pool_default_size)-(pool_default_size/2) so 2 if pool_default_size=3
 #public_network: 0.0.0.0/0
 #cluster_network: "{{ public_network }}"
 #osd_mkfs_type: xfs
 #osd_mkfs_options_xfs: -f -i size=2048
 #osd_mount_options_xfs: noatime,largeio,inode64,swalloc
-#osd_mon_heartbeat_interval: 30
-
-# CRUSH
-#pool_default_crush_rule: 0
-#osd_crush_update_on_start: "true"
-
-# Object backend
 #osd_objectstore: filestore
 
 # xattrs. by default, 'filestore xattr use omap' is set to 'true' if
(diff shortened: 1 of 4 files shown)


leseb:
@andrewschoen looks like we are seeing the same error again on the Ubuntu VM :(

leseb:
@jdurgin, @andrewschoen, @alfredodeza feel free to chime :)

andrewschoen:
test this please

andrewschoen:
@leseb this looks ok to me, it certainly cleans things up. @jdurgin had mentioned to me that some of these defaults currently in ceph.conf are good for older versions of ceph, do we perhaps want to support version-specific configuration files?

andrewschoen:
@leseb the trusty failure was because the CI was picking up a node without an extra device attached. I think I've got that cleared up, but we will probably need to run the tests one more time.

vasukulkarni:
I think the rbd_client_log/dir should also go away, with selinux enabled this can create issues if they are not in default directories and policy exists only for default directories(for hammer/jewel), for other distro's it isn't an issue.

andrewschoen:
test this please

jdurgin:
looks good in terms of getting rid of settings that are suboptimal for jewel.

For other versions, I do think having a per-version config might make sense - e.g. the osd recovery settings commonly had to be adjust in hammer, which is probably where a lot of these came from. Maybe those could just go in a sample playbook as overrides or something? I'm not too familiar with the structure of things in ansible.

leseb:
Ok I'll try to add an example section for those who are looking at deploying Hammer then.

andrewschoen:
Test this please

leseb:
test this please

andrewschoen:
@leseb what else needs to be done for this one? Can I help?

leseb:
@andrewschoen I just need to rebase and then add some doc to address the Hammer use case I guess.

group_vars/all.sample
```diff
@@ -223,10 +209,14 @@ dummy:
 
 #rbd_client_log_path: /var/log/ceph
 #rbd_client_log_file: "{{ rbd_client_log_path }}/qemu-guest-$pid.log" # must be writable by QEMU and allowed by SELinux or AppArmor
+<<<<<<< HEAD
```
andrewschoen: I think we want to keep ``rbd_client_admin_socket_path``, but lose the rest of the ``rbd_*`` options here.

roles/ceph-common/defaults/main.yml
```diff
@@ -215,10 +201,14 @@ rbd_client_directory_mode: null
 
 rbd_client_log_path: /var/log/ceph
 rbd_client_log_file: "{{ rbd_client_log_path }}/qemu-guest-$pid.log" # must be writable by QEMU and allowed by SELinux or AppArmor
+<<<<<<< HEAD
```
andrewschoen: I think we want to keep ``rbd_client_admin_socket_path``, but lose the rest of the ``rbd_*`` options here.

leseb:
test this please

leseb:
test this please

roles/ceph-common/defaults/main.yml
leseb: sorry, forgot that during the rebase. fixed!

roles/ceph-common/templates/ceph.conf.j2
```diff
@@ -19,77 +12,14 @@ auth supported = none
...
-debug_mon = 0/0
-debug_paxos = 0/0
-debug_rgw = 0/0
-{% endif %}
-{% if enable_debug_global %}
-debug ms = {{ debug_global_level }}
-{% endif %}
-
 [client]
-rbd cache = {{ rbd_cache }}
```
mattt416: Looks like the defaults for `rbd_cache` will also need to be removed.

roles/ceph-common/templates/ceph.conf.j2
```diff
@@ -19,77 +12,14 @@ auth supported = none
...
-rbd cache writethrough until flush = true
```
mattt416: There is a default for `rbd_cache_writethrough_until_flush` despite it not actually getting used.

mattt416:
Setting | ceph-ansible | ceph
--- | --- | ---
cephx require signatures | true | false
cephx cluster require signatures | true | false
osd pool default pg num | 128 | 8
osd pool default pgp num | 128 | 8
rbd concurrent management ops | 20 | 10
rbd default map options | rw |
rbd default features | 3
rbd default format | 2 
mon osd down out interval | 600 | 300
mon osd min down reporters | 7 | 1
mon clock drift allowed | 0.15 | 0.5
mon clock drift warn backoff | 30 | 5
mon osd report timeout | 900 | 300 
filestore merge threshold | 40 | 10
filestore split multiple | 8 | 2
osd op threads | 8 | 2
filestore op threads | 8 | 2
osd recovery max active 5 | 15
osd max backfills | 2 | 10
osd recovery op priority | 2 | 63
osd disk thread ioprio class | idle | ''

leseb:
Hi @mattt416,

You raising a really good point. However I'm not really sure how to provide a compatible and non-breaking change.
What could be done for the next release of ceph-ansible, is to highlight this as a huge change.
So users getting the next ceph-ansible version will hopefully read the release note :).
The other option would be to keep previous default in `override_config` but I'm not such a big fan of this.

Any other idea?

alfredodeza:
test this please

leseb:
test this please

mattt416:
Hi @leseb,

TBH I didn't even realise you were using tags!  Having those in place helps a lot because it means you can be mindful of breaking changes when the major version is bumped.

We will discuss internally what to do for rpc-openstack -- whether we override locally to bring back a similar configuration or try to run with a more vanilla setup and only override things when we have a specific need.

Thanks!

--Matt

leseb:
test this please

leseb:
test this please

leseb:
test this please

leseb:
@mattt416 how did the discussion go?
I probably need to keep some examples for Hammer and before in the meantime.

leseb:
@jdurgin I'm currently documenting some of the variables that we would recommend to keep if we deploy a cluster before Jewel. Could you tag some of the variables we need highlight?

Thanks!

jdurgin:
@leseb I think the main ones are the recovery settings highlighted already in this pr. Others are pretty hardware-dependent, or not too important I think. You could point at the old config before this pr so folks could compare perhaps.

leseb:
@jdurgin good point, will do thanks!

PR #696, opened 2016-04-07
Deduplicate RBD client directory creation

### LINKED ISSUE #687: RBD client admin socket path created twice
The RBD client admin socket path is created multiple different times in wildly divergent ways:

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/installs/install_on_redhat.yml#L150-L160
https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/installs/install_on_debian.yml#L49-L59
https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/main.yml#L203-L209

The first two create it only if `rbd_client_directories` is true (it's false by default), with a configurable user/group (`qemu:libvirtd` by default), and with 0755 permissions. The third always creates it, uses an unconfigurable user/group (`ceph:ceph` on infernalis, `root:root` otherwise), and different permissions (`0770` on infernalis, `1777` otherwise). I'd be happy to fix this mess, but I have no idea which is correct (or if the actual correct option is some mish-mash of the two). The first two also create `rbd_client_log_path` with the same metadata, which also may or may not be correct.

stpierre:
Instead of creating the RBD client socket path three different places
in three different ways, this creates it once. Ceph on OpenStack users
have the option to customize the permissions of the RBD client
directories.

Fixes #687

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -182,11 +182,41 @@ dummy:
 #rbd_cache: "true"
 #rbd_cache_writethrough_until_flush: "true"
 #rbd_concurrent_management_ops: 20
+
 #rbd_client_directories: false # this will create rbd_client_log_path and rbd_client_admin_socket_path directories with proper permissions
-#rbd_client_directory_user: qemu
-#rbd_client_directory_group: libvirtd
-#rbd_client_log_file: /var/log/rbd-clients/qemu-guest-$pid.log # must be writable by QEMU and allowed by SELinux or AppArmor
+
+# Permissions for the rbd_client_log_path and
+# rbd_client_admin_socket_path. Depending on your use case for Ceph
+# you may want to change these values. The default, which is used if
+# any of the variables are unset or set to a false value (like `null`
+# or `false`) is to automatically determine what is appropriate for
+# the Ceph version with non-OpenStack workloads -- ceph:ceph and 0770
+# for infernalis releases, and root:root and 1777 for pre-infernalis
+# releases.
+#
+# For other use cases, including running Ceph with OpenStack, you'll
+# want to set these differently:
+#
+# For OpenStack on RHEL, you'll want:
+#   rbd_client_directory_owner: "qemu"
+#   rbd_client_directory_group: "libvirtd" (or "libvirt", depending on your version of libvirt)
+#   rbd_client_directory_mode: "0755"
+#
+# For OpenStack on Ubuntu or Debian, set:
+#    rbd_client_directory_owner: "libvirt-qemu"
+#    rbd_client_directory_group: "kvm"
+#    rbd_client_directory_mode: "0755"
+#
+# If you set rbd_client_directory_mode, you must use a string (e.g.,
+# 'rbd_client_directory_mode: "0755"', *not*
+# 'rbd_client_directory_mode: 0755', or Ansible will complain: mode
+# must be in octal or symbolic form
+#rbd_client_directory_owner: null
+#rbd_client_directory_group: null
+#rbd_client_directory_mode: null
+
 #rbd_client_log_path: /var/log/rbd-clients/
+#rbd_client_log_file: "{{ rbd_client_log_path }}/qemu-guest-$pid.log" # must be writable by QEMU and allowed by SELinux or AppArmor
 #rbd_client_admin_socket_path: /var/run/ceph/rbd-clients # must be writable by QEMU and allowed by SELinux or AppArmor
 #rbd_default_features: 3
 #rbd_default_map_options: rw
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -174,11 +174,41 @@ debug_mds_level: 20
 rbd_cache: "true"
 rbd_cache_writethrough_until_flush: "true"
 rbd_concurrent_management_ops: 20
+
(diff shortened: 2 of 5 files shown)


stpierre:
Test failure looks unrelated.

roles/ceph-common/defaults/main.yml
```diff
@@ -174,9 +174,39 @@ debug_mds_level: 20
 rbd_cache: "true"
 rbd_cache_writethrough_until_flush: "true"
 rbd_concurrent_management_ops: 20
+
 rbd_client_directories: false # this will create rbd_client_log_path and rbd_client_admin_socket_path directories with proper permissions
-rbd_client_directory_user: qemu
-rbd_client_directory_group: libvirtd
+
+# Permissions for the rbd_client_log_path and
+# rbd_client_admin_socket_path. depending on your use case for Ceph
```
leseb: s/depending/Depending/

roles/ceph-common/defaults/main.yml
```diff
@@ -174,9 +174,39 @@ debug_mds_level: 20
 rbd_cache: "true"
 rbd_cache_writethrough_until_flush: "true"
 rbd_concurrent_management_ops: 20
+
 rbd_client_directories: false # this will create rbd_client_log_path and rbd_client_admin_socket_path directories with proper permissions
```
leseb: this variable is not used anymore, we should remove it.

roles/ceph-common/defaults/main.yml
```diff
@@ -174,9 +174,39 @@ debug_mds_level: 20
...
+# If you set rbd_client_directory_mode, you must use a string (e.g.,
+# 'rbd_client_directory_mode: "0755"', *not*
+# 'rbd_client_directory_mode: 0755', or Ansible will complain: mode
+# must be in octal or symbolic form
+rbd_client_directory_user: null
+rbd_client_directory_group: null
+rbd_client_directory_mode: null
+
 rbd_client_log_file: /var/log/rbd-clients/qemu-guest-$pid.log # must be writable by QEMU and allowed by SELinux or AppArmor
 rbd_client_log_path: /var/log/rbd-clients/
```
leseb: we need to create this one too.

roles/ceph-common/defaults/main.yml
```diff
@@ -174,9 +174,39 @@ debug_mds_level: 20
...
+#
+# If you set rbd_client_directory_mode, you must use a string (e.g.,
+# 'rbd_client_directory_mode: "0755"', *not*
+# 'rbd_client_directory_mode: 0755', or Ansible will complain: mode
+# must be in octal or symbolic form
+rbd_client_directory_user: null
+rbd_client_directory_group: null
+rbd_client_directory_mode: null
+
 rbd_client_log_file: /var/log/rbd-clients/qemu-guest-$pid.log # must be writable by QEMU and allowed by SELinux or AppArmor
```
leseb: I think this one should be:
`rbd_client_log_file: {{ rbd_client_log_path }}/qemu-guest-$pid.log`

roles/ceph-common/defaults/main.yml
```diff
@@ -174,9 +174,39 @@ debug_mds_level: 20
...
+# For OpenStack on Ubuntu or Debian, set:
+#    rbd_client_directory_user: "libvirt-qemu"
+#    rbd_client_directory_group: "kvm"
+#    rbd_client_directory_mode: "0755"
+#
+# If you set rbd_client_directory_mode, you must use a string (e.g.,
+# 'rbd_client_directory_mode: "0755"', *not*
+# 'rbd_client_directory_mode: 0755', or Ansible will complain: mode
+# must be in octal or symbolic form
+rbd_client_directory_user: null
```
leseb: to keep things consistent:
s/rbd_client_directory_user/rbd_client_directory_owner/

leseb:
@stpierre several comments :)

roles/ceph-common/defaults/main.yml
stpierre: I thought we were going to keep it? In #687 you say:

"I think the right approach would be to disable the creation of the directories by default."

I've added the `when:` clause back in to directory creation. If you want me to remove it instead, I can do that.

stpierre:
Updated with all of your suggestions unless otherwise noted.

roles/ceph-common/defaults/main.yml
leseb: Argh you're right, let's stick we disable by default :).

group_vars/all.sample
```diff
@@ -182,9 +182,39 @@ dummy:
...
+# or `false`) is to automatically determine what is appropriate for
+# the Ceph version with non-OpenStack workloads -- ceph:ceph and 0770
+# for infernalis releases, and root:root and 1777 for pre-infernalis
+# releases.
+#
+# For other use cases, including running Ceph with OpenStack, you'll
+# want to set these differently:
+#
+# For OpenStack on RHEL, you'll want:
+#   rbd_client_directory_user: "qemu"
```
leseb: this one hasn't been updated, can you run `generate_group_vars_sample.sh` again?

stpierre: Done.

leseb:
test this please

leseb:
test this please

leseb:
test this please

stpierre:
The test failure seems unrelated.

leseb:
Indeed.

PR #701, opened 2016-04-08
look at monitor_interface in hostvars when choosing interface or address

andrewschoen:
This fixes a bug where monitor_interface might be set in your inventory
file and not by using group_vars or --extra-vars causing the template to
use the default address of 0.0.0.0 instead of the defined
monitor_interface.


alfredodeza:
This fixed the OSD not joining because of the bad mon for me. The ceph.conf file before:

    mon addr = 0.0.0.0

After:

    mon addr = 192.168.111.101

Issue #670, opened 2016-03-31
non-zero exit status from ceph-disk is not raised


alfredodeza:
TASK: [ceph-osd | activate osd(s) when device is a disk] **********************
    ok: [dhcp46-150.lab.eng.blr.redhat.com] 
    => (item=[
        {
            'changed': False, 
            'end': '2016-03-31 22:22:58.741350', 
            'failed': False, 
            'stdout': u'', 
            'cmd': "echo
            '/dev/vdb' | egrep '/dev/([hsv]d[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'",
            'rc': 1, 
            'start': '2016-03-31 22:22:58.730897', 
            'item': u'/dev/vdb',
            'warnings': [], 
            'delta': '0:00:00.010453', 
            'invocation': {
                'module_name': u'shell', 
                'module_complex_args': {}, 
                'module_args': u"echo '/dev/vdb' | egrep '/dev/([hsv]d[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'"},
            'stdout_lines': [], 
            'failed_when_result': False, 
            'stderr': u''}, u'/dev/vdb'
        ])
    
    => {"changed": false, "cmd": ["ceph-disk", "activate", "/dev/vdb1"], "delta":
    "0:09:58.695832", "end": "2016-03-31 22:34:15.195373", "failed": false,
    "failed_when_result": false, 
    "item": [{"changed": false, "cmd": "echo '/dev/vdb' | egrep '/dev/([hsv]d[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'",
    "delta": "0:00:00.010453", "end": "2016-03-31 22:22:58.741350", "failed":
    false, 
    "failed_when_result": false, 
    "invocation": {
        "module_args": "echo '/dev/vdb' | egrep '/dev/([hsv]d[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'",
        "module_complex_args": {}, 
        "module_name": "shell"}, 
    "item": "/dev/vdb", 
    "rc": 1, 
    "start": "2016-03-31 22:22:58.730897", 
    "stderr": "", 
    "stdout": "",
[...]

PR #703, opened 2016-04-11
if ceph-disk fails to activate an OSD then bubble up the error

alfredodeza:
Fixes issue #670

### DIFF
--- a/roles/ceph-osd/tasks/activate_osds.yml
+++ b/roles/ceph-osd/tasks/activate_osds.yml
@@ -20,10 +20,18 @@
     - devices
   changed_when: false
   failed_when: false
+  register: activate_osd_disk
   when:
     not item.0.get("skipped") and
     item.0.get("rc", 0) != 0 and
     not osd_auto_discovery
+
+- name: fail if ceph-disk cannot create an OSD
+  fail:
+    msg: "ceph-disk failed to create an OSD"
+  when:
+    " 'ceph-disk: Error: ceph osd create failed' in item.get('stderr', '') "
+  with_items: "{{activate_osd_disk.results}}"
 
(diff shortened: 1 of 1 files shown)


roles/ceph-osd/tasks/activate_osds.yml
```diff
@@ -20,11 +20,18 @@
     - devices
   changed_when: false
   failed_when: false
+  register: activate_osd_disk
   when:
     not item.0.get("skipped") and
     item.0.get("rc", 0) != 0 and
     not osd_auto_discovery
 
+- name: Fail if ceph-disk cannot create an OSD
```
leseb: just to stay consistent can you do: `s/Fail/fail/`?
Thanks!

The rest looks good :)

alfredodeza:
test this please

roles/ceph-osd/tasks/activate_osds.yml
```diff
@@ -20,11 +20,18 @@
...
+  register: activate_osd_disk
   when:
     not item.0.get("skipped") and
     item.0.get("rc", 0) != 0 and
     not osd_auto_discovery
 
+- name: fail if ceph-disk cannot create an OSD
+  fail:
+    msg: "ceph-disk failed to create an OSD"
+  when: "ceph-disk: Error: ceph osd create failed" in item.get("stderr", "")
```
andrewschoen: why not ``activate_osd_disk.get('stderr', '')``?  I don't think you need ``with_items`` here do you?

If you want all the lines of stderr I think you can use ``activate_osd_disk.stderr_lines``.

alfredodeza: you can't because `activate_osd_disk` is actually a list because it gets captured from a loop.

PR #704, opened 2016-04-11
Minor fixes that I uncovered while testing.

jimcurtis:


Vagrantfile
```diff
@@ -61,6 +61,7 @@ ansible_provision = proc do |ansible|
       ceph_mon_docker_subnet: "#{SUBNET}.0/24",
       ceph_osd_docker_extra_env: "CEPH_DAEMON=OSD_CEPH_DISK,OSD_JOURNAL_SIZE=100",
       ceph_osd_docker_devices: settings['disks'],
+      ceph_docker_on_openstack: OSVM,
```
leseb: We should not assume that we are running this on openstack by default.
So either you create a new condition for this or you update your `group_vars/all` file.

roles/ceph-osd/tasks/docker/main.yml
```diff
@@ -20,11 +20,11 @@
 
 - include: pre_requisite.yml
 
+- include: dirs_permissions.yml
```
leseb: why?

leseb:
Just fixed the docker image hardcoded path, please rebase.

Vagrantfile
jimcurtis: The condition for this is at line #22

roles/ceph-osd/tasks/docker/main.yml
jimcurtis: dirs_permissions.yml creates the directories that are need by fetch_configs.yml

Issue #718, opened 2016-04-19
Error while running purge-cluster.yml


ksingh7:
Hi , i encountered  some problems while purging Ceph cluster using purge-cluster.yml playbook

```
TASK [check for anything running ceph] *****************************************
fatal: [ceph-node1]: FAILED! => {"changed": true, "cmd": "ps awux | grep -v grep | grep -q -- ceph-", "delta": "0:00:00.085822", "end": "2016-04-19 10:37:51.001261", "failed": true, "failed_when_result": true, "rc": 0, "start": "2016-04-19 10:37:50.915439", "stderr": "grep: write error: Broken pipe", "stdout": "", "stdout_lines": [], "warnings": []}
changed: [ceph-node3]
changed: [ceph-node2]
```
The shell command execution is not consistent on all nodes

I have written a fix for this , would be submitting PR in sometime.

ksingh7:
PR #719

ksingh7:
Retested on a fresh set of nodes and it worked fine, so good to close this issue.

PR #719, opened 2016-04-19
Fix for #718

ksingh7:
Problem: Task that check for running Ceph processes fails against some nodes.
Due to that the failed node does not purges completely.

With this fix i have replaced that task with a new one "pkill -9 ceph" where it kills all running
ceph processes. The change has been tested and works fine than before

Signed-off-by: karan singh <karasing@redhat.com>

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -236,10 +236,11 @@
       ansible_distribution == 'Ubuntu' and
       rbdmirror_group_name in group_names
 
-  - name: check for anything running ceph
-    shell: "ps awux | grep -v grep | grep -q -- ceph-"
-    register: check_for_running_ceph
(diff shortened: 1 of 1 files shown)


ksingh7:
can someone help in fixing jenkins 
https://jenkins.ceph.com/job/ceph-ansible-pull-requests/77/consoleText

leseb:
👎 if the check fails it is for a good reason. Using `kill -9` when using systemd is bad idea; so we should gracefully stop services.
Can you check which service didn't stop and verify why?

Perhaps it is an issue with the stop sequence in the `purge-cluster.yml`

ksingh7:
Yeah , i should not nuke the processes :)
Anyways , Retested on a fresh set of nodes and it worked fine, so good to close this PR

PR #720, opened 2016-04-21
correctly set mon addr when monitor_interface is defined in group_vars

andrewschoen:
This fixes the ceph.conf template so that it will look for an inventory
defined value for monitor_interface or for monitor_interface defined in
a group_vars file.


andrewschoen:
Sorry for all the bugs in this template :(

PR #721, opened 2016-04-21
change defaults for rbd_client_directories and rbd_client_admin_socket_path

andrewschoen:
Changing these defaults will ensure that the rbd_client_directories are being created by default and will avoid the following warning when running ceph commands:

2016-04-19 15:20:05.128327 7fc0e4033700 -1 asok(0x7fc0dc001680)
    AdminSocketConfigObs::init: failed: AdminSocket::bind_and_listen:
    failed to bind the UNIX domain socket to
    '/var/run/ceph/rbd-clients/ceph-client.admin.20455.140466301441968.asok':
    (2) No such file or directory

See: https://bugzilla.redhat.com/show_bug.cgi?id=1328594

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -177,7 +177,7 @@ rbd_cache: "true"
 rbd_cache_writethrough_until_flush: "true"
 rbd_concurrent_management_ops: 20
 
-rbd_client_directories: false # this will create rbd_client_log_path and rbd_client_admin_socket_path directories with proper permissions
+rbd_client_directories: true # this will create rbd_client_log_path and rbd_client_admin_socket_path directories with proper permissions
 
 # Permissions for the rbd_client_log_path and
 # rbd_client_admin_socket_path. Depending on your use case for Ceph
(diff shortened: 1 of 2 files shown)


leseb:
@andrewschoen I don't think this is due to the rbdmirror but more because `/var/run/ceph/rbd-clients` does not get created automatically. 

`rbd_client_directories` default is false.

andrewschoen:
@leseb ok, I see if I set ``rbd_client_directories`` to ``true`` then I do not get this warning. Should we change the default value for ``rbd_client_directories``?

leseb:
@andrewschoen well probably and we should also stick the default to `/var/run/ceph`.
Do you want to PR this?

alfredodeza:
+1 on using the default `/var/run/ceph`

andrewschoen:
@leseb yeah, I can do that work

andrewschoen:
@leseb ok, I went ahead and made those changes in this PR.

leseb:
@andrewschoen cool let's wait for jenkins :)

andrewschoen:
test this please

andrewschoen:
We keep having issues with trusty, not related to this PR though. Maybe this next run will fair better.

roles/ceph-common/defaults/main.yml
```diff
@@ -211,7 +211,7 @@ rbd_client_directory_mode: null
 
 rbd_client_log_path: /var/log/rbd-clients/
```
leseb: I think we should stick with `/var/log/ceph` as well here.

andrewschoen: Ok, I've got that changed as well.

andrewschoen:
test this please

PR #731, opened 2016-04-25
bind-mounts /dev in osd containers

### LINKED ISSUE #730: docker: need to mount /dev in osd container (instead of specific device path)
When a partition (e.g. `/dev/sdb1`) doesn't exist, passing `/dev/sdb` is not sufficient (we need to bind-mount `/dev`).

ivotron:
This is needed in order to make ceph-disk work. Fixes #730


leseb:
@ivotron thanks!

PR #733, opened 2016-04-26
docker: fixes for OpenStack VMs

jimcurtis:
These were the changes I had to make to get things to work on my OpenStack VMs with the rhcs docker image.   In the case of the Vagrantfile change, there is already a conditional that sets OSVM based on the box and that defaults to false so I am using OSVM to set the ceph_docker_on_openstack variable.  In the case of the roles/ceph-osd/tasks/docker/main.yml, this change is necessary in new VM creation cases to make sure that dirs_permisssions.yml is run before the fetch because dirs_permissions creates the directories that we are copying into as part of the fetch.  The roles/ceph-restapi/tasks/docker/dirs_permissions.yml change is to use the same logic to set the ceph version fact as we do in all the other roles.  The logic that was there did not work for me.

### DIFF
--- a/roles/ceph-osd/tasks/docker/main.yml
+++ b/roles/ceph-osd/tasks/docker/main.yml
@@ -20,10 +20,12 @@
 
 - include: pre_requisite.yml
 
+# NOTE (jimcurtis): dirs_permissions.yml must precede fetch_configs.yml 
+# because it creats the directories needed by the latter.
+- include: dirs_permissions.yml
+
 - include: fetch_configs.yml
   when: not osd_containerized_deployment_with_kv
-
-- include: dirs_permissions.yml
 
 - include: selinux.yml
   when: ansible_os_family == 'RedHat'
--- a/roles/ceph-osd/tasks/docker/main.yml
+++ b/roles/ceph-osd/tasks/docker/main.yml
@@ -21,7 +21,7 @@
 - include: pre_requisite.yml
 
 # NOTE (jimcurtis): dirs_permissions.yml must precede fetch_configs.yml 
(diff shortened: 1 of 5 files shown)


roles/ceph-osd/tasks/docker/main.yml
```diff
@@ -20,11 +20,13 @@
 
 - include: pre_requisite.yml
 
+# NOTE (jimcurtis): dirs_permissions.yml must precede fetch_configs.yml 
+# because it creats the directories needed by the latter.
```
leseb: s/creats/creates

roles/ceph-restapi/tasks/docker/dirs_permissions.yml
```diff
@@ -9,12 +9,7 @@
 - set_fact:
     after_hamer=True
   when:
-    ceph_version not in (firefly or giant or hammer)
-
-- set_fact:
-    after_hamer=False
-  when:
-    ceph_version in (firefly or giant or hammer)
```
leseb: why do you remove this?

roles/ceph-osd/tasks/docker/main.yml
jimcurtis: Fixed typo

roles/ceph-restapi/tasks/docker/dirs_permissions.yml
jimcurtis: I copied what was done in dirs_permisssions.yml for ceph-mon and ceph-osd.  I assumed that that was the way to do it correctly.

PR #734, opened 2016-04-26
cleanup /tmp/rh-storage-repo so newer repos are actually installed

bengland2:
I got burned when I upgraded to RHCS 1.3.2 .iso image because old image still remained in /tmp/rh-storage-repo and so ceph-ansible skipped the step of copying from the (new) .iso image to /tmp/rh-storage-repo.   This fix clears out the old contents of /tmp/rh-storage-repo to ensure that new .iso image really gets used.

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -412,6 +412,9 @@
     when:
       ansible_pkg_mgr == 'dnf'
 
+  - name: purge RPM cache in /tmp
+    file: path=/tmp/rh-storage-repo state=absent
+
   - name: clean apt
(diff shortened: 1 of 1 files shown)


purge-cluster.yml
```diff
@@ -412,6 +412,9 @@
     when:
       ansible_pkg_mgr == 'dnf'
 
+  - name: purge RPM cache in /tmp
+    file: path=/tmp/rh-storage-repo state=absent
```
leseb: please use:

```
file:
  path:
  state:
```

Thanks!

PR #735, opened 2016-04-26
change rbd_client_admin_socket_path default

### LINKED ISSUE #278: AdminSocketConfigObs::init: failed: rbd-clients
After a fresh deployment on two nodes once I launch % ceph -w I always get the following error:

2015-06-10 18:34:06.599400 7f02fdce8700 -1 asok(0x7f02f8000bf0) AdminSocketConfigObs::init: failed: AdminSocket::bind_and_listen: failed to bind the UNIX domain socket to '/var/run/ceph/rbd-clients/ceph-client.admin.290250.139650727415008.asok': (2) No such file or directory

any suggestions? Thanks.

rootfs:
fix #278

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -211,7 +211,7 @@ rbd_client_directory_mode: null
 
 rbd_client_log_path: /var/log/ceph
 rbd_client_log_file: "{{ rbd_client_log_path }}/qemu-guest-$pid.log" # must be writable by QEMU and allowed by SELinux or AppArmor
-rbd_client_admin_socket_path: /var/run/ceph # must be writable by QEMU and allowed by SELinux or AppArmor
+rbd_client_admin_socket_path: /var/lib/ceph # must be writable by QEMU and allowed by SELinux or AppArmor
 rbd_default_features: 3
 rbd_default_map_options: rw
 rbd_default_format: 2


leseb:
👎 
This is unfortunate that `/var/run` is on `tmpfs`. But putting those files in `/var/lib/ceph` is going to be messy. I believe people running RedHat systems should either change `rbd_client_admin_socket_path` or add a hook to re-create the directory once you boot the host.

Thus I'm more in favour of a doc patch where we advise people to set another path :).

rootfs:
will open a doc pr later

leseb:
thanks!

fultonj:
If /var/run were tmpfs, wouldn't, wouldn't the mount command return it as that type? [1]. 
On RHEL7.2, /run is tmpfs but not /var/run AFAICT. 

[1]
[root@overcloud-controller-0 ~]# mount  | grep tmpfs
devtmpfs on /dev type devtmpfs (rw,nosuid,seclabel,size=65789380k,nr_inodes=16447345,mode=755)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev,seclabel)
tmpfs on /run type tmpfs (rw,nosuid,nodev,seclabel,mode=755)
tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,seclabel,mode=755)
tmpfs on /run/user/990 type tmpfs (rw,nosuid,nodev,relatime,seclabel,size=13161980k,mode=700,uid=990,gid=986)
tmpfs on /run/user/991 type tmpfs (rw,nosuid,nodev,relatime,seclabel,size=13161980k,mode=700,uid=991,gid=988)
tmpfs on /run/user/1000 type tmpfs (rw,nosuid,nodev,relatime,seclabel,size=13161980k,mode=700,uid=1000,gid=1000)
[root@overcloud-controller-0 ~]# mount  | grep tmpfs | grep var
[root@overcloud-controller-0 ~]# 
[root@overcloud-controller-0 ~]# cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.2 (Maipo)
[root@overcloud-controller-0 ~]#

dvusboy:
On systems using `systemd`, `/run` is a `tmpfs` and `/var/run` is a symbolic link to `/run` managed by `systemd-tmpfiles`. So, `/var/run` == `/run`.

dvusboy:
Back to the problem of vanishing `/var/run/ceph/rbd-clients`, I think what we can do, in a `systemd` way, is to use `tmpfiles` configuration to manage `/run/ceph/rbd-clients`.

Issue #736, opened 2016-04-27
`is_jewel` variable


alfredodeza:
In jewel a package split was done and installing `ceph` means pulling in every package: ceph-osd, ceph-mon, ceph-mds, etc.... regardless of the role.

We have discussed before having more granularity on these types of variables so this ticket will track progress towards the following:

* `is_hammer`
* `is_before_hammer`
* `is_after_hammer`
* `is_infernalis`
* `is_after_infernalis`
* `is_before_infernalis`
* `is_jewel`
* `is_after_jewel`
* `is_before_jewel`

The package split should be tracked separately, the related BZ to this is: https://bugzilla.redhat.com/1322774

alfredodeza:
Looking through the source it appears we have a mix of:

* `after_hammer`
* `is_ceph_infernalis`

This ticket should normalize them as per the initial description.

PR #737, opened 2016-04-27
Define readable vars for Ceph version comparisons

alfredodeza:
Fixes issue #736 

Adds vars all the way up to kraken

### DIFF
--- a/roles/ceph-common/handlers/main.yml
+++ b/roles/ceph-common/handlers/main.yml
@@ -8,8 +8,8 @@
   when:
     socket.rc == 0 and
     ansible_distribution != 'Ubuntu' and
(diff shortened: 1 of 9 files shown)


alfredodeza:
@bengland2 I think you once mentioned these vars as being confusing. Can you also take a look and let me know your thoughts on the new vars?

leseb:
@alfredodeza thanks for that, you need to set `include: facts.yml` after the package is installed.

alfredodeza:
@leseb of course! so silly :( will fix

alfredodeza:
@leseb re-pushed and now the tests are happy!

leseb:
Thanks!

PR #738, opened 2016-04-27
do not install the 'ceph' package on rpm systems for jewel

andrewschoen:
Starting with jewel we will no longer want to install the ceph package,
because it is a meta package that installs all ceph packages. We only
need to install the specific packages for the type of node being
configured, either a MON or OSD.

This depends on https://github.com/ceph/ceph-ansible/pull/737

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -63,6 +63,13 @@ ceph_origin: 'upstream' # or 'distro'
 #
 ceph_use_distro_backports: false # DEBIAN ONLY
 
+ceph_mon_packages:
+  - ceph
+  - ceph-mon
+
+ceph_osd_packages:
+  - ceph
+  - ceph-osd
 
 # STABLE
 ########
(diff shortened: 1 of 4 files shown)


ktdreyer:
:+1:

roles/ceph-common/defaults/main.yml
```diff
@@ -63,6 +63,13 @@ ceph_origin: 'upstream' # or 'distro'
 #
 ceph_use_distro_backports: false # DEBIAN ONLY
 
+ceph_mon_packages:
+  - ceph
```
leseb: I'm confused, you mentioned that the `ceph` package is actually installing all the Ceph packages.
So shouldn't we only have the `ceph-mon`package?
Am I missing something?

andrewschoen: @leseb the ``tasks/vars/after_infernalis_vars.yml`` file is overriding those default values. For anything before jewel we still need ``ceph`` and ``ceph-mon`` or ``ceph-osd``.

leseb: My bad, thanks for the clarification!

leseb:
Waiting for #737 to be fixed :)

alfredodeza:
this should not be merged yet, this relies on #737 but those vars are set after installation.

I think that these packages should *assume* that they will be needed on `rh_storage`, and maybe allow a `use_split_packages: true` setting for others that want to be granular about the installs

andrewschoen:
I'm closing this in favor of https://github.com/ceph/ceph-ansible/pull/739

PR #739, opened 2016-04-27
Adds a use_server_package_split var

andrewschoen:
This will allow a user to conditionally install the ceph package on rpm
based systems. Installing this package is not required or wanted in
versions passed infernalis.

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -38,6 +38,10 @@ dummy:
 # "state=latest".
 #upgrade_ceph_packages: False
 
+# If this is false then the 'ceph' package will be installed on rpm systems, which
+# is not needed for versions after infernalis.
+#use_server_package_split: true
+
 # /!\ EITHER ACTIVE ceph_stable OR ceph_stable_ice OR ceph_dev /!\
 
 #debian_package_dependencies:
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -30,6 +30,10 @@ check_firewall: True
 # "state=latest".
 upgrade_ceph_packages: False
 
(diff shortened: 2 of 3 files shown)


group_vars/all.sample
```diff
@@ -38,6 +38,10 @@ dummy:
 # "state=latest".
 #upgrade_ceph_packages: False
 
+# If this is false then the 'ceph' package will be installed on rpm systems, which
+# is not needed for versions after hammer.
```
ktdreyer: Shoot, this should technically be "after infernalis". My bad.

roles/ceph-common/defaults/main.yml
```diff
@@ -30,6 +30,10 @@ check_firewall: True
 # "state=latest".
 upgrade_ceph_packages: False
 
+# If this is false then the 'ceph' package will be installed on rpm systems, which
+# is not needed for versions after hammer.
```
ktdreyer: Technically "after infernalis". My bad.

ktdreyer:
"package_split" is a little generic. "server_split" ? "server_package_split" ?

group_vars/all.sample
andrewschoen: Ok, I'll fix these

alfredodeza:
this looks like a sensible approach to me (because we can't tell before installing what to do otherwise)

andrewschoen:
@ktdreyer how about ``use_package_name_split``?

ktdreyer:
The thing is that other packages might be split in the future.

andrewschoen:
@ktdreyer I've changed it to ``use_server_package_split``

leseb:
@andrewschoen @ktdreyer are you guys good with this one?

ktdreyer:
:+1: please merge

PR #744, opened 2016-05-01
take over an existing cluster with ceph ansible

leseb:
Introducing a playbook helper to control a ceph cluster that was not
deployed with ceph ansible.
The procedure is rather simple. If the cluster was deployed with the
following project there won’t be any issue:

* Ceph Deploy
* Puppet Ceph
* Chef Ceph
* Any other deployment tool that relies on ceph-disk

The procedure comes as fellow:

1. Install Ansible and add your monitors and osds hosts in it. For more
detailed information you can read the Ceph Ansible Wiki
2. Set  generate_fsid: false in group_vars
3. Get your current cluster fsid with ceph fsid and set cluster_fsid
accordingly in group_vars
4. Run the playbook called: take-over-existing-cluster.yml like this
ansible-playbook take-over-existing-cluster.yml.
5. Eventually run Ceph Ansible to validate everything by doing:
ansible-playbook site.yml.

### DIFF
--- a/take-over-existing-cluster.yml
+++ b/take-over-existing-cluster.yml
@@ -1,0 +1,42 @@
+---
+# NOTE (leseb):
+# The playbook aims to make ceph-ansible taking over a cluster
+# that was not configured with it.
+#
+# The procedure comes as fellow:
+#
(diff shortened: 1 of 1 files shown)


leseb:
This is one is just for you @ircolle!

PR #747, opened 2016-05-02
Add support for install from Ubuntu Cloud Archive

smatzek:
This adds support to allow the install of Ceph from the
Ubuntu Cloud Archive.  The Ubuntu Cloud Archive provides newer
release of Ceph than the normal Ubuntu distro repository.

Signed-off-by: Samuel Matzek <smatzek@us.ibm.com>

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -34,7 +34,7 @@ upgrade_ceph_packages: False
 # is not needed for versions after infernalis.
 use_server_package_split: true
 
-# /!\ EITHER ACTIVE ceph_stable OR ceph_stable_ice OR ceph_dev /!\
+# /!\ EITHER ACTIVE ceph_stable OR ceph_stable_ice OR ceph_stable_uca OR ceph_dev /!\
 
 debian_package_dependencies:
   - python-pycurl
@@ -129,6 +129,16 @@ ceph_stable_rh_storage_iso_install: false # usually used when nodes don't have a
 #ceph_stable_rh_storage_iso_path:
 ceph_stable_rh_storage_mount_path: /tmp/rh-storage-mount
 ceph_stable_rh_storage_repository_path: /tmp/rh-storage-repo # where to copy iso's content
+
+
+# UBUNTU CLOUD ARCHIVE
+# This allows the install of Ceph from the Ubuntu Cloud Archive.  The Ubuntu Cloud Archive
+# usually has newer Ceph releases than the normal distro repository.
+#
+ceph_stable_uca: false
+#ceph_stable_repo_uca: "http://ubuntu-cloud.archive.canonical.com/ubuntu"
+#ceph_stable_openstack_release_uca: liberty
+#ceph_stable_release_uca: "{{ansible_lsb.codename}}-updates/{{ceph_stable_openstack_release_uca}}"
 
 # DEV
 # ###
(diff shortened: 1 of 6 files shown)


leseb:
@smatzek I'm not strongly against, however I'd like to read a bit more about your motivations with this one. Is the purpose to unify repos along with an OpenStack deployment?

smatzek:
The motivation is two-fold:
1. Some shops may require the use of packages from Canonical for service contracts and support statements.  They may also want newer versions of Ceph than are in their distro.  If we take Ubuntu Trusty/14.04 as an example, the distro 'updates' apt repository contains Ceph Giant, whereas if you use the Ubuntu Cloud Archive you can get Ceph Hammer and Ceph Jewel.

2. Ceph.com [1], does not contain debian packages compiled for ppc64le whereas the Ubuntu Cloud Archive does [2].

[1] http://download.ceph.com/debian-hammer/dists/trusty/
[2] http://ubuntu-cloud.archive.canonical.com/ubuntu/dists/trusty-updates/mitaka/

alfredodeza:
test this please

roles/ceph-common/tasks/installs/debian_ceph_repository.yml
```diff
@@ -38,3 +38,15 @@
     state: present
   changed_when: false
   when: ceph_stable_ice
+
+- name: add Ubuntu Cloud Archive key package
```
leseb: please remove all the capitals :)

roles/ceph-common/tasks/installs/debian_ceph_repository.yml
```diff
@@ -38,3 +38,15 @@
...
+  apt:
+    pkg: ubuntu-cloud-keyring
+  when: ceph_stable_uca
+
+- name: add Ubuntu Cloud Archive repository
```
leseb: ditto

leseb:
@alfredodeza thanks for fixing the CI!
@smatzek would you mind adding the following check?

If an user actives UCA on an OS other than Ubuntu we fail?

smatzek:
Thanks for the feedback.  I've tested the latest check where a user activates UCA against a RHEL 7.2 target and an Ubuntu 14.04 target and it works as expected.

roles/ceph-common/defaults/main.yml
```diff
@@ -130,6 +130,16 @@ ceph_stable_rh_storage_iso_install: false # usually used when nodes don't have a
...
 ceph_stable_rh_storage_repository_path: /tmp/rh-storage-repo # where to copy iso's content
 
+
+# UBUNTU CLOUD ARCHIVE
+# This allows the install of Ceph from the Ubuntu Cloud Archive.  The Ubuntu Cloud Archive
+# usually has newer Ceph releases than the normal distro repository.
+#
+ceph_stable_uca: false
+ceph_stable_repo_uca: "http://ubuntu-cloud.archive.canonical.com/ubuntu"
+ceph_stable_openstack_release_uca: liberty
```
leseb: I think those 2 should not be enabled. Otherwise it'll always try to install from Ubuntu.

roles/ceph-common/defaults/main.yml
```diff
@@ -130,6 +130,16 @@ ceph_stable_rh_storage_iso_install: false # usually used when nodes don't have a
...
+ceph_stable_release_uca: "{{ansible_lsb.codename}}-updates/{{ceph_stable_openstack_release_uca}}"
```
leseb: I think this one should not be enabled. Otherwise it'll always try to install from Ubuntu.

leseb:
test this please

leseb:
@smatzek ok now it is another issue that we need to fix with the CI.

PR #749, opened 2016-05-03
Unify formatting of when conditional

SamYaple:
This is purely a refactor. Converts when 'and' conditionals into lists
rather than multiline strings. This does not work for nested
conditionals, but those can be formated with indents.

Moves one line when statements onto the same line as the when command
itself.

A small logic bug was found in ceph-osd/tasks/check_devices.yml which
which was also fixed.

Signed-off-by: Sam Yaple <sam@yaple.net>

### DIFF
--- a/roles/ceph-agent/tasks/pre_requisite.yml
+++ b/roles/ceph-agent/tasks/pre_requisite.yml
@@ -3,8 +3,7 @@
   command: subscription-manager identity
   register: subscription
   changed_when: false
-  when:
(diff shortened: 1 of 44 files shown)


SamYaple:
failure is related to new xenial job and is unrelated to this patch

leseb:
@SamYaple sorry for the pain, would you mind rebasing again?
I'll merge this rightaway :)
Thanks!

leseb:
@SamYaple Thanks for your patience :)

PR #748, opened 2016-05-03
ceph: use jewel release

leseb:
Signed-off-by: Sébastien Han <seb@redhat.com>

PR #750, opened 2016-05-03
Get the correct OSD ID when the cluster name includes numbers

andrewschoen:
If the ceph cluster name includes numbers, the grep used to find the OSD
IDs from /var/lib/ceph/osd/ would also return the numbers that were in
the cluster name.

For example, if the cluster was named 'mine123' and there was only one
OSD on the node, then the task that finds the OSD IDs would return
'123' and '0'.

### DIFF
--- a/roles/ceph-osd/tasks/activate_osds.yml
+++ b/roles/ceph-osd/tasks/activate_osds.yml
@@ -69,7 +69,7 @@
     is_before_infernalis
 
 - name: get osd id (for or after infernalis)
-  shell: "ls /var/lib/ceph/osd/ |grep -oh '[0-9]*'"
+  shell: "ls /var/lib/ceph/osd/ | grep '-' | cut -d '-' -f 2"
   changed_when: false
   failed_when: false
   register: osd_id


alfredodeza:
This looks OK to me!

alfredodeza:
test this please

roles/ceph-osd/tasks/activate_osds.yml
```diff
@@ -69,7 +69,7 @@
     is_before_infernalis
 
 - name: get osd id (for or after infernalis)
-  shell: "ls /var/lib/ceph/osd/ |grep -oh '[0-9]*'"
+  shell: "ls /var/lib/ceph/osd/ | grep '-' | cut -d '-' -f 2"
```
leseb: The first `grep` is not needed as cut already looks for the second field :).
So you can just keep the cut.

alfredodeza:
test this please

roles/ceph-osd/tasks/activate_osds.yml
alfredodeza: what if there are non-dashed files though? the grep ensures just dashed files will be used:


    $ touch foo bar osd-1 osd-2
    $ ls . | cut -d '-' -f 2
    bar
    foo
    1
    2
    $ ls . | grep '-' | cut -d '-' -f 2
    1
    2

leseb: @andrewschoen oh I see :), valid point!

leseb:
test this please

alfredodeza:
Of course we are failing on Xenial now because of lack of Xenial builds for Infernalis:

    E:Failed to fetch http://download.ceph.com/debian-infernalis/dists/xenial/main/binary-amd64/Packages

alfredodeza:
We can't win :(

leseb:
Yeah noticed that, but Jewel will fix this in #748 
This won't be the only thing that will be failing though, we need to rework some stuff for systemd.

PR #755, opened 2016-05-04
Set init_system fact and reference in ceph-osd role

### LINKED ISSUE #741: ceph-osd/tasks/scenarios/osd_directory.yml using fact not availabe in ansible 1.9.x
Hi All,

I get a failure when I try to use an OSD directory when using ansible 1.9.x.  I believe commit https://github.com/ceph/ceph-ansible/commit/210e4d4a015b0039c7ba1fd1427e3df846bdd430 introduced this and from what I can tell `ansible_service_mgr` is not available in ansible 1.9.x.  I'm not overly familiar w/ systemd -- what is the best way to check for this in the absence of `ansible_service_mgr` fact?

Thanks,
Matt

mattt416:
The ceph-osd role currently uses ansible_service_mgr, which is a fact
only available on ansible 2.x and greater.  This commit sets a similar
fact called init_system which will store the contents of /proc/1/comm
(systemd, init, etc.) and then references it ceph-osd instead.

Closes #741

### DIFF
--- a/roles/ceph-common/tasks/facts.yml
+++ b/roles/ceph-common/tasks/facts.yml
@@ -8,6 +8,15 @@
 - set_fact:
     ceph_version: "{{ ceph_version.stdout.split(' ')[2] }}"
 
+# NOTE(mattt): On ansible 2.x we can use ansible_service_mgr instead
+- name: check init system
+  slurp:
+    src: /proc/1/comm
+  register: init_system
+
+- set_fact:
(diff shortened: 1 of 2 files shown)


mattt416:
Does anyone know if this will work w/ docker?  Do we have a docker CI job that will fail if not?

leseb:
@mattt416 we don't have a Docker CI yet :-(
The patch looks good though, thanks!

Issue #676, opened 2016-04-03
init system checks


leseb:
We should start checking for systemd instead of distros. For example Ubuntu 15.04 is going to have systemd and no upstart anymore.
The `when` conditions should be modified in consequence.
Something like:

```
when: 
  is_systemd and
  {{ ansible_distribution_version | version_compare('15.04', '<') }}
```

To check for systemd, we can use a fact and test `grep -sq systemd /proc/1/comm`.

PR #758, opened 2016-05-04
Adds an use_systemd fact

andrewschoen:
This adds a helper fact that uses the ``init_system`` fact to determine if we should be using systemd to control services or not.

### DIFF
--- a/roles/ceph-common/handlers/main.yml
+++ b/roles/ceph-common/handlers/main.yml
@@ -17,7 +17,7 @@
       state: restarted
   when:
     socket.rc == 0 and
-    ansible_distribution != 'Ubuntu' and
+    use_systemd and
     mon_group_name in group_names and
     is_after_hammer
 
@@ -26,6 +26,7 @@
   when:
     socket.rc == 0 and
     ansible_distribution == 'Ubuntu' and
+    not use_systemd and
     mon_group_name in group_names
 
 - name: restart ceph osds
@@ -45,7 +46,7 @@
     state: restarted
   when:
     socket.rc == 0 and
-    ansible_distribution != 'Ubuntu' and
+    use_systemd and
     osd_group_name in group_names and
     is_after_hammer
 
@@ -57,6 +58,7 @@
   when:
     socket.rc == 0 and
     ansible_distribution == 'Ubuntu' and
+    not use_systemd and
     osd_group_name in group_names
 
 - name: restart ceph mdss on ubuntu
@@ -64,6 +66,7 @@
   when:
     socket.rc == 0 and
     ansible_distribution == 'Ubuntu' and
+    not use_systemd and
     mds_group_name in group_names
 
 - name: restart ceph mdss
@@ -71,6 +74,7 @@
   when:
     socket.rc == 0 and
     ansible_distribution != 'Ubuntu' and
+    not use_systemd and
     mds_group_name in group_names and
     ceph_stable and
     ceph_stable_release in ceph_stable_releases
@@ -81,7 +85,7 @@
       state: restarted
   when:
     socket.rc == 0 and
-    ansible_distribution != 'Ubuntu' and
+    use_systemd and
     mds_group_name in group_names and
     ceph_stable and
     ceph_stable_release not in ceph_stable_releases
@@ -91,6 +95,7 @@
   when:
     socketrgw.rc == 0 and
     ansible_distribution == 'Ubuntu' and
+    not use_systemd and
     rgw_group_name in group_names
 
 - name: restart ceph rgws
@@ -115,7 +120,7 @@
     state: restarted
   when:
     socketrgw.rc == 0 and
-    ansible_distribution != 'Ubuntu' and
+    use_systemd and
     rgw_group_name in group_names and
     is_after_hammer
 
--- a/roles/ceph-common/handlers/main.yml
+++ b/roles/ceph-common/handlers/main.yml
@@ -74,10 +74,9 @@
   when:
     socket.rc == 0 and
     ansible_distribution != 'Ubuntu' and
-    not use_systemd and
+    use_systemd and
     mds_group_name in group_names and
-    ceph_stable and
-    ceph_stable_release in ceph_stable_releases
+    is_before_infernalis
 
 - name: restart ceph mdss with systemd
   service:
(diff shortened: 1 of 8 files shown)


roles/ceph-common/tasks/main.yml
```diff
@@ -144,6 +144,10 @@
     (rbd_client_directory_mode is not defined or
      not rbd_client_directory_mode)
 
+- set_fact:
+    ubuntu_use_systemd: "{{ ansible_distribution_version | version_compare('16.04', '>=') }}"
```
ktdreyer: This could actually be a bit lower, since Ubuntu 15.04 was the first to ship systemd as the init system

roles/ceph-common/handlers/main.yml
```diff
@@ -91,13 +94,14 @@
   when:
     socketrgw.rc == 0 and
     ansible_distribution == 'Ubuntu' and
+    not ubuntu_use_systemd and
     rgw_group_name in group_names
 
 - name: restart ceph rgws
   command: /etc/init.d/radosgw restart
```
alfredodeza: This doesn't look like a systemd call though

andrewschoen: I think we need to make sure that Ubuntu nodes that want systemd *do not* make this call, no?

roles/ceph-common/tasks/main.yml
andrewschoen: Sure, I can make that change

andrewschoen:
test this please

mattt416:
Hi @andrewschoen , in https://github.com/ceph/ceph-ansible/commit/ed2b7757d43ba7af96d97a9f345aba38e4410bd2 we added a fact called `init_system` which _should_ return either systemd or init.  Could we use this instead of setting another fact?

alfredodeza:
@mattt416 what happens in a system that has more than one init system? (e.g. upstart and systemd like Ubuntu) We had to do odd things in ceph-deploy to accommodate: https://github.com/ceph/ceph-deploy/pull/385/commits/b796a45b119afb9186301a868c25a9ba70642891

ktdreyer:
@alfredodeza , there can only be one thing running at `/proc/1/comm` though, right? So I think that Matt's `init_system` will work?

andrewschoen:
I'm fine with using ``init_system`` if that can accomplish the same thing. What I don't like so much is having to say ``init_system == 'systemd'`` in multiple places. Perhaps I could add another fact, not a variable in defaults, like ``use_systemd: "{{ init_system == 'systemd' }}"``. Thoughts?

mattt416:
Hi all, I'm not an authority on any of this init stuff, but submitted https://github.com/ceph/ceph-ansible/commit/ed2b7757d43ba7af96d97a9f345aba38e4410bd2 as ceph-osd role to unblock a failure I was hitting.  If anyone spots any issues w/ that commit then let me know and I'll update!

@andrewschoen, your `use_systemd` variable sounds logical.  No objections there.

ktdreyer:
`use_systemd: "{{ init_system == 'systemd' }}"` sounds good to me @andrewschoen

alfredodeza:
+1 to the `use_systemd`

andrewschoen:
Ok, let me change this to create a 'use_systemd' variable instead.

andrewschoen:
test this please

roles/ceph-common/tasks/facts.yml
```diff
@@ -17,6 +17,9 @@
 - set_fact:
     init_system={{ init_system.content | b64decode }}
 
+- set_fact:
+    use_systemd={{ init_system == 'systemd\n' }}
```
andrewschoen: I don't like including the '\n' here, but I couldn't find a good filter provided by jinja2 to remove it. The logs in the CI tests proved it's needed though.

andrewschoen:
test this please

alfredodeza:
this is the first passing PR for Xenial! woot woot

andrewschoen:
I was able to get the xenial tests passing again with this commit https://github.com/ceph/ceph-ansible/pull/758/commits/324471c29777ad59bc257fc4b3f483f20253f2f3

ktdreyer:
:+1: LGTM

roles/ceph-common/tasks/facts.yml
mattt416: Odd, I'm not seeing a newline when running this on a 14.04 box w/ init:

https://gist.github.com/mattt416/54194ce68183da2bc8f42d6c87885314

Also, you could use the replace filter to strip off the newline, but as you can see in my gist I had to replace('\\n','') for some reason.

andrewschoen:
test this please

roles/ceph-common/handlers/main.yml
```diff
@@ -57,20 +58,23 @@
...
     ansible_distribution == 'Ubuntu' and
+    not use_systemd and
     mds_group_name in group_names
 
 - name: restart ceph mdss
   command: service ceph restart mds
   when:
     socket.rc == 0 and
     ansible_distribution != 'Ubuntu' and
+    not use_systemd and
```
andrewschoen: I think this is wrong. Isn't this command still systemd, just before infernalis?

roles/ceph-common/handlers/main.yml
```diff
@@ -57,20 +58,23 @@
...
     mds_group_name in group_names and
     ceph_stable and
     ceph_stable_release in ceph_stable_releases
```
andrewschoen: Could the last two lines here be replaced with ``is_before_infernalis``?

roles/ceph-common/handlers/main.yml
```diff
@@ -81,7 +85,7 @@
       state: restarted
   when:
     socket.rc == 0 and
-    ansible_distribution != 'Ubuntu' and
+    use_systemd and
     mds_group_name in group_names and
     ceph_stable and
     ceph_stable_release not in ceph_stable_releases
```
andrewschoen: Could these two lines be replaced with ``is_after_hammer``?

andrewschoen:
@leseb Could you please help me understand better my questions on the mds handlers? Everything else on this PR looks ok to me, and we're getting passing Xenial tests now. :)

roles/ceph-common/handlers/main.yml
andrewschoen: When this code was first introduced (https://github.com/ceph/ceph-ansible/commit/e6a97274786ea1b1d51a46a592ca624fcaf625a7), it would have kept this command from being run on infernalis. As it is currently this command would run on infernalis if ``ceph_stable`` was true. I think we do want to use ``is_before_infernalis`` here.

You can see here that infernalis is not in ``ceph_stable_releases`` at that commit:

https://github.com/ceph/ceph-ansible/blob/e6a97274786ea1b1d51a46a592ca624fcaf625a7/roles/ceph-common/defaults/main.yml#L44-L56

leseb:
@andrewschoen what's the issue with the mds?
Thanks for addressing #676

PR #762, opened 2016-05-07
install: remove epel

leseb:
We just add epel to conviently install Ansible. However we don't keep it
as it could disrupt ceph's installation and dependancies.


ktdreyer:
`yum remove epel-release` is sufficient; no need to run `rpm -q`

PR #770, opened 2016-05-10
Adding "jewel" to list of stable versions (ceph_stable_releases)

vincepii:


leseb:
This is not needed.  `ceph_stable_releases` is needed by a couple of check, Jewel should not be part of it.

PR #777, opened 2016-05-12
don't recreate /etc/ceph with wrong mode

rootfs:
this fixes the wrong permission on /etc/ceph
@leseb


dang:
This is correct.  dir_permissions.yml creates this correctly now.

PR #778, opened 2016-05-12
use the right interface name variable in mon; remove one line inline files

rootfs:
@leseb

### DIFF
--- a/roles/ceph-common/templates/ceph.conf.j2
+++ b/roles/ceph-common/templates/ceph.conf.j2
@@ -81,7 +81,7 @@ rgw socket path = /tmp/radosgw-{{ hostvars[host]['ansible_hostname'] }}.sock
 log file = /var/log/ceph/{{ cluster }}-rgw-{{ hostvars[host]['ansible_hostname'] }}.log
 rgw data = /var/lib/ceph/radosgw/{{ cluster }}-rgw.{{ hostvars[host]['ansible_hostname'] }}
 {% if radosgw_frontend  == 'civetweb' %}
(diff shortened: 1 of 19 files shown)


roles/ceph-common/templates/ceph.conf.j2
```diff
@@ -30,10 +30,12 @@ host = {{ hostvars[host]['ansible_fqdn'] }}
 host = {{ hostvars[host]['ansible_hostname'] }}
 {% endif %}
 # we need to check if monitor_interface is defined in the inventory per host or if it's set in a group_vars file
-{% if (hostvars[host]['monitor_interface'] is defined and hostvars[host]['monitor_interface'] != "interface") or monitor_interface != "interface" %}
-{% include 'mon_addr_interface.j2' %}
+{% set interface = ["ansible_",ceph_mon_docker_interface]|join %}
```
andrewschoen: I'm pretty sure this would break installs that are not using Docker. We use either ``monitor_interface`` or ``monitor_address``, this change would break that.

Also, the includes are necessary because if you keep the default of ``monitor_interface`` with is just 'interface' and use ``monitor_address`` instead then jinja2 fails to validate this template because ``hostvars[host]['ansible_interface']`` is not defined. However, using the includes allows jinja2 to validate this template and only render the file it needs to.

andrewschoen:
test this please

andrewschoen:
test this please

roles/ceph-common/templates/ceph.conf.j2
rootfs: That makes sense. We need another if-else to break these cases.

Issue #779, opened 2016-05-12
do not create symlinks for systemd


alfredodeza:
The rbd-mirror role does this in one of its tasks:

    - name: enable systemd unit file for the rbd mirror service (systemd after hammer)
      file:
        src: /usr/lib/systemd/system/ceph-rbd-mirror@.service
        dest: "/etc/systemd/system/multi-user.target.wants/ceph-rbd-mirror@{{ ceph_rbd_mirror_local_user }}.service"
        state: link
      changed_when: false
      failed_when: false
      when:
        - use_systemd
        - is_after_hammer

That should not be needed, if it is needed then the packaging of rbd-mirror/ceph is at fault

alfredodeza:
Hrmn, it looks like we do this all over the place. @leseb is there a specific reason we were not using:

    - service: name=ceph-rbd-mirror enabled=yes

ktdreyer:
(Note that on Ubuntu Xenial, `/usr/lib/systemd` does not even exist, so I'm a big :+1: to seeing this manage-symlinks-by-hand thing go away)

leseb:
I recently discovered that this wasn't needed since the packages are taking care of this.
I'm for removing those symlinks.

I actually recently removed it from the mon, the only issue is that the ansible "service" module doesn't seem to enable the services properly... So I had to do this: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/start_monitor.yml#L30-L36

@alfredodeza do you mind taking care of the PR?
Thanks!

ktdreyer:
In general if you see things that the packaging *should* be doing, *please* file tickets instead of adding workarounds in ceph-ansible! then we all benefit :)

leseb:
@ktdreyer it's not an issue with the packaging, it was a mistake :). However somehow this made the job quite well.

ktdreyer:
Ok, thanks

alfredodeza:
@leseb thank you for the pointers, I will take care of it.

leseb:
@alfredodeza Thanks!

PR #781, opened 2016-05-13
Do not create symlinks for systemd manually

andrewschoen:
Instead use 'systemctl enable' to create the symlinks.

This fixes: https://github.com/ceph/ceph-ansible/issues/779

### DIFF
--- a/roles/ceph-mds/tasks/pre_requisite.yml
+++ b/roles/ceph-mds/tasks/pre_requisite.yml
@@ -71,10 +71,7 @@
   when: not use_systemd
 
 - name: enable systemd unit file for mds instance (for or after infernalis)
-  file:
-    src: /usr/lib/systemd/system/ceph-mds@.service
-    dest: /etc/systemd/system/multi-user.target.wants/ceph-mds@{{ mds_name }}.service
-    state: link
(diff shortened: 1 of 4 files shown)


ktdreyer:
Can we just wipe out these bits entirely? why are we running `systemctl enable` when the service module does the same thing immediately afterwards?

leseb:
I think we should investigate a bit this. I looked into that a bit and it seems that issuing `systemctl enable` allows services to start where with only the service module nothing is happening...

leseb:
Argh this error is annoying, can anyone tell me which path we should be using?
I have the impression that sometimes with have `/etc/default/ceph` and sometimes `/etc/default/ceph/ceph`...

andrewschoen:
@leseb ``/etc/default/ceph`` is the correct path. This was a bug fixed in 10.2.1 https://github.com/ceph/ceph/commit/791eba81a5467dd5de4f1680ed0deb647eb3fb8b

I can make a PR to change that.

ktdreyer:
> I can make a PR to change that.

For posterity that fix is https://github.com/ceph/ceph-ansible/pull/784

leseb:
test this please

PR #784, opened 2016-05-13
fix the /etc/default/ceph path on debian when setting cluster name

andrewschoen:
In ceph version 10.2.1 a bug was fixed that corrects the path:

https://github.com/ceph/ceph/commit/791eba81a5467dd5de4f1680ed0deb647eb3fb8b


alfredodeza:
Looks OK to me

dcwangmit01:
The fix worked for me.

ktdreyer:
:+1:

PR #792, opened 2016-05-16
ceph-rgw: implement keystonve api version and ssl verify options

flaviodsr:
Jewel introduced rgw support for keystone v3 and ssl verfication.
This change enable this options on the ceph.conf file.

Signed-off-by: Flávio Ramalho <flaviosr@lsd.ufcg.edu.br>

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -26,6 +26,7 @@ dummy:
 #mds_group_name: mdss
 #restapi_group_name: restapis
 #rbdmirror_group_name: rbdmirrors
+#client_group_name: clients
 
 # If check_firewall is true, then ansible will try to determine if the
 # Ceph ports are blocked by a firewall. If the machine running ansible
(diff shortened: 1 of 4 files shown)


leseb:
please update `roles/ceph-common/defaults/main.yml` and then execute  `generate_group_vars_sample.sh`.

mattt416:
These defaults are already upstream ceph defaults -- can a deployer not just override with config template if they need a differing configuration?

leseb:
Yes I think we can use `config_template` for that.
I probably should have removed this section `radosgw_keystone` too.
I just thought it was handy to have it.

flaviodsr:
Done.

Could you guys explain what does "override with config template" mean?

leseb:
@flaviodsr please have a look at https://github.com/ceph/ceph-ansible#configuring-ceph

flaviodsr:
Got it!

Thanks @leseb.

leseb:
So it seems that we all agree that we should be closing this ;)

PR #801, opened 2016-05-19
ceph-osd: fix typo when calling sgdisk second time around

alfredodeza:


leseb:
Perhaps this is the right time to use the Ansible loop?

 `with_sequence: count=2`.

leseb:
Oh wait we already have a loop, that's gonna tricky.

PR #804, opened 2016-05-20
Allow running on local dev docker images

dang:
Docker makes it difficult to use images that are not on signed
registries.  This is a problem for developers, who likely won't have
access to a registry with proper signed certificates.

This allows the ability to use any docker image on the machine running
vagrant/ansible.  The way it works is that the image in question is
exported locally, then sent to each target box and imported there.

### DIFF
--- a/roles/ceph-common/tasks/docker/fetch_image.yml
+++ b/roles/ceph-common/tasks/docker/fetch_image.yml
@@ -1,0 +1,27 @@
+---
+# Normal case - pull image from registry
+- name: pull ceph daemon image
+  command: "docker pull {{ ceph_docker_username }}/{{ ceph_docker_imagename }}"
+  changed_when: false
+  failed_when: false
+  when: ceph_docker_dev_image is undefined or not ceph_docker_dev_image
+
+# Dev case - export local dev image and send it across
+- name: export local ceph dev image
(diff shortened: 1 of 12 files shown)


leseb:
@dang did you try to use the `command` module instead of `shell`, I suspect some of the tasks (to not say all of them) could work.
`shell` module brings some security concerns, so I'd like to avoid them as much as we can.

roles/ceph-common/tasks/docker/fetch_image.yml
```diff
@@ -0,0 +1,27 @@
+---
+# Normal case - pull image from registry
+- name: pull ceph daemon image
+  shell: "docker pull {{ ceph_docker_username }}/{{ ceph_docker_imagename }}"
```
SamYaple: there is an official ansible docker module for pulling and what not

dang: For both of these, I didn't actually add or change the existing code at all, just moved it to a common module.  I could fix those up, but maybe that should be a different commit?

(Sorry for the delay, my mail was being filtered.)

SamYaple: if this is just a refactor I agree with a separate commit.

dang:
Okay, converted shell to command.  I'll post a separate PR once this is in to change the open-coded docker commands to the plugin as much as possible.

PR #805, opened 2016-05-22
The when conditional was improperly being checked

SamYaple:
The scenarios were not being accurately compared to ensure that:
* A single scenario was choosen
* ONLY a single scenario was choosen

A custom filter can handle this job far better than extending the when
conditional any further. Additionally, it allows us to add new
scenarios in the future with greater ease.

### DIFF
--- a/filter_plugins/ceph-ansible.py
+++ b/filter_plugins/ceph-ansible.py
@@ -1,0 +1,13 @@
+from ansible import errors
+
+def only_one_true(a, *args):
+    params = list(args)
+    params.append(a)
+    it = iter(params)
+    return any(it) and not any(it)
+
+class FilterModule(object):
+    def filters(self):
+        return {
+            'only_one_true': only_one_true
+        }
--- a/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
+++ b/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
@@ -79,12 +79,7 @@
   when:
     - osd_group_name is defined
     - osd_group_name in group_names
-    - (journal_collocation and raw_multi_journal)
-      or (journal_collocation and osd_directory)
-      or (raw_multi_journal and osd_directory)
-      or (bluestore and journal_collocation)
-      or (bluestore and raw_multi_journal)
-      or (bluestore and osd_directory)
+    - journal_collocation | only_one_true(raw_multi_journal, osd_directory, bluestore)
 
 - name: verify devices have been provided
   fail:


SamYaple:
I am open to changing this. This patch does show a working implementation though

leseb:
We just encountered this issue with @guits, the fix looks good, however it'd be nice to have this properly handled by ansible. I'm fine with starting a custom module that can be improved later.

leseb:
About the issue you rose, I've always liked having everything being checked from the get go so moving the checks to ceph-osd is "fine" but I'd rather have the user being notified that something is not configure properly.
Moving the OSD variables is not really sophisticated but would work too.
If we set false at the end of each variable I don't think we will ever pass the check as at least one of the variable should be true.

I guess the only valid option is to move the check to ceph-osd. :)
Would you mind taking care of that?

Let me know if you want to do this here on in another PR.
Thanks!

SamYaple:
The default-to-false will properly fail if none are set as well. The filter only returns true if there is at least one true variable set. If none are set, they are all false.

leseb:
Thanks!

filter_plugins/ceph-ansible.py
```diff
@@ -0,0 +1,13 @@
+from ansible import errors
+
+def only_one_true(a, *args):
+    params = list(args)
+    params.append(a)
+    it = iter(params)
+    return any(it) and not any(it)
```
alfredodeza: this is always going to be `False`

filter_plugins/ceph-ansible.py
```diff
@@ -0,0 +1,13 @@
+from ansible import errors
+
+def only_one_true(a, *args):
+    params = list(args)
```
alfredodeza: this is already an iterable, why casting to a list here?

filter_plugins/ceph-ansible.py
```diff
@@ -0,0 +1,13 @@
...
+    params.append(a)
+    it = iter(params)
+    return any(it) and not any(it)
+
+class FilterModule(object):
+    def filters(self):
+        return {
+            'only_one_true': only_one_true
```
alfredodeza: I think this could basically be a plain `any`:

    return {'any': any}

Which is the same as "only one true" 

https://docs.python.org/2/library/functions.html#any

filter_plugins/ceph-ansible.py
SamYaple: This will not always be false. try it out.

The first any(it) iterators through the listiterator (think generator) and stops when it finds a True. The next one iterators through the _remaining_ listiterator items looking for another True, if it finds it this function returns false, if it does not find it (there was only a single True) this function returns True

it = iter([False, True, False])
any(it) and not any(it)  # This returns True

it = iter([False, False, False])
any(it) and not any(it)  # This returns False

it = iter([False, True, True])
any(it) and not any(it)  # This returns False

filter_plugins/ceph-ansible.py
SamYaple: just leftover code as I was playing with it. will remove

filter_plugins/ceph-ansible.py
SamYaple: It cannot. The issue here is that we need to make sure that there is a True in the list, and only a _single_ true. What you purpose would ensure there is _a_ scenario set, but not a _single_ scenario set.

filter_plugins/ceph-ansible.py
mattt416: TIL :)

leseb:
Guys, what do we do with this?

leseb:
@SamYaple @alfredodeza up? I'm seing more and more issues with this one so I'd like to fix it asap :)

filter_plugins/ceph-ansible.py
```diff
@@ -0,0 +1,13 @@
+from ansible import errors
+
+def only_one_true(a, *args):
```
alfredodeza: `a` would be included in `*args`, and this function doesn't seem to do anything particular to `a` except append it to `*args`.

This could just be:

    def only_one_true(*args)

filter_plugins/ceph-ansible.py
```diff
@@ -0,0 +1,13 @@
...
+    params = list(args)
+    params.append(a)
```
alfredodeza: no need for this if the function signature is `*args`

filter_plugins/ceph-ansible.py
```diff
@@ -0,0 +1,13 @@
...
+    it = iter(params)
```
alfredodeza: `params` is already an iterable

filter_plugins/ceph-ansible.py
alfredodeza: I really don't think we need the extra function. To ensure the use case you propose, this could be:

    { 'only_one_true':  lambda *args: args.count(True) == 1 }

If we insist in a function, this would look like:

   def only_one_true(*args):
        return args.count(True) == 1

alfredodeza:
This needs a bit more work. If we are going to get some Python in, we should also get thorough docstrings to document anything new.

filter_plugins/ceph-ansible.py
SamYaple: this is not the case, a is not in *args but this is a syntax thing with ansible and passing to a filter.

if the statement is as follows:

var1 | only_one_true(var2, var3)

then var1 = a and *args contains var2, var3. There are other ways to perform this check.

filter_plugins/ceph-ansible.py
SamYaple: commented above. there are a few ways to do this and it all involves the syntax we use with the filter outside of this file

filter_plugins/ceph-ansible.py
SamYaple: please read my previous comments on this subject. It is an iterable, but not a listiterator (which is more or less a generator) and iter(list()) turns it into a listiterator. This is definitely needed.

SamYaple:
@leseb I am going to submit a more simple, less flexible patchset to fix thsi is in the short term. well rearrange the variables and/or checks and then merge this one later.

filter_plugins/ceph-ansible.py
alfredodeza: Could you please point out where in Ansible this is documented behavior? I've looked at `filter_plugins` in the source tree and haven't seen anything that requires appending `a` to `*args` or similar.

SamYaple: I had edited the previous comment right after I posted it. I missed the point you were trying to make. You are correct here.

SamYaple:
Here is a PR to fix this in the short term. https://github.com/ceph/ceph-ansible/pull/831

PR #809, opened 2016-05-24
ceph-common: specify file path for repo key

alfredodeza:

Resolves: rhbz#1339068


roles/ceph-common/tasks/installs/install_rh_storage_on_debian.yml
```diff
@@ -5,7 +5,7 @@
 # intelligent way to determine the location of the key.
 - name: install the rh ceph storage repository key
   apt_key:
-    data: "{{ ceph_stable_rh_storage_repository_path }}/MON/release.asc"
+    file: "{{ ceph_stable_rh_storage_repository_path }}/MON/release.asc"
```
leseb: correct data should be used with lookups

PR #812, opened 2016-05-25
some fixes for container deployment

rootfs:
in containerized deployment, use the right interface name variable in mon; remove one line inline files

cleanup on docker osd tasks

@alfredodeza @leseb

### DIFF
--- a/group_vars/all.docker.sample
+++ b/group_vars/all.docker.sample
@@ -27,7 +27,7 @@ dummy:
 #ceph_osd_docker_username: ceph
 #ceph_osd_docker_imagename: daemon
 #ceph_osd_docker_extra_env: "CEPH_DAEMON=OSD_CEPH_DISK_ACTIVATE" # comma separated variables
-#ceph_osd_docker_prepare_env: "CEPH_DAEMON=OSD_CEPH_DISK_PREPARE,OSD_FORCE_ZAP=1" # comma separated variables
+#ceph_osd_docker_prepare_env: "OSD_FORCE_ZAP=1" # comma separated variables
 #ceph_osd_docker_devices:
 # - /dev/sdb
 # - /dev/sdc
--- a/group_vars/osds.sample
+++ b/group_vars/osds.sample
@@ -131,7 +131,7 @@ dummy:
 #osd_containerized_deployment_with_kv: false
 #kv_type: etcd
 #kv_endpoint: 127.0.0.1
-#ceph_osd_docker_prepare_env: ""
+#ceph_osd_docker_prepare_env: "OSD_FORCE_ZAP=1"
 #ceph_osd_docker_username: ceph
 #ceph_osd_docker_imagename: daemon
 #ceph_osd_docker_extra_env: "CEPH_DAEMON=OSD_CEPH_DISK" # comma separated variables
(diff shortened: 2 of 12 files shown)


roles/ceph-osd/tasks/docker/start_docker_osd.yml
```diff
@@ -15,25 +15,31 @@
...
-    image: "{{ ceph_osd_docker_username }}/{{ ceph_osd_docker_imagename }}"
-    name: "{{ ansible_hostname }}-osd-prepare-{{ item.0 | regex_replace('/', '') }}"
-    net: host
-    pid: host
-    state: started
-    privileged: yes
-    env: "OSD_DEVICE={{ item.0 }},{{ ceph_osd_docker_prepare_env }}"
-    volumes: "/var/lib/ceph:/var/lib/ceph,/etc/ceph:/etc/ceph,/dev:/dev,/run:/run"
+  shell: |   
+    docker run -ti --net=host \
```
leseb: `run -ti`?
why not `run -d`?

rootfs: the problem I see is that docker module doesn't watch container exit status, so the task may return before the command is finished (same with `docker -d`). Then when next task (osd activate) is started, the osd is still being prepared and that causes problem.

group_vars/all.docker.sample
```diff
@@ -27,7 +27,7 @@ dummy:
 #ceph_osd_docker_username: ceph
 #ceph_osd_docker_imagename: daemon
 #ceph_osd_docker_extra_env: "CEPH_DAEMON=OSD_CEPH_DISK_ACTIVATE" # comma separated variables
-#ceph_osd_docker_prepare_env: "CEPH_DAEMON=OSD_CEPH_DISK_PREPARE,OSD_FORCE_ZAP=1" # comma separated variables
+#ceph_osd_docker_prepare_env: "OSD_FORCE_ZAP=1" # comma separated variables
```
leseb: can you also edit `roles/ceph-osd/defaults/main.yml` and run `generate_group_vars_sample.sh`?

roles/ceph-osd/tasks/docker/start_docker_osd.yml
leseb: Ok I understand. Makes sense.

group_vars/all.docker.sample
rootfs: done, thanks!

rootfs:
tested containerized deployment on centos and RHEL Atomic.
@leseb @alfredodeza PTAL

group_vars/osds.sample
```diff
@@ -131,7 +131,7 @@ dummy:
 #osd_containerized_deployment_with_kv: false
 #kv_type: etcd
 #kv_endpoint: 127.0.0.1
-#ceph_osd_docker_prepare_env: ""
+#ceph_osd_docker_prepare_env: "CEPH_DAEMON=OSD_CEPH_DISK_PREPARE"
```
leseb: why do you have a different content now? 
both  `group_vars/all.docker.sample`  and `group_vars/osds.sample` should share the same content.

rootfs: re-generated

rootfs:
@leseb  @dang PTAL thanks

leseb:
test this please

PR #821, opened 2016-05-31
Ambiguous "version is only supported on RHEL 7.1"

gcharot:
Change from "only supported on RHEL 7.1" to "only supported on RHEL >= 7.1" for RHS


leseb:
@Thanks!

PR #823, opened 2016-05-31
ceph-common: ceph-base is not part of rhstorage, remove it

alfredodeza:
Solves regression when fixing https://bugzilla.redhat.com/show_bug.cgi?id=1339096

### DIFF
--- a/roles/ceph-common/tasks/installs/install_rh_storage_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_rh_storage_on_redhat.yml
@@ -45,9 +45,9 @@
   when:
     - mds_group_name in group_names
 
-- name: install red hat storage ceph base
+- name: install red hat storage ceph-common
   yum:
-    name: "ceph-base"
+    name: "ceph-common"
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
   when:
     - client_group_name in group_names
--- a/roles/ceph-common/tasks/installs/install_rh_storage_on_debian.yml
+++ b/roles/ceph-common/tasks/installs/install_rh_storage_on_debian.yml
@@ -68,3 +68,9 @@
     pkg: ceph-mds
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
   when: mds_group_name in group_names
+
+- name: install red hat storage ceph-common
+  apt:
+    pkg: ceph-common
+    state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
+  when: client_group_name in group_names


leseb:
Should we try to install something else instead? If we don't clients won't have any packages installed.
I believe the package does exist on Jewel right?
So the issue impacts everything before RHCS 2.0.

Am i right?

alfredodeza:
@leseb we never did install it in the first place (before the "rh storage on redhat" install task) and we haven't installed that for debian either.

leseb:
Yes but we only install it when the the client role is enabled. So this will only fail if the client role is enabled.
I looked at the bz, this regression is weird, this shouldn't be happening.

alfredodeza:
@leseb I guess this is happening because there is no `client_group_name` defined in downstream. Maybe the fix is to define that so this skips properly? 

But even then, we would implicitly allow the definition of `client_group_name` which would cause the error of `ceph-base` not being available

alfredodeza:
@leseb what about both: define `client_group_name` but also remove this from the tasks, there is no point in making ansible skip something that is not even provided

leseb:
Yes we should check and see if we need to update `ceph-common` vars to add `client_group_name`
I know but aren't you sure this is "just" an issue with 1.3 and that ceph-base is available in 2.0?
If the package does not exist in the end, then yes let's remove it but then we might need a replacement for it.
We have been having a lot of requests for a client role, so I'd like to keep it working :)

alfredodeza:
@leseb you are right, `ceph-base` does exist. But there is no need to install it explicitly as it is being pulled by ceph-osd, ceph-mds, and ceph-mon

ktdreyer:
> We have been having a lot of requests for a client role, so I'd like to keep it working :)

"ceph-base" is the shared set of files among the various servers (mon/osd/mds), and does not relate to the client (that package is "ceph-common")

( confusing names :fireworks: )

roles/ceph-common/tasks/installs/install_rh_storage_on_redhat.yml
```diff
@@ -45,9 +45,9 @@
   when:
     - mds_group_name in group_names
 
-- name: install red hat storage ceph base
+- name: install red hat storage ceph-common
   yum:
-    name: "ceph-base"
+    name: "ceph-common"
```
leseb: will this package be available in Jewel and later?

ktdreyer: Yes, ceph-common's been around for a long time (pre-firefly).

leseb: Cool thanks Ken

leseb:
@alfredodeza thanks for updating this, if you guys are good with this, so am I :)

PR #828, opened 2016-05-31
docker: enable rgw container on Atomic host environment

jimcurtis:
This will allow RGW to be started by systemd when it is an Atomic host environment.  This is what is done for OSD and MON.


leseb:
`roles/ceph-rgw/defaults/main.yml` needs to be updated as well, then you need to run `generate_group_vars_sample.sh`.

jimcurtis:
Made the updates as requested.

PR #831, opened 2016-06-01
The when conditional was improperly being checked

SamYaple:
The scenarios were not being accurately compared to ensure that:
* A single scenario was choosen
* ONLY a single scenario was choosen

This solution does not scale for long, but that can be addressed in a
different patchset.

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -331,3 +331,17 @@ docker: false
 # by ceph.conf.j2 template. so it should always be defined
 mon_containerized_deployment_with_kv: false
 
+
+##################
+# Temporary Vars #
+##################
+# NOTE(SamYaple): These vars are set here to they are defined before use. They
+# should be removed after a refactor has properly seperated all the checks into
+# the appropriate roles.
+
+journal_collocation: False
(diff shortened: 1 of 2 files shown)


leseb:
Looks good but don't we have the same problem with the other OSD related tasks?

leseb:
I just tested this, and it looks we also need to set `default(False)` to all the next tasks.

SamYaple:
instead of setting default(False) to all of those I purpose a temporary vars file that will difeault all the variables to false. we can remove it in a later refactor

leseb:
@SamYaple please go for it :)

leseb:
Thanks @SamYaple, I think we also need (for more complex scenarios):

* osd_directories
* osd_auto_discovery
* devices
* raw_multi_journal

SamYaple:
- raw_multi_journal is set to false in this patch already
- osd_auto_discovery is set to false in this patch already
- devices is checked with "devices is not defined"
- osd_directories is checked with "osd_directories is not defined"

That covers all of the things you commented on. If thats it, this patch is good as is.

leseb:
It is indeed, I guess it was just me not looking properly. Thanks again!

PR #483, opened 2016-01-15
help ansible find repo template from installs subdirectory

bengland2:
here is a PR with just the change to the path for the repo template - if you run without this you get an error like this.  Ansible seems to be checking ../templates, assuming that all yaml files will run from inside tasks/ subdirectory and not a directory nested inside it.

fatal: [rhs-cli-01] => input file not found at /mnt/oldroot/home/ben/ceph/dm-cache/ceph-ansible/roles/ceph-common/tasks/templates/redhat_storage_repo.j2 or /mnt/oldroot/home/ben/ceph/dm-cache/ceph-ansible/redhat_storage_repo.j2

BTW a similar change would be needed for RHCS 1.2 (Inktank Ceph Enterprise).

PR #833, opened 2016-06-02
ceph-common: uniformize template path calls

leseb:
In order to align all Ansible versions, we now use the full path for the
template. We rely on `role_path` variable. Now all the tasks using
the template module have a uniform syntax.

Might fix issue raised in #483

### DIFF
--- a/roles/ceph-common/tasks/installs/install_rgw_on_debian.yml
+++ b/roles/ceph-common/tasks/installs/install_rgw_on_debian.yml
@@ -104,7 +104,7 @@
 
 - name: install default httpd.conf
   template:
-    src: ../../templates/httpd.conf
+    src: "{{ role_path }}/templates/httpd.conf"
     dest: /etc/apache2/httpd.conf
     owner: root
     group: root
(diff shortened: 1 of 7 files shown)


leseb:
@ksingh7 this should fix your issues with Ansible 2.X

ksingh7:
@leseb looks clean 👍  will try this out

roles/ceph-common/tasks/installs/install_rgw_on_debian.yml
```diff
@@ -104,7 +104,7 @@
 
 - name: install default httpd.conf
   template:
-    src: ../../templates/httpd.conf
+    src: "{{ playbook_dir }}/roles/ceph-common/templates/httpd.conf"
```
SamYaple: Here are the three options I know of to specify a template in the templates directory. I would suggest the first one for looks.

- "httpd.conf"
- "{{ role_path }}/templates/httpd.conf"
- "{{ playbook_dir }}/roles/ceph-common/templates/httpd.conf"

leseb: right this one is not needed :)

Issue #819, opened 2016-05-30
yum breaks because purge-cluster.yml does not removes /etc/yum.repos.d/rh_storage.repo file


ksingh7:
``purge-cluster.yml``  is not removing ``/etc/yum.repos.d/rh_storage.repo`` file due to which yum breaks 

```
[root@rhcs1 ~]# yum install nmap
Loaded plugins: search-disabled-repos
file:///tmp/rh-storage-repo/MON/repodata/repomd.xml: [Errno 14] curl#37 - "Couldn't open file /tmp/rh-storage-repo/MON/repodata/repomd.xml"
Trying other mirror.


 One of the configured repositories failed (Red Hat Storage Ceph - local packages for Ceph),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:

     1. Contact the upstream for the repository and get them to fix the problem.

     2. Reconfigure the baseurl/etc. for the repository, to point to a working
        upstream. This is most often useful if you are using a newer
        distribution release than is supported by the repository (and the
        packages for the previous distribution release still work).

     3. Disable the repository, so yum won't use it by default. Yum will then
        just ignore the repository until you permanently enable it again or use
        --enablerepo for temporary usage:

            yum-config-manager --disable rh_storage_mon

     4. Configure the failing repository to be skipped, if it is unavailable.
        Note that yum will try to contact the repo. when it runs most commands,
        so will have to try and fail each time (and thus. yum will be be much
        slower). If it is a very temporary problem though, this is often a nice
        compromise:

            yum-config-manager --save --setopt=rh_storage_mon.skip_if_unavailable=true

failure: repodata/repomd.xml from rh_storage_mon: [Errno 256] No more mirrors to try.
file:///tmp/rh-storage-repo/MON/repodata/repomd.xml: [Errno 14] curl#37 - "Couldn't open file /tmp/rh-storage-repo/MON/repodata/repomd.xml"
[root@rhcs1 ~]#
[root@rhcs1 ~]# cd /etc/yum.repos.d/
[root@rhcs1 yum.repos.d]# ll
total 208
-rw-r--r--. 1 root root    957 May 29 19:54 epel.repo
-rw-r--r--. 1 root root   1056 Mar 31 00:05 epel-testing.repo
-rw-r--r--. 1 root root 196631 May 29 17:35 redhat.repo
-rw-r--r--. 1 root root    739 May 29 20:03 rh_storage.repo

[...]

PR #834, opened 2016-06-02
Fix for #819

ksingh7:
This commit removes /etc/yum.repos.d/rh_storage.repo file
for RedHat OS with RHCS ISO install during cluster purge


leseb:
Why do we need to change cache settings?

ksingh7:
@leseb my bad didn't noticed there are 2 commits going with this PR
So dont merge this one ( delete this PR if you can )

I have created a new PR here #838

PR #836, opened 2016-06-02
include the red hat storage apt-key as a file

andrewschoen:
This will allow nodes to install rhcs that do
not have access to the internet.

Resolves: rhbz#1337601


leseb:
Nice one thanks!

PR #839, opened 2016-06-02
Remove Centos7 dependency yum-plugin-priorities

dcwangmit01:
Currently, the ceph-ansible scripts fail when targeting Centos7 machines, due to the installation of yum-plugin-priorities, which forces older package dependencies (provided by default distribution repositories) to take precedence over newer ceph packages as provided by the ceph-release-{version}.noarch.rpm.  This is true even when setting the ceph yum repo priority=1.

See this [cautionary note](https://wiki.centos.org/PackageManagement/Yum/Priorities#head-38b91468cc607d0243f463489c2334bf40bfaaee) from the upstream maintainer of yum.

The error causes newer package versions from the ceph repository to be silently ignored in favor of older package versions from the default distribution repos.  The error looks like the following, repeated many times.

```
Error: Package: 1:ceph-base-10.2.1-0.el7.x86_64 (Ceph)
           Requires: librbd1 = 1:10.2.1-0.el7
           Available: 1:librbd1-0.80.7-2.el7.x86_64 (core-0)
```
The above type of errors make ansible fail, because the correct packages cannot be installed.

Tested on

```
ansible 2.1.0.0
```

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -339,3 +339,17 @@ dummy:
 #mon_containerized_deployment_with_kv: false
 
 
+##################
+# Temporary Vars #
+##################
+# NOTE(SamYaple): These vars are set here to they are defined before use. They
+# should be removed after a refactor has properly seperated all the checks into
+# the appropriate roles.
+
+#journal_collocation: False
+#raw_multi_journal: False
+#osd_directory: False
+#bluestore: False
+
+#osd_auto_discovery: False
+
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -53,7 +53,6 @@ dummy:
 #centos_package_dependencies:
 #  - python-pycurl
 #  - hdparm
-#  - yum-plugin-priorities.noarch
 #  - epel-release
 #  - ntp
 #  - python-setuptools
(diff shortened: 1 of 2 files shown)


leseb:
Please rebase your patch :)

dcwangmit01:
test this please

group_vars/all.sample
```diff
@@ -243,7 +242,7 @@ dummy:
 
 ## OSD options
 #
-#journal_size: 0 # OSD journal size in MB
```
leseb: do you mind updating `roles/ceph-common/defaults/main.yml` to reflect that?

group_vars/all.sample
```diff
@@ -132,7 +131,7 @@ dummy:
 #ceph_stable_ice_password: # htaccess password
 
 # ENTERPRISE VERSION RED HAT STORAGE (from 1.3)
-# This version is only supported on RHEL >= 7.1
```
leseb: do you mind updating `roles/ceph-common/defaults/main.yml` to reflect that?
then run `generate_group_vars_sample.sh`.
Thanks!

dcwangmit01:
@leseb Most of those changes from group_vars/all.sample were not mine, but from prior checkins without re-running generate_group_vars_sample.sh.

I've migrated the comments from group_vars/all.sample back into the main.yml file as you've requested.

To clarify, I've separated this change into 3 commits:
1) Remove yum-plugin-priorities
2) Move the comments from all.sample back into main.yml
3) Run generate_group_vars_sample.sh

I've also rebased as of 12:30PST.

Thanks!
-dave

dcwangmit01:
test this please

dcwangmit01:
test this please

dcwangmit01:
test this please

leseb:
@dcwangmit01 sorry for the pain and thanks for fixing issues that weren't yours.

Issue #754, opened 2016-05-04
Checking Firewall on Monitor Ports fails


dmesser:
Running with ansible 2.0.1.0 on RHEL7.2 the following tasks fails when check_firewall is set to true:

ceph.ceph-common: check if monitor port is not filtered
fatal: [ceph-node-a-1]: FAILED! => {"failed": true, "msg": "'ansible.vars.hostvars.HostVars object' has no attribute u'groups.mons'"}
fatal: [ceph-node-a-2]: FAILED! => {"failed": true, "msg": "'ansible.vars.hostvars.HostVars object' has no attribute u'groups.mons'"}
fatal: [ceph-node-a-3]: FAILED! => {"failed": true, "msg": "'ansible.vars.hostvars.HostVars object' has no attribute u'groups.mons'"}

Disabling check_firewall avoids the issue.

leseb:
do you mind giving this one another try?

ksingh7:
@leseb hitting this 

```
TASK [ceph.ceph-common : include] **********************************************
included: /usr/share/ceph-ansible/roles/ceph-common/tasks/./checks/check_firewall.yml for rhcs2, rhcs1, rhcs3

TASK [ceph.ceph-common : check if nmap is installed] ***************************
ok: [rhcs1]
ok: [rhcs3]
ok: [rhcs2]

TASK [ceph.ceph-common : inform that nmap is not present] **********************
skipping: [rhcs2]
skipping: [rhcs3]
skipping: [rhcs1]

TASK [ceph.ceph-common : check if monitor port is not filtered] ****************
fatal: [rhcs3]: FAILED! => {"failed": true, "msg": "'ansible.vars.hostvars.HostVars object' has no attribute u'groups.mons'"}
fatal: [rhcs1]: FAILED! => {"failed": true, "msg": "'ansible.vars.hostvars.HostVars object' has no attribute u'groups.mons'"}
fatal: [rhcs2]: FAILED! => {"failed": true, "msg": "'ansible.vars.hostvars.HostVars object' has no attribute u'groups.mons'"}
	to retry, use: --limit @site.retry

PLAY RECAP *********************************************************************
rhcs1                      : ok=5    changed=0    unreachable=0    failed=1
rhcs2                      : ok=5    changed=0    unreachable=0    failed=1
rhcs3                      : ok=5    changed=0    unreachable=0    failed=1

```
Temp fix check_firewall: false

leseb:
[...]

PR #844, opened 2016-06-04
Various fixes for check_firewall task

matthewrees:
The check to see if nmap is available was being run on the remote host, whereas the checks to see if the applicable Ceph ports were open were being run as local_actions.

Also included are various related fixes now that the above functionality has been corrected (tested on Ansible 2.x only).

Closes: https://github.com/ceph/ceph-ansible/issues/754

### DIFF
--- a/roles/ceph-common/tasks/checks/check_firewall.yml
+++ b/roles/ceph-common/tasks/checks/check_firewall.yml
@@ -1,6 +1,6 @@
 ---
 - name: check if nmap is installed
-  command: "command -v nmap"
+  local_action: shell command -v nmap
   changed_when: false
   failed_when: false
   register: nmapexist
@@ -17,7 +17,7 @@
   local_action: shell set -o pipefail && nmap -p 6789 {{ item }} {{ hostvars[item]['ansible_' + monitor_interface]['ipv4']['address'] }} | grep -sqo filtered
   changed_when: false
   failed_when: false
-  with_items: groups.{{ mon_group_name }}
+  with_items: "{{ groups[mon_group_name] }}"
   register: monportstate
   when:
     - check_firewall
@@ -30,16 +30,16 @@
   with_items: monportstate.results
   when:
     - check_firewall
-    - item.has_key('rc') and item.rc == 0
+    - item is defined and item.has_key('rc') and item.rc == 0
     - mon_group_name is defined
     - mon_group_name in group_names
     - nmapexist.rc == 0
 
-- name: check if osd and mds range is not filtered
+- name: check if osd and mds range is not filtered (osd hosts)
   local_action: shell set -o pipefail && nmap -p 6800-7300 {{ item }} {{ hostvars[item]['ansible_default_ipv4']['address'] }} | grep -sqo filtered
   changed_when: false
   failed_when: false
-  with_items: groups.{{ osd_group_name }}
+  with_items: "{{ groups[osd_group_name] }}"
   register: osdrangestate
   when:
     - check_firewall
@@ -52,16 +52,16 @@
   with_items: osdrangestate.results
   when:
     - check_firewall
-    - item.has_key('rc') and item.rc == 0
+    - item is defined and item.has_key('rc') and item.rc == 0
     - osd_group_name is defined
     - osd_group_name in group_names
     - nmapexist.rc == 0
 
-- name: check if osd and mds range is not filtered
+- name: check if osd and mds range is not filtered (mds hosts)
   local_action: shell set -o pipefail && nmap -p 6800-7300 {{ item }} {{ hostvars[item]['ansible_default_ipv4']['address'] }} | grep -sqo filtered
   changed_when: false
   failed_when: false
-  with_items: groups.{{ mds_group_name }}
+  with_items: "{{ groups[mds_group_name] }}"
   register: mdsrangestate
   when:
     - check_firewall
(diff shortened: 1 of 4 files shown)


roles/ceph-common/tasks/checks/check_firewall.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: check if nmap is installed
-  command: "command -v nmap"
+  local_action: shell command -v nmap
```
leseb: using the `command` module should work here

leseb:
test this please

roles/ceph-common/tasks/checks/check_firewall.yml
matthewrees: My first draft of this PR used the command module and it was failing, hence my use of shell instead. Testing it again now shows that it works...I can only blame myself :(

leseb: It's fine thanks ;)

leseb:
test this please

ASBishop:
I'm getting the following error (cntl0 is one of my mon hosts)

TASK [ceph.ceph-common : check if monitor port is not filtered] ****************
fatal: [cntl0]: FAILED! => {"failed": true, "msg": "'dict object' has no attribute u'ansible_interface'"}

ASBishop:
My environment:

% ansible-playbook --version
ansible-playbook 2.0.2.0

matthewrees:
@ASBishop What is the value of your "monitor_interface" variable, as it looks like it is set to "interface" based on the output you provided, ie: the attribute it is referring to should look look something like "ansible_eth0" as opposed to "ansible_interface"

ASBishop:
@matthewrees I define monitor_address and do not define monitor_interface.

matthewrees:
@ASBishop would you mind giving it another try with the latest commit?

matthewrees:
@ASBishop Bearing in mind that you might need to set check_firewall to true manually now as we have disabled it by default. Thanks!

ASBishop:
@matthewrees Sorry, but now I get a similar error, but with the monitor_address. 

TASK [ceph.ceph-common : check if monitor port is not filtered] ****************
fatal: [cntl0]: FAILED! => {"failed": true, "msg": "'dict object' has no attribute u'A.B.C.D'"}

where A.B.C.D is the actual address of the host.

matthewrees:
@ASBishop Sorry, silly mistake, hopefully this will be the last time!

ASBishop:
@matthewrees OK, the syntax problem is fixed, but it exposed two new problems.

I have 3 mons, and it looks like the with_items:  "{{ groups[mon_group_name] }}" causes it to generate a local_action command string that always references the host name and address of the first mon. In my test environment, I manually closed the first mon's port, and left it open on my other two mons. I expected the first mon host to fail and the second two to pass.

For example, if I my mons are moe, larry and curly,  the "failed" output message will report this:

failed: [moe] ...
failed:[larry] ...
failed:[curly] ...

But the command string will show that nmap was always testing moe and moe's IP address.

The second problem is I think nmap needs to run as root. I use a non-root local account, and if I manually run nmap the command will fail ("Host seems down" but the error message does not contain the word "filtered". This will cause a false positive. The test will always fail, but will never be detected as a failure.

ASBishop:
Sorry, ignore my comment about the second problem (nmap needs to run as root). While that appears to be true when I run it manually, it does not causes a false positive or else the tasks would succeed.

matthewrees:
@ASBishop I've made some fixes to how nmap is called and filtered, and have also tested the firewall checks against monitor_address instead of monitor_interface. Could you test it again and let me know?

PS. If you are an IRC user you are welcome to reach me in the Ceph IRC channel as the user "secate" if you'd like to debug this in real time.

ASBishop:
@matthewrees Progress! It now correctly probes all of the hosts (not just the first). However, the combination of using with_items to loop over the hosts and run_once to only perform the loop once causes any errors to be reported against the first host (the lucky run_once host). Using my previous example, let's say moe's and curly's port is open but larry's port is not. The test will properly fail, but it will claim it was moe who failed.

With this in mind, I humbly offer the following alternative that uses inventory_hostname. It eliminates the need for with_items and run_once.

- name: check if monitor port is not filtered
  local_action: shell set -o pipefail && nmap -p 6789 {{ hostvars[inventory_hostname]['ansible_' + monitor_interface]['ipv4']['address'] if hostvars[inventory_hostname]['ansible_' + monitor_interface] is defined else hostvars[inventory_hostname]['monitor_address'] }} | grep -sq -e filtered -e '0 hosts up'
  changed_when: false
  failed_when: false
  register: monportstate
  when:
    - check_firewall
    - nmapexist.rc == 0

- name: fail if monitor port is filtered
  fail:
      msg: "Please allow port 6789 on your firewall"
  when:
    - check_firewall
    - nmapexist.rc == 0
    - monportstate.rc == 0

Obviously the same pattern would need to be used for the other port tests.

I currently do not have IRC access, but will look for you when I get it back.

matthewrees:
@ASBishop Thanks! I should have noticed a cleanup was possible after all the changes, done.

ASBishop:
@matthewrees Whoops, there's a typo in the shell command that probes the monitor port (6689 should be 6789).

ASBishop:
Huzzah! Everything works for me now. I did just a quick test on mons, osds and rgws (sorry, no mds) and things behave as I expected. No false positives or false negatives.

leseb:
test this please

leseb:
Thanks guys, nice collaboration! Let's see what the CI has to say about this and then we can merge this :)

Issue #845, opened 2016-06-08
Option to enable ntp


font:
After the ansible playbook completed successfully a `ceph -s` indicated a `HEALTH_WARN` due to a clock skew being detected on the mon instances. This warning was valid as ceph detected a clock skew larger than the default threshold of 0.05 seconds:

`[root@ceph-mon0 ~]# ceph health detail`
`HEALTH_WARN clock skew detected on mon.ceph-mon1, mon.ceph-mon2; Monitor clock skew detected`
`mon.ceph-mon1 addr 192.168.121.156:6789/0 clock skew 0.318922s > max 0.05s (latency 0.000729135s)`
`mon.ceph-mon2 addr 192.168.121.157:6789/0 clock skew 0.240514s > max 0.05s (latency 0.000675196s)`

I noticed that the ntp package is installed as part of the playbook, but ntpd is not enabled nor started. Once I started ntpd all was well.

Can I add an option to the playbook to enable/start the ntp service?

leseb:
Yes you can, please submit a PR :)
Thanks!

maxtortime:
I think only one server syncs with ntp server and another servers follow synced server.
But I don't know fixing what code.

leseb:
This is a bit too much for this repository, remember we only deploy Ceph and don't take over the world :).
In your infrastructure you should have a NTP playbook that configures a NTP server and followers.

The least we can do in our repo is making sure NTP is started on all the node.

mattt416:
Agreed -- I'd prefer to see the enablement of ntp left out of this repo.

PR #848, opened 2016-06-15
Add option to enable ntp

font:
This fixes: #845 only for non-containerized deployments

##### Questions/Comments

1. As it stands we install ntp on all nodes and this will enable it on all nodes as well. Is it necessary on all nodes?
2. I filed a feature idea on ansible's packaging modules in order to simplify the task of querying for the state of a package across different package managers. It would've helped simplify this code.
3. Do we want to install and enable ntp on container hosts? If so, I will enable the containerized part pending the answer to question 1.

### DIFF
--- a/roles/ceph-common/tasks/checks/check_ntp_debian.yml
+++ b/roles/ceph-common/tasks/checks/check_ntp_debian.yml
@@ -1,0 +1,7 @@
+---
+- name: check ntp installation on debian
+  command: dpkg -s ntp
+  register: ntp_pkg_query
+  ignore_errors: true
+  changed_when: false
+  when: ansible_os_family == 'Debian'
--- a/roles/ceph-common/tasks/checks/check_ntp_redhat.yml
+++ b/roles/ceph-common/tasks/checks/check_ntp_redhat.yml
@@ -1,0 +1,7 @@
+---
+- name: check ntp installation on redhat
+  command: rpm -q ntp
+  register: ntp_pkg_query
+  ignore_errors: true
+  changed_when: false
+  when: ansible_os_family == 'RedHat'
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -71,6 +71,16 @@
   tags:
     - package-install
 
+- include: ./misc/ntp_redhat.yml
+  when:
+    - ansible_os_family == 'RedHat'
+    - ntp_service_enabled
+
+- include: ./misc/ntp_debian.yml
+  when:
+    - ansible_os_family == 'Debian'
+    - ntp_service_enabled
+
 - include: facts.yml
 
 - set_fact:
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -64,6 +64,9 @@ dummy:
 #  - ntp
 #  - python-setuptools
 
+# Enable the ntp service by default to avoid clock skew on
+# ceph nodes
+#ntp_service_enabled: true
 
 # The list of ceph packages needed for debian.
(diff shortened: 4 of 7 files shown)


roles/ceph-common/tasks/checks/check_ntp_debian.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: check ntp installation on debian
+  shell: dpkg -s ntp
```
leseb: please use `command` module.

roles/ceph-common/tasks/checks/check_ntp_redhat.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: check ntp installation on redhat
+  shell: rpm -q ntp
```
leseb: please use `command` module.

leseb:
@font Answers to your questions:

1. Well since I don't want to add a NTP role into this repo this is the only thing we can do for now.
Enabling NTP on all the machines is the only thing we can do to avoid errors.
If people have different env, then will setup for own time servers/config if necessary.

2. Can you share your feature idea? 

3. For containers, I think the general recommendation is to `-v /etc/localtime:/etc/localtime:ro`.
Of course NTP runs on the host :)
What do you think?

roles/ceph-common/tasks/main.yml
```diff
@@ -71,6 +71,16 @@
   tags:
     - package-install
 
+- include: ./misc/ntp_redhat.yml
+  when:
+      - ansible_os_family == 'RedHat'
```
leseb: please use a 2 space indent

roles/ceph-common/tasks/main.yml
```diff
@@ -71,6 +71,16 @@
...
+      - ntp_service_enabled
+
+- include: ./misc/ntp_debian.yml
+  when:
+      - ansible_os_family == 'Debian'
```
leseb: ditto

font:
@leseb Follow-up comments to your answers:

1. I don't think we necessarily need to add an NTP role to support per-node-type installation and starting of the ntp service. We could just apply the ntp installation and starting tasks to the respective roles i.e. ceph-mon, ceph-osd, etc. instead of in ceph-common. What would you prefer? I'll leave it up to you.
2. Feature idea: https://github.com/ansible/ansible-modules-core/issues/3917
3. Yes, we'd install and run NTP on the host for sure. I will work on that. As far as exporting /etc/localtime, it seems like you'd want the timezone of the host synced to the container to allow the user to define whatever timezone they'd like propagated to the container. For distributed systems, I'm always inclined to have one consistent UTC timezone across all nodes so I think just installing and starting ntp on the host would suffice. But that assumes everybody would want their hosts using UTC as their timezone which is probably not the case. Do you agree? If so, I will also work on mounting the /etc/localtime volume as read-only for the container.

leseb:
@font My answers:

1. Yup this is exactly what we need to do. Just install NTP from ceph-common and start it. There is no need to do that on each role since they all depend on ceph-common

2. I like this idea

3. Syncing the time, even for containers is crucial, so first let's make sure the hosts have the right time (and hopefully the same for all of them in the end). Then after this let's sync the time in the container using host's value.

leseb:
test this please

leseb:
test this please

font:
@leseb Can we merge this since it works for non-containerized deployments? I will create a new PR for the containerized portion when it is ready.

leseb:
@font yes sure sorry I was waiting for jenkins to run again :)

PR #851, opened 2016-06-16
Enable rhcs repos according to the released version

alfredodeza:
Similar to what it was done for `ice` versions. This will help for the upcoming new version, enabling the repositories for the configured version (defaulting to 1.3)

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1346812

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -135,6 +135,8 @@ ceph_stable_ice_kmod: 3.10-0.1.20140702gitdc9ac62.el7.x86_64
 # on RHEL 7.
 #
 ceph_stable_rh_storage: false
+# This will affect how/what repositories are enabled depending on the desired
+# version. The next version will use "2" not "2.0" which would not work.
 ceph_stable_rh_storage_version: 1.3 # next version is 2
 ceph_stable_rh_storage_cdn_install: false # assumes all the nodes can connect to cdn.redhat.com
 ceph_stable_rh_storage_iso_install: false # usually used when nodes don't have access to cdn.redhat.com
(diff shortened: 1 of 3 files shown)


ktdreyer:
:+1: LGTM (untested)

leseb:
@alfredodeza `roles/ceph-common/defaults/main.yml` and ` group_vars/all.sample` are not identical. Please update `roles/ceph-common/defaults/main.yml` and then run `generate_group_vars_sample.sh`.

roles/ceph-common/defaults/main.yml
```diff
@@ -135,6 +135,7 @@ ceph_stable_ice_kmod: 3.10-0.1.20140702gitdc9ac62.el7.x86_64
 # on RHEL 7.
 #
 ceph_stable_rh_storage: false
+ceph_stable_rh_storage_version: 1.3 # next version is 2
```
leseb: perhaps we can add a link to the redhat doc to see all the versions available.

alfredodeza: The problem here is that "1.3" will cover "1.3.1" as well as "1.3.2" because it is the same repo. Adding a link that shows all versions available (e.g. https://access.redhat.com/documentation/en/red-hat-ceph-storage/ ) will make this break if someone goes strictly by released version.

roles/ceph-common/defaults/main.yml
```diff
@@ -135,6 +135,9 @@ ceph_stable_ice_kmod: 3.10-0.1.20140702gitdc9ac62.el7.x86_64
 # on RHEL 7.
 #
 ceph_stable_rh_storage: false
+# This will affect how/what repositories are enabled depending on the desired
+# version. The next version will use "2" not "2.0" which would not work.
+#ceph_stable_rh_storage_version: 1.3 # next version is 2
```
leseb: it should be uncommented otherwise they won't be declared :)

andrewschoen:
test this please

PR #853, opened 2016-06-20
ceph-agent: install the correct package for the agent in Ubuntu

alfredodeza:
See: https://bugzilla.redhat.com/show_bug.cgi?id=1348130

### DIFF
--- a/roles/ceph-agent/tasks/pre_requisite.yml
+++ b/roles/ceph-agent/tasks/pre_requisite.yml
@@ -7,7 +7,7 @@
 
 - name: install dependencies
   apt:
-    pkg: calamari-server
+    pkg: rhscon-agent
     state: present
   when: ansible_os_family == 'Debian'
   tags:


cdwertmann:
ceph-ansible now fails on my Ubuntu 14.04 hosts. Where can I get the ```rhscon-agent``` package for Ubuntu?

leseb:
@cdwertmann are you installing the ceph-agent role?

alfredodeza:
@cdwertmann why are you trying to install the agent? :) There is no public access to this package yet, and you probably don't need it for Ceph functionality.

ktdreyer:
@cdwertmann I'm sorry there is no "rhscon-agent" package available publicly. There is an equivalent of an early Ubuntu Xenial package of the software, though, here: https://launchpad.net/~kdreyer-redhat/+archive/ubuntu/skynet . The big difference is that the package is named "skynet", not "rhscon-agent".

cdwertmann:
Thank you for your replies. It seems that I don't need the storage agent to run ceph, so I'm skipping that role now.

PR #858, opened 2016-06-22
Add a custom yum repo ability

dang:
Add the ability to use a custom yum repo, rather than just upstream,
RHEL, and distro.  This allows ansible to be used for internal testing.

### DIFF
--- a/roles/ceph-common/tasks/installs/debian_ceph_repository.yml
+++ b/roles/ceph-common/tasks/installs/debian_ceph_repository.yml
@@ -49,3 +49,11 @@
     state: present
   changed_when: false
   when: ceph_stable_uca
+
+- name: add custom repo
+  apt_repository:
+    repo: "deb {{ ceph_custom_repo }} {{ ansible_lsb.codename }} main"
+    state: present
+  changed_when: false
+  when: ceph_custom
+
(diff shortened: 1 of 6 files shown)


leseb:
@dang please add Debian support as well.
I think we can also remove https://github.com/ceph/ceph-ansible/blob/master/localrepo-site.yml.sample since it does the same thing.

dang:
I'm not currently equipped to add debian support (no machines I can test on, extremely low priority in my day job, never set up a custom debian repo before) so it'll be a while before I can add that.

leseb:
It's not different from what you have with RedHat. Just add the same `get_url` in `roles/ceph-common/tasks/installs/debian_ceph_repository.yml` and make it point to `/etc/apt/sources.list.d/`.

roles/ceph-common/tasks/installs/debian_ceph_repository.yml
```diff
@@ -49,3 +49,11 @@
     state: present
   changed_when: false
   when: ceph_stable_uca
+
+- name: add custom repo
+  apt_repository:
+    repo: "deb {{ ceph_custom_repo }} {{ ansible_lsb.codename }} main"
+    state: present
+  changed_when: false
+  when: ceph_dev
```
leseb: ceph_custom

dang:
Oops.

leseb:
test this please
 Hide all checks

PR #859, opened 2016-06-22
Add DNF for RedHat

dang:


leseb:
test this please

roles/ceph-common/tasks/installs/install_on_redhat.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- name: install dependencies
+- name: install redhat dependencies yum
```
leseb: perhaps we can stay consistent with your last change and keep using "via yum"

same goes for the next tasks.

dang:
Ping

PR #862, opened 2016-06-24
Copies admin keyring to gateway node

smanjara:
Fixes:
 https://bugzilla.redhat.com/show_bug.cgi?id=1330835


leseb:
👎  just enable `copy_admin_key` and the key will be copied to `/etc/ceph/`.

PR #865, opened 2016-06-28
Fixing containerized deployment with kv

PiotrProkop:
Adding option to change etcd port and docker image tag. Removing symbolic links, fixing enabling of systemd unit files. Migrating from docker module to shell module to ensure osd disk prepare finishes before launching next task.

### DIFF
--- a/roles/ceph-osd/defaults/main.yml
+++ b/roles/ceph-osd/defaults/main.yml
@@ -123,6 +123,7 @@ osd_containerized_deployment: false
 osd_containerized_deployment_with_kv: false
 kv_type: etcd
 kv_endpoint: 127.0.0.1
+kv_port: 4001
 ceph_osd_docker_prepare_env: "OSD_FORCE_ZAP=1"
 ceph_osd_docker_username: ceph
 ceph_osd_docker_imagename: daemon
--- a/roles/ceph-osd/tasks/docker/start_docker_osd.yml
+++ b/roles/ceph-osd/tasks/docker/start_docker_osd.yml
@@ -99,7 +99,7 @@
 
 - name: run the ceph osd docker image
   docker:
-    image: "{{ ceph_osd_docker_username }}/{{ ceph_osd_docker_imagename }}:{{ ceph_osd_docker_tag }}"
+    image: "{{ ceph_osd_docker_username }}/{{ ceph_osd_docker_imagename }}:{{ ceph_osd_docker_image_tag }}"
     name: "{{ ansible_hostname }}-osd-{{ item | regex_replace('/', '') }}"
     net: host
     pid: host
@@ -115,7 +115,7 @@
 
 - name: run the ceph osd docker image with kv
   docker:
-    image: "{{ ceph_osd_docker_username }}/{{ ceph_osd_docker_imagename }}:{{ ceph_osd_docker_tag }}"
+    image: "{{ ceph_osd_docker_username }}/{{ ceph_osd_docker_imagename }}:{{ ceph_osd_docker_image_tag }}"
     name: "{{ ansible_hostname }}-osd-{{ item | regex_replace('/', '') }}"
     net: host
     pid: host
--- a/roles/ceph-osd/tasks/docker/start_docker_osd.yml
+++ b/roles/ceph-osd/tasks/docker/start_docker_osd.yml
@@ -42,7 +42,7 @@
 
 - name: prepare ceph osd disk with kv_store
   shell: |
-    docker run  --net=host --rm  \
+    docker run  --net=host   \
     --pid=host \
     --privileged=true \
     --name="{{ ansible_hostname }}-osd-prepare-{{ item.0 |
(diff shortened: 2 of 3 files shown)


roles/ceph-osd/defaults/main.yml
```diff
@@ -123,9 +123,11 @@ osd_containerized_deployment: false
 osd_containerized_deployment_with_kv: false
 kv_type: etcd
 kv_endpoint: 127.0.0.1
+kv_port: 4001
 ceph_osd_docker_prepare_env: "OSD_FORCE_ZAP=1"
 ceph_osd_docker_username: ceph
 ceph_osd_docker_imagename: daemon
+ceph_osd_docker_tag: latest
```
dang: I'm not sure tag needs a separate config field; it can just be included in the imagename.  I do this regularly, and it works fine.  (FWIW, I also don't think imagename needs to be separate from username...)

roles/ceph-osd/tasks/docker/start_docker_osd.yml
```diff
@@ -20,7 +20,7 @@
 # starting the next task
 - name: prepare ceph osd disk
   shell: |   
-    docker run  --net=host \
+    docker run  --net=host --rm \
```
dang: It's been useful for me during development to leave this behind, so I can view logs if it fails.  I can also see that it's useful to make it go away, since this is a one-time action on a potentially long time machine.  I'm not sure what the solutions is, since I'm not convinced that adding *another* config knob is correct.  So I guess I'm asking: how useful do people find both situations: removing it and not removing it?

roles/ceph-osd/tasks/docker/start_docker_osd.yml
```diff
@@ -40,17 +40,24 @@
...
-    image: "{{ ceph_osd_docker_username }}/{{ ceph_osd_docker_imagename }}"
-    name: "{{ ansible_hostname }}-osd-prepare-{{ item | regex_replace('/', '') }}"
-    net: host
-    pid: host
-    state: running
-    privileged: yes
-    volumes: "{{ item.0 }}:{{ item.0 }}"
-    env: "OSD_DEVICE={{ item.0 }},KV_TYPE={{kv_type}},KV_IP={{kv_endpoint}},{{ ceph_osd_docker_prepare_env }}"
-  ignore_errors: true
+  shell: |
```
dang: This should be command, if possible, and not shell.

roles/ceph-osd/tasks/docker/start_docker_osd.yml
```diff
@@ -64,23 +71,15 @@
   become: true
   template:
     src: "{{ role_path }}/templates/ceph-osd.service.j2"
-    dest: /var/lib/ceph/ceph-osd@.service
+    dest: /etc/systemd/system/ceph-osd@.service
```
dang: This won't work for multiple OSDs.  That's why it was set up to copy once and symlink.  If you want to copy instead of symlink, you will need to copy once per OSD and use the naming scheme from the symlink task, I think.

roles/ceph-osd/tasks/docker/start_docker_osd.yml
```diff
@@ -117,18 +116,3 @@
     - ansible_os_family != 'CoreOS'
     - not osd_containerized_deployment_with_kv
```
dang: Here you're completely losing the ability to ron on !redhat !coreos.  Is that intentional?

roles/ceph-osd/tasks/docker/start_docker_osd.yml
PiotrProkop: It will worj=k that is why this unit file is templated with ceph-osd@.service. When you do systemctl enable then systemd automatically makes those symlinks.

roles/ceph-osd/tasks/docker/start_docker_osd.yml
PiotrProkop: My bad just deleted too much.

roles/ceph-osd/defaults/main.yml
PiotrProkop: I think one should try to make things as clear as possible and separating each part of the docker images is such thing.

roles/ceph-osd/tasks/docker/start_docker_osd.yml
leseb: I'd like to keep it for debug purpose. If this fails, there is no way to tell what exactly failed.
Agree with @dang's initial thoughts

PiotrProkop:
@leseb removed --rm.

roles/ceph-osd/tasks/docker/start_docker_osd.yml
leseb: @PiotrProkop can you please address this comment or just respond to it? thanks!

PiotrProkop: @leseb i just used convention which was present in this role. I can change it to command if you want.

leseb:
@PiotrProkop would you mind rebasing this? Thanks!

PiotrProkop:
@leseb done

PR #866, opened 2016-06-28
Adding option to change kv_port and docker tag

PiotrProkop:
And removing symbolic links and fixing systemctl enable on CoreOS.

### DIFF
--- a/roles/ceph-mon/defaults/main.yml
+++ b/roles/ceph-mon/defaults/main.yml
@@ -69,6 +69,7 @@ ceph_mon_docker_interface: eth0
 #ceph_mon_docker_subnet: # subnet of the ceph_mon_docker_interface
 ceph_mon_docker_username: ceph
 ceph_mon_docker_imagename: daemon
+ceph_osd_docker_tag: latest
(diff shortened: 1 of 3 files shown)


roles/ceph-mon/defaults/main.yml
```diff
@@ -69,6 +69,7 @@ ceph_mon_docker_interface: eth0
 #ceph_mon_docker_subnet: # subnet of the ceph_mon_docker_interface
 ceph_mon_docker_username: ceph
 ceph_mon_docker_imagename: daemon
+ceph_osd_docker_tag: latest
```
leseb: Why osd and not mon? I mean you're modifying the mon role so I was expecting to see this variable being named ceph_mon_docker_tag instead

leseb:
@rootfs could you please have a look at the systems stuff? Thanks!

roles/ceph-mon/defaults/main.yml
PiotrProkop: Yeah sure it was a typo.

PR #871, opened 2016-07-06
ceph-common: ensure that destination iso paths exist before copying

alfredodeza:
Fixes issue: https://bugzilla.redhat.com/show_bug.cgi?id=1338551


ktdreyer:
LGTM :+1:

ksingh7:
👍  much needed

PR #875, opened 2016-07-11
ceph-common: update names for local RHCS repos

ktdreyer:
Use more descriptive names for the local RHCS Yum repositories.

### DIFF
--- a/roles/ceph-common/templates/redhat_storage_repo.j2
+++ b/roles/ceph-common/templates/redhat_storage_repo.j2
@@ -1,6 +1,6 @@
 # {{ ansible_managed }}
 [rh_storage_mon]
-name=Red Hat Ceph Storage - local packages for Ceph
+name=Red Hat Ceph Storage - local packages for Ceph monitor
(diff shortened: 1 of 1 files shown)


roles/ceph-common/templates/redhat_storage_repo.j2
```diff
@@ -1,22 +1,22 @@
...
+name=Red Hat Ceph Storage - local packages for Ceph OSD
 baseurl=file://{{ ceph_stable_rh_storage_repository_path }}/OSD
 enabled=1
 gpgcheck=1
 type=rpm-md
 priority=1
 
 [rh_storage_tools]
-name=Red Hat Ceph Storage - assorted tools
+name=Red Hat Ceph Storage - local packages for Ceph client, MDS, and RGW
```
leseb: rbd-mirror too?

ktdreyer: rbd-mirror is in there, you're right. It was getting a little long so I didn't list everything :)

leseb: fine :)

PR #876, opened 2016-07-11
ceph-common: rm "type=rpm-md" from RHCS repos

ktdreyer:
As far as I can tell, this is a SUSE-ism, not needed for RHCS.


leseb:
It's something I picked up from one of our ISO install ^^

ktdreyer:
Yeah, apparently this has been copied & pasted in a bunch of places :( eg https://github.com/ceph/ceph/pull/10248

Issue #585, opened 2016-03-01
msg: Error when trying to enable ceph: rc=1 Failed to execute operation: No such file or directory


snerdish:
Getting the following in TASK: [ceph-osd | start and add that the OSD service to the init sequence] **** 

TASK: [ceph-osd | start and add that the OSD service to the init sequence] **** 
failed: [foo.bar] => {"failed": true}
msg: Error when trying to enable ceph: rc=1 Failed to execute operation: No such file or directory

Seem to be some issues with systemd unit file management atm.

leseb:
Yes, it looks like it cannot find the systemd unit files

leseb:
Oh I see you're using the directory scenario, can you check if the OSD gets started anyway?
If so, we can probably remove this task.

snerdish:
Looks like the OSD gets started.

PR #877, opened 2016-07-12
ceph-osd: Use the use_systemd fact when starting directory-scenario OSDs

aluminous:
Fixes an issue where the directory scenario of the ceph-osd role fails to detect systemd.

Previously, the "init_system" fact was used. In our deployments (CentOS 7), this fact contains a trailing newline. This change switches to the use_systemd fact. This appears to be more consistent with the rest of the project as well.

I believe this is the same issue as #585


leseb:
This trailing line issue should have been fixed by #867 but yes the change brings more consistency.

leseb:
@aluminous thanks!

PR #881, opened 2016-07-12
ceph-common: explicitly create the directory portion of the iso path

alfredodeza:
This will help if the path to the iso exists in the originating server but not
in the remote paths. This issue is not seen if using /tmp/file.iso but does
show up when using nested paths.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1355762

### DIFF
--- a/roles/ceph-common/tasks/pre_requisites/prerequisite_rh_storage_iso_install.yml
+++ b/roles/ceph-common/tasks/pre_requisites/prerequisite_rh_storage_iso_install.yml
@@ -6,7 +6,13 @@
   with_items:
     - "{{ ceph_stable_rh_storage_mount_path }}"
     - "{{ ceph_stable_rh_storage_repository_path }}"
-    - "{{ ceph_stable_rh_storage_iso_path }}"
+
+- name: ensure destination iso directory exists
+  file:
+    path: "{{ ceph_stable_rh_storage_iso_path | dirname }}"
+    state: directory
+    recurse: yes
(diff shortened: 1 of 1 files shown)


roles/ceph-common/tasks/pre_requisites/prerequisite_rh_storage_iso_install.yml
```diff
@@ -6,7 +6,13 @@
...
     - "{{ ceph_stable_rh_storage_mount_path }}"
     - "{{ ceph_stable_rh_storage_repository_path }}"
-    - "{{ ceph_stable_rh_storage_iso_path }}"
+
+- name: ensure destination iso directory exists
+  file:
+    path: "{{ ceph_stable_rh_storage_iso_path | dirname }}"
+    state: directory
+    recurse: yes
+  when: "'{{ ceph_stable_rh_storage_iso_path | dirname }}' != '/'"
```
leseb: Why not `/`? It's not usual but nothing prevents us to do so right?

alfredodeza: My inclination was not to even attempt it if not necessary. This is totally not an issue if the path is `/tmp/`  (but it was when the path was `/root/iso/`)

I don't mind removing the when clause if you feel it is redundant

leseb: No I'm fine with the change, thanks!

PR #883, opened 2016-07-13
Allow generating FSID for docker

dang:
The docker case wasn't able to generate an FSID, it required it to be
set.  Allow generating it.

The better solution (which is in my TODO list) is to factor these bits + config generation in ceph-common and have the docker case include the common case.  However, this is a downstream QE PR (https://bugzilla.redhat.com/show_bug.cgi?id=1356105) , so I'd like to get a fix in ASAP.


ktdreyer:
@dang who can review/merge this?

dang:
@ktdreyer I think only @leseb can merge.

PR #888, opened 2016-07-19
Added support for --check flag in ceph-common

khappone:
The ceph-common role fails when you run ansible with --check. Adding
always_run to a few tasks makes the check go through easier (although
it's not foolproof).


leseb:
Thanks @khappone

PR #892, opened 2016-07-20
Add ability to purge containerized Ceph cluster

font:
This removes containers, container images, packages, configuration files
and all the data. Removal of container image tasks are tagged with
'remove_img' to skip removal if desired.

### DIFF
--- a/purge-docker-cluster.yml
+++ b/purge-docker-cluster.yml
@@ -1,43 +1,447 @@
 ---
 # This playbook purges a containerized Ceph cluster
-# It removes: packages, configuration files and ALL THE DATA
+# It removes: packages, containers, configuration files and ALL THE DATA
 
-- hosts:
-  - mons
-  - osds
+- name: confirm whether user really meant to purge the cluster
+
+  hosts:
+    - localhost
+
+  vars_prompt:
+    - name: ireallymeanit
+      prompt: Are you sure you want to purge the cluster?
+      default: 'no'
+      private: no
 
   tasks:
+  - name: exit playbook, if user did not mean to purge cluster
+    fail:
+      msg: >
+        "Exiting purge-docker-cluster playbook, cluster was NOT purged.
+         To purge the cluster, either say 'yes' on the prompt or
+         or use `-e ireallymeanit=yes` on the command line when
+         invoking the playbook"
+    when: ireallymeanit != 'yes'
 
-  - name: install the latest version of gdisk
-    package:
-      name: gdisk
-      state: present
 
-  - name: collect ceph containers
-    shell: docker ps -aq --filter="ancestor=ceph/daemon"
-    register: containers
+- name: retrieve variables
 
-  - name: delete ceph containers
-    shell: docker rm -f {{ item }}
-    with_items: containers.stdout_lines
+  hosts:
+    - all
+    - localhost
 
-  - name: purge ceph directories
+  tasks:
+    - include_vars: roles/ceph-common/defaults/main.yml
+    - include_vars: roles/ceph-mon/defaults/main.yml
+    - include_vars: roles/ceph-osd/defaults/main.yml
+    - include_vars: roles/ceph-mds/defaults/main.yml
+    - include_vars: roles/ceph-rgw/defaults/main.yml
+    - include_vars: roles/ceph-nfs/defaults/main.yml
+    - include_vars: roles/ceph-restapi/defaults/main.yml
+    - include_vars: group_vars/all
+
+    - name: check if it is Atomic host
+      stat: path=/run/ostree-booted
+      register: stat_ostree
+
+    - name: set fact for using Atomic host
+      set_fact:
+        is_atomic: "{{ stat_ostree.stat.exists }}"
+
+
+- name: purge ceph mds cluster
+
+  hosts:
+    - mdss
+
+  become: true
+
+  tasks:
+  - name: disable ceph mds service
+    service:
+      name: "ceph-mds@{{ ansible_hostname }}"
+      state: stopped
+      enabled: no
+    ignore_errors: true
+
+  - name: remove ceph mds container
+    docker:
+      image: "{{ ceph_mds_docker_username }}/{{ ceph_mds_docker_imagename }}:{{ ceph_mds_docker_image_tag }}"
+      name: "{{ ansible_hostname }}"
+      state: absent
+    ignore_errors: true
+
+  - name: remove ceph mds service
+    file:
+        path: /etc/systemd/system/ceph-mds@.service
+        state: absent
+
+  - name: remove ceph mds image
+    docker_image:
+      state: absent
+      name: "{{ ceph_mds_docker_username }}/{{ ceph_mds_docker_imagename }}"
+      tag: "{{ ceph_mds_docker_image_tag }}"
+    tags:
+      remove_img
+
+- name: purge ceph rgw cluster
+
+  hosts:
+    - rgws
+
+  become: true
+
+  tasks:
+  - name: disable ceph rgw service
+    service:
+      name: "ceph-rgw@{{ ansible_hostname }}"
+      state: stopped
+      enabled: no
+    ignore_errors: true
+
+  - name: remove ceph rgw container
+    docker:
+      image: "{{ ceph_rgw_docker_username }}/{{ ceph_rgw_docker_imagename }}:{{ ceph_rgw_docker_image_tag }}"
+      name: "{{ ansible_hostname }}"
+      state: absent
+    ignore_errors: true
+
+  - name: remove ceph rgw service
+    file:
+        path: /etc/systemd/system/ceph-rgw@.service
+        state: absent
+
+  - name: remove ceph rgw image
+    docker_image:
+      state: absent
+      name: "{{ ceph_rgw_docker_username }}/{{ ceph_rgw_docker_imagename }}"
+      tag: "{{ ceph_rgw_docker_image_tag }}"
+    tags:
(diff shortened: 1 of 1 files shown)


purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
+         To purge the cluster, either say 'yes' on the prompt or
+         or use `-e ireallymeanit=yes` on the command line when
+         invoking the playbook"
+    when: ireallymeanit != 'yes'
+
+
+- name: retrieve variables
+
+  hosts:
+    - all
```
leseb: Using 'all' is a bit dangerous, this will get applied to the entire infrastructure. So I think we should restrain this to `mon_group_name`, `osd_group_name`, `mds_group_name` etc...

purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
+
+    - name: set fact for using Atomic host
+      set_fact:
+        is_atomic: "{{ stat_ostree.stat.exists }}"
+
+
+- name: purge ceph mds cluster
+
+  hosts:
+    - mdss
```
leseb: We should use `mds_group_name` because `mdss` might not exist.

purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
+      state: absent
+      name: "{{ ceph_mds_docker_username }}/{{ ceph_mds_docker_imagename }}"
+      tag: "{{ ceph_mds_docker_image_tag }}"
+    tags:
+      remove_img
+
+- name: purge ceph rgw cluster
+
+  hosts:
+    - rgws
```
leseb: ditto

purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
+
+  - name: remove ceph rgw image
+    docker_image:
+      state: absent
+      name: "{{ ceph_rgw_docker_username }}/{{ ceph_rgw_docker_imagename }}"
+      tag: "{{ ceph_rgw_docker_image_tag }}"
+    tags:
+      remove_img
+
+  - name: remove libnss3-tools on redhat
```
leseb: hum what the relationship with containers?

purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
-    package:
-      name: gdisk
-      state: present
+- name: purge ceph nfs cluster
 
-  - name: collect ceph containers
-    shell: docker ps -aq --filter="ancestor=ceph/daemon"
-    register: containers
+  hosts:
+    - nfss
```
leseb: `nfs_group_name`

purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
+      name: "{{ ceph_nfs_docker_username }}/{{ ceph_nfs_docker_imagename }}"
+      tag: "{{ ceph_nfs_docker_image_tag }}"
+    tags:
+      remove_img
+
+
+- name: purge ceph osd cluster
+
+  hosts:
+    - osds
```
leseb: `osd_group_name`

purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
+      name: "{{ ceph_osd_docker_username }}/{{ ceph_osd_docker_imagename }}"
+      tag: "{{ ceph_osd_docker_image_tag }}"
+    tags:
+      remove_img
+
+
+- name: purge ceph mon cluster
+
+  hosts:
+    - mons
```
leseb: `mon_group_name`

purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
+      name: "{{ ceph_mon_docker_username }}/{{ ceph_mon_docker_imagename }}"
+      tag: "{{ ceph_mon_docker_image_tag }}"
+    tags:
+      remove_img
+
+
+- name: remove installed packages
+
+  hosts:
+    - all
```
leseb: once again using `all` is dangerous :)

purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
+
+  tasks:
+  - name: stop docker service
+    service:
+      name: docker
+      state: stopped
+      enabled: no
+    when: not is_atomic
+
+  - name: remove pip and docker on ubuntu
```
leseb: There is no reasons to remove docker engine here, the purpose of this playbook should be to remove ceph containers and/or images. The machine can have more containers running and we don't really want to uninstall docker-engine :)

purge-docker-cluster.yml
```diff
@@ -1,43 +1,447 @@
...
+      state: absent
+    when:
+      ansible_version['full'] | version_compare('2.1.0.0', '>=') and
+      not is_atomic
+
+
+- name: purge ceph directories
+
+  hosts:
+    - all
```
leseb: ditto

purge-docker-cluster.yml
font: I wasn't sure exactly. It's just that in https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-rgw/tasks/main.yml#L5 we `include: openstack-keystone.yml` based on this `radosgw_keystone` conditional variable. If this isn't container specific we should probably have an additional `not rgw_containerized_deployment` check and I can remove it from this playbook.

purge-docker-cluster.yml
font: You're right - if we never installed it to begin with we definitely don't want to remove it. The problem is that we may have installed it to begin with so we'd be leaving extra stuff behind by not fully undoing all the steps we did. Unfortunately, we don't have an easy way to check if we installed it initially without leaving some sort of trail behind.

The idea is that we'd give the user the option of skipping these steps by passing the `--skip-tags=with_pkg` the way we do when we run the `ansible-playbook --skip-tags=with_pkg site-docker.yml` command. Unfavorably, it does require the user to know that these steps are going to happen and to use the tag to skip them.

On a side note: it'd be nice to have the opposite behavior supported in Ansible such that we _always_ skip these steps unless the user purposefully requests them. I'm thinking something like a `--no-skip-tags=with_pkg` option. I suppose the same thing could be done with a conditional that the user defines using `--extra-args`, but then you have to add the conditional check in each of these tasks unless you're using roles. I should look into seeing if anybody has made a feature request like this. Ok I looked and there is https://github.com/ansible/ansible/issues/11045 which would come in handy here.

Anyhow, what would you like to see happen here?

purge-docker-cluster.yml
font: Good point. I assumed the user would have only these ceph cluster hosts in their inventory file, but I know that's not the case. I will make the necessary changes.

font: Good point. I assumed the user would have only these ceph cluster hosts in their inventory file, but I know that's not the case. However, how do we know what the variables *_group_name are without including some variable files first? In other words, how do we restrict the hosts without knowing what the hosts are yet? Unless you are thinking we define the *_group_name variables in this file the way the purge-cluster.yml playbook does it. But that seems to be duplicating variables and restricting the user's flexibility.

purge-docker-cluster.yml
font: I will update this to use the respective variable. The reason I went this route is because I saw a similar thing being done in site.yml.sample and site-docker.yml.sample. It seems like we should update hosts in those playbooks to reference the respective `"{{ *_group_name }}"` variable as well.

leseb: That's true, we should change the `site.yml` to reflect that as well :)
I' can give it a try.

purge-docker-cluster.yml
leseb: This is not container specific so yes we should add this `when` condition.

purge-docker-cluster.yml
leseb: I understand your point, perhaps the easiest way to do it is to ask the user to remove the packages or not, so something like asking explicitly.
The problem is that we can not detect arguments passed in the command from within the playbook...

So I'd say just enforce this at the beginning, like a 2 steps check:

1. Are you sure you want to purge?
2. If `--skip-tags=with_pkg` is not set **docker packages and more** will uninstalled, do you want to continue?

This second one is tricky because the user might be set the `--skip-tags=with_pkg` already. However since we can run the playbook with `--extra-args` they can also skip this step by saying "yes i know what i'm doing with packages".

Voilà voilà.

purge-docker-cluster.yml
leseb: Indeed, I believe it would make sense at some point to have a dedicated role to purge cluster since purge playbook have been starting to have dependancies on ceph-common. Moreover we have two scenarios to purge (bare metal and containers) so it's worth it.

For now, let's stick with `*_group_name` even we don't know all of them in advance.

font:
@leseb thanks for your feedback. I believe I have addressed all comments. I have used the *_group_name but also had to define it in place as well. This at least gives the user the ability to override it on the command line to whatever group names they've chosen. This is probably the best we can do without adding roles to the purge playbooks.

In addition, I think the setup of this playbook should not make it difficult to split out each play into new roles in the future - perhaps something like ceph-mon-purge, ceph-osd-purge, ceph-mds-purge, etc.

leseb:
@font Perhaps instead of creating new roles, we could simply add the purge functionality inside the roles themselves. Just create a new var to activate this and that's all. What do you think?

font:
@leseb Yes I agree, that's probably the cleanest way similar to the way we're handling containerized deployment.

PR #900, opened 2016-07-25
purge-cluster: remove -q flag from grep to prevent broken pipes

alfredodeza:
See: https://bugzilla.redhat.com/show_bug.cgi?id=1339576

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -261,7 +261,7 @@
       rbdmirror_group_name in group_names
 
   - name: check for anything running ceph
-    shell: "ps awux | grep -v grep | grep -q -- ceph-"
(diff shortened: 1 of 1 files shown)


purge-cluster.yml
```diff
@@ -261,7 +261,7 @@
       rbdmirror_group_name in group_names
 
   - name: check for anything running ceph
-    shell: "ps awux | grep -v grep | grep -q -- ceph-"
+    shell: "ps awux | grep -v grep | grep -- ceph-"
```
leseb: in the meantime can we do `grep -- [c]eph-` to remove the previous `grep`?

leseb:
Sorry for the late merge.

PR #906, opened 2016-07-26
remove references to Inktank Ceph Storage

ktdreyer:
ICE is beyond end-of-life, and this Inktank GPG key is considered untrusted (https://access.redhat.com/blogs/766093/posts/2176181)

RIP ICE, we will not miss you.


leseb:
RIP ICE 😢

PR #908, opened 2016-07-28
Fix purge-cluster.yml fail by removing unnecessary ceph_stable_rh_storage_iso_install check.

jsaintrocc:
In the `purge rh_storage.repo file in /etc/yum.repos.d` task, there is a check for `ceph_stable_rh_storage_iso_install` which, in many cases, is undefined and will cause the playbook to fail.  Given that the task will only delete the repo if it is present there is no need to check `ceph_stable_rh_storage_iso_install`.

### DIFF
--- a/purge-cluster.yml
+++ b/purge-cluster.yml
@@ -451,5 +451,4 @@
       path: /etc/yum.repos.d/rh_storage.repo
       state: absent
     when:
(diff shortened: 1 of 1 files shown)


jsaintrocc:
retest this please

PR #909, opened 2016-07-28
Fix bad civetweb port in rgw frontends

jsaintrocc:
The civetweb config the ceph.conf.j2 incorrectly added `port=host:port` instead of just `port=port`.  RadosGW only works with `port=port`.


leseb:
@jsaintrocc hum it used to work with `host:port` the goal was to have a specific IP for rgw to listen on.

jsaintrocc:
Seems to work but radosgw isn't happy about it.  However looking at the [civetweb documentation](https://github.com/civetweb/civetweb/blob/master/docs/UserManual.md) this is valid syntax.  I'll put in a bug with the main ceph project to resolve on their end.

```
[root@pistorergw2-cc38-h02 ceph]# ceph --version
ceph version 10.2.2 (45107e21c568dd033c2f0a3107dec8f0b0e58374)
[root@pistorergw2-cc38-h02 ceph]# /usr/bin/radosgw -f --cluster ceph --name client.rgw.pistorergw2-cc38-h02 --setuser ceph --setgroup ceph
error parsing int: 10.251.1.163:8080: The option value '10.251.1.163:8080' seems to be invalid
```

jsaintrocc:
Ah and for reference I commented on an already filed bug for rgw [Here](http://tracker.ceph.com/issues/16581)

leseb:
Thanks @jsaintrocc!

PR #910, opened 2016-07-28
Add option to enable ntp

### LINKED ISSUE #845: Option to enable ntp

font:
This fixes #845 for containerized deployments. We now also mount the
/etc/localtime volume in the containers in order to synchronize the host
timezone with the container timezone.

### DIFF
--- a/roles/ceph-common/tasks/checks/check_ntp_atomic.yml
+++ b/roles/ceph-common/tasks/checks/check_ntp_atomic.yml
@@ -4,4 +4,3 @@
   register: ntp_pkg_query
   ignore_errors: true
   changed_when: false
-  when: ansible_os_family == 'RedHat'
--- a/roles/ceph-common/tasks/checks/check_ntp_atomic.yml
+++ b/roles/ceph-common/tasks/checks/check_ntp_atomic.yml
@@ -1,0 +1,7 @@
+---
+- name: check ntp installation on atomic
+  command: rpm -q chrony
+  register: ntp_pkg_query
+  ignore_errors: true
+  changed_when: false
+  when: ansible_os_family == 'RedHat'
--- a/roles/ceph-common/tasks/misc/ntp_atomic.yml
+++ b/roles/ceph-common/tasks/misc/ntp_atomic.yml
@@ -1,6 +1,6 @@
 ---
 - include: ../checks/check_ntp_atomic.yml
-  when: ansible_os_family == 'RedHat'
+  when: is_atomic
 
 - name: start the ntp service
   service:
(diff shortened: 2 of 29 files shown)


roles/ceph-common/tasks/checks/check_ntp_atomic.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: check ntp installation on atomic
+  command: rpm -q chrony
+  register: ntp_pkg_query
+  ignore_errors: true
+  changed_when: false
+  when: ansible_os_family == 'RedHat'
```
leseb: The `when` is a bit redundant with the `when` in the include.
Moreover, shouldn't we check for Atomic specifically?

font: The is_atomic is checked before including ntp_atomic.yml->check_ntp_atomic.yml so that should be covered there. I wanted to avoid duplicating the is_atomic register that is already done in the role's .../docker/main.yml. I thought about passing in the is_atomic variable but ultimately opted for this instead. The when check here is in case anybody wants to include this file individually.

If you'd like, I could see about just using is_atomic alone either by requiring callers to pass it in, or registering it here as well despite it being duplicated.

font:
test this please

roles/ceph-common/tasks/checks/check_ntp_atomic.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: check ntp installation on atomic
+  command: rpm -q chrony
+  register: ntp_pkg_query
+  ignore_errors: true
+  changed_when: false
+  when: is_atomic
```
leseb: the `when` is not needed here since there is already one on the `include` below.

font: I left it in here in case anybody in the future included this file by itself. By the same logic, the `include` below is not needed because it is already checked before ntp_atomic.yml is included. I can remove it.

PR #916, opened 2016-08-02
ceph-common: client settings are for libvirt

ktdreyer:
Prior to this change, each ceph cluster node would end up with several `qemu-client-$pid.log` files owned by root. The `[client]` section would capture *all* client activity (for example the "`ceph health`" command, etc), not just librbd-in-qemu.

Restrict this section to libvirt clients only so that we don't generate these spurious log files for other Ceph client traffic.


ktdreyer:
See the corresponding change in ceph's docs, https://github.com/ceph/ceph/pull/10542

leseb:
Thanks @ktdreyer

ktdreyer:
Thank you!

PR #922, opened 2016-08-09
[WIP] Initial support for better handlers

leseb:
The idea is to not restart service when not necessary. We will try to verify which value changed, then try to inject the flag. If the flag has been successfully injected then we don't restart. If it failed we will restart the approprieate section based on what changed and thus the proper daemon.

Signed-off-by: Sébastien Han seb@redhat.com


roles/ceph-common/handlers/inject-args.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: check the diff between old and new ceph.conf
+  command: "diff --changed-group-format='%>' --unchanged-group-format='' /etc/ceph/{{ cluster }}.conf /tmp/{{ cluster }}.conf"
+  register: diffcephconf
+
+- name: inject the configuration option
+  command:
```
mattt416: Haven't ansibled in a while, is this empty command correct?  :)

leseb: Ah this PR is really WIP so it is not meant to work at all, just a way for me to save my work online just in case ;)

SamYaple:
looks good so far for WIP (minus the missing code of course) The structure is what we discussed

leseb:
Closing this, we are waiting for https://github.com/ceph/ceph/pull/10736 and we will rebase the work on top of https://github.com/ceph/ceph-ansible/pull/995/files

PR #935, opened 2016-08-16
Remove duplicate "max open files"

cdwertmann:


leseb:
Why is it duplicate?

cdwertmann:
It's also in line 14

leseb:
oops you're right, thanks!

Issue #843, opened 2016-06-03
Ceph Hammer: Unit ceph-radosgw.service failed to load (again!)


ASBishop:
I am able to reproduce the problem reported in #642. Everything works fine the first time I deploy the cluster, but if I run purge-cluster.yml and deploy again it will fail. This is when deploying ceph_stable_rh_storage on RHEL 7.2, when mons and rgws are on the same hosts.

After the install fails, I notice the systemd unit file for 'ceph' is present, but the one for 'ceph-radosgw' is missing. These files are (or should be) in /var/run/systemd/generator.late.

Upon investigation, I see the 'ceph' RPM (ceph-0.94.5-9.el7cp.x86_64.rpm) has a postinstall scriptlet that executes "/sbin/chkconfig --add ceph" and that causes systemd to re-run its generators, which in turn creates the unit file.

However, the 'ceph-radosgw' RPM (ceph-radosgw-0.94.5-9.el7cp.x86_64.rpm) does not run chkconfig in its postinstall scriptlet. My guess is the first time a cluster is deployed, something else causes the systemd generators to run some time after ceph-radosgw is installed. However, this does not appear to happen after a purge and redeploy.

The fact that the ceph-radosgw RPM isn't running chckconfig on installation may well be a bug to file with Red Hat. Meanwhile, I found a workaround, and that is to add the following task at the very end of install_on_redhat.yml:

- name: trigger creation of ceph-radosgw systemd unit file
  command: chkconfig --add ceph-radosgw
  when:
    - rgw_group_name in group_names

leseb:
@ASBishop thanks for investigating.
Adding @ktdreyer to this thread for the potential package issue.

Did you check if other packages (mon, osd) also have this `chkconfig --add ceph*`?
Thanks!

ASBishop:
The mon and osd functionality are governed by the 'ceph' daemon. The packages are dependent on the ceph RPM, which we know calls chkconfig. So mon and osd are OK.

ceph-radosgw runs its own daemon, and therefore the RPM should call chkconfig.

I have no experience with mds (not using it).

leseb:
Closing this, feel free to re-open if the issue is not solved.

ASBishop:
Was there a change that should have solved the issue?

leseb:
Not really, can you tell me if you're still having this issue?

ASBishop:
[...]

PR #938, opened 2016-08-16
Fix pre-infernalis RPM installation of ceph-radosgw

ASBishop:
For pre-infernalis installation of ceph-radosgw from RPM, run 'chkconfig'
to ensure systemd's ceph-radosgw.service is created.

This fixes issue #843.

### DIFF
--- a/roles/ceph-rgw/tasks/pre_requisite.yml
+++ b/roles/ceph-rgw/tasks/pre_requisite.yml
@@ -41,6 +41,14 @@
     group: "{{ key_group }}"
   when: cephx
 
+- name: ensure ceph-radosgw systemd unit file is present
+  command: chkconfig --add ceph-radosgw
+  args:
+    creates: /var/run/systemd/generator.late/ceph-radosgw.service
+  when:
+    - ansible_os_family == "RedHat"
+    - is_before_infernalis
+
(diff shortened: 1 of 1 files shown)


roles/ceph-rgw/tasks/pre_requisite.yml
```diff
@@ -41,6 +41,12 @@
     group: "{{ key_group }}"
   when: cephx
 
+- name: trigger creation of ceph-radosgw systemd unit file
+  command: chkconfig --add ceph-radosgw
```
leseb: is the command idempotent?

ASBishop: Yes, I believe it is, although I'm not an expert in this area. If you prefer, it could be made conditional on the ceph-radosgw.service file not being present.

leseb: I'm fine with running this command all the time, I just want to make sure that if we run the command multiple times it will always report 0 is success and/or already exist :).
The conditional check works for me too.

ASBishop:
I pushed another version that modifies the command to ensure its idempotent.

leseb:
@ASBishop Better, thanks! :)

PR #941, opened 2016-08-18
NFS fixups

dang:
- Move fsal_rgw config to ceph-common, as it's shaered with ceph-rgw
- Update all.docker.sample with NFS config


roles/ceph-common/defaults/main.yml
```diff
@@ -360,3 +360,5 @@ dmcrypt_journal_collocation: False
 dmcrypt_dedicated_journal: False
 
 osd_auto_discovery: False
+
```
leseb: Can you add a comment to explain what the variable does? Then run: `generate_group_vars_sample.sh`.

PR #945, opened 2016-08-22
docker: gather all the fact

leseb:
If we don't do this we won't be able to get the ip address of the docker
interface.

### DIFF
--- a/site-docker.yml.sample
+++ b/site-docker.yml.sample
@@ -1,6 +1,9 @@
 ---
 # Defines deployment design and assigns role to server groups
 
+- hosts: all
+  tasks: []
+
 - hosts: mons
   become: True
   roles:


site-docker.yml.sample
```diff
@@ -1,6 +1,9 @@
 ---
 # Defines deployment design and assigns role to server groups
 
+- hosts: all
```
font: Do we need `all` in this case? Should we restrict it to just the `mons` or the `mons`, `osds`, `mdss`, `rgws`, `nfss`, and `restapis`?

leseb: I should have done that yes, I'm not even following my own advices haha. Stupidly copied `site.yml.sample`.
Thanks @font

site-docker.yml.sample
```diff
@@ -1,6 +1,9 @@
...
+  tasks: []
+
 - hosts: mons
   become: True
```
font: If we're gathering facts in the beginning we can probably set `gather_facts` to `false` for the rest of the plays so we don't gather facts twice and speed up the processing.

leseb: Yup, do you mind sending a quick PR for that? I'm busy fixing the docker stuff :)

font: @leseb you bet.

leseb: Thanks!

PR #947, opened 2016-08-22
Restrict fact gathering to mons and update ceph.conf

font:
- Gather facts only for mons before processing ceph-mon role serially in
  containerized playbook sample
- Updated ceph.conf in order to generate a valid ceph.conf

### DIFF
--- a/roles/ceph-common/templates/ceph.conf.j2
+++ b/roles/ceph-common/templates/ceph.conf.j2
@@ -32,19 +32,19 @@ mon host = {% for host in groups[mon_group_name] %}{{ hostvars[host]['ansible_' 
 {% if mon_containerized_deployment %}
 fsid = {{ fsid }}
 {% if groups[mon_group_name] is defined %}
-{% for host in groups[mon_group_name] %}
+mon host = {% for host in groups[mon_group_name] -%}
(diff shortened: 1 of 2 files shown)


roles/ceph-common/templates/ceph.conf.j2
```diff
@@ -32,19 +32,19 @@ mon host = {% for host in groups[mon_group_name] %}{{ hostvars[host]['ansible_'
 {% if mon_containerized_deployment %}
 fsid = {{ fsid }}
 {% if groups[mon_group_name] is defined %}
-{% for host in groups[mon_group_name] %}
+mon host = {% for host in groups[mon_group_name] -%}
```
font: Before these changes we were generating the `ceph.conf` `mon host` variable as:
```
mon_host = XXX.XXX.XXX.XXX
,mon_host = XXX.XXX.XXX.XXX
,mon_host = XXX.XXX.XXX.XXX
```

Now we generate it as:
```
mon host = XXX.XXX.XXX.XXX,XXX.XXX.XXX.XXX,XXX.XXX.XXX.XXX
```

leseb: That was one of my fixes :)

PR #957, opened 2016-08-25
Correct naming of referenced file

gbraad:
Trivial fix. Referenced file exists, but naming of the link was not consistent with the actual file on disk.


leseb:
Thanks @gbraad

PR #963, opened 2016-08-29
ceph-disk activate must be used on the OSD partition

cdwertmann:
When using `journal_collocation`, the `ceph-disk activate` must activate the OSD partition, which is the first partition on the device.

### DIFF
--- a/roles/ceph-osd/tasks/activate_osds.yml
+++ b/roles/ceph-osd/tasks/activate_osds.yml
@@ -3,7 +3,7 @@
 # partition.
 
 - name: automatically activate osd disk(s) without partitions
-  command: ceph-disk activate "/dev/{{ item.key }}"
+  command: ceph-disk activate "/dev/{{ item.key }}p1"
   ignore_errors: true
   with_dict: ansible_devices
   when:


cdwertmann:
If the device is used, I get the following error:

```
failed: [ceph-osd-4] (item={'value': {u'scheduler_mode': u'', u'sectorsize': u'512', u'vendor': None, u'sectors': u'3907029168', u'host': u'Non-Volatile memory controller: Intel Corporation PCIe Data Center SSD (rev 01)', u'rotational': u'0', u'removable': u'0', u'support_discard': u'512', u'model': u'INTEL SSDPE2ME020T4', u'size': u'1.82 TB', u'holders': [], u'partitions': {}}, 'key': u'nvme7n1'}) => {"changed": true, "cmd": ["ceph-disk", "activate", "/dev/nvme7n1"], "delta": "0:00:00.193891", "end": "2016-08-29 15:49:49.444343", "failed": true, "item": {"key": "nvme7n1", "value": {"holders": [], "host": "Non-Volatile memory controller: Intel Corporation PCIe Data Center SSD (rev 01)", "model": "INTEL SSDPE2ME020T4", "partitions": {}, "removable": "0", "rotational": "0", "scheduler_mode": "", "sectors": "3907029168", "sectorsize": "512", "size": "1.82 TB", "support_discard": "512", "vendor": null}}, "rc": 1, "start": "2016-08-29 15:49:49.250452", "stderr": "ceph-disk: Cannot discover filesystem type: device /dev/nvme7n1: Line is truncated: ", "stdout": "", "stdout_lines": [], "warnings": []}

File: roles/ceph-osd/tasks/activate_osds.yml [hunk 1]
```diff
@@ -3,7 +3,7 @@
 # partition.
 
 - name: automatically activate osd disk(s) without partitions
-  command: ceph-disk activate "/dev/{{ item.key }}"
+  command: ceph-disk activate "/dev/{{ item.key }}p1"
```
leseb: this is only valid for NVMe drives I suppose, not generic enough and this will break current and new deployments.

leseb:
@cdwertmann any updates?

cdwertmann:
I ran out of time with my deployment, so I had to revert back to an earlier commit of ceph-ansible where everything worked smoothly with my NVMe drives. I didn't have a chance to research this issue further.

leseb:
@cdwertmann what the sha of the commit you got back to?

cdwertmann:
I went back to e92d9fbce0519b307b297519ddb629d837e04132 which just works without any modifications to the code

leseb:
hum I don't see why this commit could break anything but looking further ahead it seems that this one could: https://github.com/ceph/ceph-ansible/commit/b4fe16c31fd15119de91b42bd5f7306368fd9f63

I'm looking into this

leseb:
@cdwertmann can you give me your sign-off? I'm going to update this and I'd like to add you in the commit credit.

cdwertmann:
e92d9fbce0519b307b297519ddb629d837e04132 just happened to be a known working codebase for me, so the issue I've had must have been introduced at a later commit.

I assume that #1006 contains your proposed fix? Unfortunately I cannot test it because my cluster went into production, but you can add my name (Christoph Dwertmann) to the commit credit if you like.

leseb:
@cdwertmann correct #1006 has the fix, I just added your name to the commit message.

PR #970, opened 2016-09-02
adding cluster name so it will work with all cluster name

racpatel:
rolling_update.yml will not work if cluster name is not 'ceph'. Adding --cluster <cluster_name> will solve this problem

Fixes issue #969


leseb:
Please rebase your patch :)

racpatel:
submitted another PR  #972, hence closing this one

PR #972, opened 2016-09-06
rolling_update.yml will not work if cluster name is other than ceph

racpatel:
rolling_update.yml will not work if cluster name is other than  'ceph'. Adding --cluster will solve this problem

 Fixes issue #969


leseb:
Thanks @racpatel

PR #978, opened 2016-09-15
Add floating_ip_pool openstack

FranBarrera:
Signed-off-by: FranBarrera <FranBarrera6@gmail.com>


FranBarrera:
Without floating-ip, vagrant can't connect to instances

PR #981, opened 2016-09-19
ceph-osd: get full link path when testing if a partition

leseb:
this allows us to test devices set with persistent naming such as
/dev/disk/by-*


leseb:
@bengland2

leseb:
test this please

andrewschoen:
test this please

bengland2:
can you fill in context on why are we doing this and what specifically you want to test?  Is it [this bz](https://bugzilla.redhat.com/show_bug.cgi?id=1375378)?

I think instead of 
    echo '$(readlink -f {{ item }})' | ...
you could just write:
    readlink -f {{ item }} | ...
since readlink prints out either the input device name or the device name referred to by the input softlink.  But it amounts to the same thing AFAIK.  I backported your fix to an older version of ceph-ansible and tried it there, and it seems to work.

leseb:
@bengland2 you're right thanks, I added your suggestion and more context as well. This is not 100% connected to the BZ you linked but at least that's the one that gave me the idea.

PR #982, opened 2016-09-19
vagrant: add support for Linode VMs

batrick:
This will be used in upcoming performance testing.

### DIFF
--- a/vagrant_variables.yml.linode
+++ b/vagrant_variables.yml.linode
@@ -1,0 +1,42 @@
+---
+
+vagrant_box: 'linode'
+vagrant_box_url: 'https://github.com/displague/vagrant-linode/raw/master/box/linode.box'
+
+ssh_username: 'vagrant'
+ssh_private_key_path: '~/.ssh/id_rsa'
+
+cloud_distribution: 'CentOS 7'
+cloud_datacenter: 'newark'
+
+# Memory for each Linode instance, this determines price! See Linode plans.
+memory: 2048
+
+# The private network on Linode, you probably don't want to change this.
+subnet: 192.168.0
+
+# DEFINE THE NUMBER OF VMS TO RUN
+mon_vms: 3
+osd_vms: 3
+mds_vms: 1
+rgw_vms: 0
+nfs_vms: 0
+rbd_mirror_vms: 0
+client_vms: 0
+iscsi_gw_vms: 0
+
+# Deploy RESTAPI on each of the Monitors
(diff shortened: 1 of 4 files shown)


vagrant_variables.yml.linode
```diff
@@ -0,0 +1,51 @@
...
+cloud_distribution: 'CentOS 7'
+cloud_datacenter: 'newark'
+
+# Memory for each Linode instance, this determines price! See Linode plans.
+memory: 2048
+
+# The private network on Linode, you probably don't want to change this.
+subnet: 192.168.0
+
+# The interface the monitors should listen on, this is currently ignored for linode.
```
leseb: Why don't we simply remove this then?

vagrant_variables.yml.linode
```diff
@@ -0,0 +1,51 @@
...
+
+# The sync directory changes based on vagrant box
+# Set to /home/vagrant/sync for Centos/7, /home/{ user }/vagrant for openstack and defaults to /vagrant
+# vagrant_sync_dir: /home/vagrant/sync
+
+os_tuning_params:
+  - { name: kernel.pid_max, value: 4194303 }
+  - { name: fs.file-max, value: 26234859 }
+
+ceph_conf_overrides:
```
leseb: is it a left over?

vagrant_variables.yml.linode
batrick: Okay, will do.

vagrant_variables.yml.linode
batrick: Not really. I thought it'd be useful to have an example `vagrant_variables.yml.X` with the new ceph_conf_overrides support. I will remove the `mds debug` but I'd like to keep the `fuse default permissions` override (as it's generally a large performance improvement for cephfs).

leseb: Hum `ceph_conf_overrides` can be used and should from `group_vars/all`, so I'd like to keep it that way. When it comes to optimizations, I have a local branch for "profiles". These use `ceph_conf_overrides` and provide some guidance on how to tune/tweak specific deployment. The fuse option falls down into this so it's a good candidate once I push the branch, until then I suggest we don't do anything.

batrick: Oh, that makes a lot more sense. Thanks for clarifying! I will remove it.

leseb: Np, thanks!

PR #998, opened 2016-09-23
vagrant: add label_prefix setting

batrick:
This enables running multiple clusters concurrently in the same Linode
account. Linode does not allow machines to have the same label.


leseb:
Since this variable is called by Vagrant (no matter which  provider you use) you need to add this to all the `vagrant_variable` files otherwise this will break other Vagrant installations.

batrick:
@leseb how about the alternative I've added?

leseb:
Sorry I was traveling, thanks!

PR #1000, opened 2016-09-28
Small fix in config template

Oloremo:
1) Remove duplication of "osd crush chooseleaf type = 0"
2) Add osd_pool_default_size = 1 for a single node deployment

Signed-off-by: Proskurin Kirill <kproskurin@mirantis.com>

### DIFF
--- a/roles/ceph-common/templates/ceph.conf.j2
+++ b/roles/ceph-common/templates/ceph.conf.j2
@@ -56,9 +56,6 @@ public_network = {{ public_network }}
 {% if cluster_network is defined %}
 cluster_network = {{ cluster_network }}
 {% endif %}
-{% if common_single_host_mode is defined %}
-osd crush chooseleaf type = 0
-{% endif %}
 
 [client.libvirt]
 admin socket = {{ rbd_client_admin_socket_path }}/$cluster-$type.$id.$pid.$cctid.asok # must be writable by QEMU and allowed by SELinux or AppArmor


leseb:
It's not because we use a single node that we want to have a single replica per pool. So we should keep the crush chooseleaf type to 0.

Oloremo:
@leseb I dont get it. In master we have 2 instances of "osd crush chooseleaf type = 0"

First one:
https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/templates/ceph.conf.j2#L15-L17

Second one:
https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/templates/ceph.conf.j2#L59-L61

They are both in [global] section, so if you deploy ceph with "common_single_host_mode" it will put it twice into config and ceph will print a warning what this var ressigned.

And for a single node deployment we need to osd_pool_default_size and min size to 1, or cluster status will be WARN. So I putted it into the "common_single_host_mode" condition.

Did I get something wrong?

leseb:
Oh right, there are 2 declarations. I didn't notice that. I'm fine with that chance. However you can deploy 3 OSDs on one machine and still use 3 replicas on the same box. When setting pool default size to one you're forcing a single replica per pool.

Oloremo:
@leseb well, imo, single node cluster used for simple test\dev scenarios and with this change it will work out of the box with cluster OK status. And you can scale up it later if needed.

leseb:
Well you never know, some of the developers might want to verify that the repliacation is working correctly or something else. So we can not force users into a certain direction. If they only deploy one disk then Ceph won't be happy but they can easily fix it and use config overrides to add pool default size to 1.

Oloremo:
@leseb Removed osd_pool part.

leseb:
Thanks!

PR #1004, opened 2016-10-04
Debian installation: fix Ansible compilation and correct syntax for Ansible v2

analbeard:
Fixes an issue with Ansible compilation where no submodules are available, and corrects syntax for Ansible v2.

### DIFF
--- a/install-ansible.sh
+++ b/install-ansible.sh
@@ -42,6 +42,7 @@ if [[ "Debian" =~ $os_VENDOR ]]; then
   pip install PyYAML jinja2 paramiko
   git clone https://github.com/ansible/ansible.git
   cd ansible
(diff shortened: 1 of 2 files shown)


install-ansible.sh
```diff
@@ -42,6 +42,7 @@ if [[ "Debian" =~ $os_VENDOR ]]; then
   pip install PyYAML jinja2 paramiko
   git clone https://github.com/ansible/ansible.git
   cd ansible
+  git submodule update --init --recursive
```
leseb: what's that supposed to do? update all the submodule from the ansible repo?

analbeard: That's correct - I was running into this: https://github.com/leucos/ansible-tuto/issues/92

PR #1005, opened 2016-10-05
Fix ubuntu apt nfs ganesha

OrFriedmann:


roles/ceph-common/tasks/installs/install_on_debian.yml
```diff
@@ -39,7 +39,14 @@
     default_release: "{{ ceph_stable_release_uca | default(ansible_distribution_release) }}{{ '-backports' if ceph_origin == 'distro' and ceph_use_distro_backports else ''}}"
   when: mds_group_name in group_names
 
-- name: install NFS gateway
+- name: install jemalloc
+  apt:
+    pkg: libjemalloc1
+    state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
+    update_cache: yes
+  when: nfs_group_name in group_names
```
leseb: I think you want to on the OSD nodes not on the nfs ones

roles/ceph-common/tasks/installs/debian_ceph_repository.yml
```diff
@@ -46,7 +46,11 @@
 
 - name: add gluster nfs ganesha repo
   apt_repository:
-    repo: "ppa:gluster/nfs-ganesha"
+    repo: "{{ item }}"
     state: present
+  with_items:
+    - ppa:gluster/libntirpc
```
leseb: same as below, jemalloc is needed by osds not nfs, if nfs is not enabled this repo won't get added

OrFriedmann:
https://bugzilla.redhat.com/show_bug.cgi?id=1350202
When the playbook finish to run I saw the error from the link that requires jemalloc for nfs ganesha.

leseb:
Ok sounds good then.

PR #1011, opened 2016-10-06
Make keystone ssl verify optional and add keystone v3 support

Oloremo:
But enabled by default.

Signed-off-by: Proskurin Kirill <kproskurin@mirantis.com>

### DIFF
--- a/group_vars/all.sample
+++ b/group_vars/all.sample
@@ -285,12 +285,20 @@ dummy:
 #radosgw_civetweb_num_threads: 50
 #radosgw_keystone: false # activate OpenStack Keystone options full detail here: http://ceph.com/docs/master/radosgw/keystone/
 #radosgw_keystone_url: # url:admin_port ie: http://192.168.0.1:35357
+#radosgw_keystone_api_ver: 2
+#radosgw_keystone_admin_domain: default
+#radosgw_keystone_admin_project: admin
+#radosgw_keystone_use_admin_token: true
+#radosgw_keystone_use_admin_cred: false
 #radosgw_keystone_admin_token: password
+#radosgw_keystone_admin_user: admin
+#radosgw_keystone_admin_password: password
 #radosgw_keystone_accepted_roles: Member, _member_, admin
 #radosgw_keystone_token_cache_size: 10000
 #radosgw_keystone_revocation_internal: 900
 #radosgw_s3_auth_use_keystone: "true"
 #radosgw_nss_db_path: /var/lib/ceph/radosgw/ceph-radosgw.{{ ansible_hostname }}/nss
+#radosgw_keystone_verify_ssl: true
 # Toggle 100-continue support for Apache and FastCGI
 # WARNING: Changing this value will cause an outage of Apache while it is reinstalled on RGW nodes
 #http_100_continue: false
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -283,6 +283,7 @@ radosgw_keystone_token_cache_size: 10000
 radosgw_keystone_revocation_internal: 900
 radosgw_s3_auth_use_keystone: "true"
 radosgw_nss_db_path: /var/lib/ceph/radosgw/ceph-radosgw.{{ ansible_hostname }}/nss
+radosgw_keystone_verify_ssl: true
 # Toggle 100-continue support for Apache and FastCGI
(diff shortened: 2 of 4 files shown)


Oloremo:
Failed playbook test not related to my commit. :-(

leseb:
Is there any cases where we don't do SSL? I'm trying to understand the purpose of the PR.
Why don't we simply add this option into the ceph.conf template?
The entire  `openstack-keystone.yml` file is about configured the SSL for the rgw/keystone so I'm confused.

Oloremo:
@leseb oh you're right about whole file thing, I, somehow, miss it. 
I need this option for 2 reasons:
1) Well, it's possible, so why not give users a choice? 
2) In running OS in docker containers via k8s... And I have external ceph setup, so, I can easlly access keystone directory. And even if I could - I dont want to, right now I'm running sslless setup, which is default for keystone.

I could update patch to skip this whole file if you're convinced what this feature is needed.

Oloremo:
And this setup requires keystone-manage pki_setup which is:
```
Deprecated: keystone-manage pki_setup is deprecated as of Mitaka in favor of not using PKI tokens and may be removed in 'O' release.
2016-10-06 13:30:29.010 1314 WARNING keystone.cmd.cli [-] keystone-manage pki_setup is not recommended for production use.
```

leseb:
Ok here's what I propose:

* we don't add  `rgw_keystone_verify_ssl` to the ceph.conf since it defaults to `true` in ceph already
* we replace this condition https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-rgw/tasks/main.yml#L8 with `when: radosgw_keystone_verify_ssl`

What do you think?

Oloremo:
@leseb but in this case we skip SSL bootstrap(openstack-keystone.yml) but keep `rgw keystone verify ssl` true

Oloremo:
Ops, I accidentally added a v3 support commit here, wanted to create a separate PR. Sorry, I hope it's ok, if it's not, ill try to separate them... somehow.

leseb:
v3 support is fine, you can leave it in this PR since there are 2 different commits

Oloremo:
@leseb Does both patches looks ok for you? Or I need to improve something?

leseb:
Almost good, have a look at this one, https://github.com/ceph/ceph-ansible/pull/1018
The only issue at the moment is that it duplicates the rgw sections...

Oloremo:
@leseb Not an ansible expert, so I dont really follow here. Which duplications? 

Like "radosgw_keystone_api_ver" 2 or 3? But it's a based on condition. And I think it's better to state value explicitly, even the defaults one, so operator dont need to guess which api version is in use.

The thing I dont really like is the "{% if radosgw_keystone_use_admin_token %}" condition followed by "{% elif radosgw_keystone_use_admin_cred %}". It could be changed to just "else" I think. 

About #1018, not sure how it helps.

leseb:
@Oloremo By duplicate I meant, with #1018 we end up having 2 rgw sections...
I believe that we can close this PR in favour of #1018. If we want to do any rgw/keystone we just need to enable `radosgw_keystone`. What do you think?

Oloremo:
@leseb #1018 doesn't allow to define(or Im missing something?):
radosgw_keystone_api_ver
radosgw_keystone_admin_domain
radosgw_keystone_admin_project
radosgw_keystone_verify_ssl
radosgw_keystone_admin_user
radosgw_keystone_admin_password

Which is the whole point of this patch. Using "admin token" is deprecated in keystone and not secure.

verify_ssl needs to be disabled for deployment in containers + pki setup is deprecated too.

I dont really care how this is going to be added, I just want to make it happen. :-)

leseb:
@Oloremo #1018 is just an example, you can easily add the variables you mentioned and also disable the one you don't want.

Oloremo:
@leseb oh ok. Not sure why #1018 is better, but whatever works for you. :-)

leseb:
@Oloremo #1018 is better because we don't add anything in the ceph.conf template, we don't want to edit it anymore since we have `ceph_conf_overrides` :). I just updated #1018 with more guidance, let me know how that looks :) Thanks!

Oloremo:
@leseb yes, now it's way clearer. In this case you could close this PR in favor of #1018

leseb:
@Oloremo thanks I'm closing this, please review #1018 to make sure I didn't miss anything. Thanks!

PR #1014, opened 2016-10-07
Allow to propagate cluster name to ceph-fetch-keys

9seconds:
`ceph-fetch-keys` role currently works only if cluster name is `ceph`.
This commit allows to set custom cluster name in `defaults` in the same
fashion as other roles do.


leseb:
Could you please run `generate_group_vars_sample.sh` and push the appropriate file?
Thanks

9seconds:
@leseb done. Please check

leseb:
Thanks @9seconds

Issue #952, opened 2016-08-23
MAC address get concatenated in OSDs with vagrant-libvirt 0.0.35


rassie:
I've tried setting up CEPH using `ceph-ansible` with `vagrant-libvirt`. However, I've got errors setting up OSDs about MAC address on the `eth2` interface being different from what was expected. I've debugged a bit and found out that those MAC addresses are two MACs each, concatenated (so the error message is correct). The problem is that `vagrant-libvirt` 0.0.35 uses an XPath expression to query the MAC from `libvirt`'s XML: `"/domain/devices/interface[source[@network='#{@network_name}']]/mac/@address"` (change came in in commit https://github.com/vagrant-libvirt/vagrant-libvirt/commit/ee7afab369dc7e80128ff35a2ff11f6eab23d368). However, this is how the XML for an OSD looks like:

```
    <interface type='network'>
      <mac address='52:54:00:c9:8b:13'/>
      <source network='ceph-ansible0' bridge='virbr2'/>
      <target dev='vnet3'/>
      <model type='virtio'/>
      <alias name='net1'/>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x08' function='0x0'/>
    </interface>
    <interface type='network'>
      <mac address='52:54:00:7a:9b:4c'/>
      <source network='ceph-ansible0' bridge='virbr2'/>
      <target dev='vnet4'/>
      <model type='virtio'/>
      <alias name='net2'/>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x09' function='0x0'/>
    </interface>
```

As you can see, both interfaces (`eth1` and `eth2`) have the same source network name, so that the XPath expression actually returns two addresses, which is not caught by `vagrant-libvirt` (so a bug report is due there as well).

So the question is basically (apart from `vagrant-libvirt`'s parsing): is this (probably default) network naming correct? If not, could this be changed?

leseb:
I personally don't use `vagrant-libvirt` and I'm not sure why both interfaces have the same network name. If you can find the appropriate syntax to change the network name then I'm fine putting a change in the `Vagrantfile`.

font:
[...]

PR #1015, opened 2016-10-07
Create separate subnet for public/cluster networks

font:
- All Ceph instances now communication using public subnet and
  additionally OSDs communicate with each other using private cluster
  subnet
- Workaround for
  https://github.com/vagrant-libvirt/vagrant-libvirt/issues/645
- Fix for #952 to avoid concatenated MAC addresses caused by
  vagrant-libvirt bug.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,7 +19,8 @@ RESTAPI        = settings['restapi']
 NRBD_MIRRORS   = settings['rbd_mirror_vms']
 CLIENTS        = settings['client_vms']
 NISCSI_GWS     = settings['iscsi_gw_vms']
-SUBNET         = settings['subnet']
+PUBLIC_SUBNET  = settings['public_subnet']
+CLUSTER_SUBNET = settings['cluster_subnet']
 BOX            = settings['vagrant_box']
 BOX_URL        = settings['vagrant_box_url']
 SYNC_DIR       = settings['vagrant_sync_dir']
@@ -68,9 +69,9 @@ ansible_provision = proc do |ansible|
   end
 
   ansible.extra_vars = {
-      cluster_network: "#{SUBNET}.0/24",
+      cluster_network: "#{CLUSTER_SUBNET}.0/24",
       journal_size: 100,
-      public_network: "#{SUBNET}.0/24",
+      public_network: "#{PUBLIC_SUBNET}.0/24",
   }
   if settings['ceph_install_source'] == 'dev' then
     ansible.extra_vars['ceph_dev'] = true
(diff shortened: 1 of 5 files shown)


Vagrantfile
```diff
@@ -114,11 +115,11 @@ ansible_provision = proc do |ansible|
     # Use monitor_address_block instead of monitor_interface:
     ansible.extra_vars.delete(:monitor_interface)
     ansible.extra_vars = ansible.extra_vars.merge({
-      cluster_network: "#{SUBNET}.0/16",
+      cluster_network: "#{CLUSTER_SUBNET}.0/16",
       devices: ['/dev/sdc'], # hardcode leftover disk
       journal_collocation: 'true',
-      monitor_address_block: "#{SUBNET}.0/16",
-      public_network: "#{SUBNET}.0/16",
+      monitor_address_block: "#{PUBLIC_SUBNET}.0/16",
```
leseb: ditto

Vagrantfile
```diff
@@ -114,11 +115,11 @@ ansible_provision = proc do |ansible|
...
+      public_network: "#{PUBLIC_SUBNET}.0/16",
```
leseb: ditto

Vagrantfile
```diff
@@ -114,11 +115,11 @@ ansible_provision = proc do |ansible|
     # Use monitor_address_block instead of monitor_interface:
     ansible.extra_vars.delete(:monitor_interface)
     ansible.extra_vars = ansible.extra_vars.merge({
-      cluster_network: "#{SUBNET}.0/16",
+      cluster_network: "#{CLUSTER_SUBNET}.0/16",
```
leseb: hum let's ask @batrick for the /16

batrick: It should be /16 since the private LAN on Linode is anywhere in 192.168.0.0/16.  It would probably be a good idea to fix this so SUBNET is actually a subnet rather than a partial address...

font: @batrick I think the reason it's a partial address is because it simplifies setting static IP addresses as the host part is just appended farther down in the Vagrantfile.

batrick: I don't follow what you mean. For Linode, we aren't assigning any IP addresses. For background, my use-case with Linode is performance testing and I don't want the ceph daemons listening on any publicly accessible addresses. So all of the daemons should be listening on 192.168.0.0/16 (and ceph-ansible/vagrant cannot control what those addresses eventually are).

font: I've updated the PR to use the same private subnet 192.168.0.0/16 for both public and cluster networks. Linode, and OpenStack for that matter, just have a slightly different network configuration.

leseb:
Thanks @font

PR #1022, opened 2016-10-10
Updated to use xenial by default.  Added OSD specific controller.

jsaintrocc:
Found a Vagrant box image for xenial that works well.  The ubuntu and bento ones don't include python by default and have other issues as well.  geerlingguy/ubuntu1604 image is the 2nd most popular one on the atlas repo with 144K downloads.  The developer looks pretty legit and has the code posted in his git repo.  @leseb is this image ok for us to use?

I've added code to create a SCSI controller as part of the vagrant provisioning.  This makes us a little less dependent on the controller the box has built in.  Also it means we can rely on the controller name being a constant.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -25,7 +25,6 @@ BOX            = settings['vagrant_box']
 BOX_URL        = settings['vagrant_box_url']
 SYNC_DIR       = settings['vagrant_sync_dir']
 MEMORY         = settings['memory']
-STORAGECTL     = settings['vagrant_storagectl']
 ETH            = settings['eth']
 DOCKER         = settings['docker']
 USER           = settings['ssh_username']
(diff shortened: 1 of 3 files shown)


leseb:
I'm not strongly against it but I've been using `bento/ubuntu-16.04` without any issues before. Can you elaborate on the issues you had?
I think creating our own storage controller makes sense though so we always get it right.

jsaintrocc:
The problem I run into with using 'bento/ubuntu-16.04' is that there is a 5 min 3s (during initial boot) delay for 'A start job is running for Raise network interfaces'.  This exceeds the default 5 min timeout and slows things down too much.

jsaintrocc:
Ah.  I discovered it was an issue with me using Virtualbox 5.0.X.  Upgrading to 5.1.X gets rid of the long delay in booting 'bento/ubuntu-16.04'.  I've updated the PR to use the bento box.  Thanks to Chris Levi for giving me a clue.

PR #1029, opened 2016-10-12
site.yml.sample: Attempt to install python2

badone:
Some modern distros ship without python2 so we need to install it blindly,
before any module is run. Fedora > 23 also requires some aditional packages be
installed.

### DIFF
--- a/site.yml.sample
+++ b/site.yml.sample
@@ -1,52 +1,97 @@
 ---
 # Defines deployment design and assigns role to server groups
 
+- hosts:
+  - mons
+  - agents
+  - osds
+  - mdss
+  - rgws
+  - nfss
+  - restapis
+  - rbdmirrors
+  - clients
+  - iscsigws
+  gather_facts: false
+  tasks:
+    # If we can't get python2 installed before any module is used we will fail
+    # so just try what we can to get it installed
+    - name: check for python2
+      stat:
+        path: /usr/bin/python
+      ignore_errors: yes
+      register: systempython2
+    - name: install python2 for Debian based systems
+      raw: sudo apt-get -y install python-simplejson
+      ignore_errors: yes
+      when: systempython2.stat.exists is undefined or systempython2.stat.exists == false
+    # Try to install python2 on Fedora > 23
+    - name: install python2 for Fedora
+      raw: sudo dnf -y install python creates=/usr/bin/python
+      ignore_errors: yes
+      when: systempython2.stat.exists is undefined or systempython2.stat.exists == false
+    - name: gathering facts
+      setup:
+    - name: install required packages for Fedora > 23
+      raw: sudo dnf -y install python2-dnf libselinux-python ntp
+      when: ansible_distribution == 'Fedora' and ansible_distribution_major_version|int >= 23
+
 - hosts: mons
+  gather_facts: false
   become: True
   roles:
   - ceph-mon
 
 - hosts: agents
+  gather_facts: false
   become: True
   roles:
   - ceph-agent
 
 - hosts: osds
+  gather_facts: false
   become: True
   roles:
   - ceph-osd
 
 - hosts: mdss
+  gather_facts: false
   become: True
   roles:
   - ceph-mds
 
 - hosts: rgws
+  gather_facts: false
   become: True
   roles:
   - ceph-rgw
 
 - hosts: nfss
+  gather_facts: false
   become: True
   roles:
   - ceph-nfs
 
 - hosts: restapis
+  gather_facts: false
   become: True
   roles:
   - ceph-restapi
 
 - hosts: rbdmirrors
+  gather_facts: false
   become: True
   roles:
   - ceph-rbd-mirror
 
 - hosts: clients
+  gather_facts: false
   become: True
   roles:
   - ceph-client
 
 - hosts: iscsigws
+  gather_facts: false
   become: True
   roles:
   - ceph-iscsi-gw


badone:
This also works with the latest f25 beta cloud image.

https://dl.fedoraproject.org/pub/fedora/linux/development/25/CloudImages/x86_64/images/Fedora-Cloud-Base-Vagrant-25-20161012.n.0.x86_64.vagrant-libvirt.box

site.yml.sample
```diff
@@ -1,52 +1,88 @@
 ---
 # Defines deployment design and assigns role to server groups
 
+- hosts: all
```
leseb: `all` can potentially be harmful if we have other nodes being managed by ansible (other than ceph nodes), so we should explicitly list all the ceph roles.

site.yml.sample
```diff
@@ -1,52 +1,88 @@
 ---
 # Defines deployment design and assigns role to server groups
 
+- hosts: all 
+  gather_facts: false
+  tasks:
+    # If we can't get python2 installed before any module is used we will fail
+    # so just try what we can to get it installed
+    - name: Check for python2
```
leseb: please the remove the capital letter: is `name: check for python2`.

site.yml.sample
```diff
@@ -1,52 +1,88 @@
...
+      stat:
+        path: /usr/bin/python
+      ignore_errors: yes 
+      register: systempython2
+    # Try to install python2 on Ubuntu Xenial and possibly others
+    - name: Install python2 for Ubuntu
```
leseb: "for Debian based systems"

site.yml.sample
```diff
@@ -1,52 +1,88 @@
...
+      raw: sudo apt-get -y install python-simplejson
+      ignore_errors: yes 
+      when: systempython2.stat.exists is undefined or systempython2.stat.exists == false
+    # Try to install python2 on Fedora > 23
+    - name: Install python2 for Fedora
```
leseb: ditto

site.yml.sample
```diff
@@ -1,52 +1,88 @@
...
+      raw: sudo dnf -y install python creates=/usr/bin/python
+      ignore_errors: yes 
+      when: systempython2.stat.exists is undefined or systempython2.stat.exists == false
+    - name: Gathering facts
+      setup:
+    - name: Install required packages for Fedora > 23
```
leseb: ditto

site.yml.sample
```diff
@@ -1,52 +1,88 @@
...
+  gather_facts: false
+  tasks:
+    # If we can't get python2 installed before any module is used we will fail
+    # so just try what we can to get it installed
+    - name: Check for python2
+      stat:
+        path: /usr/bin/python
+      ignore_errors: yes 
+      register: systempython2
+    # Try to install python2 on Ubuntu Xenial and possibly others
```
leseb: I think we can remove this comment since the name of the task is pretty explicit

site.yml.sample
```diff
@@ -1,52 +1,88 @@
...
+    - name: Install python2 for Ubuntu
+      raw: sudo apt-get -y install python-simplejson
+      ignore_errors: yes 
+      when: systempython2.stat.exists is undefined or systempython2.stat.exists == false
```
leseb: add `and ansible_os_family == 'Debian'` otherwise this will run on Fedora and others too

site.yml.sample
```diff
@@ -1,52 +1,88 @@
...
+    # Try to install python2 on Fedora > 23
+    - name: Install python2 for Fedora
+      raw: sudo dnf -y install python creates=/usr/bin/python
+      ignore_errors: yes 
+      when: systempython2.stat.exists is undefined or systempython2.stat.exists == false
```
leseb: add  `and ansible_distribution == 'Fedora'`?

site.yml.sample
badone: No can do I'm afraid. This is running before we have run "setup" and defined these variables. Unfortunately that's the point, if we run "setup" withput python2 installed we fail so my intention was just attempt to brute force it into place regardless of OS/distro and ignore errors doing so.

site.yml.sample
```diff
@@ -1,52 +1,87 @@
 ---
 # Defines deployment design and assigns role to server groups
 
+- hosts: mons agents osds mdss rgws nfss restapis rbdmirrors clients iscsigws
```
leseb: This is not going to work, this be written like this:

```
- hosts:
  - mons
  - agents
...
```

leseb:
Almost good with this, why are you disabling fact collection?
I believe this won't work without gathering facts...

badone:
@leseb we gather facts on line 34. I figured once was enough?

leseb:
Well perhaps I'm getting blind but I don't see where we are gathering facts in this file: https://github.com/badone/ceph-ansible/blob/e6994aecd3d56afd596484251a50ac26d954f062/site.yml.sample

badone:
@leseb ah, I must misunderstand how this works. How I thought it was working was that I disable gathering facts on line 15 and then, when I call "setup" on line 34 that then does the actual fact gathering. I needed to call it there so I can use ansible_distribution and ansible_distribution_major_version in that block. Sorry if this is not how it works but I saw this code in multiple resources that talked about disabling and re-enabling "gathering facts"

font:
@leseb @badone Since we're trying to bootstrap ourselves to solve a chicken and egg problem this does seem like somewhat of a hack, but it should work. The `setup` module can be used to force gathering facts even if `gather_facts` is set to `false` for the play. This should then make the facts available for all subsequent plays.

@badone Did you have an environment where you tested this on to verify?

badone:
@font yes, I tested it using the latest f24 and f25 official vagrant images.

leseb:
Alright I missed the call of the `setup` module, my bad. Let's go ahead with this.

PR #1035, opened 2016-10-16
Ensure extras repo enabled on CentOS

verdurin:
This is to address the problem I found in issue #1030, in cases where (for whatever reason) the CentOS `Extras` repo is not enabled.

### DIFF
--- a/roles/ceph-mon/tasks/docker/pre_requisite.yml
+++ b/roles/ceph-mon/tasks/docker/pre_requisite.yml
@@ -40,7 +40,6 @@
     state: present
     enabled: yes
   when:
-    - ansible_os_family == 'RedHat'
     - ansible_distribution == 'CentOS'
   tags:
     with_pkg
--- a/roles/ceph-mon/tasks/docker/pre_requisite.yml
+++ b/roles/ceph-mon/tasks/docker/pre_requisite.yml
@@ -33,6 +33,18 @@
   tags:
     with_pkg
 
+# ensure extras enabled for docker
+- name: enable extras on centos
+  yum_repository:
+    name: extras
+    state: present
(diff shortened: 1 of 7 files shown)


roles/ceph-mon/tasks/docker/pre_requisite.yml
```diff
@@ -33,6 +33,18 @@
...
     with_pkg
 
+# ensure extras enabled for docker
+- name: enable extras on centos
+  yum_repository:
+    name: extras
+    state: present
+    enabled: yes
+  when:
+    - ansible_os_family == 'RedHat'
```
leseb: Since CentOS is a red hat based system, I think this condition is not necessary

leseb:
This is potentially an issue for osd, mds, rgw and other containerized versions rigth?

verdurin:
Yes, it just always happened first for `mon` in my testing.

Happy to add it to the others, if you prefer.

leseb:
well it happens to mons first because they are the one behind deployed first, I suspect that this only fixes mon but you will likely have the same problem for the other machines? (unless there are running on the same machines as the mons).

Can you validate this? Thanks

verdurin:
In the case of the small cluster I tested on, they're all on the same machines.

leseb:
Alright, anyway this will hit the other daemons if they run on another machine so please update them. Thanks!

alfredodeza:
This PR caused all current builds to be broken because of 9de16e4d

The linked PR ( #1039 ) still does not fix master. Please double check all runs from Jenkins. The ones where devices are not present are safe to ignore. But the failure introduced by this PR was clear from the PR and has been present in every subsequent PR.

PR #1042, opened 2016-10-19
Change enabling of extras repo for CentOS to use ini_file

verdurin:
Fix the problems caused by my previous attempt to make sure the `extras` repo is enabled on `CentOS`.

Signed-off-by: Adam Huffman <bloch@verdurin.com>

### DIFF
--- a/roles/ceph-mds/tasks/docker/pre_requisite.yml
+++ b/roles/ceph-mds/tasks/docker/pre_requisite.yml
@@ -39,10 +39,11 @@
   failed_when: false
 
 - name: enable extras repo for centos
-  yum_repository:
-    name: extras
-    state: present
-    enabled: yes
+  ini_file:
+    dest: /etc/yum.repos.d/CentOS-Base.repo
+    section: extras
+    option: enabled
+    value: 1
   when: ansible_distribution == 'CentOS'
   tags:
     with_pkg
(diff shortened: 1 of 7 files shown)


leseb:
I can not test this at the moment, are we good this time? :)

whatevsz:
I can't test it right now, maybe tomorrow. Will report back :)

alfredodeza:
This PR was merged with broken results from Jenkins.

verdurin:
On Thu, Oct 20, 2016 at 05:14:14AM -0700, Leseb wrote:
> I can not test this at the moment, are we good this time? :)
> 

We should be, except for the fact that:

tags:
   with_pkg

needs to be fixed throughout these files to be:

tags:
   - with_pkg

or:

tags: with_pkg

Is it okay to make this change for everything, not just the small part I
added?

whatevsz:
OK, so I looked into it. You missed one `yum_repository` task in `roles/ceph-osd/tasks/pre_requisite.yml` (line 9).

I fixed the tags with

```
find -name 'pre_requisite.yml' -exec sed -i 's/with_pkg/- with_pkg' {} \;
```
and it works well. You also have to modify `infrastructure-playbooks/purge-docker-cluster.yml` line 412.

The `ini_file` module behaves a bit oddly: It only sets `enabled` to `1` if such a key already exists. Is this desired behaviour? It doesn't matter, because the default is `1`, just something that surprised me ...

PR #1047, opened 2016-10-25
rolling_update: fix wrong indent

leseb:
Fixing: https://bugzilla.redhat.com/show_bug.cgi?id=1388295
Also add some notes in the README on how to run infrastructure
playbooks.

Signed-off-by: Sébastien Han seb@redhat.com

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,16 @@ site.yml
 site-docker.yml
 *.pyc
 *.sw?
+purge-cluster.yml
+shrink-mon.yml
+cluster-maintenance.yml
+purge-docker-cluster.yml
+shrink-osd.yml
+cluster-os-migration.yml
+purge-multisite.yml
+switch-from-non-containerized-to-containerized-ceph-daemons.yml
+localrepo-site.yml.sample
+rgw-standalone.yml
+take-over-existing-cluster.yml
+osd-configure.yml
+rolling_update.yml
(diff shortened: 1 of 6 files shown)


leseb:
@font any issues with the removal of all the `../`? I've found it easier to simply move the file to the upper directory, what do you think?

font:
I'm okay with it, but I suspect many will not know that they need to copy the playbooks into ceph-ansible's root directory and won't read the README. I'm curious as to why https://bugzilla.redhat.com/show_bug.cgi?id=1388295 is experiencing this issue. Perhaps using the `"{{ playbook_dir }}"` instead of `..` would help.

After taking a closer look at the bug, it looks like there may be some spacing issues that are triggering the problem. The bug reports:

```
> ansible-playbook rolling_update.yml 
ERROR: Syntax Error while loading YAML script, rolling_update.yml
Note: The error may actually appear before this position: line 31, column 1


vars:
^
```
I think this error may be due to the spacing where we are implicitly starting another play without proper indentation as shown here https://github.com/ceph/ceph-ansible/blob/master/infrastructure-playbooks/rolling_update.yml#L31-L59. If we indent just that play over and add `- name: ...` then it may fix the bug.

font:
@leseb I've opened up PR https://github.com/ceph/ceph-ansible/pull/1049 to try to address https://bugzilla.redhat.com/show_bug.cgi?id=1388295 as well as running with relative paths. Let me know what you think.

PR #1051, opened 2016-10-25
infrastructure-playbooks: fix syntax errors in all playbooks

andrewschoen:
Signed-off-by: Andrew Schoen <aschoen@redhat.com>

PR #1053, opened 2016-10-26
Fix tag list throughout Docker playbooks

verdurin:
This should (finally) fix the issues with the original PR for enabling CentOS extras. Apologies for the disruption this caused.

Signed-off-by: Adam Huffman <bloch@verdurin.com>

### DIFF
--- a/infrastructure-playbooks/purge-docker-cluster.yml
+++ b/infrastructure-playbooks/purge-docker-cluster.yml
@@ -409,7 +409,7 @@
   become: true
 
   tags:
-    with_pkg
+    - with_pkg
 
   tasks:
   - name: check if it is Atomic host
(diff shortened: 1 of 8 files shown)


alfredodeza:
Hi @verdurin we have added syntax checking for ansible playbooks for each PR and it looks like this PR failed that one. Please ensure that is fixed.

verdurin:
@alfredodeza the syntax check reports a problem in a file that this PR doesn't touch. Are you sure it's a problem here and not elsewhere?

alfredodeza:
@verdurin we got the syntax failures fixed with #1051 that just got merged. Can you rebase this PR?

andrewschoen:
It looks like all of these ``pre_requisite.yml`` files have all of the tasks tagged with ``with_pkg``. If that's the case you could just tag the include task that imports these files and then every task inside the file will inherit that tag.

verdurin:
Okay, I'll do that in a separate PR. Thanks for your patience with this.

PR #1068, opened 2016-11-03
moving to ansible v2.2 compatibility

leseb:

### DIFF
--- a/roles/ceph-mon/tasks/docker/checks.yml
+++ b/roles/ceph-mon/tasks/docker/checks.yml
@@ -13,7 +13,7 @@
 - name: stat for ceph config and keys
   stat:
     path: "{{ item }}"
-  with_items: ceph_config_keys
+  with_items: "{{ ceph_config_keys }}"
   changed_when: false
   failed_when: false
   register: statleftover
@@ -22,6 +22,6 @@
   fail:
     msg: "looks like no cluster is running but ceph files are present, please remove them"
   with_together:
-    - ceph_config_keys
-    - statleftover.results
+    - "{{ ceph_config_keys }}"
+    - "{{ statleftover.results }}"
   when: item.1.stat.exists == true
--- a/contrib/splitup.yml
+++ b/contrib/splitup.yml
@@ -45,14 +45,14 @@
       command: git subtree split --prefix=roles/{{ item }} -b {{ item }} --squash
       args:
         chdir: "{{ basedir }}"
-      with_items: roles
+      with_items: "{{ roles }}"
 
     - name: adds remote github repos for the splits
       tags: split
       command: git remote add {{ item }} git@github.com:{{ github }}-{{ item }}.git
       args:
         chdir: "{{ basedir }}"
-      with_items: roles
+      with_items: "{{ roles }}"
 
     - name: adds upstream remote
       tags: update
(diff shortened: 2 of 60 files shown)


leseb:
@ktdreyer @alfredodeza @andrewschoen 
@samdoran do we also need to do the `with_together` loops?

andrewschoen:
test this please

andrewschoen:
Looks good to me, I've changed the CI to use ansible==2.2 for that test I just started up.

andrewschoen:
All the tests seem to be failing at:

```
TASK [ceph-mon : stat for ceph config and keys] ********************************
task path: /home/jenkins-build/build/workspace/ceph-ansible-pull-requests/DIST/ceph_ansible_pr_centos7/ceph-ansible/roles/ceph-mon/tasks/docker/checks.yml:13
fatal: [localhost]: FAILED! => {
    "failed": true, 
    "msg": "'ceph_config_keys' is undefined"
}
```

https://jenkins.ceph.com/job/ceph-ansible-pull-requests/792/DIST=ceph_ansible_pr_centos7/console

roles/ceph-mon/tasks/docker/checks.yml
```diff
@@ -13,7 +13,7 @@
 - name: stat for ceph config and keys
   stat:
     path: "{{ item }}"
-  with_items: ceph_config_keys
+  with_items: "{{ ceph_config_keys }}"
```
samdoran: You can also use `default([])` inline. This has the effect of [skipping the task](http://docs.ansible.com/ansible/playbooks_conditionals.html#loops-and-conditionals) if the variable is undefined.

```yaml
with_items: " {{ cepth_config_keys | default([]) }}"
```

samdoran: Very true!

leseb:
@andrewschoen , @ktdreyer @alfredodeza Alright, it's a critical move, once we have this we need to branch out for "1.9.4 to 2.1.x" and explain this won't bring more features into the other branch, just bug fixes.

jfcoz:
Signed-off-by: Julien Francoz <julien@francoz.net>

leseb:
thanks @jfcoz

ktdreyer:
It would be a good idea to create a "version-1" branch at this point to maintain Ansible 1.9 compatibility, and then master can be tagged with "v2.0", "v2.1", etc.

leseb:
@ktdreyer I just pushed a branch called: ansible-1.9, how does that sound for you?
Can we merge this now?

ktdreyer:
@leseb, would you please clarify what tags we should expect to see on each branch, going forward?

For example, will "1.0.z" tags be on the ansible-1.9 branch? If so, what tags will be on master?

leseb:
@ktdreyer well going forward the ansible-1.9 branch will only receive bug fixes so I haven't thought about doing any tags for it. I'd like to keep our focus on 2.x, Ansible 2.x has been around for more than almost a year now, and most of the distro are shipping it so I believe it's fair.

In order to not confuse people and try to stay consistent I think next's master tag should be v2.x. What do you think?

andrewschoen:
I did a quick test and the ``ansible-1.9`` branch isn't compatible with ansible 1.9.4.

I get:

```
ERROR: yum_repository is not a legal parameter in an Ansible task or handler
```

ktdreyer:
> I think next's master tag should be v2.x. What do you think?

That's great, thanks for explaining.

leseb:
@andrewschoen thanks for testing, do you mind having a closer look and submit a fix directly into the branch?
Thanks!

@ktdreyer np! Let's wait a bit until we have a working 1.9 branch then we will merge this one.

PR #1072, opened 2016-11-04
Update .gitignore to ignore emacs auto-save files

mcv21:
We're using this to deploy a ceph cluster, and it would be nice to not have git keep telling me about emacs auto-save files :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ rgw-standalone.yml
 take-over-existing-cluster.yml
 osd-configure.yml
 rolling_update.yml
+*~


alfredodeza:
Actually, it might be nicer if we could ignore `.gitignore` so that you can add all kinds of ignores that are specific to your needs?

mcv21:
I think providing a useful `.gitignore` is useful; not sure that then ignoring it will avoid confusion. Maybe see if you start getting a lot of PRs of this form?

alfredodeza:
I would prefer not to get custom ignores in it. Custom ignores (custom for us in this case, not you) should either go in your global gitignore unless the current .gitignore is also ignored and custom ignores don't need to make it to the repo.

Can't you add this ignore in your global settings? Or do you need to see those emacs files in other projects?

leseb:
I don't have a strong opinion on this one and it's not like we are changing it everyday.
If this patch makes it easier for developers to send PRs then I'm for it.

mcv21:
[FWIW, Ansible itself has *~ in .gitignore ; I don't want to annoy you by arguing endlessly about this, but emacs is pretty widely-used, and it's been a while since a project I forked didn't ignore *~]

leseb:
@mcv21 this is mainly due to the fact I'm a VIM user I guess :).
@alfredodeza is that a big issue for you?

If I were an emac user I'd have done that ignore already :).

alfredodeza:
I would only add this if it is useful to have them visible in other projects/environments (I don't think that is the case). 

For ignores that should be across the board for every project (not specific to this one) I suggest adding them to the user's personal global gitignore file (regardless of Vim/Emacs)

whatevsz:
Quoting from [the official git docs](https://git-scm.com/docs/gitignore):
> 
> Which file to place a pattern in depends on how the pattern is meant to be used.
> 
> * Patterns which should be version-controlled and distributed to other repositories via clone (i.e., files that all developers will want to ignore) should go into a .gitignore file.
> 
> * Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the $GIT_DIR/info/exclude file.
> 
> * Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

So I'm with @alfredodeza here: Either put the emacs ignore patterns into the global gitignore, or into `$GIT_DIR/info/exclude` (which is local to your repo and not shared with others).

leseb:
Alright looks like we have a quorum, sorry about that @mcv21

PR #1075, opened 2016-11-05
ensure there is a unique container name for RGW container

castlemilk:
Both the  monitor and RGW roles create a container with name {{ ansible_hostname}}, this will cause a clash if they are deployed on the same host.

Proposed correction is to add a unique identifier to the container name for the RGW container, alternatively the monitor container could also have a unique name appended such as {{ ansible_hostname }}-mon

### DIFF
--- a/roles/ceph-rgw/templates/ceph-rgw.service.j2
+++ b/roles/ceph-rgw/templates/ceph-rgw.service.j2
@@ -4,8 +4,8 @@ After=docker.service
 
 [Service]
 EnvironmentFile=-/etc/environment
-ExecStartPre=-/usr/bin/docker stop {{ ansible_hostname }}
-ExecStartPre=-/usr/bin/docker rm {{ ansible_hostname }}
+ExecStartPre=-/usr/bin/docker stop {{ ansible_hostname }}-rgw
+ExecStartPre=-/usr/bin/docker rm {{ ansible_hostname }}-rgw
 ExecStart=/usr/bin/docker run --rm --net=host \
    {% if not rgw_containerized_deployment_with_kv -%}
    -v /var/lib/ceph:/var/lib/ceph \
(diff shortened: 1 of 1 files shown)


leseb:
Well if you create multiple RGWs on the same host they will all be named  `{{ ansible_hostname}}-rgw`. Where is your unique identifier?

castlemilk:
Well if you'd like to account for that case would something like {{ ansible_hostname }}-rgw-{{ ansible_hostname | to_uuid }} be appropriate? I can see why you'd like to avoid this so you can easily reference a container when executing commands on containers i.e. _#docker exec {ansible_hostname} ceph -s_. But then you're back to probably not requiring a name and can use the container ID.

Alternatively you could build out some conditional logic which checks to see if your playbook will be deploying more than one container on the target host and take the necessary action to ensure unique names. This way you could keep your friendly names in particular deployments. 

Do you make assumptions about deployment topologies that I'm not aware of? I'm still getting familiar. Is it unusual to put a rgw and mon on the same host? Is it advised to put a rgw container where the osd containers are?

Thanks,
Ben

dang:
I don't think you can have multiple RGWs on the same host.  RGW creates a REST server, that runs on a globally configured port, and the second one will fail to start because the first one has the port open already.

castlemilk:
It's possible to standup multiple RGW's on the same host and changing _ceph_rgw_civetweb_port_ for each one, then potentially load balancing across them with something like HAproxy.

leseb:
It's fine having multiple rgw on the same machine. Please go with the UUID solution to preserve the uniqueness. It should be possible to re-use https://github.com/ceph/ceph-ansible/blob/master/infrastructure-playbooks/rgw-standalone.yml so we can run multiple containers of rgw, although this might break the idempotency since the "global/default" deployment won't know about the other rgws, it will only know the initial one. This might lead to potential issues for upgrades.

leseb:
@castlemilk any plan on taking this one further or should we close it?

PR #1077, opened 2016-11-06
Support containerized rolling update

font:
- Update rolling update playbook to support containerized deployments
  for mons, osds, mdss, and rgws
- Skip checking if existing cluster is running when performing a rolling
  update
- Fixed bug where we were failing to start the mds container because it
  was missing the admin keyring. The admin keyring was missing because
  it was not being pushed from the mon host to the ansible host due to
  the keyring not being available before running the copy_configs.yml
  task include file. Now we forcefully wait for the admin keyring to be
  generated before continuing with the copy_configs.yml task include file
- Skip pre_requisite.yml when running on atomic host. This technically
  no longer requires specifying to skip tasks containing the with_pkg tag
- Add missing variables to all.docker.sample
- Misc. cleanup

Signed-off-by: Ivan Font <ifont@redhat.com>

This was tested on CentOS Atomic host.

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -28,6 +28,7 @@
          invoking the playbook"
     when: ireallymeanit != 'yes'
 
+
 - name: gather facts and check the init system
   vars:
     mon_group_name:       mons
@@ -62,6 +63,7 @@
       command: grep -sq systemd /proc/1/comm
       register: is_systemd
 
+
 - name: upgrade ceph mon cluster
 
   vars:
@@ -80,7 +82,7 @@
     - name: set mon_host_count
       set_fact: mon_host_count={{ groups.mons | length }}
 
-    - debug: msg="WARNING - upgrading a Ceph cluster with only one monitor node ({{ inventory_hostname }})"
+    - debug: msg="WARNING - upgrading a ceph cluster with only one monitor node ({{ inventory_hostname }})"
       when: mon_host_count | int == 1
 
     - name: stop ceph mons with upstart
@@ -101,7 +103,9 @@
         name: ceph-mon@{{ ansible_hostname }}
         state: stopped
         enabled: yes
-      when: is_systemd
+      when:
+        - is_systemd
+        - not mon_containerized_deployment
 
   roles:
     - ceph-mon
@@ -125,7 +129,18 @@
         name: ceph-mon@{{ ansible_hostname }}
         state: started
         enabled: yes
-      when: is_systemd
+      when:
+        - is_systemd
+        - not mon_containerized_deployment
+
+    - name: restart containerized ceph mons with systemd
+      service:
+        name: ceph-mon@{{ ansible_hostname }}
+        state: restarted
+        enabled: yes
+      when:
+        - is_systemd
+        - mon_containerized_deployment
 
     - name: set mon_host_count
       set_fact: mon_host_count={{ groups.mons | length }}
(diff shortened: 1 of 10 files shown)


infrastructure-playbooks/rolling_update.yml
```diff
@@ -110,10 +101,21 @@
         enabled: yes
       when: is_systemd
 
-    - name: select a running monitor
+    - name: set mon_host_count
+      set_fact: mon_host_count={{ groups.mons | length }}
+
```
leseb: can we prompt a message if `mon_host_count=1`? Something like: "are you sure you want to continue?"

infrastructure-playbooks/rolling_update.yml
```diff
@@ -123,6 +125,17 @@
       retries: "{{ health_mon_check_retries }}"
       delay: "{{ health_mon_check_delay }}"
       delegate_to: "{{ mon_host }}"
+      when: not mon_containerized_deployment
+
+    - name: waiting for the containerized monitor to join the quorum...
+      shell: |
+        docker exec {{ hostvars[mon_host]['ansible_hostname'] }} ceph -s  --cluster {{ cluster }} | grep quorum | sed 's/.*quorum//' | egrep -sq {{ ansible_hostname }}
```
leseb: remove the extra space between `-s` and `--cluster`.

infrastructure-playbooks/rolling_update.yml
```diff
@@ -156,41 +162,46 @@
...
+      when:
+        - is_systemd
+        - not osd_containerized_deployment
+
+    - name: restart containerized ceph osds with systemd
+      service:
+        name: ceph-osd@{{ item | basename }}
+        state: restarted
+        enabled: yes
+      with_items: ceph_osd_docker_devices
```
leseb: please use ansible 2.2 syntax= `"{{ ceph_osd_docker_devices }}"`

infrastructure-playbooks/rolling_update.yml
font: @leseb How would you like this handled? I don't think Ansible supports prompting at the task level nor having conditional prompts like this.

Also, what is the Ceph concern with updating the only monitor host? Some kind of inadvertent impact to your cluster? I can print a debug warning and/or add some sort of generic single mon host message when we initially prompt the user to confirm. But aside from that, I'm not sure what else we can do.

leseb: Oops I missed that, yes we cannot do a prompt on a `when`, a debug message seems enough then.

font:
test this please

leseb:
please rebase

font:
@leseb done

infrastructure-playbooks/rolling_update.yml
```diff
@@ -76,15 +79,11 @@
   become: True
 
   pre_tasks:
-    - include_vars: roles/ceph-common/defaults/main.yml
```
font: These are just redundant because we're using roles. Defining the roles will automatically include these default files.

leseb: Correct, however we currently have a bug.
Discovered by @guits yesterday, the `ansible.cfg` file is honoured on 2.2, see: https://github.com/ansible/ansible/issues/18090

We won't fix this with this commit, we just need to be aware of that and potentially offer a fix in a separate PR.

font: That appears to only be a problem for relative paths without `role` as the directory name. It may be a problem if we're running the `rolling_update.yml` playbook from within `infrastructure-playbooks` as I have not tried that recently, but I thought your most recent suggestion via this [commit](https://github.com/ceph/ceph-ansible/blob/b0989c700f9a4651b2cae978653d255ef0e0b4d9/infrastructure-playbooks/README.md) was to copy them into the root directory and run it from there.

leseb: I think this commit was before your change that introduced the `ansible.cfg` thing.

PR #1080, opened 2016-11-07
tests: remove validation yaml playbooks (placeholders only)

alfredodeza:
I've left the rados ones since those seem to actually do something. Once we port those to actual tests they should be removed as well.


leseb:
Nah please remove everything, I think we need to use teuthology on some portion of the ceph functional test to validate the deployment, I don't wan to re-invent the well.

leseb:
@alfredodeza please remove rados as well

PR #1081, opened 2016-11-07
Fix with_ 'bare variable' handling

albertomurillo:
In ansible 2.2 variable names should be templated {{ }}
when used as an argument to a with_ statement.

Otherwise they will be used as plain text variables.

Signed-off-by: Alberto Murillo Silva <alberto.murillo.silva@intel.com>


albertomurillo:
This is a dup of https://github.com/ceph/ceph-ansible/pull/1068 which i didnt spot earlier :)

PR #1086, opened 2016-11-08
testing: adds 3 new functional testing scenarios

andrewschoen:
This adds a few more testing scenarios:

- a 3 mon xenial cluster that is useful for testing ceph.conf creation.

- a 1 mon, 1 osd xenial cluster testing the journal collocation scenario

- a 1 mon, 1 osd centos7 cluster testing the journal collocation scenario

The 3 mon xenial scenario I used to test some reformatting of the ceph.conf template, which is included here as well. I also fixed a bug in ``mon hosts`` where if ``monitor_address`` was set for the current node then that value would be used for all monitors.

These can be run locally by ``tox -rv`` at the root of ceph-ansible, the only caveat being these vagrant boxes only support virtualbox.


roles/ceph-common/defaults/main.yml
```diff
@@ -244,6 +244,7 @@ rbd_client_admin_socket_path: /var/run/ceph # must be writable by QEMU and allow
 monitor_interface: interface
 monitor_address: 0.0.0.0
 mon_use_fqdn: false # if set to true, the MON name used will be the fqdn in the ceph.conf
+monitor_address_block: false
```
alfredodeza: what is this for?

alfredodeza: Ah, I see it defined later, but I'm still not sure what is going on with that :(

tests/functional/centos/7/mon-osd/hosts
```diff
@@ -0,0 +1,5 @@
+[mons]
+mon0 monitor_interface=enp0s8
```
alfredodeza: excellent. Do you know if it is possible to add comments here? It would be nice to say why `enp0s8` is needed

roles/ceph-common/defaults/main.yml
andrewschoen: That's a feature that was added to the ceph.conf. However, it fails on an OSD to create the ceph.conf unless this is defaulted.

tests/functional/centos/7/mon-osd/hosts
andrewschoen: There probably is, I can look into it. It'd be nice to comment the tox.ini as well with some details on each scenario.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = {ansible2.1,ansible2.2}-{initial-members}
+envlist = {ansible2.1}-{xenial-conf-tests,xenial-mon-osd,centos7-mon-osd}
```
andrewschoen: Yeah, I thought of that as well. The ``envlist`` is just used with you don't pass the ``-e`` flag though, we can be descriptive with our local testing and CI testing using ``-e`` for now.

PR #1088, opened 2016-11-10
Correct syntax of rgw frontends line for civetweb

mcv21:
The correct configuration of the civetweb port is:
rgw frontents = "civetweb port=80"

rather than ipaddress:port. Our civetwebs wouldn't work until this was
corrected.

### DIFF
--- a/roles/ceph-common/templates/ceph.conf.j2
+++ b/roles/ceph-common/templates/ceph.conf.j2
@@ -114,7 +114,7 @@ rgw socket path = /tmp/radosgw-{{ hostvars[host]['ansible_hostname'] }}.sock
 log file = /var/log/ceph/{{ cluster }}-rgw-{{ hostvars[host]['ansible_hostname'] }}.log
 rgw data = /var/lib/ceph/radosgw/{{ cluster }}-rgw.{{ hostvars[host]['ansible_hostname'] }}
 {% if radosgw_frontend  == 'civetweb' %}
-rgw frontends = civetweb port={{ radosgw_civetweb_bind_ip }}:{{ radosgw_civetweb_port }} num_threads={{ radosgw_civetweb_num_threads }}
+rgw frontends = "civetweb port={{ radosgw_civetweb_port }}"
 {% endif %}
 {% if radosgw_keystone %}
 rgw keystone url = {{ radosgw_keystone_url }}


leseb:
Ceph version?

mcv21:
Sorry, should have said - jewel.

OrFriedmann:
@mcv21 Did you try to `curl <bind_ip>:<port>` (from the ceph.conf) and it didn't worked?
if you bind to only one of the ips of the rgw you would not be able to interact with rgw on another ip and on another port.

mcv21:
Well, I used `telnet ip port` but yes, I tested that, from the machine with the rgw on.

leseb:
It works for me on Jewel Ubuntu 16.04:

```
root@ceph-rgw0:~# systemctl status ceph-radosgw@rgw.ceph-rgw0
● ceph-radosgw@rgw.ceph-rgw0.service - Ceph rados gateway
   Loaded: loaded (/lib/systemd/system/ceph-radosgw@.service; enabled; vendor preset: enabled)
   Active: active (running) since Mon 2016-11-14 14:02:18 UTC; 1min 57s ago
 Main PID: 24182 (radosgw)
   CGroup: /system.slice/system-ceph\x2dradosgw.slice/ceph-radosgw@rgw.ceph-rgw0.service
           └─24182 /usr/bin/radosgw -f --cluster ceph --name client.rgw.ceph-rgw0 --setuser ceph --setgroup ceph

Nov 14 14:02:18 ceph-rgw0 systemd[1]: Started Ceph rados gateway.
Nov 14 14:02:24 ceph-rgw0 radosgw[24182]: error parsing int: 10.0.2.15:8080: The option value '10.0.2.15:8080' seems to be invalid


root@ceph-rgw0:~# grep front /etc/ceph/ceph.conf
rgw frontends = civetweb port=10.0.2.15:8080 num_threads=50

root@ceph-rgw0:~# netstat -ptnl | grep radosgw
tcp        0      0 10.0.2.15:8080          0.0.0.0:*               LISTEN      24182/radosgw

root@ceph-rgw0:~# curl http://10.0.2.15:8080
<?xml version="1.0" encoding="UTF-8"?><ListAllMyBucketsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Owner><ID>anonymous</ID><DisplayName></DisplayName></Owner><Buckets></Buckets></ListAllMyBucketsResult>
```


So yes you have the misleading message from radosgw saying that  `ip:port` seems to be an invalid configuration, however it still works.

mcv21:
I think the error message is correct - we should be passing an int to civetweb's port argument. So even if sometimes it works despite the incorrect syntax, isn't correcting the syntax (which will stop the error message) the way to go?

leseb:
@mcv21 well for me it's not *sometimes* it just works, even with the error message...

alfredodeza:
jenkins test this please

alfredodeza:
jenkins test this please

alfredodeza:
jenkins test this please

mcv21:
Right, I think I understand better now.

The error message comes from src/rgw/rgw_main.cc line 392 in 16.04 [l456 in ceph git](https://github.com/ceph/ceph/blob/master/src/rgw/rgw_main.cc#L456) where ceph calls get_val (string,int,int) which produces the error message observed (and fails). 

Subsequently, rgw_civetweb_frontend.cc line 39 [l53 in ceph git](https://github.com/ceph/ceph/blob/master/src/rgw/rgw_civetweb_frontend.cc#L53) uses the string version of get_val which succeeds and stores the string value verbatim in what will become the listening_ports argument to civetweb; which is happy with a string of the form you use.

Sorry for the noise; I think the issue remains that you need some way of specifying just a port (or port list), though?

leseb:
@mcv21 thanks for the details.

True for the port-only, I was actually thinking to use a variable for the whole content so users can specify what they want. Something like:

`rgw frontends: "civetweb port={{ radosgw_civetweb_bind_ip }}:{{ radosgw_civetweb_port }} num_threads={{ radosgw_civetweb_num_threads }}"`

Since it's var you can easily remove `radosgw_civetweb_port` and `radosgw_civetweb_num_threads` if you want.

I think we can address this in #1104

leseb:
I think we reached an agreement on this one so I'm closing this PR and will followup.

PR #1090, opened 2016-11-10
Fix user key creation.

whatevsz:
With `ansible-playbook 2.2.0.0`, user key creation fails because ansible interprets the `with_items` key as a string instead of a jinja variable. This pull request fixes that issue.

Please be aware that I'm not confident that this is backwards-compatible with older versions of `ansible-playbook`. Can someone confirm or deny this?

UPDATE:

Running `grep -r 'with_items: [^"]'` on the source tree reveals that this is a common issue. I'd fix this if desired.


whatevsz:
I just saw that this duplicates #1068, so please ignore.

PR #1092, opened 2016-11-14
Use new ceph vagrant boxes

alfredodeza:
And disable the sync folder on all the functional testing (optionally set on vagrant_variables.yml)


vagrant_variables.yml
```diff
@@ -0,0 +1,70 @@
+---
```
andrewschoen: Did you mean to check in this change to ``vagrant_variables.yml.sample``?

alfredodeza: ouch. Will fix

leseb:
Are all the new boxe available under this "ceph" namespace?

PR #1093, opened 2016-11-14
rolling update: stop services before upgrading and start afterwards

andrewschoen:
Services are restarted twice on upgrades because of the restarts in the ``ceph-common`` role and at the end of the upgrade process by the ``rolling_update.yml`` playbook. This sometimes exposes a bug which leaves the OSD process running but not marked up.

See:

    https://bugzilla.redhat.com/show_bug.cgi?id=1394928
    https://bugzilla.redhat.com/show_bug.cgi?id=1391675
    https://bugzilla.redhat.com/show_bug.cgi?id=1394929

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -353,6 +353,33 @@
   serial: 1
   become: True
 
+  pre_tasks:
+    - include_vars: roles/ceph-common/defaults/main.yml
+    - include_vars: roles/ceph-rgw/defaults/main.yml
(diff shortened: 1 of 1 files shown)


ktdreyer:
For reference, the exact bug this exposes is http://tracker.ceph.com/issues/17899

alfredodeza:
I don't see any `--cluster "{{ cluster }}"` usage here. Are we OK not supporting it? (Maybe it wasn't before this)

alfredodeza:
I don't think we do anything with service handlers and custom clusters. This looks OK to me.

PR #1102, opened 2016-11-17
Fix jinja defaulting syntax

Logan2211:
a397922 introduced a syntax error by attempting to default an unquoted
string, which causes execution failures on some ansible versions with:

Failed to template {{ ceph_rhcs_mount_path }}: Failed to template {{ ceph_stable_rh_storage_mount_path | default(/tmp/rh-storage-mount) }}: template error while templating string: unexpected '/'


ktdreyer:
@Logan2211 , which Ansible version(s) hit this error, for the record?

Logan2211:
@ktdreyer tested on stable-1.9 and 2.1.1.0. Only stable-1.9 is showing the fail.

Looking at the include line, I think this bug is being exposed because of 1.9's poor static include handling, so the file is being handled as a dynamic include instead, which then fails due to the broken jinja.

leseb:
test this please

alfredodeza:
jenkins test this please

alfredodeza:
jenkins test this please

alfredodeza:
jenkins test this please

leseb:
jenkins test this please

leseb:
@Logan2211 please rebase :)

PR #1104, opened 2016-11-17
remove apache support for RGW

ktdreyer:
libfcgi is dead upstream (http://tracker.ceph.com/issues/16784)
    
The RGW developers intend to remove libfcgi support entirely before the Luminous release.
    
Since libfcgi gets little-to-no developer attention or testing, remove it entirely from ceph-ansible.


roles/ceph-common/defaults/main.yml
```diff
@@ -277,7 +277,6 @@ mds_max_mds: 3
 ## Rados Gateway options
 #
 #radosgw_dns_name: your.subdomain.tld # subdomains used by radosgw. See http://ceph.com/docs/master/radosgw/config/#enabling-subdomain-s3-calls
-radosgw_frontend: civetweb # supported options are 'apache' or 'civetweb', also edit roles/ceph-rgw/defaults/main.yml
```
andrewschoen: The ``radosgw_frontend`` var is being used in the the ceph.conf template. Removing this will cause the template to fail to render.

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/templates/ceph.conf.j2#L116

andrewschoen:
It's a bit hard to see with all the output, but the tests did fail here: https://jenkins.ceph.com/job/ceph-ansible-pull-requests/8/consoleFull

With:
```

TASK [ceph.ceph-common : generate ceph configuration file] *********************
task path: /home/jenkins-build/build/workspace/ceph-ansible-pull-requests/roles/ceph-common/tasks/main.yml:210
[DEPRECATION WARNING]: ansible.utils.unicode.to_bytes is deprecated.  Use 
ansible.module_utils._text.to_bytes instead.
This feature will be removed in 
version 2.4. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: ansible.utils.unicode.to_unicode is deprecated.  Use 
ansible.module_utils._text.to_text instead.
This feature will be removed in 
version 2.4. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
fatal: [mon0]: FAILED! => {"failed": true, "msg": "'radosgw_frontend' is undefined"}
	to retry, use: --limit @/home/jenkins-build/build/workspace/ceph-ansible-pull-requests/site.yml.retry
```

roles/ceph-common/defaults/main.yml
ktdreyer: Thanks! I rebased and re-pushed with the `ceph.conf.j2` fix.

andrewschoen:
test this please

leseb:
Sorry to be picky but can we have a more meaningful commit message to explain why we are removing this? I think a bit more background would be valuable for users.

I also just hope that there is not too many users relying on this feature and that most of the people moved to civetweb...

roles/ceph-common/defaults/main.yml
leseb: Can we actually keep this and do some refactor?
I was thinking to use a variable to bring more flexibility to users, so we could have something like this:

`rgw frontends: "civetweb port={{ radosgw_civetweb_bind_ip }}:{{ radosgw_civetweb_port }} num_threads={{ radosgw_civetweb_num_threads }}"`

Potentially users can add more civetweb options and also decide to only specify a port and not a bind_ip.
So for we need to keep the existing variables:

* `radosgw_civetweb_bind_ip``
* `radosgw_civetweb_port``
* `radosgw_civetweb_num_threads``

For backward compatibility.

Of course we can address that into a new PR but I just don't want to delete and re-introduce something.

What do you guys think?

ktdreyer: I don't think we should add new features like that until we have CI tests beforehand.

ktdreyer:
@leseb I've updated the description to provide more context, and linked to a relevant tracker ticket.

If users are relying on libfcgi, they need to switch to civetweb ASAP. There are unresolved questions about its security in the distros, for example: https://bugs.launchpad.net/ubuntu/+source/libfcgi/+bug/933417/comments/5

leseb:
Thanks @ktdreyer :)

PR #1106, opened 2016-11-17
add RPM packaging for CI

ktdreyer:
`make rpm` will build a ceph-ansible RPM and place it in the current
working directory.

This will allow us to run this command in Jenkins for every branch.


leseb:
Cool stuff thanks @ktdreyer

PR #1109, opened 2016-11-21
rolling_update: clarify "serial" usage

leseb:
Prior to this commit the serial variable was poorly documented. Now we
are making clear that this value should be left untouched as the rolling
update mechanism should happen serially.

Solves: bz-1396742

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -1,6 +1,10 @@
 ---
 # This playbook does a rolling update for all the Ceph services
-# Change the value of 'serial:' to adjust the number of server to be updated.
+#
+# The value of 'serial:' adjusts the number of servers to be updated simultaneously.
+# We recommend a value of 1, which means hosts of a group (e.g: monitor) will be
+# upgraded one by one. It is really crucial for the update process to happen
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/rolling_update.yml
```diff
@@ -1,6 +1,10 @@
 ---
 # This playbook does a rolling update for all the Ceph services
-# Change the value of 'serial:' to adjust the number of server to be updated.
+#
+# The value of 'serial:' adjusts the number of server to be updated simultaneously.
```
alfredodeza: s/number of server/number of servers/

infrastructure-playbooks/rolling_update.yml
```diff
@@ -1,6 +1,10 @@
...
+# We recommend a value of 1, which means hosts of a group (ie: monitor) will be
+# upgraded one by one. It is really crucial for the update process to happen
+# in a serialized fashion. DO NOT CHANGE THIS VALUE.
```
alfredodeza: if we are encouraging serialized behavior why is this an option? Is there a way to hard code it?

infrastructure-playbooks/rolling_update.yml
leseb: Done

PR #1110, opened 2016-11-21
contrib: add a new script to update remote repos

leseb:
This script allows us to split ceph-ansible role directory into
repositories that can be pushed.
This script is an enhanced version of splitup.yml that was using git
subtree. I haven't found any straighforward way to push tag using git
subtree where git filter-branch seems easier and more practical.

Now we can not only push commits from master but also all the relative
tags. It is nice for projects that want to consume our roles from the
Ansible Galaxy, pointing to a specific tags looking for a particular
release.


contrib/splitup.sh
```diff
@@ -0,0 +1,47 @@
...
+
+for ROLE in $(ls -1 roles | grep -v ceph.ceph-common); do
+  check_existing_remote $ROLE
+  for BRANCH in master ansible-1.9; do
+    if [[ ! -f .git/refs/heads/$BRANCH ]]; then
+      git checkout -t origin/$BRANCH
+    else
+      git checkout $BRANCH
+      git reset --hard origin/$BRANCH
+    fi
```
Logan2211: I found L32-36 can be simplified to a single line: `git checkout -B $BRANCH origin/$BRANCH`.

-B will either checkout the remote branch, or it will hard reset the existing local branch to match the remote branch.

contrib/splitup.sh
```diff
@@ -0,0 +1,47 @@
...
+  done
+  for TAG in $(git tag | egrep '^v[2-9].[0-9].[0-9]$'); do
+    git filter-branch -f --prune-empty --subdirectory-filter roles/ceph-$ROLE $TAG
+  done
+  git push -f --all $ROLE
+  git push -f --tags $ROLE
+done
+
+# let's bring everything back to normal
+reset_hard_origin
```
Logan2211: I think you may need to run this around line 29 each time the loop starts so the tree will be reset without the previous filtered history applied.

contrib/splitup.sh
leseb: Done thanks!

contrib/splitup.sh
```diff
@@ -0,0 +1,45 @@
...
+function reset_hard_origin {
+  git checkout $LOCAL_BRANCH
+  git fetch origin
+  git reset --hard origin/master
+}
+
+# MAIN
+pull_origin
+
+for ROLE in $(ls -1 roles | grep -v ceph.ceph-common); do
```
ErwanAliasr1: I'd suggest using ls -d -1 as you are expecting directories. That avoid making wrong stuff if someone drop a file here.

contrib/splitup.sh
```diff
@@ -0,0 +1,45 @@
...
+  check_existing_remote $ROLE
+  reset_hard_origin
+  for BRANCH in master ansible-1.9; do
```
ErwanAliasr1: Please put that code is a function that indicate what you are trying to achieve here.

ErwanAliasr1:
As per the local discussion we had together and the changes we put, that looks +1 for me if tests are ok.

contrib/splitup.sh
```diff
@@ -0,0 +1,51 @@
...
+set -xe
+
+# VARIABLES
+BASEDIR=$(dirname "$0")
+LOCAL_BRANCH=$(cd $BASEDIR && git rev-parse --abbrev-ref HEAD)
+
+
+# FUNCTIONS
+function check_existing_remote {
+  if ! git remote show $1 &> /dev/null; then
```
Logan2211: This conditional is broken-- it always evaluates to true so the `git add` line causes a fatal error if the remote already exists, stopping the script since it is in `set -e`

Maybe use:
```
function check_existing_remote {
  git remote get-url $1 2>&1 > /dev/null
  if [[ $? -ne 0 ]]; then
    git remote add $1 git@github.com:/ceph/ansible-ceph-$1.git
  fi
}
```

Note this breakage shouldn't affect the execution of this script against a fresh repo in a CI build, but someone else trying to use this against the same repo multiple times would run into problems.

contrib/splitup.sh
```diff
@@ -0,0 +1,51 @@
...
+  # so we always make sure we 'push' to a remote and 'filter' a role
+  REMOTE=$ROLE
+  check_existing_remote $REMOTE
+  reset_hard_origin
+  # first we filter branches by rewritting master with the content of roles/ceph-$ROLE
+  # this gives us a new commit history
+  for BRANCH in master ansible-1.9; do
+    git checkout -B $BRANCH origin/$BRANCH
+    git filter-branch -f --prune-empty --subdirectory-filter roles/ceph-$ROLE $BRANCH
+    git push -f $REMOTE $BRANCH
```
Logan2211: Just something to consider but since the push here should always be iterative you shouldn't need a force push here. imo if a force push is needed here, it is probably something that should fail the job and get some human review.

contrib/splitup.sh
```diff
@@ -0,0 +1,51 @@
...
+  done
+  # then we filter tags starting from version 2.0 and push them
+  for TAG in $(git tag | egrep '^v[2-9].[0-9]*.[0-9]*$'); do
+    git filter-branch -f --prune-empty --subdirectory-filter roles/ceph-$ROLE $TAG
+    git push -f $REMOTE $TAG
```
Logan2211: Same as last comment

contrib/splitup.sh
leseb: are you sure about this?

```
#!/bin/bash
set -xe

function check_existing_remote {
  if ! git remote show $1 &> /dev/null; then
    echo "adding remote"
  fi
}

check_existing_remote ceph-agent
check_existing_remote a
```


Returns:

```
+ check_existing_remote ceph-agent
+ git remote show ceph-agent
+ check_existing_remote a
+ git remote show a
+ echo 'adding remote'
adding remote
```

Logan2211: Nevermind I was trying with a repo that doesn't exist:
+ check_existing_remote ceph-test
+ git remote show ceph-test
+ git remote add ceph-test git@github.com:/ceph/ansible-ceph-test.git
+ check_existing_remote ceph-test
+ git remote show ceph-test
ERROR: Repository not found.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
+ git remote add ceph-test git@github.com:/ceph/ansible-ceph-test.git
fatal: remote ceph-test already exists.

It works fine if the repo exists :)

alfredodeza:
jenkins test xenial-cluster

alfredodeza:
jenkins test xenial-cluster

alfredodeza:
jenkins test xenial-cluster

PR #1112, opened 2016-11-22
fix python3 compatibility in jinja2 template.

zhsj:
rename iterkeys to keys


leseb:
does this have any impact on the ansible version we are using?

zhsj:
@leseb no impact, dict.keys is available in both python2/3 always.

leseb:
Thanks!

Issue #1114, opened 2016-11-23
NTP checks not usable with '--check'


D3N14L:
The registered return codes of the "check ntp installation on [atomic|debian|redhat]" tasks are required in the following execution of the ceph-common playbook, but the tasks are not run in check mode. I is safe to run them always, because they are only reading installed packages. I will provide a pull request adding the missing always_run parameters to the tasks.

D3N14L:
this issue is fixed then :)

PR #1115, opened 2016-11-23
Make NTP checks usable with '--check'

D3N14L:
As those task are only reading the installed packages and their output
is required in later tasks it is safe to run them also in check mode.


D3N14L:
This pull request fixes issue #1114

leseb:
Thanks!

PR #1117, opened 2016-11-24
Improve --check compatibility for ceph-common, ceph-mon and ceph-rgw roles

D3N14L:
Adding a few `always_run: true` lines on tasks that execute reading operations to fill variables. Making it possible to run a `ansible-playbook --check` without running into errors on undefined variables.

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -169,12 +169,14 @@
   shell: "stat /var/run/ceph/*.asok > /dev/null 2>&1"
   changed_when: false
   failed_when: false
+  always_run: true
   register: socket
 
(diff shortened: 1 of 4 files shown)


D3N14L:
Checks failed. Looks like Jenkins has a problem:
`The following script '/usr/bin/vagrant up --no-provision --provider=libvirt' failed`

leseb:
Thanks for this, did you also check the other roles?

D3N14L:
Additional to the ones in the commit I only checked the ceph-osd role. I worked without modification (at least for my setup). I will try to have a more systematic look at the roles and provide additional pull requests.

leseb:
@D3N14L alright, thanks!

PR #1118, opened 2016-11-24
[shrink-mon]: force playbook to fail if there is only one mon

### LINKED ISSUE #1083: shrink-mon: add a warning if there is only one mon

guits:
The playbook will fail if only 1 mon is in the cluster
and advise to use the `purge-cluster` playbook instead.

Fix #1083

### DIFF
--- a/infrastructure-playbooks/shrink-mon.yml
+++ b/infrastructure-playbooks/shrink-mon.yml
@@ -30,6 +30,12 @@
   - include_vars: roles/ceph-common/defaults/main.yml
   - include_vars: group_vars/all.yml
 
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/shrink-mon.yml
```diff
@@ -30,6 +30,12 @@
   - include_vars: roles/ceph-common/defaults/main.yml
   - include_vars: group_vars/all.yml
 
+  - name: exit playbook, if only one mon is present in cluster
+    fail:
+      msg: "You are about to shrink the only mon present in the cluster.
```
leseb: let's use "monitor" instead of "mon", thanks!

PR #1120, opened 2016-11-26
Add multi-cluster support to Vagrantfile

jsaintrocc:
This allows you to setup a second cluster, with non-conflicting IPs, by changing the cluster_num variable from 0 to 1.   Very useful for testing multi-cluster functionality.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -29,6 +29,7 @@ MEMORY         = settings['memory']
 ETH            = settings['eth']
 DOCKER         = settings['docker']
 USER           = settings['ssh_username']
+CLUSTER_NUM    = settings['cluster_num']
 
 ASSIGN_STATIC_IP = !(BOX == 'openstack' or BOX == 'linode')
 DISABLE_SYNCED_FOLDER = settings.fetch('vagrant_disable_synced_folder', false)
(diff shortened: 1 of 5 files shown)


leseb:
@jsaintrocc Thanks!

leseb:
jenkins test this please

leseb:
jenkins test this please

alfredodeza:
This is failing our tests because we rely on the IP settings to avoid the conflicts. Our test scenarios "bump" the IP address so there is never a conflict on the `vagrant_variables.yml` file.

I think that the use case for this PR is solved with changing the IPs. No need to have another variable.

jsaintrocc:
@alfredodeza If you mean change the subnets the problem is that the clusters won't be able to communicate with one another.  By adding the variable you are creating the clusters in the same subnet so you can test stuff like multisite replicating etc...

jsaintrocc:
@alfredodeza I stand corrected.  I tested changing the IP subnets of a 2nd cluster and surprisingly VirtualBox lets one subnet talk to the other (at least on OS X).  I didn't VirtualBox would let host-only networks could talk to one another.  Do you know if this is the case for VirtualBox in Windows?

leseb:
@jsaintrocc I'm not running Windows. I'm not sure to follow, are you saying we can run 2 clusters side by side just by changing the IPs? Won't this rebuild the VMs with another subnet instead?

I think I'm missing something here.

jsaintrocc:
@leseb You would need to have to separate copies of the ceph-ansible repo.  That way each is maintaining it's own vagrant/ceph config etc...

leseb:
@jsaintrocc gotcha, so if we need 2 copies, why do we need this patch then? :)

PR #1121, opened 2016-11-27
Reviewed all roles and site.yml.sample and added "always_run: true" where appropriate

D3N14L:
I went through all the roles and looked out for tasks that may safely be executed in (or are required to also run in) check mode.


leseb:
Thanks @D3N14L

PR #1123, opened 2016-11-28
contrib: add a goto basedir function

leseb:
This way if the script is not executed from the git directory we jump
into it, do our stuff and then go back to the dir where the user was.


alfredodeza:
This looks OK to me

PR #1127, opened 2016-11-28
testing dmcrypt OSD scenarios

andrewschoen:
This adds functional testing scenarios for both the ``dmcrypt_dedicated_journal`` and the ``dmcrypt_journal_collocation`` OSD scenarios.

This also renames the ``centos7-mon-osd`` scenario to ``raw-multi-journal`` to better define what that scenario is testing.


tests/functional/centos/7/raw-multi-journal/scenario.py
```diff
@@ -0,0 +1,19 @@
...
+# tests to get. Each one of these can get requested as a py.test fixture to
+# validate information.
+nodes = {
+    'mon0': {
+        'username': 'vagrant',
+        'components': ['mon', 'mon_initial_members']
+    },
+    'osd0': {
+        'username': 'vagrant',
+        'components': [],
```
alfredodeza: you may want to add specific tests for this scenario(s). For collocated journals for example you would need to ensure that the partitions are there on the device, and probably other things as well

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = {ansible2.2}-{xenial-conf-tests,xenial-mon-osd,xenial-cluster,centos7-mon-osd,centos7-cluster}
+envlist = {ansible2.2}-{xenial-conf-tests,xenial-mon-osd,xenial-cluster,raw-multi-journal,centos7-cluster,dmcrypt-dedicated-journal,dmcrypt-journal-collocation}
```
alfredodeza: can we do `journal` and `journal-collocated` ? I think that we should assume dedicated if not specified, and always specify 'collocated' when needed

andrewschoen: I've named them this way so they match up with the name of the actual config setting, but I can change them if you feel strongly. If you want to assume a dedicated journal on these tests then we might want to change ``xenial-cluster`` and ``centos7-cluster`` to use a dedicated journal and change the ``raw-multi-journal`` scenario to ``journal-collocated``.

andrewschoen: Ok, I've got all these renamed now. I've also made ``centos7-cluster`` and ``xenial-cluster`` use a dedicated journal.

PR #1128, opened 2016-11-29
Adding choose_disk module

ErwanAliasr1:
This PR is about adding a module that allow detecting disks by their features rather than by their path.

### DIFF
--- a/choose_disk.yml
+++ b/choose_disk.yml
@@ -1,0 +1,17 @@
+---
+# Defines deployment design and assigns role to server groups
+
+- hosts: localhost
+  gather_facts: false
+  vars:
+      lookup_disks: "{'storage_disks': {'model': 'SAMSUNG MZ7LN512', 'rotational': '0', 'count': 2 }}"
+  tasks:
+    # If we can't get python2 installed before any module is used we will fail
+    # so just try what we can to get it installed
+    - name: gathering facts
+      setup:
+    - name: choose disk
+      choose_disk:
+        facts: "{{ansible_devices}}"
+        disks: "{{lookup_disks}}"
+
--- a/choose_disk.yml
+++ b/choose_disk.yml
@@ -3,10 +3,22 @@
 
 - hosts: localhost
   gather_facts: false
+  become: True
   vars:
-      lookup_disks: "{'storage_disks': {'model': 'SAMSUNG MZ7LN512', 'rotational': '0', 'count': 2 }}"
+      fsid: "e8ba1fe2-dbf6-4b15-b0ca-2c5ffd97f59e"
+      public_network: 172.16.41.0/24
+      cluster_network: 172.16.41.0/24
+      journal_size: 100
+      journal_collocation: true
+      devices: []
+      cluster: ceph
+      osd_objectstore: filestore
+      osd_group_name: osds
+      #lookup_disks: "{'storage_disks': {'model': 'SAMSUNG MZ7LN512', 'rotational': '0',  'ceph_type' : 'data', 'count': 2 }}"
+      lookup_disks: "{'storage_disks': {'vendor': '0x1af4', 'ceph_type' : 'data', 'size': 'gte(30 GB)', 'count': 2 }}"
+      legacy_devices: []
 #      legacy_lookup_disks: "/dev/sda /dev/sdb /dev/sdc"
-  tasks:
+  pre_tasks:
     # If we can't get python2 installed before any module is used we will fail
     # so just try what we can to get it installed
     - name: gathering facts
@@ -16,4 +28,10 @@
         facts: "{{ansible_devices}}"
         disks: "{{lookup_disks}}"
 #       legacy_disks: "{{legacy_lookup_disks}}"
-
+    - debug: var=storage_devices
+    - debug: var=journal_devices
+    - debug: var=legacy_devices
+    - debug: var=legacy_raw_journal_devices
+    - debug: var=devices_to_activate
+  roles:
+    - ceph-osd
--- a/choose_disk.yml
+++ b/choose_disk.yml
@@ -14,10 +14,17 @@
       cluster: ceph
       osd_objectstore: filestore
       osd_group_name: osds
-      #lookup_disks: "{'storage_disks': {'model': 'SAMSUNG MZ7LN512', 'rotational': '0',  'ceph_type' : 'data', 'count': 2 }}"
-      lookup_disks: "{'storage_disks': {'vendor': '0x1af4', 'ceph_type' : 'data', 'size': 'gte(30 GB)', 'count': 2 }}"
-      legacy_devices: []
-#      legacy_lookup_disks: "/dev/sda /dev/sdb /dev/sdc"
+#      devices: {'storage_disks': {'vendor': '0x1af4', 'ceph_type' : 'data', 'size': 'gte(30 GB)', 'count': 2 }}
+#      devices:
+#       storage_disks:
+#          model: 'SAMSUNG MZ7LN512'
+#          rotational: '0'
+#          ceph_type: 'data'
+#          count: 2
+#     devices: [ '/dev/sda', '/dev/sdb', '/dev/sdc' ]"
+#     devices:
+#       - /dev/sda
+#       - /dev/sdb
   pre_tasks:
     # If we can't get python2 installed before any module is used we will fail
     # so just try what we can to get it installed
@@ -25,9 +32,7 @@
       setup:
     - name: choose disk
       choose_disk:
-        facts: "{{ansible_devices}}"
-        disks: "{{lookup_disks}}"
-#       legacy_disks: "{{legacy_lookup_disks}}"
+        vars: "{{vars}}"
     - debug: var=storage_devices
     - debug: var=journal_devices
     - debug: var=legacy_devices
--- a/choose_disk.yml
+++ b/choose_disk.yml
@@ -5,6 +5,7 @@
   gather_facts: false
   vars:
       lookup_disks: "{'storage_disks': {'model': 'SAMSUNG MZ7LN512', 'rotational': '0', 'count': 2 }}"
+#      legacy_lookup_disks: "/dev/sda /dev/sdb /dev/sdc"
   tasks:
     # If we can't get python2 installed before any module is used we will fail
     # so just try what we can to get it installed
@@ -14,4 +15,5 @@
       choose_disk:
         facts: "{{ansible_devices}}"
         disks: "{{lookup_disks}}"
+#       legacy_disks: "{{legacy_lookup_disks}}"
 
--- a/choose_disk.yml
+++ b/choose_disk.yml
@@ -36,7 +36,6 @@
     - debug: var=storage_devices
     - debug: var=journal_devices
     - debug: var=legacy_devices
-    - debug: var=legacy_raw_journal_devices
     - debug: var=devices_to_activate
   roles:
     - ceph-osd
--- a/library/choose_disk.py
+++ b/library/choose_disk.py
@@ -627,10 +627,10 @@ def main():
 
     if legacy is True:
         # Reporting devices & raw_journal_devices for compatiblity
-        module.exit_json(msg=message, changed=changed, ansible_facts=dict(legacy_devices=ceph_data, journal_devices=journal, devices_to_activate=to_activate))
+        module.exit_json(msg=message, changed=changed, ansible_facts=dict(legacy_devices=ceph_data, journal_devices=journal, devices_to_activate=to_activate, storage_devices=[]))
     else:
         # Reporting storage_devices & journal_devices
-        module.exit_json(msg=message, changed=changed, ansible_facts=dict(storage_devices=ceph_data, journal_devices=journal, devices_to_activate=to_activate))
+        module.exit_json(msg=message, changed=changed, ansible_facts=dict(storage_devices=ceph_data, journal_devices=journal, devices_to_activate=to_activate, legacy_devices=[]))
 
 
 if __name__ == '__main__':
--- a/library/choose_disk.py
+++ b/library/choose_disk.py
@@ -461,7 +461,7 @@ def fake_device(legacy_devices, ceph_type):
     devices = {}
     count = 0
     for device in legacy_devices:
-        devices["%s_%d" % (ceph_type, count)] = {"bdev": os.path.dirname(device)+"/"+os.path.basename(device)}
+        devices["%s_%d" % (ceph_type, count)] = {"bdev": os.path.join(os.path.dirname(device), os.path.basename(device))}
         count = count + 1
 
     return devices
--- a/library/choose_disk.py
+++ b/library/choose_disk.py
@@ -335,15 +335,6 @@ def setup_logging():
     logger.info("############")
 
 
-def success(message):
-    global module
-    info(message)
-    logger.info("#######")
-    logger.info("# End #")
-    logger.info("#######")
-    module.exit_json(msg=message)
-
-
 def fatal(message):
     global module
     error(message)
@@ -370,7 +361,8 @@ def main():
     }
 
     module = AnsibleModule(
-        argument_spec=fields
+        argument_spec=fields,
+        supports_check_mode=True
     )
 
     physical_disks = select_only_free_devices(module.params["facts"])
@@ -405,8 +397,23 @@ def main():
 
     if len(matched_devices) < len(lookup_disks):
         fatal("Could only find %d of the %d expected devices\n" % (len(matched_devices), len(lookup_disks)))
-    else:
-        success("All searched devices were found")
+
+    ceph_count = 0
+    for matched_device in matched_devices:
+        if "ceph" in matched_devices[matched_device]:
+            ceph_count = ceph_count + 1
+
+    changed = True
+    logger.info("%d/%d disks already configured" % (ceph_count, len(matched_devices)))
+    if ceph_count == matched_devices:
+        changed = False
+
+    message = "All searched devices were found"
+    info(message)
+    logger.info("#######")
+    logger.info("# End #")
+    logger.info("#######")
+    module.exit_json(msg=message, changed=changed, ansible_facts=dict(devices=matched_devices))
 
 if __name__ == '__main__':
         main()
--- a/library/choose_disk.py
+++ b/library/choose_disk.py
@@ -1,0 +1,146 @@
+#!/usr/bin/python
+
+from ansible.module_utils.basic import AnsibleModule
+import json
+import os
+
+DOCUMENTATION = '''
+---
+module: choose_disk
+author: Erwan Velu <erwan@redhat.com>
+short_description: Choose disks based on their features
+description:
+    To be completed
+'''
+
+fout = open("/tmp/ansible", "a")
+
+
+def find_match(physical_disks, lookup_disks):
+    ''' Find a set of matching devices in physical_disks
+    '''
+    matched_devices = {}
+    exclude_list = []
+    for disk in lookup_disks:
+        fout.write("Working on %s\n" % disk)
+        for physical_disk in physical_disks:
+            if physical_disk in exclude_list:
+                fout.write("Skipping %s as per the exclude list\n" % physical_disk)
+                continue
+            current_physical_disk = physical_disks[physical_disk]
+            current_lookup = lookup_disks[disk]
+            match_count = 0
+            for feature in current_lookup:
+                if feature not in current_physical_disk:
+                    continue
+                elif current_lookup[feature] != current_physical_disk[feature]:
+                    match_count = match_count
+                    # nomatch
+                else:
+                    fout.write("  %s : match for %s\n" % (physical_disk, feature))
+                    match_count = match_count + 1
+                    continue
+
+            if match_count == len(current_lookup):
+                fout.write("  full match (%d/%d) for %s\n" % (match_count, len(current_lookup), physical_disk))
+                matched_devices[physical_disk] = physical_disks[physical_disk]
+                exclude_list.append(physical_disk)
+            elif match_count > 0:
+                fout.write("  partial match for %s with %d/%d\n" % (physical_disk, match_count, len(current_lookup)))
+            else:
+                fout.write("  no match for %s\n" % (physical_disk))
+
+    return matched_devices
+
+
+def expand_disks(lookup_disks):
+    '''
+    Read the disks structure and expand them according to the count directive
+    If no count is provided, a default value of 1 is taken.
+    '''
+    final_list = {}
+    for disk in lookup_disks:
+        count = 0
+        if 'count' not in lookup_disks[disk]:
+            count = 1
+        if 'count' in lookup_disks[disk]:
+            count = int(lookup_disks[disk]['count'])
+            del lookup_disks[disk]['count']
+
+        for n in range(0, count, 1):
+            final_list["%s_%d" % (disk, n)] = lookup_disks[disk]
+
+    return final_list
+
+
+def select_only_free_devices(physical_disks):
+    ''' Don't keep that have partitions '''
+    selected_devices = {}
+    for physical_disk in physical_disks:
+        current_physical_disk = physical_disks[physical_disk]
+
+        # Don't consider devices that doesn't have partitions
+        if 'partitions' not in current_physical_disk:
+            continue
+        # Don't consider the device if partition list is not empty,
+        if len(current_physical_disk['partitions']) > 0:
+            continue
+
+        selected_devices[physical_disk] = physical_disks[physical_disk]
+
+    return selected_devices
+
+
+def get_block_devices_persistent_name(physical_disks):
+    ''' Replace the short name (sda) by the persistent naming 'by-id' '''
+    directory = "/dev/disk/by-id/"
+
+    # If the directory doesn't exist, reports the list as-is
+    if not os.path.isdir(directory):
+        return physical_disks
+
+    final_list = {}
+    matching_list = {}
+    for f in os.listdir(directory):
+        device_name = os.readlink(directory + f).split("/")[-1]
+        if device_name in physical_disks:
+            if device_name not in matching_list:
+                matching_list[device_name] = [f]
+            else:
+                matching_list[device_name].append(f)
+
+    for physical_disk in physical_disks:
+        if physical_disk in matching_list:
+            current_index = sorted(matching_list[physical_disk])[0]
+        else:
+            current_index = physical_disk
+
+        final_list[current_index] = physical_disks[physical_disk]
+        final_list[current_index]["prefix"] = "%s" % (directory)
+
+    return final_list
+
+def main():
+    fields = {
+        "facts": {"required": False, "type": "dict"},
+        "disks": {"required": True, "type": "dict"},
+    }
+
+    module = AnsibleModule(
+        argument_spec=fields
+    )
+
+    physical_disks = module.params["facts"]
+    lookup_disks = expand_disks(module.params["disks"])
+
+    matched_devices = find_match(physical_disks, lookup_disks)
+
+    if len(matched_devices) < len(lookup_disks):
+        message = module.fail_json(msg="Could only find %d of the %d expected devices \n" % (len(matched_devices), len(lookup_disks)))
+        fout.write(message)
+        module.fail_json(msg=message)
+    else:
+        module.exit_json(msg="Success")
+
+if __name__ == '__main__':
+        main()
--- a/library/choose_disk.py
+++ b/library/choose_disk.py
@@ -350,9 +350,9 @@ def expand_disks(lookup_disks, ceph_type="", module=None):
     return final_disks
 
 
-def is_ceph_disk(partition):
+def disk_label(partition):
     '''
-    Check if a partition is containing some ceph structures
+    Reports if a partition is containing some ceph structures
     '''
     stdout = subprocess.check_output(["lsblk", "-no", "PARTLABEL", "%s" % partition])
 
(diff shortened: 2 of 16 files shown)


library/choose_disk.py
```diff
@@ -0,0 +1,436 @@
...
+
+def _gt(left, right):
+    return float(left) > float(right)
+
+
+def _lt(left, right):
+    return float(left) < float(right)
+
+
+_REGEXP = re.compile(r'^([^(]+)'          # function name
```
alfredodeza: this could be much simpler by relying only on the operands themselves: `>`, `<`. Chances are that more people will understand (and remember) a plain operand like `<` than using a custom name and syntax like `_lt()`. 

In the case of simple operands, there wouldn't be a need to parse with a regex, check if the operand exists in globals, and then execute the parsed value. It would be enough to check for existence of the operand and apply it directly. A simplistic example would look like:

library/choose_disk.py
```diff
@@ -0,0 +1,436 @@
...
+                if infinite is False:
+                    break
+            # We were unable to find all part of the required features
+            elif match_count > 0:
+                info("  %50s partially matched with %d/%d items" % (physical_disk, match_count, len(current_lookup)))
+            else:
+                info("  %50s no devices matched" % (physical_disk))
+
+    final_list = {}
+    for matched_device in matched_devices.keys():
```
alfredodeza: by default a loop on a dictionary will use the keys

library/choose_disk.py
```diff
@@ -0,0 +1,436 @@
...
+        for n in range(0, len(matched_devices[matched_device]), 1):
```
alfredodeza: by default `range` will use `1` as the step, there is no need to specify it here

library/choose_disk.py
```diff
@@ -0,0 +1,436 @@
...
+            name = matched_device
+            if len(matched_devices[matched_device]) > 1:
+                name = "%s_%03d" % (matched_device, n)
```
alfredodeza: I am not sure what this naming convention (`matched_device_n`) is for? Is this an internal bookkeeping?

tests/test_choose_disk.py
```diff
@@ -0,0 +1,151 @@
+import imp
+import inspect
+import library.choose_disk as choose_disk
+import mock
+import os
+from nose.tools import assert_equals, nottest
```
ErwanAliasr1: will have to rewrite all the tests :(

library/choose_disk.py
ErwanAliasr1: that have the benefits of being extandable and support stuff like between(x,y). The real call is lt() or gt() the underscore here is to avoid any overloading between a potential python function and our 'language'.

library/choose_disk.py
ErwanAliasr1: fixed.

library/choose_disk.py
ErwanAliasr1: True. Was used during debugging and never been removed. Fixed.

library/choose_disk.py
ErwanAliasr1: Internal numbering in case of 'infinite' matching.

library/choose_disk.py
```diff
@@ -0,0 +1,436 @@
...
+    directory = "/dev/disk/by-id/"
+
+    info('Finding persistent disks name')
+    # If the directory doesn't exist, reports the list as-is
+    if not os.path.isdir(directory):
+        info(' Cannot open %s' % directory)
+        return physical_disks
+
+    final_list = {}
+    matching_list = {}
```
ErwanAliasr1: That was more about the semantic than the data structure type. I renamed them properly to avoid any confusion.

library/choose_disk.py
```diff
@@ -0,0 +1,436 @@
...
+    for f in os.listdir(directory):
+        device_name = os.readlink(directory + f).split("/")[-1]
```
ErwanAliasr1: does it worth the change ?

tests/test_choose_disk.py
andrewschoen: I agree, I think ``py.test`` is the way to go here. From looking at the tests I think you might only need to change the ``assert`` lines.

tests/test_choose_disk.py
```diff
@@ -0,0 +1,151 @@
+import imp
+import inspect
+import library.choose_disk as choose_disk
```
andrewschoen: nitpick: why not ``from library import choose_disk`` ?

library/choose_disk.py
```diff
@@ -0,0 +1,436 @@
...
+        if physical_disks[physical_disk]["bdev"] not in bdev_matched:
+            bdev_unmatched.append(physical_disks[physical_disk]["bdev"])
+
+    info("Unmatched devices : %3d" % len(bdev_unmatched))
+    for bdev in sorted(bdev_unmatched):
+        info(" %s" % bdev)
+
+
+def info(message):
+    global logger
```
andrewschoen: @ErwanAliasr1 I don't think you'd need to pass the logger to every call though. Anywhere you're calling ``info("some message")`` this should be used instead: ``logger.info("some message")``. Doing so would remove the need for all of these logging functions.

library/choose_disk.py
```diff
@@ -0,0 +1,479 @@
...
+
+
+def error(message):
+    global logger
+    if logger:
+        logger.error(message)
+
+
+def setup_logging():
+    global logger
```
andrewschoen: This shouldn't be needed

library/choose_disk.py
```diff
@@ -0,0 +1,479 @@
...
+    hdlr.setFormatter(formatter)
+    logger.addHandler(hdlr)
+    logger.setLevel(logging.INFO)
+    logger.info("############")
+    logger.info("# Starting #")
+    logger.info("############")
+
+
+def fatal(message):
+    global module
```
andrewschoen: I think we should avoid the use of ``global`` of just pass ``module`` into this function.

library/choose_disk.py
```diff
@@ -0,0 +1,479 @@
...
+        if "journal" in device["ceph_type"]:
+            journal.append(device["bdev"])
+
+    changed = True
+    logger.info("%d/%d disks already configured" % (ceph_count, len(matched_devices)))
+    if ceph_count == len(matched_devices):
+        changed = False
+
+    message = "All searched devices were found"
+    info(message)
```
andrewschoen: ``logger.info(message)``

library/choose_disk.py
ErwanAliasr1: @andrewschoen done.

library/choose_disk.py
ErwanAliasr1: fixed

library/choose_disk.py
```diff
@@ -0,0 +1,455 @@
...
+    ignored_devices = []
+
+    logger.info("Looking for matches")
+    # Inspecting every disk we search for
+    for disk in sorted(lookup_disks):
+
+        current_lookup = dict(lookup_disks[disk])
+        infinite = False
+        current_type = ""
+        if "infinite" in current_lookup.keys():
```
alfredodeza: there is no need to use `keys()` here since `item in dict` will look at a dict's keys

library/choose_disk.py
```diff
@@ -0,0 +1,455 @@
+#!/usr/bin/python
+
+from ansible.module_utils.basic import AnsibleModule
+import json
```
alfredodeza: doesn't look like you are using this in the module

tests/test_choose_disk.py
```diff
@@ -0,0 +1,159 @@
+import imp
+import inspect
+import mock
+import os
+from nose.tools import assert_equals, nottest
```
alfredodeza: `nose` is a test runner/framework that is no longer maintained. I would suggest using `py.test` here which is what we use on most of our projects

library/choose_disk.py
alfredodeza: There are lots of issues of using `global` and I am not seeing why it is necessary here (other than to avoid passing it into functions).

Using globals in this way will obfuscate where things are coming from, making it harder to debug when/if something fails. When there are side-effects it also makes it hard to determine what/where is causing them.

Refactoring gets affected as well, because the scope becomes unclear. The namespace gets polluted, and there is potential for overwriting or redefining the global unknowingly. 

Linters will have a hard time detecting some issues associated with assignment too.

A global shouldn't be used just because it avoids the inconvenience of passing a variable around.

library/choose_disk.py
ErwanAliasr1: True. Deleted.

library/choose_disk.py
ErwanAliasr1: fixed

library/choose_disk.py
```diff
@@ -0,0 +1,455 @@
...
+
+def find_match(physical_disks, lookup_disks):
+    ''' Find a set of matching devices in physical_disks
+    '''
+    matched_devices = {}
+    ignored_devices = []
+
+    logger.info("Looking for matches")
+    # Inspecting every disk we search for
+    for disk in sorted(lookup_disks):
```
ErwanAliasr1: Same answer as previously ;)

library/choose_disk.py
```diff
@@ -0,0 +1,453 @@
...
+    for physical_disk in sorted(physical_disks):
+        if physical_disks[physical_disk]["bdev"] not in bdev_matched:
+            bdev_unmatched.append(physical_disks[physical_disk]["bdev"])
+
+    logger.info("Unmatched devices : %3d" % len(bdev_unmatched))
+    for bdev in sorted(bdev_unmatched):
+        logger.info(" %s" % bdev)
+
+
+def setup_logging(logger):
```
andrewschoen: You shouldn't need to pass ``logger`` into this method.

library/choose_disk.py
```diff
@@ -0,0 +1,453 @@
...
+    hdlr.setFormatter(formatter)
+    logger.addHandler(hdlr)
+    logger.setLevel(logging.INFO)
+    logger.info("############")
+    logger.info("# Starting #")
+    logger.info("############")
+
+
+def fatal(message, module):
+    if logger:
```
andrewschoen: This isn't necessary as ``logger`` is defined at module scope and will always be available to this method.

library/choose_disk.py
```diff
@@ -0,0 +1,453 @@
...
+def main():
+    module = None
+    legacy = False
+    matched_devices = None
+    lookup_disks = None
+    disks = "disks"
+    devices = "devices"
+    raw_journal_devices = "raw_journal_devices"
+
+    setup_logging(logger)
```
andrewschoen: You shouldn't need to pass ``logger`` into this method.

library/choose_disk.py
```diff
@@ -0,0 +1,453 @@
...
+            infinite = True
+            del current_lookup["infinite"]
+
+        # We cannot keep the disk type as a feature to lookup
+        if "ceph_type" in current_lookup.keys():
+            current_type = current_lookup["ceph_type"]
+            del current_lookup["ceph_type"]
+
+        if len(ignored_devices) == len(physical_disks):
+            logger.info(" Skipping %s as no more free devices to match" % (disk))
```
andrewschoen: When using the logging module directly it's safer to use it's built-in method for string interpolation instead of the ``"%s" % var`` approach.  This is safer because if you use a variable there that does not interpolate correctly the code will throw an exception and halt execution. If you use the logging modules built-in way then it will catch that exception and report it as part of the logging message without stopping code execution.

That method would look like:

```
logger.info(" Skipping %s as no more free devices to match", disk)
```

Here's an example the exposes the problem I mentioned:

```
import logging

logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
logger = logging.getLogger()

logger.info('normal logging line')
logger.info('normal logging line, with some correct args: %s %s', 'arg1', 'arg2')
logger.info('normal logging line, with some incorrect args: %s %s %s %s', 'arg1', 'arg2')
logger.info('execution does not halt with the previous invalid line')
logger.info('invalid formatting without logging: %s %s' % 'arg1')
logger.info('this line will never show because the previous line is broken')
```

Could you please all the logging lines to use the approach I mentioned?

library/choose_disk.py
```diff
@@ -0,0 +1,453 @@
...
+                        # Check if the associated function exists
+                        if new_operator in globals():
+                            # and assign operands with the new values
+                            operator = new_operator
+                            right = arguments.group(2)
+                        else:
+                            fatal("Unsupported %s operator in : %s" % (new_operator, right), module)
+
+                # Let's check if (left <operator> right) is True meaning the match is done
+                if globals()[operator](convert_units(left), convert_units(right)):
```
andrewschoen: I'm really not a fan of how you've created a DSL to handle these operators as I think it unnecessarily obfuscates and complicates this code. However, if we must keep it then we should avoid using ``globals()`` here for all the same reasons we've mentioned before about global scope.

A better approach would be to define a dictionary constant that maps the operator to the appropriate method. Something like:

```
OPERATORS = {
    "=": _equal,
    ">": _gt,
    "<": _lt, 
}

# and usage like

if OPERATORS[operator](convert_units(left), convert_units(right)):

```

ErwanAliasr1: Got it . Pushed like that.

ktdreyer:
It seems like this feature would be perfect for ceph-disk, where it can be used across the whole community of Ceph operators and Config Mgmt systems beyond Ansible

library/choose_disk.py
ErwanAliasr1: All the code have been updated accordingly

library/choose_disk.py
ErwanAliasr1: Fixed.

library/choose_disk.py
ErwanAliasr1: fixed

library/choose_disk.py
ErwanAliasr1: fixed

library/choose_disk.py
ErwanAliasr1: Yes indeed. It thought it would have been more trivial but that was a good feature to add to clean the code & review it. So done for between() ;)

library/choose_disk.py
ErwanAliasr1: fixed

library/choose_disk.py
ErwanAliasr1: fixed too

library/choose_disk.py
```diff
@@ -0,0 +1,455 @@
...
+
+
+def get_keys_by_ceph_order(physical_disks):
+    '''
+    Return a list of keys where ceph disks are reported first
+    while keeping the list sorted
+    '''
+    ceph_disks = []
+    non_ceph_disks = []
+    for physical_disk in sorted(physical_disks):
```
ErwanAliasr1: I added a comment to explain that better.

ErwanAliasr1:
jenkins test this please

library/choose_disk.py
```diff
@@ -0,0 +1,638 @@
+#!/usr/bin/python
+
```
leseb: I know it's cosmetic but shouldn't we add a LICENSE and AUTHOR header?

library/choose_disk.py
```diff
@@ -0,0 +1,638 @@
...
+                     r'(?:\s*,\s*(.+))?'  # remaining optional arguments
+                     r'\)$')              # last parenthesis
+
+
+logger = logging.getLogger('choose_disk')
+
+
+def to_bytes(value):
+    '''
+    Convert storage units into bytes to ease comparaison between sizex
```
leseb: s/sizex/size?

library/choose_disk.py
```diff
@@ -0,0 +1,638 @@
...
+                name = "%s_%03d" % (matched_device, n)
+            final_disks[name] = matched_devices[matched_device][n]
+
+    return final_disks
+
+
+def expand_disks(lookup_disks, ceph_type="", module=None):
+    '''
+    Read the disks structure and expand them according to the count directive
+    If no count is provided, a default value of 1 is taken.
```
leseb: Is it still true? It looks like we fail if no `count` is provided.

library/choose_disk.py
```diff
@@ -0,0 +1,638 @@
...
+            for partition in sorted(current_physical_disk['partitions']):
+                disk_type = is_ceph_disk("/dev/" + partition)
+                if disk_type:
+                    if ceph_disk:
+                        ceph_disk += " + "
+                    # Saving the ceph type (journal and/or data)
+                    ceph_disk += disk_type
+
+            if not ceph_disk:
+                logger.info(' Ignoring %10s : Device have exisiting partitions', physical_disk)
```
leseb: s/Device have exisiting partitions/Device has existing partitions/

library/choose_disk.py
ErwanAliasr1: Updated.

library/choose_disk.py
ErwanAliasr1: fixed

library/choose_disk.py
ErwanAliasr1: fixed

library/choose_disk.py
ErwanAliasr1: Nice catch. Updated the faulty patch with this removal.

leseb:
jenkins test this please

tests/test_choose_disk.py
alfredodeza: this hasn't been addressed. This test tool (nose) is no longer maintained.

choose_disk.yml
```diff
@@ -0,0 +1,41 @@
...
+      fsid: "e8ba1fe2-dbf6-4b15-b0ca-2c5ffd97f59e"
+      public_network: 172.16.41.0/24
+      cluster_network: 172.16.41.0/24
+      journal_size: 100
+      journal_collocation: true
+      devices: []
+      cluster: ceph
+      osd_objectstore: filestore
+      osd_group_name: osds
+#      devices: {'storage_disks': {'vendor': '0x1af4', 'ceph_type' : 'data', 'size': 'gte(30 GB)', 'count': 2 }}
```
alfredodeza: why are these commented out? can they be removed or are they here as an example?

library/choose_disk.py
```diff
@@ -0,0 +1,651 @@
...
+            # Checking what features are matching
+            for feature in current_lookup:
+                if feature not in current_physical_disk:
+                    continue
+
+                # Assign left and right operands
+                right = current_lookup[feature]
+                left = current_physical_disk[feature]
+
+                def evaluate_operator(left, right):
```
alfredodeza: doesn't seem like this function should be nested in here, by making it a closure it is making it harder to unit test (the inputs are currently forced to be passed onto the parent function). It is better to have it outside so that testing is improved.

library/choose_disk.py
```diff
@@ -0,0 +1,651 @@
...
+                    return OPERATORS[operator](left, right)
+
+                # Let's check if (left <operator> right) is True meaning the match is done
+                if evaluate_operator(left, right):
+                    logger.debug("  %s : match  %s %s", physical_disk, left, right)
+                    match_count = match_count + 1
+                    continue
+                else:
+                    # comparaison is false meaning devices doesn't match
+                    logger.debug("  %s : no match  %s %s", physical_disk, left, right)
```
alfredodeza: why are there leading spaces in some logging statements like this one? can we make them consistent?

library/choose_disk.py
```diff
@@ -0,0 +1,651 @@
...
+    '''
+    stdout = subprocess.check_output(["lsblk", "-no", "PARTLABEL", "%s" % partition])
+
+    if "ceph data" in stdout:
+        return "data"
+
+    if "ceph journal" in stdout:
+        return "journal"
+
+    return ""
```
alfredodeza: the naming (and docstring) of this function is misleading since it implies a boolean response (vs. a string describing what this is). I would expect to have the return values changed or the function signature to be updated

library/choose_disk.py
```diff
@@ -0,0 +1,651 @@
...
+    if not os.path.isdir(directory):
+        logger.info(' Cannot open %s', directory)
+        return physical_disks
+
+    final_disks = {}
+    matching_devices = {}
+
+    # Searching in the /dev/disk/by-id/ which are the disks we look for
+    for f in os.listdir(directory):
+        device_name = os.readlink(directory + f).split("/")[-1]
```
alfredodeza: Python has a native way of handling traversing of paths while keeping files, dirs, and parent dirs available with `os.walk`:

    Generate the file names in a directory tree by walking the tree either top-down or bottom-up. For each directory in the tree rooted at directory top (including top itself), it yields a 3-tuple (dirpath, dirnames, filenames).

It would be preferable to just consume the native way rather than re-implementing it this way

library/choose_disk.py
```diff
@@ -0,0 +1,651 @@
...
+def fake_device(legacy_devices, ceph_type):
+    '''
+    In case of legacy block device names, let's create an internal faked
+    entry with a 'bdev' entry filled with the actual path. This will be used to
+    make a match later on.
+    '''
+    devices = {}
+    count = 0
+    for device in legacy_devices:
+        devices["%s_%d" % (ceph_type, count)] = {"bdev": os.path.dirname(device)+"/"+os.path.basename(device)}
```
alfredodeza: paths that are getting joined like this should use the `os.path.join`, a more robust implementation of what this line is trying to do here. From the docs:

    Join two or more pathname components, inserting '/' as needed.
    If any component is an absolute path, all previous path components
    will be discarded.  An empty last part will result in a path that
    ends with a separator.

library/choose_disk.py
```diff
@@ -0,0 +1,651 @@
...
+    logger.info("Unmatched devices : %3d", len(bdev_unmatched))
+    for bdev in sorted(bdev_unmatched):
+        logger.info(" %s", bdev)
+
+
+def setup_logging():
+    '''
+    Preparing the logging system
+    '''
+    hdlr = logging.FileHandler('/var/log/choose_disk.log')
```
alfredodeza: this will require that current user to have permissions in `/var/log/choose_disk.log` or `/var/log`. If that hasn't been set before it will imply the need for super user privileges, while the rest of ceph-ansible doesn't require it. This should be optional, with a default that doesn't force a user to run with super user privileges

library/choose_disk.py
```diff
@@ -0,0 +1,651 @@
...
+
+
+def get_var(module, variable):
+    '''
+    Extract the ansible variable from the playbook
+    If variable doesn't exist, let's return None
+    '''
+    if variable in module.params["vars"]:
+        return module.params["vars"][variable]
+    return None
```
alfredodeza: no need to return `None` here as functions that do not return anything will implicitly have `None` returned to the caller

tests/test_choose_disk.py
```diff
@@ -0,0 +1,197 @@
...
+from library import choose_disk
+
+host_dir = "tests/unit/hosts/"
+
+
+@nottest
+def get_hosts_to_test():
```
alfredodeza: another bonus of using `py.test` is that this wouldn't need to exist as there is support for fixtures to be loaded per test function (or test method)

choose_disk.yml
```diff
@@ -0,0 +1,41 @@
+---
+# Defines deployment design and assigns role to server groups
```
andrewschoen: What is the purpose of this playbook? Is this only for testing? If so, I don't think we need it included here.

tests/test_choose_disk.py
andrewschoen: +1 ``py.test`` would be much preferred here, ``py.test`` is also what we're using in the CI tests.

tests/test_choose_disk.py
ErwanAliasr1: Done in my last commiy.

library/choose_disk.py
ErwanAliasr1: Unnested it in last commit + unit testing.

library/choose_disk.py
ErwanAliasr1: fixed in latest commit.

library/choose_disk.py
ErwanAliasr1: fixed in latest commit.

library/choose_disk.py
ErwanAliasr1: Fixed in latest commit

choose_disk.yml
ErwanAliasr1: include here = in the repository or in this directory ?

I think it's pretty useful to get sample code to see how to use it

library/choose_disk.py
ErwanAliasr1: That is made to get a more readable output as it kind of create some visual blocks.

choose_disk.yml
ErwanAliasr1: Yes, If left them here as an example as there is many way to use the 'devices' parameter.

tests/test_choose_disk.py
```diff
@@ -0,0 +1,203 @@
...
+    result = choose_disk.expand_disks(eval("{'storage_disks': {'model': 'SAMSUNG MZ7LN512', 'rotational': '1', 'count': 1, 'ceph_type': 'data' }}"))
+    assert result == eval("{'storage_disks_000': {'model': 'SAMSUNG MZ7LN512', 'rotational': '1', 'ceph_type': 'data' }}")
+
+
+def test_expand_disks_explict_count_2():
+    """
+    expand_disks - test expand disk with explicit count=2
+    """
+    result = choose_disk.expand_disks(eval("{'storage_disks': {'model': 'SAMSUNG MZ7LN512', 'rotational': '1', 'count': 2, 'ceph_type': 'data' }}"))
+    assert result == eval("{'storage_disks_000': {'model': 'SAMSUNG MZ7LN512', 'rotational': '1', 'ceph_type': 'data' }, 'storage_disks_001': {'model': 'SAMSUNG MZ7LN512', 'rotational': '1', 'ceph_type': 'data' }}")
```
alfredodeza: why is this calling `eval` ?

library/choose_disk.py
alfredodeza: is there some relation between the visual blocks and the logging statements? For example, if there is a log line that has two leading spaces vs. another one that has none (or just 1), would that mean some kind of hierarchy? The reason why I was asking for normalizing is because it is easier for parsing. If there is a need for better visuals maybe create a different logger?

ErwanAliasr1: Yes, it's a kind of way to organize the visual output to get everything readable. A kind of tabbing to add a kind of visual dependency with the upper displayed items. Is that really critical for a first merge ?

tests/test_choose_disk.py
ErwanAliasr1: Because it's wrong ;)
Fixed

alfredodeza:
The code base looks OK now, thanks for all the changes. This will be a tremendous help on solving the issue of dealing with many different devices on large-enough clusters.

However, I am -1 on merging this and making it a part of ceph-ansible:

* it is significantly large: 
  - updates, changes, fixes will be harder than just dealing with plain ansible
  - it will be very hard to move it into something else (like a module as it was suggested) later

* The Python code will place the maintenance burden directly on the ceph-ansible maintainers. Although I don't foresee a disconnect of contribution if the code lived separately, I do see current maintainers will have to carry the load for fixes and features

This could very well live in a few  places with little to no modifications to the current code:

* As an Ansible module, which can be optionally installed and consumed directly by ceph-ansible
* As a Python module that can take ansible facts or other input to detect disks
* As a ceph-disk addition, where ceph-ansible could rely directly on Ceph for better device detection/filtering

My inclination would be on ceph-disk as it would be something that could be used outside of the smaller ceph-ansible audience. The impact would be wider and would make ceph-disk a lot better. Of course it would need to address the issue of facts/discovery but it doesn't look like that would be much more work than what it is currently presented here.

ErwanAliasr1:
@alfredodeza, please find my position regarding your comment :
- it is significantly large" : that's as big as the config_template.
- "fixes will be harder than just dealing with plain ansible" : we have also some python stuff into ceph-ansible, that's not a new thing
- "The Python code will place the maintenance burden directly on the ceph-ansible maintainers" : yes that have been developed to increase ceph-ansible capabilities & features. For bugs & features, regarding that code, I take care of it. Nevertheless, I tried to produced a well splitted/documented code to let others working on it. I also kept all the history of development to ease a future git blame.
- Moving to ceph-disk, that's clearly a "no" in the current state. That would be a too long path to get something functional.
- Moving to an ansible/python module: why not but what is the real benefit of it ? That tool have been developed & designed for ceph and ansible.  That's not generic at all.

To get the feature in the product asap and show all the benefits I suggest we keep it here and move it elsewhere if others willing to use it.

andrewschoen:
jenkins test this please

ircolle:
@dachary - what are your thoughts on moving this functionality into ceph-disk?

dachary:
@ircolle after reading the description it looks like something that depends on ceph-disk rather than something that should be implemented as part of ceph-disk. If functions implemented in ceph-disk are useful to implement this feature and they are not exposed by the CLI, ceph-disk can also be used as a module.

That being said there may be reasons why this should be a ceph-disk feature but I don't see them right now.

ircolle:
@leseb - any idea why this is failing so many tests?

leseb:
@ircolle, it's failing simply because we are missing one variable in the code.
I'll take care of this with @ErwanAliasr1 early next week, it's nothing.

andrewschoen:
@leseb @ErwanAliasr1 I think a new testing scenario that executes ``choose_disk`` module is needed as well.

leseb:
@andrewschoen oh yes definitely, that's part of the plan :).

leseb:
Since @ErwanAliasr1 is not available, we need to move ahead with this one. Since this was pushed from @ErwanAliasr1's repo I can not rebase this and/or add more commits. I don't have the choice to close this and re-open it: https://github.com/ceph/ceph-ansible/pull/1558 so all the contributors of this repo can collaborate on fixing this.

bengland2:
I'm really interested in this feature because it solves an important deployment problem in large-scale Ceph installations, including OpenStack sites that rely on Ceph.  @fultonj have you seen this?  It's a lot like what I proposed in bz [1438590](https://bugzilla.redhat.com/show_bug.cgi?id=1438590) where OpenStack would select OSD devices by rule using introspection data.

leseb:
This now lives here: https://github.com/ceph/ceph-ansible/pull/1558, we will resurrect this for 3.1

PR #1135, opened 2016-11-30
Add gluster nfs ganesha repo only for Ubuntu

guits:
According to the official documentation, adding PPA is not available for Debian.

### DIFF
--- a/roles/ceph-common/tasks/installs/debian_ceph_repository.yml
+++ b/roles/ceph-common/tasks/installs/debian_ceph_repository.yml
@@ -52,5 +52,6 @@
     - ppa:gluster/libntirpc
     - ppa:gluster/nfs-ganesha
   changed_when: false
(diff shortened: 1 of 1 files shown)


roles/ceph-common/tasks/installs/debian_ceph_repository.yml
```diff
@@ -52,5 +52,6 @@
     - ppa:gluster/libntirpc
     - ppa:gluster/nfs-ganesha
   changed_when: false
-  when: nfs_obj_gw or nfs_file_gw
-
+  when:
+    - (nfs_obj_gw or nfs_file_gw)
+      and not ansible_distribution == "Debian"
```
leseb: This should be written as:

```
when:
  - (nfs_obj_gw or nfs_file_gw)
  - not ansible_distribution == "Debian"
```

PR #1138, opened 2016-11-30
tests: adds a docker_cluster scenario

andrewschoen:
This tests containerized deployment of 1mon, 1osd, 1mds and 1rgw node on a centos atomic host. There are also a bunch of little fixes I had to make to get this to run.

Let's not merge this until I get the CI updated to run this scenario so we can verify it's going to work correctly.


andrewschoen:

andrewschoen:

andrewschoen:

roles/ceph-mon/tasks/docker/fetch_configs.yml
```diff
@@ -21,7 +21,7 @@
 
 - name: try to fetch ceph config and keys
   copy:
-    src: "{{ playbook_dir }}/{{ fetch_directory }}/docker_mon_files/{{ item.0 }}"
+    src: "{{ fetch_directory }}/docker_mon_files/{{ item.0 }}"
```
font: What's the reason for this change? We need to be careful because now we'll rely on the relative path `fetch/docker_mon_files/` instead of the absolute path provided by `playbook_dir`. Will this impact running playbooks from within `infrastructure-playbooks`?

tests/functional/centos/7/docker-cluster/vagrant_variables.yml
```diff
@@ -0,0 +1,69 @@
...
+rbd_mirror_vms: 0
+client_vms: 0
+iscsi_gw_vms: 0
+
+# Deploy RESTAPI on each of the Monitors
+restapi: true
+
+# INSTALL SOURCE OF CEPH
+# valid values are 'stable' and 'dev'
+ceph_install_source: stable
```
andrewschoen: Will remove

roles/ceph-mon/tasks/docker/fetch_configs.yml
andrewschoen: The issue arises when ``fetch_directory`` is set to an absolute path, which is what CI does (and the ceph-installer project). Including ``playbook_dir`` here removes the users ability to set ``fetch_directory`` to anything other than a path relative to the playbook.

I didn't test this from a playbook from within ``infrastructure-playbooks/``, but isn't our current guidance there to move those playbooks up to the project root anyway?

tests/functional/centos/7/docker-cluster/group_vars/all
```diff
@@ -0,0 +1,25 @@
...
+mds_containerized_deployment: True
+rgw_containerized_deployment: True
+ceph_mon_docker_interface: eth1
+ceph_mon_docker_subnet: "192.168.15.0/24"
+ceph_osd_docker_extra_env: "CEPH_DAEMON=OSD_CEPH_DISK_ACTIVATE,OSD_JOURNAL_SIZE=100"
+ceph_docker_on_openstack: False
+public_network: "192.168.15.0/24"
+cluster_network: "192.168.16.0/24"
+journal_size: 100
+journal_collocation: true
```
andrewschoen: I had to add this and ``devices`` because of the mandatory variable check here https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/main.yml#L2

Do we want to not include that file when ``osd_containerized_deployment: True`` instead?

leseb: I think at some point `ceph_osd_docker_devices` should  use the content of `devices`, not sure why we have 2 variables... @font do you remember?

leseb: Same goes for `journal_size` we should have:

`ceph_osd_docker_extra_env: "CEPH_DAEMON=OSD_CEPH_DISK_ACTIVATE,OSD_JOURNAL_SIZE="{{ journal_size}}""`

tests/functional/centos/7/docker-cluster/vagrant_variables.yml
```diff
@@ -0,0 +1,69 @@
+---
+
+# DEPLOY CONTAINERIZED DAEMONS
+docker: True
+
+# DEFINE THE NUMBER OF VMS TO RUN
+mon_vms: 1
```
leseb: I think it's important we test 3 mons, just as a realistic scenario :)

tests/functional/centos/7/docker-cluster/group_vars/all
andrewschoen: Can we set ``journal_size`` and then use it in the same group_vars file? I'm remembering having trouble with something like that in the past.

leseb: This shouldn't be an issue, we have this https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/defaults/main.yml#L258 which kinda does the same thing.

tests/functional/centos/7/docker-cluster/vagrant_variables.yml
leseb: Great thanks!

tests/functional/centos/7/docker-cluster/group_vars/all
andrewschoen: Awesome, I'll do that then

tests/functional/centos/7/docker-cluster/vagrant_variables.yml
andrewschoen: I'll have to make the change in ceph-build for this first, but I don't think it'll be much trouble.

roles/ceph-mon/tasks/docker/fetch_configs.yml
font: That's the current guidance, but only because of https://github.com/ansible/ansible/issues/18090. Once that bug is fixed we should be able to run playbooks from within `infrastructure-playbooks`.

tests/functional/centos/7/docker-cluster/group_vars/all
font: @leseb We've always had 2 variables since I've joined. :) I'm not sure why either. I think it's for the same reason many things are split between containerized and non-containerized but don't necessarily have to be that way. I agree we should consolidate to `devices`.

roles/ceph-mon/tasks/docker/fetch_configs.yml
andrewschoen: Ok, that's good to know. I still think having ``playbook_dir`` here is incorrect though, after https://github.com/ansible/ansible/issues/18090 it won't be an issue and until then we have a workaround.

tests/functional/centos/7/docker-cluster/group_vars/all
leseb: Ook perhaps it was just me then haha

roles/ceph-mon/tasks/docker/fetch_configs.yml
font: I agree, let's remove it here so we allow more flexibility to the user. It may or may not break running playbooks from within another directory, but if it does, perhaps we can make changes to the `ansible.cfg` in that directory to accommodate. In addition, maybe we can just define the default `fetch_directory` variables to include `playbook_dir` instead.

andrewschoen:
docker_cluster failed because I forgot to change scenario.py to know that it's testing for 3 mons now instead of one, will fix. The playbook run itself completed correctly.

roles/ceph-mon/tasks/docker/fetch_configs.yml
andrewschoen: Including ``playbook_dir`` in the default ``fetch_directory`` is a great idea.

tests/functional/centos/7/docker-cluster/group_vars/all
```diff
@@ -0,0 +1,25 @@
...
+ceph_osd_docker_extra_env: "CEPH_DAEMON=OSD_CEPH_DISK_ACTIVATE,OSD_JOURNAL_SIZE={{ journal_size }}"
+ceph_docker_on_openstack: False
+public_network: "192.168.15.0/24"
+cluster_network: "192.168.16.0/24"
+journal_collocation: true
+ceph_rgw_civetweb_port: 8080
+ceph_osd_docker_devices:
+  - '/dev/sda'
+  - '/dev/sdb'
+devices:
```
font: Is this to prepare for the changes in the future?

tests/functional/centos/7/docker-cluster/group_vars/all
```diff
@@ -0,0 +1,25 @@
...
+  - '/dev/sda'
+  - '/dev/sdb'
+os_tuning_params:
```
font: This isn't used in the containerized deployment roles.

tests/functional/centos/7/docker-cluster/group_vars/osds
```diff
@@ -0,0 +1,4 @@
+---
+os_tuning_params:
```
font: Same comment as above.

tests/functional/centos/7/docker-cluster/group_vars/all
andrewschoen: This is there because checks fail in https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/check_mandatory_vars.yml without it currently. Once we have this PR in refactoring these tasks to use ``devices`` instead of ``ceph_osd_docker_devices`` will be easier.

tests/functional/centos/7/docker-cluster/group_vars/all
andrewschoen: removed

Issue #1145, opened 2016-12-05
'pool_default_pg_num' is undefined for 'client' role


eduardegorov:
ceph-ansible: 0f3ae4c
ansible: 2.2.0.0

If `client` role is assigned to a host with default group vars, this error pops up:
```
TASK [ceph-client : create pools] **********************************************
task path: /root/tfs/OpenStack2Configs/roles/ceph-client/tasks/create_users_keys.yml:2^[[0m
fatal: [n1]: FAILED! => {"failed": true, "msg": "[{u'name': u'test', u'pgs': u'{{ pool_default_pg_
num }}'}, {u'name': u'test2', u'pgs': u'{{ pool_default_pg_num }}'}]: 'pool_default_pg_num' is undefined"}
fatal: [n2]: FAILED! => {"failed": true, "msg": "[{u'name': u'test', u'pgs': u'{{ pool_default_pg_
num }}'}, {u'name': u'test2', u'pgs': u'{{ pool_default_pg_num }}'}]: 'pool_default_pg_num' is undefined"}
fatal: [n3]: FAILED! => {"failed": true, "msg": "[{u'name': u'test', u'pgs': u'{{ pool_default_pg_
num }}'}, {u'name': u'test2', u'pgs': u'{{ pool_default_pg_num }}'}]: 'pool_default_pg_num' is undefined"}
fatal: [n4]: FAILED! => {"failed": true, "msg": "[{u'name': u'test', u'pgs': u'{{ pool_default_pg_
num }}'}, {u'name': u'test2', u'pgs': u'{{ pool_default_pg_num }}'}]: 'pool_default_pg_num' is undefined"}
```

`pool_default_pg_num` is defined in `mon` role variables only. For my case, this can be workarounded by:
```diff
diff --git a/roles/ceph-client/tasks/main.yml b/roles/ceph-client/tasks/main.yml
index c2b363f..32f5743 100644
--- a/roles/ceph-client/tasks/main.yml
+++ b/roles/ceph-client/tasks/main.yml
@@ -2,3 +2,4 @@
 - include: pre_requisite.yml
 - include: create_users_keys.yml
   when: user_config
+  static: False
```

But, of course, the main issue is not solved and might annoy newcomers if they would try to add test pools as well. Should we move this `pool_default_pg_num` to all.yml or calculate pgs for test/test2 pools using http://ceph.com/pgcalc/ formula in order to avoid HEALTH_WARN messages (since these pools are not for production purposes anyway)?

[...]

PR #1146, opened 2016-12-05
mon: pool creation and pgs

leseb:
Since we introduced config_overrides we removed a lot of options from
the default template. In some cases, like mds pool, openstack pools etc
we need to know the amount of PGs required. The idea here is to skip the
task if ceph_conf_overrides.global.osd_pool_default_pg_num is not define
in your `group_vars/all.yml`.

Closes: #1145

### DIFF
--- a/roles/ceph-client/defaults/main.yml
+++ b/roles/ceph-client/defaults/main.yml
@@ -7,8 +7,8 @@ fetch_directory: fetch/
 
 user_config: false
 pools:
-  - { name: test, pgs: "{{ pool_default_pg_num }}" }
-  - { name: test2, pgs: "{{ pool_default_pg_num }}" }
+  - { name: test, pgs: "{{ ceph_conf_overrides.global.osd_pool_default_pg_num }}" }
+  - { name: test2, pgs: "{{ ceph_conf_overrides.global.osd_pool_default_pg_num }}" }
 
 keys:
   - { name: client.test, value: "mon 'allow r' osd 'allow class-read object_prefix rbd_children, allow rwx pool=test'" }
--- a/roles/ceph-mon/defaults/main.yml
+++ b/roles/ceph-mon/defaults/main.yml
@@ -15,7 +15,6 @@ monitor_secret: "{{ monitor_keyring.stdout }}"
 cephx: true
 
 # CephFS
-pool_default_pg_num: 128
 cephfs_data: cephfs_data
 cephfs_metadata: cephfs_metadata
 cephfs: cephfs
@@ -42,16 +41,16 @@ calamari: false
 openstack_config: false
 openstack_glance_pool:
   name: images
-  pg_num: "{{ pool_default_pg_num }}"
+  pg_num: "{{ osd_pool_default_pg_num }}"
 openstack_cinder_pool:
   name: volumes
-  pg_num: "{{ pool_default_pg_num }}"
+  pg_num: "{{ osd_pool_default_pg_num }}"
 openstack_nova_pool:
   name: vms
-  pg_num: "{{ pool_default_pg_num }}"
+  pg_num: "{{ osd_pool_default_pg_num }}"
 openstack_cinder_backup_pool:
   name: backups
-  pg_num: "{{ pool_default_pg_num }}"
+  pg_num: "{{ osd_pool_default_pg_num }}"
 
 openstack_pools:
   - "{{ openstack_glance_pool }}"
--- a/tests/functional/centos/7/cluster/vagrant_variables.yml
+++ b/tests/functional/centos/7/cluster/vagrant_variables.yml
@@ -10,7 +10,7 @@ mds_vms: 1
 rgw_vms: 1
 nfs_vms: 0
 rbd_mirror_vms: 0
-client_vms: 0
+client_vms: 1
 iscsi_gw_vms: 0
 
 # Deploy RESTAPI on each of the Monitors
--- a/group_vars/clients.yml.sample
+++ b/group_vars/clients.yml.sample
@@ -15,8 +15,8 @@ dummy:
 
 #user_config: false
 #pools:
-#  - { name: test, pgs: "{{ pool_default_pg_num }}" }
-#  - { name: test2, pgs: "{{ pool_default_pg_num }}" }
+#  - { name: test, pgs: "{{ ceph_conf_overrides.global.osd_pool_default_pg_num }}" }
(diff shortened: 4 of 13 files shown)


andrewschoen:
jenkins test this please

leseb:

leseb:

leseb:

leseb:

leseb:

roles/ceph-mon/defaults/main.yml
```diff
@@ -42,16 +41,16 @@ calamari: false
...
-  pg_num: "{{ pool_default_pg_num }}"
+  pg_num: "{{ osd_pool_default_pg_num }}"
 openstack_nova_pool:
   name: vms
-  pg_num: "{{ pool_default_pg_num }}"
+  pg_num: "{{ osd_pool_default_pg_num }}"
 openstack_cinder_backup_pool:
   name: backups
-  pg_num: "{{ pool_default_pg_num }}"
+  pg_num: "{{ osd_pool_default_pg_num }}"
```
andrewschoen: Won't this fail in the same way because where is ``osd_pool_default_pg_num`` defaulted?

roles/ceph-client/defaults/main.yml
```diff
@@ -7,8 +7,8 @@ fetch_directory: fetch/
 
 user_config: false
 pools:
-  - { name: test, pgs: "{{ pool_default_pg_num }}" }
-  - { name: test2, pgs: "{{ pool_default_pg_num }}" }
+  - { name: test, pgs: "{{ ceph_conf_overrides.global.osd_pool_default_pg_num }}" }
+  - { name: test2, pgs: "{{ ceph_conf_overrides.global.osd_pool_default_pg_num }}" }
```
andrewschoen: Will this break if ``ceph_conf_overrides.global`` is not defined?

leseb: Not really, I mean these will be called only if those are valid https://github.com/ceph/ceph-ansible/blob/e3fb818c663383bd987c4a29cdde155c58ac808d/roles/ceph-client/tasks/main.yml#L5-L6

If not declared the creation task is skipped and the vars never called.

Am I getting this right?

andrewschoen: @leseb Yeah, that's right. Thanks for the explanation.

roles/ceph-mon/defaults/main.yml
leseb: It shouldn't, at least it does not since we check variables (see below), I'm testing this locally.

leseb:
I need to add the tests for this.

alfredodeza:
@leseb I understand the reasoning to define things in `ceph_conf_overrides` and making that a central location instead of consuming variables from a vars file.

But in this particular case (I am sure this will not be the only one) there is a duality, because the same value is used in some Ceph CLI invocations. It is odd to see a `ceph.conf` value being consumed as a CLI value as well.

I think that this situation should be the exception to keep a value out of `ceph_conf_overrides` and define it as a variable.

andrewschoen:
@alfredodeza Yeah, I tend to agree with you here. Putting this value in ``ceph_conf_overrides.global.osd_pool_default_pg_num`` also hides it a bit from the user. How will we expect them to know to set this and that the tasks expect it? Maybe ``ceph_conf_overrides`` is something we should leave entirely to the user to define and not require any of our tasks to consume values from it.

leseb:
I understand the duality, and yes this is a bit tricky. It's just that I've been chasing what I believe are unnecessary variables. I believe if we keep these variables, users will get confused. Moreover, users will know what they do because:

* if `ceph_conf_overrides.global.osd_pool_default_pg_num` is not declared, we use the default value provided by Ceph
* if `ceph_conf_overrides.global.osd_pool_default_pg_num` exists we use this value BUT user can always change it to something else, so they have granularity.

`ceph_conf_overrides.global.osd_pool_default_pg_num` to me is the default, if this makes it easier or less confusing we can just declare an empty dict `{}`. The only problem is we will need to verify that the dict is not empty...

andrewschoen:
@leseb Maybe we could define ``ceph_conf_overrides.global.osd_pool_default_pg_num`` in the ``ceph-common`` defaults and set it's value to the output of this command https://github.com/ceph/ceph-ansible/pull/1146/files#diff-a8b173d21449acf0c155eaf1ba772872R46 ? That'd mean we'd probably need to move that command to ``ceph-common`` though. Actually, how does ``ceph-client`` get access to this default value that's set in ``ceph-mon``? That might need to go to ``ceph-common`` anyway.

leseb:
@andrewschoen to be populated this implies that the ceph-mon has run so it can set the fact. Probably not ideal I suppose... it's hard to move the logic in cpeh-common has we need the cluster to be up to run certain commands, commands that we can only run on the monitor because this is where we are fetching the default value (if we have to). A bit of a chicken and egg problem...

tests/functional/centos/7/cluster/vagrant_variables.yml
```diff
@@ -10,7 +10,7 @@ mds_vms: 1
 rgw_vms: 1
 nfs_vms: 0
 rbd_mirror_vms: 0
-client_vms: 0
+client_vms: 1
```
andrewschoen: You need to create a ``[clients]`` section in the hosts file for this scenario too.

I think we should add a client node to the xenial_cluster scenario as well, with ``user_config: False`` and ``openstack_config: False``.

leseb: Do you mean a `client.yml`?

andrewschoen: @leseb No, I mean you need to add a section like this to your hosts file:

```
[clients]
client0
```

Like I've done here: https://github.com/ceph/ceph-ansible/pull/1236/files#diff-b0b857e9bebef348b0c2b15e96769ab6R15

My other comment was about testing the inverse of what you've done for ``centos7_cluster`` in ``xenial_cluster``. Meaning, add a client vm, add it to the hosts file and then do not set ``user_config`` or ``openstack_config`` in the group_vars for ``xenial_cluster``.

leseb: Alright, then that I did :). I'll do what you suggest for `xenial_cluster`. Thanks.

leseb:

PR #1150, opened 2016-12-05
tests: use testinfra to verify cluster setup in our CI tests

andrewschoen:
Switches to using the testinfra project to write python tests that verify that state of the cluster for each testing scenario. See docs on testinfra here: https://testinfra.readthedocs.io/en/latest/

This greatly simplifies on our first iteration of python testing for our scenarios and removes the need for each scenario to include a ``scenario.py`` file.  The testinfra project also provides a nice set of pytest fixtures that can be used that I've leverage here.

I've tested this with ``ansible2.2-journal_collocation`` and ``ansible2.2-centos7_cluster`` locally, but I'm gonna depend on the CI tests that trigger with this PR to test the rest.


andrewschoen:

andrewschoen:
``ansible2.2-docker_cluster`` failed when trying to install ``net-tools`` for the new tests, it seems like that scenario will need some special treatment here.

tests/functional/tests/mon/test_mons.py
```diff
@@ -0,0 +1,29 @@
...
+    def test_mon_service_is_running(self, CephNode, Service):
+        service_name = "ceph-mon@ceph-%s" % CephNode["vars"]["inventory_hostname"]
+        assert Service(service_name).is_running
+
+    def test_mon_service_is_enabled(self, CephNode, Service):
+        service_name = "ceph-mon@ceph-%s" % CephNode["vars"]["inventory_hostname"]
+        assert Service(service_name).is_enabled
+
+    def test_can_get_cluster_health(self, CephNode, Command):
+        output = Command.check_output("sudo ceph -s")
```
leseb: This only shows a response from a monitor, we actually don't know Ceph health state from this. Moreover this won't work with clusters with names different than 'ceph'.
I think we should use `ceph --connect-timeout 5 --cluster $CLUSTER_NAME health` and look for `HEALTH_OK`. And we also fail after 5 seconds if the mon didn't respond.

tests/functional/tests/mon/test_mons.py
```diff
@@ -0,0 +1,29 @@
...
+        assert output.strip().startswith("cluster")
+
+
+class TestOSDs(object):
+
+    def test_all_osds_are_up_and_in(self, CephNode, Command):
+        output = Command.check_output("sudo ceph -s")
```
leseb: `--cluster $CLUSTER_NAME`

tests/functional/tests/osd/test_osds.py
```diff
@@ -0,0 +1,23 @@
+
+class TestOSDs(object):
+
+    def test_ceph_osd_package_is_installed(self, CephNode, Package):
+        assert Package("ceph-osd").is_installed
+
+    def test_osd_listens_on_6800(self, CephNode, Socket):
+        assert Socket("tcp://%s:6800" % CephNode["address"]).is_listening
```
leseb: This will only test a single OSD.

tests/functional/tests/osd/test_osds.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def test_osd_services_are_enabled(self, CephNode, Service):
+        # TODO: figure out way to paramaterize CephNode['osd_ids'] for this test
+        for osd_id in CephNode["osd_ids"]:
+            assert Service("ceph-osd@%s" % osd_id).is_enabled
+
+    def test_osd_are_mounted(self, CephNode, MountPoint):
+        # TODO: figure out way to paramaterize CephNode['osd_ids'] for this test
+        for osd_id in CephNode["osd_ids"]:
+            assert MountPoint("/var/lib/ceph/osd/ceph-%s" % osd_id).exists
```
leseb: `$CLUSTER_NAME-%s`

tests/functional/tests/test_install.py
```diff
@@ -1,10 +1,40 @@
...
-    def test_ceph_dir_exists(self):
-        assert os.path.isdir('/etc/ceph')
+    def test_ceph_dir_exists(self, File):
+        assert File('/etc/ceph').exists
+
+    def test_ceph_dir_is_a_directory(self, File):
+        assert File('/etc/ceph').is_directory
+
+    def test_ceph_conf_exists(self, File):
+        assert File('/etc/ceph/ceph.conf').exists
```
leseb: `/etc/ceph/$CLUSTER_NAME.conf`

tests/functional/tests/test_install.py
```diff
@@ -1,10 +1,40 @@
...
+
+    def test_ceph_conf_is_a_file(self, File):
+        assert File('/etc/ceph/ceph.conf').is_file
```
leseb: `/etc/ceph/$CLUSTER_NAME.conf`

tests/functional/tests/test_install.py
```diff
@@ -1,10 +1,40 @@
...
+
+    def test_ceph_command_exists(self, Command):
+        assert Command.exists("ceph")
+
+
+class TestCephConf(object):
+
+    def test_ceph_config_has_inital_members_line(self, CephNode, File):
+        assert File("/etc/ceph/ceph.conf").contains("^mon initial members = .*$")
```
leseb: `/etc/ceph/$CLUSTER_NAME.conf`

tests/functional/tests/test_install.py
```diff
@@ -1,10 +1,40 @@
...
+
+    def test_initial_members_line_has_correct_value(self, CephNode, File):
+        mons = ",".join("ceph-%s" % host
+                        for host in CephNode["vars"]["groups"]["mons"])
+        line = "mon initial members = {}".format(mons)
+        assert File("/etc/ceph/ceph.conf").contains(line)
```
leseb: `/etc/ceph/$CLUSTER_NAME.conf`

tests/functional/tests/test_install.py
```diff
@@ -1,10 +1,40 @@
...
+
+    def test_ceph_config_has_mon_host_line(self, CephNode, File):
+        assert File("/etc/ceph/ceph.conf").contains("^mon host = .*$")
```
leseb: `/etc/ceph/$CLUSTER_NAME.conf`

tests/functional/tests/test_install.py
```diff
@@ -1,10 +1,40 @@
...
 
-    def test_ceph_conf_exists(self):
-        assert os.path.isfile('/etc/ceph/ceph.conf')
+    def test_mon_host_line_has_correct_value(self, CephNode, File):
+        num_mons = len(CephNode["vars"]["groups"]["mons"])
+        mon_ips = []
+        for x in range(0, num_mons):
+            mon_ips.append("{}.1{}".format(CephNode["subnet"], x))
+        line = "mon host = {}".format(",".join(mon_ips))
+        assert File("/etc/ceph/ceph.conf").contains(line)
```
leseb: `/etc/ceph/$CLUSTER_NAME.conf`

tests/functional/tests/mon/test_mons.py
andrewschoen: This test was here mainly just to see if ``sudo ceph -s`` would return a proper status code and could connect to the cluster.  Most (maybe all) of our scenarios don't have a setup that would allow for the cluster to reach ``HEALTH_OK``. We might be able to raise the amount of OSDs on the ``centos7|xenial_cluster`` scenarios to achieve that though.

I don't think we should be verifying that *every* testing scenario we have can reach ``HEALTH_OK``.

tests/functional/tests/osd/test_osds.py
andrewschoen: I lifted this from https://github.com/redhat-cip/openstack-serverspec/blob/master/spec/tests/rbdosd/rbdosd_spec.rb#L9

Is there a range of ports I should be checking?

tests/functional/tests/mon/test_mons.py
leseb: Alright, I don't mind keeping the `ceph -s` but we really need support for cluster name here. I pushed https://github.com/ceph/ceph-ansible/pull/1153 this morning, we can discuss if it's the right approach.

tests/functional/tests/mon/test_mons.py
leseb: Cool

tests/functional/tests/osd/test_osds.py
leseb: 1 OSD = 4 listened ports, 2 on `public_network` and 2 on `cluster_network`. So if we have 2 OSDs, we need to check from 6800 to 6007.

tests/functional/tests/mon/test_mons.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def test_mon_service_is_running(self, CephNode, Service):
+        service_name = "ceph-mon@ceph-%s" % CephNode["vars"]["inventory_hostname"]
+        assert Service(service_name).is_running
+
+    def test_mon_service_is_enabled(self, CephNode, Service):
+        service_name = "ceph-mon@ceph-%s" % CephNode["vars"]["inventory_hostname"]
+        assert Service(service_name).is_enabled
+
+    def test_can_get_cluster_health(self, CephNode, Command):
```
alfredodeza: why is `CephNode` needed here?

tests/conftest.py
```diff
@@ -1,101 +1,47 @@
...
-def pytest_addoption(parser):
-    default = 'scenario.py'
-    parser.addoption(
-        "--scenario",
-        action="store",
-        default=default,
-        help="YAML file defining scenarios to test. Currently defaults to: %s" % default
+@pytest.fixture()
+def CephNode(Ansible, Interface, Command, request):
+    vars = Ansible.get_variables()
```
alfredodeza: this is overriding a builtin

tests/functional/tests/test_install.py
```diff
@@ -1,10 +1,40 @@
...
+        line = "mon initial members = {}".format(mons)
+        assert File("/etc/ceph/ceph.conf").contains(line)
+
+    def test_ceph_config_has_mon_host_line(self, CephNode, File):
+        assert File("/etc/ceph/ceph.conf").contains("^mon host = .*$")
 
-    def test_ceph_conf_exists(self):
-        assert os.path.isfile('/etc/ceph/ceph.conf')
+    def test_mon_host_line_has_correct_value(self, CephNode, File):
+        num_mons = len(CephNode["vars"]["groups"]["mons"])
```
alfredodeza: I think this could be part of the fixture setup. It looks useful enough to be extracted

tests/conftest.py
```diff
@@ -1,101 +1,47 @@
...
 
-def pytest_addoption(parser):
-    default = 'scenario.py'
-    parser.addoption(
-        "--scenario",
-        action="store",
-        default=default,
-        help="YAML file defining scenarios to test. Currently defaults to: %s" % default
+@pytest.fixture()
+def CephNode(Ansible, Interface, Command, request):
```
andrewschoen: It's capitalized to match the rest of the ``testinfra`` fixtures, but doesn't really need to be. I doubt we'll ever operate on a node that's not ceph.

tests/conftest.py
andrewschoen: ok, I can change

tests/conftest.py
```diff
@@ -1,101 +1,47 @@
...
+    vars = Ansible.get_variables()
+    node_type = vars["group_names"][0]
+    if not request.node.get_marker(node_type) and not request.node.get_marker('all'):
+        pytest.skip("Not a valid test for node type")
+
+    if request.node.get_marker("journal_collocation") and not vars.get("journal_collocation"):
+        pytest.skip("Skipping because scenario is not using journal collocation")
```
andrewschoen: Ok, I'll change the message.

Yeah, we probably will run into that situation but let's clean that up in a subsequent PR.

tests/functional/setup.yml
```diff
@@ -0,0 +1,11 @@
+---
+
+- hosts: all
+  gather_facts: true
+  become: yes
+  tasks:
+
+   - name: install net-tools
```
andrewschoen: Sure

tests/functional/tests/mon/test_mons.py
andrewschoen: It's not currently but will be if when I add the custom cluster name support.

andrewschoen:
@leseb I've added the custom cluster name support to this PR.

I'm finishing up some work on the OSD socket tests now and I need to get docker passing, but I think this is close. Depending on how the docker work goes I might just ignore failing tests for now and then follow up with another PR for docker tests.

leseb:
@andrewschoen alright thanks, I'll have another look tomorrow.

andrewschoen:
So it seems like some of the failures in ``docker_cluster`` were valid. 9942b5028a2093032caf155fc9eee0767e78afb3 addresses an issue where the ceph.conf would only include and address for a single mon even if there were multiple mons in the inventory.

The other failing tests are checking for packing installs (which should be skipped) or failures related to the CI tests checking the incorrect service name for containerized OSDS, ``ceph-osd@0`` instead of ``ceph-osd@sda``.

andrewschoen:
jenkins test this please

andrewschoen:

andrewschoen:

andrewschoen:

andrewschoen:
The tests are having lots of trouble with epel repos again today it seems.

andrewschoen:

PR #1155, opened 2016-12-06
take-over: use more precise ceph.conf detection

leseb:
Prior to this patch we were just looking for any *.conf file which
sometimes could results in multiple matches. The new command looks for a
.conf file that must contain [global] and 'fsid' patterns. This will
definitely get us the ceph.conf file. We can not directly use ceph.conf
because of a different cluster name.


leseb:
to be tested by @haklein

Issue #1148, opened 2016-12-05
ceph fsid regenerate on a existing cluster if fetch dir is empty 


mistur:
Hello,

Applying the playbook on a fresh ceph-ansible clone (ie fetch directory is empty) on an existing cluster with the config stores in cluster-<clustername> regenerate a new ceph fsid instead of getting the one of the cluster. 

so the fsid in /etc/ceph/<clustername>.conf is not the one of the cluster

font:
You can manually disable the fsid generation in your `group_vars/all.yml` by setting `generate_fsid: false` and setting the existing `fsid` as shown here https://github.com/ceph/ceph-ansible/blob/80c018b575b92e47ad732f70685dae1f887b4aec/group_vars/all.yml.sample#L199-L206.

mistur:
I try this but got an error when I have applied the playbook with this parameter :

fsid: "<real fsid of my clster"
generate_fsid: false

maybe I souldn't specify the fsid and comment the line instead.

but I think it could be good to protect from creating a new fsid if one already exist.

leseb:
@mistur what was the error?

mistur:
here the error

```
fatal: [iccluster002.iccluster.epfl.ch]: FAILED! => {"failed": true, "msg": "{{ cluster_uuid.stdout }}: 'dict object' has no attribute 'stdout'"}
fatal: [iccluster010.iccluster.epfl.ch]: FAILED! => {"failed": true, "msg": "{{ cluster_uuid.stdout }}: 'dict object' has no attribute 'stdout'"}
fatal: [iccluster018.iccluster.epfl.ch]: FAILED! => {"failed": true, "msg": "{{ cluster_uuid.stdout }}: 'dict object' has no attribute 'stdout'"}
```

here the change :

```--- a/cluster-cephprod/group_vars/all	2016-12-05 14:57:16.481611727 +0100
+++ b/cluster-cephprod/group_vars/all	2016-12-06 15:29:28.329141647 +0100
@@ -196,8 +196,8 @@
 # generated, you may find it useful to disable fsid generation to
 # avoid cluttering up your ansible repo. If you set `generate_fsid` to
 # false, you *must* generate `fsid` in another way.
-fsid: "{{ cluster_uuid.stdout }}"
-generate_fsid: true
+fsid: "f9dfd27f-c704-4d53-9aa0-4a23d655c7c4"
+generate_fsid: false
 
 cephx: true
 max_open_files: 131072
```

PR #1157, opened 2016-12-06
common: do not regenerate a cluster fsid if cluster exits

leseb:
This commit solves the situation where you lost your fetch directory and
you are running ansible against an existing cluster. Since no fetch
directory is present the file containing the fsid doesn't exist so we
are creating a new one. Later the ceph.conf gets updated with a wrong
fsid which causes problems for clients and ceph processes.

Closes: #1148

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -186,14 +186,22 @@
   run_once: true
   when: cephx or generate_fsid
 
-- name: generate cluster uuid
+- name: generate cluster fsid
   local_action: shell python -c 'import uuid; print(str(uuid.uuid4()))' | tee {{ fetch_directory }}/ceph_cluster_uuid.conf
     creates="{{ fetch_directory }}/ceph_cluster_uuid.conf"
   register: cluster_uuid
   become: false
-  when: generate_fsid
+  when:
+    - generate_fsid
+    - ceph_current_fsid.rc != 0
 
-- name: read cluster uuid if it already exists
+- name: reuse cluster fsid when cluster is already running
+  local_action: shell echo {{ fsid }} | tee {{ fetch_directory }}/ceph_cluster_uuid.conf
+    creates="{{ fetch_directory }}/ceph_cluster_uuid.conf"
+  become: false
+  when: ceph_current_fsid.rc == 0
+
+- name: read cluster fsid if it already exists
   local_action: command cat {{ fetch_directory }}/ceph_cluster_uuid.conf
     removes="{{ fetch_directory }}/ceph_cluster_uuid.conf"
   changed_when: false
--- a/roles/ceph-common/tasks/facts.yml
+++ b/roles/ceph-common/tasks/facts.yml
@@ -5,6 +5,18 @@
   always_run: yes
   register: ceph_version
 
+- name: is ceph running already?
+  command: ceph --connect-timeout 3 --cluster {{ cluster }} fsid
+  changed_when: false
+  failed_when: false
+  always_run: yes
+  register: ceph_current_fsid
+  delegate_to: "{{ groups.mons[0] }}"
+
+- set_fact:
+    fsid: "{{ ceph_current_fsid.stdout }}"
+  when: ceph_current_fsid.rc == 0
+
 - set_fact:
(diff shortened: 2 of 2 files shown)


leseb:
@mistur please test this. The mon key has the same problem, I'm fixing this as well.

mistur:
seem to be good for fsid, but mon key seem to not be fetch correctly

fetch/monitor_keyring.conf should be the same as /var/lib/ceph/mon/<cluster name>-<mon fqdn>/keyring, souldn't it ?

leseb:
Yes it should, it's only useful when you want to add new nodes actually

mistur:
this patch does not fix the fetch of the mon keys but only the fsid, right ?

leseb:
Correct, I can have a look into the mon key, later

mistur:
ok, so this patch fix my issue. you can close

roles/ceph-common/tasks/main.yml
```diff
@@ -200,7 +208,9 @@
   register: cluster_uuid
   become: false
   always_run: true
-  when: generate_fsid
+  when:
+    - generate_fsid
+    - ceph_current_fsid.rc != 0
```
font: Wouldn't you still want to run this task to register `cluster_uuid`? It looks like `cluster_uuid` really isn't used later on, but that doesn't mean it won't ever get used in the future.

roles/ceph-common/tasks/main.yml
```diff
@@ -191,7 +191,15 @@
...
+
+- name: generate cluster uuid when cluster exits
+  local_action: shell echo {{ fsid }} | tee {{ fetch_directory }}/ceph_cluster_uuid.conf
+    creates="{{ fetch_directory }}/ceph_cluster_uuid.conf"
+  become: false
+  when: ceph_current_fsid.rc == 0
 
 - name: read cluster uuid if it already exists
   local_action: command cat {{ fetch_directory }}/ceph_cluster_uuid.conf
```
font: Why do we remove `ceph_cluster_uuid.conf` after reading it? This comment actually applies to the next line that shows as collapsed but it appears that github does not allow you to comment on collapsed lines.

roles/ceph-common/tasks/main.yml
leseb: Right, I forgot about this one.

roles/ceph-common/tasks/main.yml
leseb: I don't exactly remember why we are doing this actually. It was a long time ago. I'll look into that tomorrow.

roles/ceph-common/tasks/main.yml
```diff
@@ -191,7 +191,15 @@
     creates="{{ fetch_directory }}/ceph_cluster_uuid.conf"
   register: cluster_uuid
   become: false
-  when: generate_fsid
+  when:
+    - generate_fsid
+    - ceph_current_fsid.rc != 0
+
+- name: generate cluster uuid when cluster exits
```
font: typo 'exits' should be 'exists'

roles/ceph-common/tasks/main.yml
```diff
@@ -191,7 +191,15 @@
     creates="{{ fetch_directory }}/ceph_cluster_uuid.conf"
   register: cluster_uuid
```
font: We register `cluster_uuid` here and below as well.

roles/ceph-common/tasks/main.yml
leseb: fixed

roles/ceph-common/tasks/main.yml
leseb: oh right that'll cause an issue if the previous one is skipped...

leseb: I actually tested several times and it doesn't seem to be a problem. Here's what I did:

* initial bootstrap: fetch dir got created with proper values
* re provision again: no problem
* remove fetch dir and run provision: fetch gets created and ceph_cluster_uuid.conf with the right value

leseb:

leseb:

leseb:

PR #1158, opened 2016-12-06
docker: consolidate ceph-ansible and ceph-docker varible

leseb:
This commit re-uses some of the existing ceph-ansible variables for a
containirzed deployment. There is no reasons why we should add new
variables for the containerized deployment.


group_vars/mons.yml.sample
```diff
@@ -80,12 +80,12 @@ dummy:
...
 #mon_containerized_default_ceph_conf_with_kv: false
-#ceph_mon_docker_interface: eth0
-#ceph_mon_docker_subnet: # subnet of the ceph_mon_docker_interface
+#ceph_mon_docker_interface: "{{ monitor_interface }}"
+#ceph_mon_docker_subnet: "{{ public_network }}"# subnet of the ceph_mon_docker_interface
 #ceph_mon_docker_username: ceph
 #ceph_mon_docker_imagename: daemon
 #ceph_mon_docker_image_tag: latest
-#ceph_mon_extra_envs: "MON_NAME={{ ansible_hostname }}" # comma separated variables
+#ceph_mon_extra_envs: "FSID={{ fsid }},MON_NAME={{ ansible_hostname }}" # comma separated variables
```
leseb: we don't know `fsid` yet so this won't work, we skip ceph-common...

font: This is true, but this variable is really only used for non-CoreOS and non-RedHat based systems. Even then, the `FSID` or `CEPH_FSID` environment variables, where `ceph_mon_extra_envs` is mostly used, aren't even consumed by the entrypoint.

Issue #1161, opened 2016-12-07
timeouts are needed for possible hangs


alfredodeza:
There are some tasks that will hit parts of Ceph that will wait for ever:

    TASK [ceph-mon : collect admin and bootstrap keys] *****************************
    task path: /home/jenkins-build/build/workspace/ceph-ansible-prs-ansible2.2-centos7_cluster/roles/ceph-mon/tasks/start_monitor.yml:64

This is stuck because it hits an issue where the bootstrapping process will try forever until it can get the keys.

A fix is now in the newest Jewel but ceph-ansible needs to support everything that doesn't have this changeset: https://github.com/ceph/ceph/pull/11749

See issue #760 

There are numerous reasons to *why* it might wait/loop forever, so the right fix here is to properly timeout on tasks that request these keys with ceph-create-keys

leseb:
As far as I remember this was a safety measure in some old releases. I'm not sure if we still need this as `ceph-create-keys` should get fired when the monitor process starts.  I remember having issues on centos 7.0 and 7.1 not triggering `ceph-create-keys`.
It's also a safe guard in case the mon process doesn't start properly. However we have this task right after "wait for client.admin key exists" that would fail after 300 seconds.

I just tried to remove the "collect admin and bootstrap keys" task and it seems to be working perfectly on 3 mons. I suspect we can remove this task.

andrelohmann:
Hi there, I know this issue is quite old, but I'm having exactly this problem again on ceph-ansible stable-3.1 deploying the monitor on an ubuntu 16.04.

root@mon1:~# ceph-create-keys --cluster ceph -i mon1 -t 30
INFO:ceph-create-keys:Talking to monitor...
Cluster connection interrupted or timed out
Cluster connection interrupted or timed out
INFO:ceph-create-keys:Cannot get or create admin key, permission denied

leseb:
@andrelohmann please look at your monitor logs to see what's going on.

PR #1162, opened 2016-12-07
mon: remove collect keys

leseb:
Once the monitor process starts it will also trigger `ceph-create-keys`
which will collect the admin key and bootstrap keys. We used to force
this command because we were having issues on some distros like centos
7.0 and 7.1 not triggering this. This is fixed on centos 7.2 and not an
issue on ubuntu 14.04 or 16.04 so we can remove this task. If the
monitor hangs or fails to start the playbook will fail right after at
the "wait for client.admin key exists" task after 300sec.

Closes: #1161

### DIFF
--- a/roles/ceph-mon/tasks/start_monitor.yml
+++ b/roles/ceph-mon/tasks/start_monitor.yml
@@ -60,9 +60,3 @@
   when:
     - use_systemd
     - ceph_release_num.{{ ceph_release }} > ceph_release_num.hammer
-
-- name: collect admin and bootstrap keys
-  command: ceph-create-keys --cluster {{ cluster }} --id {{ monitor_name }}
-  changed_when: false
-  failed_when:  false
-  when: cephx


andrewschoen:
jenkins test this please

andrewschoen:
This will be really nice as that task has locked up many playbook runs for me, I just want to see everything pass with the new CI tests.

andrewschoen:

andrewschoen:

andrewschoen:
More epel failures

batrick:
@leseb @andrewschoen: CentOS 7.2 on Linode has issues with this. The client.admin keyring is not created automatically and the ansible run fails. Any idea why? (I don't really understand why it should be done automatically by the monitors in any case...)

leseb:
@batrick oh really? I don't think we have any issues on CentOS 7.2, we are using it on the CI. Can you run `systemctl status ceph-create-keys@²hostname>`?

batrick:
````
# systemctl status ceph-create-keys@li289-151
● ceph-create-keys@li289-151.service
   Loaded: not-found (Reason: No such file or directory)
   Active: inactive (dead)
[root@li289-151 ~]# cat /etc/redhat-release 
[root@li289-151 ~]# ls /usr/lib/systemd/system/ceph^C
[root@li289-151 ~]# find / -name ceph-create-keys\*
/usr/sbin/ceph-create-keys
/usr/share/man/man8/ceph-create-keys.8.gz
````

leseb:
don't you have `/lib/systemd/system/ceph-create-keys\@.service`?

batrick:
I do not apparently...

leseb:
To be honest, I'm not sure how ceph-create-keys unit is being triggered. @alfredodeza any idea?

PR #1163, opened 2016-12-07
Exclude non-containerized check from containerized osd

jimcurtis:
This new check was added about a week ago but was missing the criteria to only include in non-containerized OSD.

### DIFF
--- a/roles/ceph-osd/tasks/main.yml
+++ b/roles/ceph-osd/tasks/main.yml
@@ -1,5 +1,7 @@
 ---
 - include: check_mandatory_vars.yml
+  when: not osd_containerized_deployment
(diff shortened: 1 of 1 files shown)


leseb:
I think this is still needed for containers. See: https://github.com/ceph/ceph-ansible/pull/1158/files
we consolidate most of the variables, so I think the only check that needs non-container are:

* https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/check_mandatory_vars.yml#L10
* https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/check_mandatory_vars.yml#L23

PR #1168, opened 2016-12-08
Fix the mons running check to use group name var

Logan2211:
mon_group_name variable can be used to override mons group, but
this task assumes the group is always 'mons'. So we need to use
the var to find the group name instead.


leseb:
Good catch thanks!

PR #1171, opened 2016-12-09
bump package version to libcephfs2

cbodley:
I'm seeing some teuthology failures when trying to install ceph packages, such as this on centos (full [teuthology.log](http://qa-proxy.ceph.com/teuthology/cbodley-2016-12-08_13:02:07-rgw-wip-cbodley-testing---basic-smithi/617718/teuthology.log)):
```
Error: Package: 1:ceph-common-11.0.2-2242.gbfb49aa.el7.x86_64 (ceph)
           Requires: libcephfs2 = 1:11.0.2-2242.gbfb49aa.el7
           Installed: 1:libcephfs2-11.0.2-2336.g27f6b03.el7.x86_64 (@ceph)
               libcephfs2 = 1:11.0.2-2336.g27f6b03.el7
           Available: 1:libcephfs2-11.0.2-2242.gbfb49aa.el7.x86_64 (ceph)
               libcephfs2 = 1:11.0.2-2242.gbfb49aa.el7
```
It appears that the existing package for `libcephfs2` was not cleaned up by the ceph-ansible's `Ensure ceph packages are not present` task, which is referring to this package as `libcephfs1`.

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -509,6 +509,7 @@
 
     ceph_remaining_packages:
       - libcephfs1
+      - libcephfs2
       - librados2
       - libradosstriper1
       - librbd1
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -66,7 +66,7 @@ debian_ceph_packages:
   - ceph-common    #|
   - ceph-fs-common #|--> yes, they are already all dependencies from 'ceph'
   - ceph-fuse      #|--> however while proceding to rolling upgrades and the 'ceph' package upgrade
-  - libcephfs1     #|--> they don't get update so we need to force them
+                   #|--> they don't get update so we need to force them
 
 # Whether or not to install the ceph-test package.
 ceph_test: False


ktdreyer:
In terms of installation: "yum install ceph-common" should pull in libcephfs1 or libcephfs2, so I don't think we want Ansible to try to install either one explicitly.

In terms of cleanup: I don't see an issue with ceph-ansible ensuring that both libcephfs1 and libcephfs2 are absent from a system.

cbodley:
> In terms of installation: "yum install ceph-common" should pull in libcephfs1 or libcephfs2, so I don't think we want Ansible to try to install either one explicitly.

the comments in `roles/ceph-common/defaults/main.yml` suggest that `libcephfs1` was there to make sure it wasn't skipped on upgrades. i guess we should either change this to `libcephfs2`, or remove `libcephfs1` from the list. this part should be unrelated to the teuthology failures, so i don't mind either way

> In terms of cleanup: I don't see an issue with ceph-ansible ensuring that both libcephfs1 and libcephfs2 are absent from a system.

okay, it makes sense to leave `libcephfs1` in that list to be safe

ktdreyer:
> there to make sure it wasn't skipped on upgrades.

Good point. In Hammer's `/debian/control`, ceph-common does depend on `python-cephfs (= ${binary:Version})`, but python-cephfs does not depend on the exact version of libcephfs1. So this is needed in Hammer, but not in Jewel or newer.

If we're adding support for Kraken now, I think it's time to drop Hammer support. The [estimated EOL date](http://docs.ceph.com/docs/master/releases/) was last month.

cbodley:
updated to preserve `libcephfs1` in the list of packages for purge-cluster, and removed the old dependency on `libcephfs1` from `debian_ceph_packages`

leseb:

Issue #895, opened 2016-07-21
Deploy ceph using partitions for journal and osds


vincepii:
We are using the roles to deploy ceph with partitions as follows:

```
raw_multi_journal: true

devices:
 - /dev/sdb3

raw_journal_devices:
 - /dev/sdb2
```

During the deployment process, the ceph osd prepare step is skipped because the `check if the device is a partition` task (which runs `"echo '{{ item }}' | egrep '/dev/([hsv]d[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'"`) registers the `ispartition` variable with a successful return code (0).

Later, the `prepare osd disk(s)` task is skipped because the condition

```
item.1.get("rc", 0) != 0
```

is not true.

We are using version `878dc47`, but it seems that these two parts are the same in master.

Do we need some pre-steps in order to use the roles with partitions?

lhoss:
and following (solved and btw similar) issue suggests that devices based on partitions should be supported https://github.com/ceph/ceph-ansible/issues/636

vincepii:
@leseb any ideas?

One thing that is changed is the

```
ignore_errors: true
```

that is gone in the activate OSD task (was there for sure ~1yr ago).

leseb:
@vincepii hum I've never really used it myself. After looking at the code, you should have ended up to this task: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/activate_osds.yml#L36-L47

Did the activation run?

lhoss:
Hi @leseb,
actually we got the deployment done by doing 2 changes, that Im sure are not optimal for other use cases:
* disabling the check mentioned in the issue description:
> Later, the prepare osd disk(s) task is skipped because the condition
>```
>item.1.get("rc", 0) != 0
>```

* by re-adding in the task activation  (which I understand was removed in a PR, to 'fail early' on unwanted failures, which I'ld endorse as the right strategy)
```
ignore_errors: true
```

As this isn't a satisfactory solution, we did not want to create a PR.
IMO the right solution is to work on a proper detection of configured ceph partitions (when working with partitions instead of disks), as the current  <parted...|egrep '1....ceph'> check clearly fails when done a partition 'device' (as obviously a partition doesn't contain a partition table).

zhsj:
Hi there,
you can try latest ceph-ansible to see whether it works well for partitions.

[...]

PR #1178, opened 2016-12-13
Add prepare osd with partition devices in raw_multi_journal

zhsj:
Just simply add a task in scenarios/raw_multi_journal.yml
I tested it locally without problem. But further discussion is welcome to see any better solution.
 
This addresses #895

### DIFF
--- a/infrastructure-playbooks/make-osd-partitions.yml
+++ b/infrastructure-playbooks/make-osd-partitions.yml
@@ -1,0 +1,49 @@
+---
+# This playbook will make custom partition layout for your osd hosts.
+# You should define `devices` variable for every host.
+#
+# For example, in host_vars/hostname1
+#
+# devices:
+# - device_name: sdb
+#   partitions:
+#   - index: 1
+#     size: 10G
+#     type: data
+#   - index: 2
+#     size: 5G
+#     type: journal
+# - device_name: sdc
+#   partitions:
+#   - index: 1
+#     size: 10G
+#     type: data
+#   - index: 2
+#     size: 5G
+#     type: journal
+#
+- vars:
+    osd_group_name: osds
+    journal_typecode: 45b0969e-9b03-4f30-b4c6-b4b80ceff106
+    data_typecode: 4fbd7e29-9d25-41b8-afd0-062c0ceff05d
+  hosts:
+  - "{{ osd_group_name }}"
+
+  tasks:
+  - name: install sgdisk(gdisk)
+    package:
+      name: gdisk
+      state: present
+
+  - name: erase all previous partitions(dangerous!!!)
+    shell: sgdisk --zap-all -- /dev/{{item.device_name}}
+    with_items: "{{ devices }}"
+
+  - name: make osd partitions
+    shell: >
+           sgdisk --new={{item.1.index}}:0:+{{item.1.size}} "--change-name={{item.1.index}}:ceph {{item.1.type}}"
+           "--typecode={{item.1.index}}:{% if item.1.type=='data' %}{{data_typecode}}{% else %}{{journal_typecode}}{% endif %}"
+           --mbrtogpt -- /dev/{{item.0.device_name}}
+    with_subelements:
+    - "{{ devices }}"
+    - partitions
--- a/roles/ceph-osd/tasks/check_devices.yml
+++ b/roles/ceph-osd/tasks/check_devices.yml
@@ -24,14 +24,26 @@
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
   static: False
 
+- name: check the journal device is partition
+  shell: "readlink -f {{ item }} | egrep '/dev/([hsv]d[a-z]{1,2}|cciss/c[0-9]d[0-9]p|nvme[0-9]n[0-9]p)[0-9]{1,2}$'"
+  with_items: "{{ raw_journal_devices }}"
+  changed_when: false
+  failed_when: false
+  always_run: true
+  register: journal_ispartition_results
+
 - name: check the partition status of the journal devices
-  shell: "parted --script {{ item }} print > /dev/null 2>&1"
-  with_items: "{{ raw_journal_devices|unique }}"
+  shell: "parted --script {{ item.1 }} print > /dev/null 2>&1"
+  with_together:
+    - "{{ journal_ispartition_results.results }}"
+    - "{{ raw_journal_devices|unique }}"
   changed_when: false
   failed_when: false
   always_run: true
   register: journal_partition_status
-  when: raw_multi_journal or dmcrypt_dedicated_journal
+  when:
+    - (raw_multi_journal or dmcrypt_dedicated_journal)
+    - item.0.rc != 0
 
 - name: fix partitions gpt header or labels of the journal devices
   shell: "sgdisk --zap-all --clear --mbrtogpt -- {{ item.1 }} || sgdisk --zap-all --clear --mbrtogpt -- {{ item.1 }}"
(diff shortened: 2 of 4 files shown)


zhsj:
ceph-disk will give a warning if partition is not set proper attributes.
So I use the following playbook to make partition:

```
---
- hosts: all
  vars:
#    devices:
#    - device_name: sdb
#      partitions:
#      - index: 1
#        size: 10G
#        type: data
#      - index: 2
#        size: 5G
#        type: journal
    journal_typecode: 45b0969e-9b03-4f30-b4c6-b4b80ceff106
    data_typecode: 4fbd7e29-9d25-41b8-afd0-062c0ceff05d


  tasks:
  - name: install sgdisk(gdisk)
    package:
      name: gdisk
      state: present
    register: pkg_result

  - name: erase all previous partitions(dangerous!!!)
    command: /usr/sbin/sgdisk --zap-all -- /dev/{{item.device_name}}
    with_items: "{{ devices }}"
    register: erase_result
    when: pkg_result|succeeded

  - name: make osd partitions
    command: |
           /usr/sbin/sgdisk --new={{item.1.index}}:0:+{{item.1.size}} "--change-name={{item.1.index}}:ceph {{item.1.type}}"
           "--typecode={{item.1.index}}:{% if item.1.type=='data' %}{{data_typecode}}{% else %}{{journal_typecode}}{% endif %}"
           --mbrtogpt -- /dev/{{item.0.device_name}}
    when: erase_result|succeeded
    with_subelements:
    - "{{ devices }}"
    - partitions
```

andrewschoen:

leseb:
@zhsj Can you also add this playbook in `infrastructure-playbooks`? Thanks!

zhsj:
@leseb plz review

infrastructure-playbooks/make-osd-partitions.yml
```diff
@@ -0,0 +1,52 @@
+---
+# This playbook will make custom partition layout for your osd hosts.
+# You should define `devices` variable for every host.
+#
+# For example, in host_vars/hostname1
+# ```
```
leseb: The backquotes are not needed in the comment

infrastructure-playbooks/make-osd-partitions.yml
```diff
@@ -0,0 +1,52 @@
...
+#     type: journal
+# - device_name: sdc
+#   partitions:
+#   - index: 1
+#     size: 10G
+#     type: data
+#   - index: 2
+#     size: 5G
+#     type: journal
+# ```
```
leseb: ditto

infrastructure-playbooks/make-osd-partitions.yml
```diff
@@ -0,0 +1,52 @@
...
+- hosts: all
```
leseb: Using `all` is dangerous, as it points to all the hosts in your infra, please see: https://github.com/ceph/ceph-ansible/blob/master/infrastructure-playbooks/purge-cluster.yml#L37-L51

And only take osds.

infrastructure-playbooks/make-osd-partitions.yml
```diff
@@ -0,0 +1,52 @@
...
+    package:
+      name: gdisk
+      state: present
+    register: pkg_result
+
+  - name: erase all previous partitions(dangerous!!!)
+    command: /usr/sbin/sgdisk --zap-all -- /dev/{{item.device_name}}
+    with_items: "{{ devices }}"
+    register: erase_result
+    when: pkg_result|succeeded
```
leseb: Why don't you fail instead?
If sgdisk is not present we should fail and tell users to install it.

roles/ceph-osd/tasks/check_devices.yml
```diff
@@ -24,14 +24,26 @@
...
+  with_together:
+    - "{{ journal_ispartition_results.results }}"
+    - "{{ raw_journal_devices|unique }}"
   changed_when: false
   failed_when: false
   always_run: true
   register: journal_partition_status
-  when: raw_multi_journal or dmcrypt_dedicated_journal
+  when:
+    - raw_multi_journal or dmcrypt_dedicated_journal
```
leseb: I'm not 100% sure but you might want to write this like this instead:
`- (raw_multi_journal or dmcrypt_dedicated_journal)`

infrastructure-playbooks/make-osd-partitions.yml
zhsj: I see, no need for this condition. Since the task will fail if previous `package` action doesn't succeed.

zhsj:
@leseb done

infrastructure-playbooks/make-osd-partitions.yml
```diff
@@ -0,0 +1,49 @@
...
+    journal_typecode: 45b0969e-9b03-4f30-b4c6-b4b80ceff106
+    data_typecode: 4fbd7e29-9d25-41b8-afd0-062c0ceff05d
+  hosts:
+  - "{{ osd_group_name }}"
+
+  tasks:
+  - name: install sgdisk(gdisk)
+    package:
+      name: gdisk
+      state: present
```
leseb: can we fail using the `fail` module so we have a message to the user?

infrastructure-playbooks/make-osd-partitions.yml
```diff
@@ -0,0 +1,49 @@
...
+
+  - name: erase all previous partitions(dangerous!!!)
+    command: /usr/sbin/sgdisk --zap-all -- /dev/{{item.device_name}}
+    with_items: "{{ devices }}"
+
+  - name: make osd partitions
+    command: |
+           /usr/sbin/sgdisk --new={{item.1.index}}:0:+{{item.1.size}} "--change-name={{item.1.index}}:ceph {{item.1.type}}"
```
zhsj: I will modify this to `shell: sgdisk` in next force push. Because I find `sgdisk` is at `/sbin/sgdisk` on Debian.

infrastructure-playbooks/make-osd-partitions.yml
leseb: Ok ok.

infrastructure-playbooks/make-osd-partitions.yml
leseb: Please do.

zhsj: Done

infrastructure-playbooks/make-osd-partitions.yml
```diff
@@ -0,0 +1,49 @@
...
+    package:
+      name: gdisk
+      state: present
+
+  - name: erase all previous partitions(dangerous!!!)
+    shell: sgdisk --zap-all -- /dev/{{item.device_name}}
+    with_items: "{{ devices }}"
+
+  - name: make osd partitions
+    shell: >
```
leseb: In this repos we use: `shell: |` instead of `shell: >`

zhsj: And a lot of `shell: >` in https://github.com/ceph/ceph-ansible/blob/fbeacb325cde245bf5f6134d85e7fb6fe3539da6/infrastructure-playbooks/cluster-os-migration.yml#L51
Although command is one line, and `>` is unnecessary.

leseb: Yes `cluster-os-migration.yml` is one of the oldest, so we never moved it to our new convention. I didn't know that `>` avoid the usage of `\n` I actually never add to force it anywhere. In your case you don't need to explicitly put `\` at the end of each line.

zhsj: The shell command is too long that put it in one line is ugly. So I break it into lines. But if using `|`, `\n` will be added to shell command, which results errors.

leseb: Alright let's keep it like this then.

PR #1179, opened 2016-12-13
RGW/Keystone integration improvements

Logan2211:
* Add support for deploying RGW with Keystone user token authentication instead of requiring the deprecated Keystone admin token be enabled.

* Allow disabling the nss PKI database which is not applicable when PKI auth is not used in Keystone.

* Add the ability to configure RGW using the Keystone v3 API.

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -277,6 +277,7 @@ radosgw_civetweb_bind_ip: "{{ ansible_default_ipv4.address }}"
 radosgw_civetweb_num_threads: 50
 radosgw_keystone: false # activate OpenStack Keystone options full detail here: http://ceph.com/docs/master/radosgw/keystone/
 #radosgw_keystone_url: # url:admin_port ie: http://192.168.0.1:35357
+radosgw_keystone_ssl: true # Can be used to disable PKI revocation checks when other token types are used.
 # for admin_token method, define radosgw_keystone_admin_token
 # for auth_token method, define _user, _password, and _tenant
 radosgw_keystone_auth_method: admin_token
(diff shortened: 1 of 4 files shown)


roles/ceph-common/defaults/main.yml
```diff
@@ -277,7 +277,14 @@ radosgw_civetweb_bind_ip: "{{ ansible_default_ipv4.address }}"
 radosgw_civetweb_num_threads: 50
 radosgw_keystone: false # activate OpenStack Keystone options full detail here: http://ceph.com/docs/master/radosgw/keystone/
 #radosgw_keystone_url: # url:admin_port ie: http://192.168.0.1:35357
-radosgw_keystone_admin_token: password
+#radosgw_keystone_api_version: 3 #default is 2
```
leseb: why are all the variables commented?

Logan2211:
@leseb: Fixed it up. Thanks

leseb:
@Logan2211 I have the impression that ` roles/ceph-common/defaults/main.yml` and group_vars/all.yml.sample` are different, did you re-run `generate_group_vars_sample.sh`?

Logan2211:
Indeed there was a whitespace difference in the group_vars sample. I was just syncing the vars by hand, didn't realize there was a script, and I guess my editor stripped the extra whitespace from the end. Should be gtg now. :)

leseb:

leseb:

leseb:

Logan2211:
Was just looking -- this is actually broken now that the commenting is removed from the ceph-common vars. It is supposed to be "either radosgw_keystone_admin_token is defined, or else all of the radosgw_keystone_admin_user/radosgw_keystone_admin_password/radosgw_keystone_admin_tenant vars are defined".. but having them all defaulted will just bypass the user auth stuff all the time.

Logan2211:
Fixed that mess up. Now there is a switch called `radosgw_keystone_auth_method` to switch between admin_token and auth_token.

PR #1181, opened 2016-12-14
Change attribute call to .get call in ceph keys

tynorth-cisco:
If previous check was not run, .stdout_lines is not a valid key on the dictionary.
To get around this, use .get("stdout_lines") instead.

Also add in a default empty list

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -35,7 +35,7 @@
     owner: "{{ key_owner }}"
     group: "{{ key_group }}"
   with_items:
-    - "{{ ceph_keys.stdout_lines }}"
+    - "{{ ceph_keys.get('stdout_lines') | default([]) }}"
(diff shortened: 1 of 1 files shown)


tynorth-cisco:
Example of error
```
TASK [ceph.ceph-mon : set keys permissions] ************************************               
fatal: [mcp1.paslab008003.mc.metacloud.in]: FAILED! => {"failed": true, "msg": "'dict object' has no attribute 'stdout_lines'"}
fatal: [mcp2.paslab008003.mc.metacloud.in]: FAILED! => {"failed": true, "msg": "'dict object' has no attribute 'stdout_lines'"}
fatal: [mcp3.paslab008003.mc.metacloud.in]: FAILED! => {"failed": true, "msg": "'dict object' has no attribute 'stdout_lines'"}
    to retry, use: --limit @/home/tynorth/Work/Codes/ansible-systems/playbooks/openstack/metapod.retry

```

PR #1185, opened 2016-12-14
bump ansible testing version to 2.2.1 for the 2.2 environment

alfredodeza:
`2.2.1` is not released yet, but this will be what we should rely on once it gets out.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -36,9 +36,6 @@ changedir=
   # creates a cluster, purges the cluster and then brings the cluster back up
   purge_cluster_collocated: {toxinidir}/tests/functional/centos/7/journal-collocation
 commands=
-  # install latest stable version of ansible 2.2 from github, this can be removed
-  # when there is a release > 2.2.1.0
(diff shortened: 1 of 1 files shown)


alfredodeza:
Everything failing is expected since 2.2.1 is not out yet.

leseb:
@alfredodeza so we should just wait for the release right?

ktdreyer:
Depending on the gdeploy situation (https://github.com/gluster/gdeploy/issues/253), it might be nice to test with both 2.2.0.0 and 2.2.1.0, so we can evaluate the feasibility of keeping parity with what they're doing.

alfredodeza:
@leseb yes, we should wait until the release is out.

@ktdreyer we can certainly add variants (we do have them available in `tox.ini` right now). But maybe we could set those to run daily (vs. for every pull request)

alfredodeza:
jenkins test this please

ktdreyer:
Marking for backport so we'll test stable-2.1 with this version of Ansible as well.

Issue #1182, opened 2016-12-14
set apt pinning with Red Hat Ceph Storage


ktdreyer:
When `ceph_stable_rh_storage` is True, every cluster node should have a `/etc/apt/preferences.d/rhcs.pref` file with the following contents:

```
Explanation: Prefer Red Hat packages
Package: *
Pin: release o=/Red Hat/
Pin-Priority: 999
```

ceph-deploy already did this when used with ice-setup, and we need to do the same thing with the ceph-ansible stack.

PR #1186, opened 2016-12-14
common: set apt pinning with Red Hat Ceph Storage

leseb:
When `ceph_stable_rh_storage` is True, every cluster node should have a
`/etc/apt/preferences.d/rhcs.pref` file with the following contents:

```
Explanation: Prefer Red Hat packages
Package: *
Pin: release o=/Red Hat/
Pin-Priority: 999
```

ceph-deploy already did this when used with ice-setup, and we need to do
the same thing with the ceph-ansible stack.

Closes: #1182

### DIFF
--- a/roles/ceph-common/tasks/installs/install_rh_storage_on_debian.yml
+++ b/roles/ceph-common/tasks/installs/install_rh_storage_on_debian.yml
@@ -25,6 +25,14 @@
     - ceph_rhcs
     - ceph_rhcs_iso_install
 
+- name: set apt pinning for red hat ceph storage
+  template:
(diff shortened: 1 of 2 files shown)


leseb:
@ktdreyer

ktdreyer:
There are some changes to `name:` in here, that would make this change harder to cherry-pick, if we had to do a cherry-pick. Want to break those out into a separate commit?

leseb:
@ktdreyer yes that'd be better :)

leseb:
@ktdreyer done!

andrewschoen:

PR #1195, opened 2016-12-15
rpm: require Ansible >= 2.2.0.0

ktdreyer:
This is the only version that our CI uses for testing, so it's the only version we can confidently say works.

Update the RPM packaging to specifically require this version of Ansible.


ktdreyer:
A side effect of this change is that yum will no longer try to pull in the "ansible19" package from EPEL in order to satisfy the dependency.

PR #1203, opened 2016-12-20
common: enable tool repo for mds install of rhcs

leseb:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1405985


andrewschoen:
Should we label PRs like this so we know we need to backport them to our downstream stable branch?

leseb:
Yes, we should.

alfredodeza:
This looks OK to me. 

Discussed this briefly with @ErwanAliasr1 and we should always target master and then backport with a cherry-pick, *unless* the patch doesn't make sense for master, in which case it would go to the stable downstream branch

ErwanAliasr1:
@alfredodeza +1

ErwanAliasr1:
@leseb I do think that the backport label should integrate the name of the targeted stable branch to ease the work & tracking. As per today it's pretty implicit, by the time it will not. So using the good habit from the beginning could be efficient.

leseb:
@ErwanAliasr1 yes I just used the generic "backport" term because we don't know the name of the release yet, I'll change that once we know :).

ktdreyer:
Reflecting on this, I think we should cherry-pick both of those commits in this PR, rather than only the first commit. It'll make future backports easier if the file is named "rhcs" in both branches

PR #1204, opened 2016-12-20
move monitor definition from ceph-mon back to ceph-common

tynorth-cisco:
ceph-common will overwrite ceph.conf on every run, so
when ceph-mon also edits the file it will cause an
unnecessary restart of ceph-mon

### DIFF
--- a/roles/ceph-common/templates/ceph.conf.j2
+++ b/roles/ceph-common/templates/ceph.conf.j2
@@ -145,3 +145,15 @@ nss db path = {{ radosgw_nss_db_path }}
 keyring = /var/lib/ceph/restapi/ceph-restapi/keyring
 log file = /var/log/ceph/ceph-restapi.log
 {% endif %}
+
+{% if groups[mon_group_name] is defined %}
+{% for host in groups[mon_group_name] %}
+{% if hostvars[host]['ansible_fqdn'] is defined and mon_use_fqdn %}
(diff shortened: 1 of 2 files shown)


leseb:
@tynorth-cisco although it makes sense, I have a quick question, are you running a Hammer cluster? Hammer is going EOL pretty soon which means we won't support it. We will actually branch and provide bug fixes. To be honest, I'm not really keen on changing the template even if this sounds like the right solution to fix this.

leseb:
@tynorth-cisco I had more thoughts about this with @ErwanAliasr1, I just created a branch just for you so you can push fixes for Hammer. Branch name is: `unsupported-2.1`. I'll merge **all** of your bug fixes **only** for Hammer and other non-supported Ceph versions.
We have to this because we can not add this patch since there is no conditional in it. I don't like the idea of changing the template as this will change all the current config files (then trigger unnecessary restarts).

How does that sound for you?

tynorth-cisco:
@leseb yeah that sounds good, thanks!

leseb:
@tynorth-cisco cool,  so please re-push this into the right branch :)

tynorth-cisco:
@leseb push directly or pull?

(sorry im just paranoid about breaking things)

leseb:
@tynorth-cisco you won't break anything, no worries. Close this PR and push the same change against the branch `unsupported-2.1`.

Issue #1207, opened 2016-12-24
Multi-site HAProxy + SSL


clwells:
I'm trying to run multi-site with HAProxy + SSL.  A single site setup works great but it looks like the multisite/secondary playbook wants keys to be sent over http (without any easy option to switch) and to the civetweb port (again, without any option to switch).

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-rgw/tasks/multisite/secondary.yml#L2

Can I PR a change that would allow this to be overridden?

PR #1208, opened 2016-12-26
[rtr] Granular Multi-site PULL Variables

clwells:
Allowing granular control of the pull host/port/protocol for installs where SSL/443 is used and direct civetweb access isn't necessarily available.

fixes: #1207

### DIFF
--- a/roles/ceph-rgw/defaults/main.yml
+++ b/roles/ceph-rgw/defaults/main.yml
@@ -22,6 +22,10 @@ cephx: true
 # allowing root to not require tty
 radosgw_user: root
 
+# Multi-site remote pull URL variables
+rgw_pullport: "{{ radosgw_civetweb_port }}"
+rgw_pullproto: "http"
+
 ##########
 # DOCKER #
 ##########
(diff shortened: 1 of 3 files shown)


roles/ceph-rgw/defaults/main.yml
```diff
@@ -22,6 +22,10 @@ cephx: true
 # allowing root to not require tty
 radosgw_user: root
 
+# Multi-site remote pull URL variables
+rgw_pullport: "{{ radosgw_civetweb_port }}"
+rgw_pullproto: "http"
```
leseb: Quotes are not needed.
what about `rgw_pull_proto` instead?

roles/ceph-rgw/defaults/main.yml
```diff
@@ -22,6 +22,10 @@ cephx: true
 # allowing root to not require tty
 radosgw_user: root
 
+# Multi-site remote pull URL variables
+rgw_pullport: "{{ radosgw_civetweb_port }}"
```
leseb: what about `rgw_pull_port` instead?

clwells: Sure, I was just keeping with the current standard: rgw_pull_host, but can change it if you like.

roles/ceph-rgw/defaults/main.yml
clwells: Same as above, just trying to keep with standard.

PR #1209, opened 2016-12-28
fix check nmapexist on debian

zhsj:
command is a shell-buitin, so `command -v nmap` must use shell module

Signed-off-by: Shengjing Zhu <zsj950618@gmail.com>


leseb:
Why? Can we see the error? `command` should work fine since there are no pipes or/and special characters.

zhsj:
@leseb unlike centos, there is no `/usr/bin/command` in debian. `command` is a shell built in. 

```
zsj@debian ~ $ which command
command: shell built-in command
```

PR #1214, opened 2017-01-03
mon: make sure osd_pool_default_size is honoured

leseb:
This patch makes sure we set the proper pool size on the rbd pool.
Usually during bootstrap the rbd pool size is not honoured so we need to
add this workaround.

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -20,7 +20,7 @@
 - name: test if initial mon keyring is in mon kv store
   command: ceph --cluster {{ cluster }} config-key get initial_mon_keyring
   changed_when: false
-  failed_when: false
+  ignore_errors: true
   always_run: true
   run_once: true
   register: is_initial_mon_keyring_in_kv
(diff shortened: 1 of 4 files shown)


leseb:
jenkins test this please

ktdreyer:
Mind linking to the upstream Ansible bug, so we'll know when it's fixed and we can revert this?

leseb:
@ktdreyer do you want me to create a bug for this? Is this what you meant?

ktdreyer:
@leseb in your "ci: fix the purge cluster scenario" commit, it says "There is an Ansible bug which makes the playbook fail when we are running a playbook from the non-git root directory."

What is the link to that bug in https://github.com/ansible/ansible ?

alfredodeza:
@leseb can you split these changes into single commits? we can't cherry-pick right now because one or more commits span changes on multiple files

leseb:
@ktdreyer added the ansible bug
@alfredodeza which commit do you want to decouple?

andrewschoen:
@leseb the bit @alfredodeza wanted in it's own commit was the copying of purge-cluster.yml from infrastructure-playbooks/ to the root of ceph-ansible.

I've just opened https://github.com/ceph/ceph-ansible/pull/1221 that includes that fix though.

leseb:
@andrewschoen alright let me remove my commit then.

andrewschoen:
@leseb is this PR still valid or did https://github.com/ceph/ceph-ansible/pull/1236 supersede it?

leseb:
@andrewschoen still valid, pushed a new version

PR #1217, opened 2017-01-05
ceph-common: always include release.yml

ktdreyer:
Prior to this change, a playbook run with `--tags` or `--skip-tags` would fail, because the ceph-common role would not include the release.yml task, and this file defines critical things like ceph_release.

Thanks @andrewschoen for help with the fix.

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -4,6 +4,8 @@
 
 # Set ceph_release
 - include: ./release.yml
+  tags:
+    - always
 
 - include: ./checks/check_firewall.yml
   when: check_firewall


alfredodeza:
if this needs to run always I don't see much value in having it as a separate file other than making main.yml smaller

andrewschoen:
@alfredodeza tagging the include does let us avoid the need to tag the 2 tasks in ``release.yml`` though.

ktdreyer:
After discussion, we agreed that just adding the tag here is the least-intrusive fix for the 2.1.z series. If we want to merge `release.yml` back into `main.yml`, we can re-organize that in a separate PR

andrewschoen:
The ``purge_cluster_collocated`` test is known to fail currently and is not caused by this PR.

ktdreyer:
`stable-2.1` PR is https://github.com/ceph/ceph-ansible/pull/1218

PR #1221, opened 2017-01-10
tests: fixes for purge_cluster_collocated

andrewschoen:
I need to remove those include_vars lines because including those default vars overwrites what we have defined here: https://github.com/ceph/ceph-ansible/blob/master/tests/functional/centos/7/journal-collocation/group_vars/all

Because the defaults are being used ``devices`` is set to ``[]`` which causes this task to skip, https://github.com/ceph/ceph-ansible/blob/master/infrastructure-playbooks/purge-cluster.yml#L395, resulting in OSDs not being properly purged.

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -287,8 +287,6 @@
      state: absent
 
   tasks:
-  - include_vars: roles/ceph-common/defaults/main.yml
-  - include_vars: roles/ceph-osd/defaults/main.yml
   - include_vars: group_vars/all.yml
     ignore_errors: true
   - include_vars: group_vars/{{ osd_group_name }}.yml
--- a/tox.ini
+++ b/tox.ini
@@ -7,6 +7,7 @@ whitelist_externals =
     vagrant
     bash
     pip
+    cp
 passenv=*
 setenv=
   ANSIBLE_SSH_ARGS = -F {changedir}/vagrant_ssh_config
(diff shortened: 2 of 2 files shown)


andrewschoen:
jenkins test ansible2.2-purge_cluster_collocated

infrastructure-playbooks/purge-cluster.yml
```diff
@@ -287,8 +287,6 @@
      state: absent
 
   tasks:
-  - include_vars: roles/ceph-common/defaults/main.yml
```
leseb: why?

andrewschoen:
jenkins test ansible2.2-purge_cluster_collocated

infrastructure-playbooks/purge-cluster.yml
```diff
@@ -287,8 +287,6 @@
...
-  - include_vars: roles/ceph-osd/defaults/main.yml
```
alfredodeza: I understand the change for testing, but for a regular use case, wouldn't a user need to have these defined here? Or it should all still work correctly without including these files/vars?

andrewschoen: As far as I can tell there is nothing in the "purge ceph osd cluster" play that would require anything defined in these defaults. The only value it's using that's even defined in these defaults is ``devices``.

I'm honestly not even sure why we're trying to include files from a relative ``group_vars`` directory. In my experience the correct way to use ansible is to have a ``group_vars`` directory relative to the inventory being used, not the playbook.

Issue #1224, opened 2017-01-13
osd_directories is read although osd_directory is false


D3N14L:
Somehow our ceph-ansible installation gets broken when using:
Ubuntu 16.04
ansible v2.2.0.0
ceph-ansible v2.1.0

I have added some debug tasks to show the variable values:
```
TASK [ceph-osd : debug] ********************************************************
ok: [blk0001_ssd] => {
    "osd_directory": false
}
ok: [blk0002_ssd] => {
    "osd_directory": false
}
ok: [blk0003_ssd] => {
    "osd_directory": false
}

TASK [ceph-osd : debug] ********************************************************
ok: [blk0001_ssd] => {
    "osd_directories": "VARIABLE IS NOT DEFINED!"
}
ok: [blk0002_ssd] => {
    "osd_directories": "VARIABLE IS NOT DEFINED!"
}
ok: [blk0003_ssd] => {
    "osd_directories": "VARIABLE IS NOT DEFINED!"
}

TASK [ceph-osd : get osd id for directory scenario] ****************************
fatal: [blk0001_ssd]: FAILED! => {"failed": true, "msg": "'osd_directories' is undefined"}
fatal: [blk0002_ssd]: FAILED! => {"failed": true, "msg": "'osd_directories' is undefined"}
fatal: [blk0003_ssd]: FAILED! => {"failed": true, "msg": "'osd_directories' is undefined"}
```
This is the task  from `ceph-osd/tasks/osd_fragment.yml` that fails:
```yaml
- name: get osd id for directory scenario
  command: cat {{ item.stdout }}/whoami
  with_items: "{{ osd_directories }}"
  changed_when: false
  failed_when: false
  always_run: true
  register: osd_id_dir_scenario
  when: osd_directory
```
I think the intention is clear - skip the task completely, if `osd_directory` is set to `false` - , but obviously the `when` line here is interpreted for each item in `osd_directories` which is undefined.

D3N14L:
My quickfix was to change:
`  with_items: "{{ osd_directories }}"`
to:
`  with_items: "{{ osd_directories|default([]) }}"`

andrewschoen:
@D3N14L I think it'd probably be safe to just default ``osd_directories`` in ``ceph-osd/defaults/main.yml`` here https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/defaults/main.yml#L136

[...]

PR #1225, opened 2017-01-13
Set empty list as default for osd_directories

D3N14L:
As described in issue #1224 leaving this variable undefined may
cause a problem during execution of the ceph-osd role.


D3N14L:
done.

andrewschoen:
Thanks!

PR #1018, opened 2016-10-07
profiles: introducing cluster profiles

leseb:
This commit introduces a new directory called "profiles" which
contains some set of variables for a particular use case. These profiles
provide guidance for certain scenarios such as:
- configuring rgw with keystone v3

Signed-off-by: Sébastien Han seb@redhat.com

PR #1226, opened 2017-01-16
Enhance RGW config options

Logan2211:
- Allow for enhanced configurability of the civetweb parameters
- Add switch to enable CNAME bucket resolution (ie. some.website.com.objects.example.com) where bucket name is "some.website.com"
- Ability to enable RGW usage logging and its associated configuration parameters which are detailed at http://docs.ceph.com/docs/jewel/man/8/radosgw/#usage-logging
- Add switch to enable S3-style static website hosting.

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -284,6 +284,10 @@ mds_max_mds: 3
 radosgw_civetweb_port: 8080 # on Infernalis we get: "set_ports_option: cannot bind to 80: 13 (Permission denied)"
 radosgw_civetweb_bind_ip: "{{ ansible_default_ipv4.address }}"
 radosgw_civetweb_num_threads: 50
+# For additional civetweb configuration options available such as SSL, logging,
+# keepalive, and timeout settings, please see the civetweb docs at
+# https://github.com/civetweb/civetweb/blob/master/docs/UserManual.md
+radosgw_civetweb_options: "port={{ radosgw_civetweb_bind_ip }}:{{ radosgw_civetweb_port }} num_threads={{ radosgw_civetweb_num_threads }}"
 radosgw_keystone: false # activate OpenStack Keystone options full detail here: http://ceph.com/docs/master/radosgw/keystone/
 #radosgw_keystone_url: # url:admin_port ie: http://192.168.0.1:35357
 radosgw_keystone_api_version: 2 # API versions 2 and 3 are supported
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -301,6 +301,12 @@ radosgw_keystone_token_cache_size: 10000
 radosgw_keystone_revocation_internal: 900
 radosgw_s3_auth_use_keystone: "true"
 radosgw_nss_db_path: /var/lib/ceph/radosgw/ceph-radosgw.{{ ansible_hostname }}/nss
+# Settings for the RGW usage logging described at http://docs.ceph.com/docs/jewel/man/8/radosgw/#usage-logging
+radosgw_usage_log: false
+radosgw_usage_log_tick_interval: 30
+radosgw_usage_log_flush_threshold: 1024
+radosgw_usage_max_shards: 32
+radosgw_usage_max_user_shards: 1
 # Rados Gateway options
 email_address: foo@bar.com
 
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -311,6 +311,9 @@ radosgw_usage_log_tick_interval: 30
 radosgw_usage_log_flush_threshold: 1024
 radosgw_usage_max_shards: 32
 radosgw_usage_max_user_shards: 1
+# Settings for static website hosting
+radosgw_static_website: false
+radosgw_dns_s3website_name: your.subdomain.tld # subdomain used by radosgw for website bucket hosting.
 # Rados Gateway options
 email_address: foo@bar.com
 
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -281,6 +281,7 @@ mds_max_mds: 3
 ## Rados Gateway options
 #
 #radosgw_dns_name: your.subdomain.tld # subdomains used by radosgw. See http://ceph.com/docs/master/radosgw/config/#enabling-subdomain-s3-calls
+radosgw_resolve_cname: false # enable for radosgw to resolve DNS CNAME based bucket names
 radosgw_civetweb_port: 8080 # on Infernalis we get: "set_ports_option: cannot bind to 80: 13 (Permission denied)"
 radosgw_civetweb_bind_ip: "{{ ansible_default_ipv4.address }}"
 radosgw_civetweb_num_threads: 50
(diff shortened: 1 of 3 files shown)


roles/ceph-common/defaults/main.yml
```diff
@@ -73,10 +73,9 @@ ntp_service_enabled: true
 # install source or architecture.
 debian_ceph_packages:
   - ceph
-  - ceph-common    #|
-  - ceph-fs-common #|--> yes, they are already all dependencies from 'ceph'
-  - ceph-fuse      #|--> however while proceding to rolling upgrades and the 'ceph' package upgrade
-                   #|--> they don't get update so we need to force them
+  - ceph-common    #|--> yes, they are already all dependencies from 'ceph'
```
leseb: did you rebase on master? I don't think this should be here.

roles/ceph-common/defaults/main.yml
```diff
@@ -282,9 +281,11 @@ mds_max_mds: 3
 ## Rados Gateway options
 #
 #radosgw_dns_name: your.subdomain.tld # subdomains used by radosgw. See http://ceph.com/docs/master/radosgw/config/#enabling-subdomain-s3-calls
+radosgw_resolve_cname: false # enable for radosgw to resolve DNS CNAME based bucket names
 radosgw_civetweb_port: 8080 # on Infernalis we get: "set_ports_option: cannot bind to 80: 13 (Permission denied)"
 radosgw_civetweb_bind_ip: "{{ ansible_default_ipv4.address }}"
 radosgw_civetweb_num_threads: 50
+radosgw_frontends: "civetweb port={{ radosgw_civetweb_bind_ip }}:{{ radosgw_civetweb_port }} num_threads={{ radosgw_civetweb_num_threads }}"
```
leseb: I think we could add some example of additional variables as well:

`request_timeout_ms=<new timeout value in milliseconds> error_log_file=/var/log/radosgw/civetweb.error.log access_log_file=/var/log/radosgw/civetweb.access.log`

roles/ceph-common/templates/ceph.conf.j2
```diff
@@ -111,7 +111,8 @@ keyring = /var/lib/ceph/radosgw/{{ cluster }}-rgw.{{ hostvars[host]['ansible_hos
 rgw socket path = /tmp/radosgw-{{ hostvars[host]['ansible_hostname'] }}.sock
 log file = /var/log/ceph/{{ cluster }}-rgw-{{ hostvars[host]['ansible_hostname'] }}.log
 rgw data = /var/lib/ceph/radosgw/{{ cluster }}-rgw.{{ hostvars[host]['ansible_hostname'] }}
-rgw frontends = civetweb port={{ radosgw_civetweb_bind_ip }}:{{ radosgw_civetweb_port }} num_threads={{ radosgw_civetweb_num_threads }}
+rgw frontends = {{ radosgw_frontends }}
```
leseb: I'm tempted to keep the "civetweb" hardcoded, we don't want any users using Apache/FastCGI anymore.

Logan2211: We could do that, and I could change it to 
```
radosgw_civetweb_options: "port={{ radosgw_civetweb_bind_ip }}:{{ radosgw_civetweb_port }} num_threads={{ radosgw_civetweb_num_threads }}"
```

roles/ceph-common/defaults/main.yml
Logan2211: Yes please see the commit comments for 7ae918620355125cbeb8419005c19885dbd3bc3e -- the package was removed earlier, however the current master has ugly formatting for the comments in ceph-common, and also the package removal was never synced to `group_vars/all.yml.sample` so if you run `generate_group_vars_sample.sh` on a clean checkout from master currently it will have this change. This commit fixes the comment formatting and brings the group_vars sample back into sync with ceph-common defaults.

Logan2211: Should I submit 7ae918620355125cbeb8419005c19885dbd3bc3e as a separate PR?

roles/ceph-common/defaults/main.yml
Logan2211: Sure, maybe it best thing would be to add a comment with a link to https://github.com/civetweb/civetweb/blob/master/docs/UserManual.md? SSL, logging, and timeouts are probably the top options we're interested in but there are a ton of others.

roles/ceph-common/defaults/main.yml
Logan2211: This is in #1227 now. This is rebased on top of #1227.

leseb: Thanks @Logan2211

leseb:

leseb:

leseb:

roles/ceph-common/defaults/main.yml
```diff
@@ -302,6 +306,15 @@ radosgw_keystone_token_cache_size: 10000
 radosgw_keystone_revocation_internal: 900
 radosgw_s3_auth_use_keystone: "true"
 radosgw_nss_db_path: /var/lib/ceph/radosgw/ceph-radosgw.{{ ansible_hostname }}/nss
+# Settings for the RGW usage logging described at http://docs.ceph.com/docs/jewel/man/8/radosgw/#usage-logging
+radosgw_usage_log: false
+radosgw_usage_log_tick_interval: 30
```
andrewschoen: Are these config options better put into the ``rgw`` section of ``ceph_conf_overrides`` and not defined here as variables that are put manually into our ceph.conf template?

@leseb thoughts?

Logan2211: There's no `rgw` section in the rendered configs. The radosgw sections are _per host_ so defining ceph_conf_overrides for rgw becomes tedious very quickly.

andrewschoen: @Logan2211 yeah, I do see that in the template. I just asked because ``ceph_conf_overrides`` apparently supports adding or overriding an ``rgw`` section. https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/defaults/main.yml#L326

Logan2211: Fair point. I wonder if defining in `[rgw]` would set baseline defaults across all of the `[client.rgw.hostname]` instances. Regardless, I still think it wouldn't be a bad thing to expand the rgw configurability in the way proposed here, but its worth exploring possible alternatives :)

andrewschoen: I suspect defining these in ``[rgw]`` would set those baseline defaults for all rgw nodes, but I don't know that for certain.

Your approach does have the advantage of exposing these config options to the user, but I know that @leseb has been pushing to reduce the amount of config variables like this and promote usage of ``ceph_config_overrides``.

leseb: Unfortunately, if I'm saying "yes" to these rgw additions in the template is because we cannot use `ceph_conf_overrides`, see this PR: https://github.com/ceph/ceph-ansible/pull/1018

Basically if we add a new section, we will end up with 2 rgw sections...

roles/ceph-common/defaults/main.yml
leseb: Can you please rebase?

roles/ceph-common/defaults/main.yml
leseb: We could potentially fix that by introducing a really disruptive change. There is no need for a rgw.$(hostname) section, we can hardcode the name of the gateway with something like `rados.gateway`. IIRC we actually used to do that...
However this is just so disruptive for current user, and I don't know how we'd maintain a backward compatibility...

Logan2211:
rebased ^

leseb:
jenkins test this please

Issue #1173, opened 2016-12-09
"wait for client.admin key" failed


davidebelloni:
Hi,
with CoreOS and etcd the following step fail:

> ...
> TASK [ceph-mon : run the ceph monitor docker image] ****************************
> 
> TASK [ceph-mon : run the ceph monitor docker image with kv] ********************
> 
> TASK [ceph-mon : wait for client.admin key exists] *****************************
> fatal: [10.1.2.20]: FAILED! => {"changed": false, "elapsed": 300, "failed": true, "msg": "Timeout when waiting for file /etc/ceph/ceph-test.client.admin.keyring"}
> fatal: [10.1.2.22]: FAILED! => {"changed": false, "elapsed": 300, "failed": true, "msg": "Timeout when waiting for file /etc/ceph/ceph-test.client.admin.keyring"}
> fatal: [10.1.2.21]: FAILED! => {"changed": false, "elapsed": 300, "failed": true, "msg": "Timeout when waiting for file /etc/ceph/ceph-test.client.admin.keyring"}
> 

My setup is:

> mon_containerized_deployment: true
> mon_containerized_deployment_with_kv: true
> mon_containerized_default_ceph_conf_with_kv: true

The problems that I see are the following:

1. with etcd the client.admin key should be on KV store?
2. with this setup which of the following conf keys should be true and which should be false?
```
mon_containerized_deployment
mon_containerized_deployment_with_kv
mon_containerized_default_ceph_conf_with_kv
```
3. the name of the cluster here is "ceph-test", but in the container the file has the following name: /etc/ceph/ceph.client.admin.keyring

Thanks

harobed:
Same problem here: http://lists.ceph.com/pipermail/ceph-ansible-ceph.com/2016-December/000007.html

Fixed if I use this revision: https://github.com/ceph/ceph-ansible/commit/7812e0b32ea4ff91cc95a5e4a13e765d5d3aed49

I think there is one regression in last commits.

davidebelloni:
Sorry but with revision 7812e0b I have the same issues.
What is you setup and configuration?

Thanks!

andrewschoen:
@davidebelloni have you tried without using a custom cluster name? IIRC, the docker deployments have issue with custom cluster names.

Logan2211:
[...]

PR #1230, opened 2017-01-17
mon: run ceph-create-keys after startup for kraken release

lae:
Followup to #1173 

As part of https://github.com/ceph/ceph/pull/9345 the ceph-mon service no longer implicitly calls `ceph-create-keys` to generate the admin client key.


roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -1,4 +1,13 @@
 ---
+- name: collect admin and bootstrap keys (for or after kraken release)
+  command: ceph-create-keys --cluster {{ cluster }} -i {{ monitor_name }}
+  args:
+    creates: /etc/ceph/{{ cluster }}.client.admin.keyring
+  failed_when: false
```
leseb: Can you add `always_run: true` and `changed_when: false`?

PR #1231, opened 2017-01-17
Purge cluster fixes

alfredodeza:
These fixes are in master but were never ported back to the stable-2.1 branch.

Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1413985

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -74,9 +74,9 @@
   - include_vars: roles/ceph-common/defaults/main.yml
   - include_vars: roles/ceph-mds/defaults/main.yml
   - include_vars: group_vars/all.yml
-    failed_when: false
+    ignore_errors: true
   - include_vars: group_vars/{{ mds_group_name }}.yml
-    failed_when: false
+    ignore_errors: true
 
   - name: stop ceph.target with systemd
     service:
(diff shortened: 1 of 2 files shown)


infrastructure-playbooks/purge-cluster.yml
```diff
@@ -408,6 +400,14 @@
      ceph_data_partlabels.rc == 0 and
      zap_block_devs
 
+  - name: get ceph journal partitions
+    shell: |
+      blkid | awk '/ceph journal/ { sub (":", "", $1); print $1 }'
+    when:
+      - ceph_journal_partlabels.rc == 0
```
leseb: With a single line, we usually do it like this: `when: ceph_journal_partlabels.rc == 0` only.
(yes I know, **I** did that...)

andrewschoen:
@leseb That sounds like a reasonable solution, I'm not sure we can trust "Since we import group_vars and defaults" though, I've seen those includes not work and cause unexpected failures.

I think would should cherry-pick these commits unmodified from master and then if we want to clean this up it should be in a separate PR against the master branch. Thoughts?

leseb:
@andrewschoen right let's just do cherry-picks here and in a separate PR we will address the refactor.

andrewschoen:
I'm going to go ahead and merge this and we'll address @leseb's concern in another PR.

alfredodeza:
@leseb agree on improving/changing things somewhere else, this is just a backport.

You do mention however that `zap_block_devs` should be used. I don't think a user should need to know about that flag when purging a cluster to be able to purge correctly. It is *nice* to have it around for granular control though.

The variable will probably need to be renamed as it implies it will be used to "zap" a "block device" but it is used as a conditional for the task: "zap ceph journal partitions"

leseb:
@alfredodeza right, let's address this in another PR. Let me have a look at that.

Issue #1241, opened 2017-01-20
docker mon setup uses short hostname for monitors


ibotty:
It uses `ansible_hostname` which makes a host like a.a.example.com become host a monitor `mon-a`. A host a.b.example.com will now also host `mon-a`. I _think_ using `ansible_fqdn` is correct, while a little verbose. Of course that won't be backwards-compatible.

Ideas?

leseb:
We can change the template by exposing `MON_NAME` to it and default it to $(hostname), then you can change this variable later. Let me quickly do that.

leseb:
@ibotty https://github.com/ceph/ceph-ansible/pull/1244

leseb:
@ibotty please do your changes here: https://github.com/ceph/ceph-ansible/blob/625ee0952e34036f85dc959a8710d973e59c8df4/roles/ceph-mon/defaults/main.yml#L79

ibotty:
Thank you!

ibotty:
Well, finally got to it again. Just setting MON_NAME is unfortunately by far not enough.

There are the following issues (which I hacked around) because `{{ansible_hostname}}` is hardcoded in quiet a few places. I can prepare a pull request that abstracts the MON_NAME (and MDS_NAME, etc.)

 * There are the `mon_use_fqdn` and `mds_use_fqdn` variables to be set,
 * `roles/ceph-mon/tasks/docker/main.yml` uses the wrong monitor socket,
 * the mon docker containers are named `{{ansible_hostname}}`. I would certainly prefer it to be `ceph-mon-$MON_NAME`, more seriously: so are the ceph-mds- and the ceph-rbd-mirror containers! When testing (or colocating for "hyperconverged storage" ;-)) multiple ceph services on one node that leads to madness!

What's your take on it? I patched most things locally for now, but would like to make it better! Should I file separate issues for the docker container naming inconsistency?

leseb:
Alright, yes I wouldn't say no to a PR that fixes the MON_NAME all over the places.
As for naming the container `ceph-mon-$MON_NAME`, I'm fine with this as well. It's just the name of the container :). I think it's fine keeping this issue for reference. Just make sure you have separate commits for everything and we should be good.

Thanks @ibotty

PR #1244, opened 2017-01-20
docker: make container name configurable

leseb:
You can now use the ceph_mon_name variable to set a different name for
your container. Obviously at your own risk. The default name will still
be the value of ansible_hostname, so the short hostname of the machine.

Closes: #1241

### DIFF
--- a/roles/ceph-mon/templates/ceph-mon.service.j2
+++ b/roles/ceph-mon/templates/ceph-mon.service.j2
@@ -23,6 +23,7 @@ ExecStart=/usr/bin/docker run --rm --name %i --net=host \
    --net=host \
    {% endif -%}
    -e CEPH_DAEMON=MON \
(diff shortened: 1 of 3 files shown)


font:
@leseb I should be able to get to this tomorrow.

roles/ceph-mon/templates/ceph-mon.service.j2
```diff
@@ -23,6 +23,7 @@ ExecStart=/usr/bin/docker run --rm --name %i --net=host \
    --net=host \
    {% endif -%}
    -e CEPH_DAEMON=MON \
+   -e MON_NAME={{ ceph_mon_name }} \
```
leseb: You're right, I forgot about that, no I don't want to expose it further to the user. Although I think we should document the options available in ceph-docker. I'm closing this.

PR #1246, opened 2017-01-23
add unique filter to openstack pool names

tynorth-cisco:
could have scenario where different openstack components would
use the same pool, but the logic would create the same pool
more than once

add unique filter to account for this


leseb:
jenkins test playbook syntax

PR #1247, opened 2017-01-24
Adds ip_version configuration option

andrewschoen:
This allows the user to set ip_version to either ipv4 or ipv6. This
resolves a bug where monitor_address is set to an ipv6 address, but the
template fails to render because it's hardcoded to look for an 'ipv4'
key in the ansible facts.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1416010

Resolves: bz#1416010

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -260,6 +260,8 @@ dummy:
 # will go to monitor_interface if both are defined.
 #monitor_interface: interface
 #monitor_address: 0.0.0.0
+# set to either ipv4 or ipv6, whichever your network is using
+#ip_version: ipv4
 #mon_use_fqdn: false # if set to true, the MON name used will be the fqdn in the ceph.conf
 #monitor_address_block: false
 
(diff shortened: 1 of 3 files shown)


leseb:
I was actually working on something similar after looking into the BZ :). I think we also need to inform (maybe in the commit message) that one needs to set `ms bind ipv6` to its configuration in order to listen on ipv6.

andrewschoen:
@leseb ok, is ``ms bind ipv6`` something that would need to be set in ``ceph_conf_overrides.global`` ? I think we probably would want to add that as a note in the ``defaults/main.yml`` and ``group_vars/all.yml.sample``.

I guess we could also just make that change in the ceph.conf if ip_version == ipv6.

leseb:
I'm more in favor of harcoding it if `ip_version == ipv6` and yes, in the global section.

andrewschoen:
@leseb me too, I'll make that change.

andrewschoen:

andrewschoen:
The ``dmcrypt_journal_collocation`` scenario has actually passed twice now, but it is failing to update github status.

https://jenkins.ceph.com/job/ceph-ansible-prs-ansible2.2-dmcrypt_journal_collocation/261/console

PR #1256, opened 2017-01-28
Using ini_file + ansible_hostname for rgw_zone setting

clwells:
Using ini_file with ansible_hostname to ensure each INI block gets the rgw_zone setting in a multi-RGW setup. Also, ansible_hostname better matches what ceph-common does for the actual hostname (ansible_host != ansible_hostname under all conditions).

Using lineinfile caused only the last stanza for RGW to get the rgw_zone setting.


leseb:
Thanks @clwells

PR #1257, opened 2017-01-28
Use ansible_hostname instead of ansible_host in handler

clwells:
Use hostname instead of host (host can be an IP) and hostname matches the default nomenclature in ceph.conf


leseb:
Thanks @clwells

Issue #1262, opened 2017-01-31
Defining 'cluster_network' in ceph_conf_overrides instead of under OSD options in 'group_vars/all.yml' results in two conflicting cluster_network addresses


squizzi:
If `ceph_conf_overrides` is used to set `cluster_network` instead of setting it under OSD options in the `group_vars/all.yml` file the default var for `cluster_network == {{ public_network }}` is still set resulting in two conflicting `cluster_network` entries in the resulting `ceph.conf` file: 

~~~
[global]
mon initial members = mon1,mon2,mon3
cluster network = 192.168.100.0/24 <--
mon host = 192.168.100.21,192.168.100.22,192.168.100.23
public network = 192.168.100.0/24
cluster_network = 10.10.100.0/24 <--
~~~

The `cluster_network` var works as expected, it's more an issue with conf overrides, but the implementation isn't smart in anyway as we just dump the overrides values into the conf.  Perhaps we need syntax checks?  Or perhaps it's just something we just need to educate users on and tell them to not do that.

I can open a PR with my first whack at this, but I don't know how you want to deal with this scenario.  

We could implement a check that just fails if users define `cluster_network` in `ceph_conf_overrides` and force them into using the `cluster_network` variable with something like: 
~~~
- name: verify that cluster_network was not defined in ceph_conf_overrides
  fail:
    msg: "define cluster_network outside of ceph_conf_overrides"
  when:
    - ceph_conf_overrides.global.cluster_network is defined 
~~~

Or we could go with a different approach and modify the `ceph.conf.j2` template so that we're only accepting either `cluster_network` or `ceph_conf_overrides.global.cluster_network`:

~~~
{% if ceph_conf_overrides.global.cluster_network is defined %}
  { cluster network = {{ ceph_conf_overrides.global.cluster_network }} }
{% elif cluster_network is defined %}
  { cluster network = {{ cluster_network }} }
~~~

leseb:
Can you try:

```
ceph_conf_overrides:
  global:
    cluster network: 10.10.100.0/24
```

That should work, your issue is because of the _ versus the space I believe, so not a bug from the override module.

squizzi:
@leseb you're right, that fixes the behavior: 

~~~
[global]
mon initial members = mon1,mon2,mon3
cluster network = 10.10.100.0/24
mon host = 192.168.100.21,192.168.100.22,192.168.100.23
public network = 192.168.100.0/24
max open files = 131072
[...]

PR #1269, opened 2017-02-01
README: Do not recommend using underscores for opts in ceph_conf_overrides

squizzi:
Adding underscores in the ceph_conf_overrides variable can result in incorrect
config options appearing.  A note has been added to clarify that using
underscores here can cause this behavior and recommending not to do it.

See: https://github.com/ceph/ceph-ansible/issues/1262

Signed-off-by: Kyle Squizzato <ksquizz@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,10 @@ ceph_conf_overrides:
      osd mkfs type: ext4
 ```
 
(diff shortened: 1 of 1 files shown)


leseb:
Thanks @squizzi

PR #1276, opened 2017-02-02
tests: if no osds are created do not error in conftest.py

andrewschoen:
If no OSDs were created the command would fail and because we were using
check_output it would throw an exception and mark all OSD tests as ERROR.
This keeps tests from running that might tell us why the OSDS were not created.


leseb:

PR #1280, opened 2017-02-03
Revert to v2.1.6 (631e6d0fbbb2ddc640a1c57dccb4d998099b5d7b)

andrewschoen:
We decided these changes were not essential for the 2.1.z release series.


andrewschoen:
The ``ansible2.2-docker_cluster`` test is failing here because it uses the ``latest`` tag and now the latest is a kraken container that this branch does not support.

PR #1281, opened 2017-02-06
purge: ability to purge client role

leseb:

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -40,6 +40,7 @@
     rgw_group_name:       rgws
     rbdmirror_group_name: rbd-mirrors
     nfs_group_name:       nfss
+    client_group_name:       clients
 
   hosts:
     - "{{ mon_group_name }}"
@@ -48,6 +49,7 @@
     - "{{ rgw_group_name }}"
     - "{{ rbdmirror_group_name }}"
     - "{{ nfs_group_name }}"
+    - "{{ client_group_name }}"
 
   become: true
 
@@ -435,6 +437,7 @@
     rgw_group_name:       rgws
     rbdmirror_group_name: rbd-mirrors
     nfs_group_name:       nfss
+    client_group_name:       clients
 
     # When set to true both groups of packages are purged.
     # This can cause problem with qemu-kvm
@@ -466,6 +469,7 @@
     - "{{ rgw_group_name }}"
     - "{{ rbdmirror_group_name }}"
     - "{{ nfs_group_name }}"
+    - "{{ client_group_name }}"
 
   gather_facts: false # Already gathered previously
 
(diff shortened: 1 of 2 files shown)


leseb:

infrastructure-playbooks/purge-cluster.yml
```diff
@@ -48,10 +49,23 @@
...
   tasks:
+  - name: test if ceph-detect-init command exist
+    command: command -v ceph-detect-init
+    changed_when: false
+    failed_when: false
+    register: ceph_detect_init_command
+
+  - name: exit playbook, if ceph-detect-init command does not exist
+    debug:
+      msg: "The ceph-detect-init command is not available, please install it."
```
ktdreyer: "please install it" -> "please install ceph-base" ?

leseb: yes!

ktdreyer:
Clients do not typically have ceph-base installed. Are we requiring `ceph-detect-init` to be present on all nodes (including clients) when we purge?

infrastructure-playbooks/purge-cluster.yml
```diff
@@ -48,10 +49,23 @@
...
   tasks:
+  - name: test if ceph-detect-init command exist
+    command: command -v ceph-detect-init
+    changed_when: false
+    failed_when: false
+    register: ceph_detect_init_command
+
+  - name: exit playbook, if ceph-detect-init command does not exist
+    debug:
+      msg: "The ceph-detect-init command is not available, please install ceph-base."
```
andrewschoen: Couldn't it also be the case that if ``ceph-detect-init`` doesn't exist then the node has already been successfully purged?

leseb: Well only if the playbook is ran multiple times then yes... It will also fail to run the command. So I guess we have the same problem in both cases.

ktdreyer:
if `ceph-detect-init` is unavailable, you can use the following logic:

- If Ubuntu 14.04, `init_system` should be "upstart"
- Otherwise, `init_system` should be "systemd"

leseb:
@ktdreyer yes I guess I could do that for Ubuntu and the same for Debian... :)

andrewschoen:

andrewschoen:
jenkins test ansible2.2-purge_cluster

ktdreyer:
alternate solution in https://github.com/ceph/ceph-ansible/pull/1284

leseb:
let me drop 598487aa85f263b59864355cdea7bf0914f47237

leseb:
@ktdreyer drop the commit, we can do a final review on this.

ktdreyer:
Mind re-titling this PR to reflect the most recent changes? The PR title goes into the Git commit log

leseb:
@ktdreyer done

PR #1243, opened 2017-01-20
Remove support of releases prior to Jewel.

guits:
According to #1216, we need to simply the code by removing the
support of anything before Jewel.

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #1283, opened 2017-02-07
fix all ceph components are installed

zhsj:
This is a logic mistake introduced by #1243

previous the logic is:

```
  when:
    - mds_group_name in group_names
    - ansible_pkg_mgr == "yum"
    - ceph_release_num.{{ ceph_release }} > ceph_release_num.infernalis
      or ceph_origin == "distro"
      or ceph_custom
```

The third condition should always be true when we dropped the
infernalis support.

### DIFF
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -74,8 +74,6 @@
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
   when:
     - mon_group_name in group_names
-      or ceph_origin == "distro"
-      or ceph_custom
 
 - name: install distro or red hat storage ceph osd
   package:
@@ -83,8 +81,6 @@
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
   when:
     - osd_group_name in group_names
-      or ceph_origin == "distro"
-      or ceph_custom
 
 - name: install distro or red hat storage ceph mds
   package:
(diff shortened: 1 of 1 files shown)


zhsj:
ping?

leseb:
If the third condition needs to be true then I think what we need is:

``` 
  when:
     - (ceph_origin == "distro" or ceph_custom)
```

zhsj:
I don't know why previous logic is
```
- ceph_release_num.{{ ceph_release }} > ceph_release_num.infernalis
      or ceph_origin == "distro"
      or ceph_custom
```

But when ceph_release is never smaller than infernalis, the whole statement is always true.
So modify to
```
- (ceph_origin == "distro" or ceph_custom)
```
is not the same logic because this statement will be false when ceph_origin set to "upstream"

zhsj:
hmm, after reading the `install_on_redhat.yml`, I guess someone is try to avoid installing ceph package when the `ceph_origin` is set to `local`. But the logic is wrong long before. Does the `local` installation is working before?

zhsj:
I think the fix for "local" installation should go into another PR, let this PR just fixes the problem that all ceph components are installed in every role.

leseb:
Why don't you add a `ceph_origin != 'local'`, so we fix the `local` installation as well?

zhsj:
@leseb actually I don't known what exactly the "local" means since I can't figure out which commit is correct for the "local"

leseb:
@zhsj I believe this commit introduced it: 08766a243a705626defd3988c6f58127efb4ae4a

leseb:

PR #1284, opened 2017-02-07
purge-cluster: do not use ceph-detect-init

andrewschoen:
This copies the method from rolling_update.yml to detect the init system
of the host. We can not always ensure that ceph-detect-init will be
present on the system.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1418980

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -31,7 +31,7 @@
          invoking the playbook"
     when: ireallymeanit != 'yes'
 
-- name: gather facts and check init system
+- name: gather facts on all hosts
 
   vars:
     mon_group_name:       mons
@@ -54,9 +54,7 @@
   become: true
 
   tasks:
-  - name: detect init system
-    command: ceph-detect-init
-    register: init_system
+    - debug: msg="gather facts on all Ceph hosts for following reference"
 
 - name: purge ceph mds cluster
 
@@ -83,16 +81,16 @@
       name: ceph-mds@{{ ansible_hostname }}
       state: stopped
       enabled: no
-    when: init_system.stdout == 'systemd'
+    when: ansible_service_mgr == 'systemd'
 
   - name: stop ceph mdss
     shell: "service ceph status mds ; if [ $? == 0 ] ; then service ceph stop mds ; else echo ; fi"
-    when: init_system.stdout == 'sysvinit'
+    when: ansible_service_mgr == 'sysvinit'
 
   - name: stop ceph mdss on ubuntu
     command: initctl stop ceph-mds cluster={{ cluster }} id={{ ansible_hostname }}
     failed_when: false
-    when: init_system.stdout == 'upstart'
+    when: ansible_service_mgr == 'upstart'
 
 
 - name: purge ceph rgw cluster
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/purge-cluster.yml
```diff
@@ -52,9 +52,22 @@
...
+
+    - name: check if sysvinit
+      stat:
+        path: /etc/rc?.d/S??ceph
+        follow: yes
+      register: is_sysvinit
+
+    - name: check if upstart
+      stat:
+        path: /var/lib/ceph/mon/ceph-{{ ansible_hostname }}/upstart
```
leseb: `{{ cluster }}-{{ ansible_hostname }}`

andrewschoen: fixed

ktdreyer:
Looks like ansible2.2-purge_dmcrypt fails with 'cluster' is undefined :(

andrewschoen:
@ktdreyer 8ab59d157cf970b4f10afa857c85f3d57ee24463 should address that failure

ktdreyer:
Sorry, I don't understand why we'd update the tests for this situation? `roles/ceph-common/defaults/main.yml` already sets `cluster: ceph`, right?

What I'm getting at: does this now introduce a behavior change, in that users will always need to set `cluster: ` in their own `group_vars` in order to make purging work?

andrewschoen:
@ktdreyer these playbooks in ``infrastructure-playbooks`` don't use the roles and should not depend on variables set in them. IMHO, it's a bug that we do include the ``defaults/main.yml`` file from the roles in this playbook (e.g. https://github.com/ceph/ceph-ansible/blob/master/infrastructure-playbooks/purge-cluster.yml#L72). This has the behavior of overriding the value of ``cluster`` if it was set in ``group_vars/all.yml``, which is the correct thing to do if you wish to use a custom cluster name.

Actually, I'm pretty sure this playbook would do the wrong thing if the node was using upstart and using a custom cluster name.

andrewschoen:
@ktdreyer and yes, I guess it would introduce a behavior change if a user was depending on the default value of ``cluster: ceph`` in ``ceph-common/defaults/main.yml`` to work while purging a cluster. They would need to set ``cluster`` in ``group_vars/all.yml`` as well before attempting to purge the cluster, even if none of their nodes are using upstart.

ktdreyer:
I see, ok. Can we default `{{cluster}}` to `ceph`, then, if it's not defined in `group_vars/all.yml` when purging?

andrewschoen:
@ktdreyer yup, that's a good approach. I've pushed a few more commits and removed the one that sets the default for ``cluster`` in ``purge_dmcrypt``.

infrastructure-playbooks/purge-cluster.yml
```diff
@@ -52,9 +52,27 @@
...
+    - debug: msg="gather facts on all Ceph hosts for following reference"
+
+    - name: set cluster value if not set
+      set_fact:
+        cluster: ceph
+      when: cluster is not defined
+
+    - name: check if sysvinit
+      stat:
+        path: /etc/rc?.d/S??ceph
```
leseb: This actually does not work. We need to use the `stat` command instead I'm afraid...

ktdreyer: I think we should drop sysvinit support in this case.

leseb: I have a PR coming in.

andrewschoen: How does it not work? Are you suggesting something like this?

```
- name: check if sysvinit
   command: stat /etc/rc?.d/S?ceph
   ...
```

leseb: yup, see: https://github.com/ceph/ceph-ansible/pull/1289

leseb:
Would be nice to rebase on top of https://github.com/ceph/ceph-ansible/pull/1289

andrewschoen:
@leseb sounds good, lets just wait for https://github.com/ceph/ceph-ansible/pull/1289 to merge and then I'll backport that logic here.

PR #1285, opened 2017-02-07
ceph-common: set ms bind ipv6 = true in ceph.conf when using ipv6

andrewschoen:
This fixes an issue with parsing the ceph.conf file when ip_version is set to
ipv6.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1419814


leseb:
Oh man, how did we possibly missed this one... :/

Issue #1242, opened 2017-01-20
ceph-disk activate for partition failing... but not


ashak:
Hi,

I've recently been trying out ceph-ansible. But I don't end up with any osd's on my hosts.

My untrained eye believes the problem is happening within:
TASK [ceph-osd : activate osd(s) when device is a partition]
Each of my hosts report the same, which looks to me like ceph-disk activate /dev/sda3 is actually failing, but the task seems to pass?

[...]

PR #1291, opened 2017-02-08
ceph-mon: Fix bug #1242

guits:
We shouldn't test directly the value of
`ceph_conf_overrides.global.osd_pool_default_pg_num` because this can
cause the playbook to fail if the key `global` is not present in
`ceph_conf_overrides`. Therefore we have to use the facts that have been defined earlier.

Fix: #1242


leseb:
@andrewschoen this makes me think, thanks to @guits that we should run ansible twice just to make sure we don't hit a problem after a first play. Can we have a special case for  `xenial_cluster` to run ansible twice? I looked at the tox file and I'm not sure how to make an exception... Thanks!

PR #1298, opened 2017-02-08
Remove Ceph OSD dependency on --pid=host

font:
Fix for issue #1297.

Do not merge this yet. I wanted to create the PR to use for testing purposes.

### DIFF
--- a/roles/ceph-osd/tasks/docker/start_docker_osd.yml
+++ b/roles/ceph-osd/tasks/docker/start_docker_osd.yml
@@ -22,7 +22,6 @@
 - name: prepare ceph osd disk
   shell: |
     docker run --net=host \
-    --pid=host \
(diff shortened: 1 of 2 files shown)


roles/ceph-osd/tasks/docker/start_docker_osd.yml
```diff
@@ -22,7 +22,6 @@
 - name: prepare ceph osd disk
   shell: |
     docker run --net=host \
-    --pid=host \
```
leseb: I think it's fine and safe removing these 2 from prepare since this container doesn't run the OSD and is ephemeral.

leseb:
We haven't done enough research and test to fix, this. It seems that kraken still has the issue. Closing this for now.

PR #1302, opened 2017-02-09
docker: use a better method to pull images

leseb:
We now use the docker_image module to pull images. We also changed the
way we declare image. Prior to this patch we must have a "user/image:tag"
format, which is incompatible with non docker-hub registry where you
usually don't have a "user". On the docker hub a "user" is also
identified as a namespace, so for Ceph the user was "ceph". Now if you
want to specify a container image you need to set one of the variables:
'ceph_$DAEMON_docker_name' where DAEMON must be replaced by mon, osd,
mds, rgw, nfs, rbd_mirror or restapi. Here is an example of how you can
set this variable:

1. For docker hub images: ceph_mon_docker_name: "ceph/daemon" will give
you the 'daemon' image of the 'ceph' user.

2. For non docker hub images: ceph_mon_docker_name: "daemon" will simply
give you the "daemon" image.

Infrastructure playbooks have been modified as well.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1420207


andrewschoen:
Why not just use one variable ``ceph_docker_name`` and then have users set it in the appropriate group_vars file to define it for the individual daemons, if it differs. You could even define it per host in the inventory.

I've had some frustration in setting up the CI because I wanted to use a custom tag for the image I was building, but to do that I had to set the same value in like 4 different variables in ceph-ansible to deploy it.

IMHO, all these ``ceph_$daemon_docker_*`` variables should be made generic and moved into ``ceph-docker-common``.

leseb:
@andrewschoen we still need the image and tag separation because some module require it, see the  `docker_image` module. 

BTW on Atomic host, docker-py version is python-docker-py-1.7.2-1.el7.noarch, this is why we are hitting https://github.com/ansible/ansible-modules-core/issues/1227 :(

@andrewschoen should we go back to the "docker pull command" in the meantime? That's **so** annoying.

andrewschoen:
@leseb I think image and tag separation is fine, lets just not create one of those variables for each daemon type.

Yeah, I suppose we should go back to ``docker pull`` in the meantime. Does ``docker_image`` have support for using a custom registry? I don't think I saw that in this PR and our ceph-docker tests depend on that functionality.

leseb:
@andrewschoen alright, I'll do that for the variables. `docker_image` supports custom registries. So we can do `name: localhost:5000/image:tag`.

leseb:
@andrewschoen any last comments for this one?

group_vars/nfss.yml.sample
```diff
@@ -85,9 +85,8 @@ dummy:
 #nfs_containerized_deployment_with_kv: false
 #kv_type: etcd
 #kv_endpoint: 127.0.0.1
-#ceph_nfs_docker_username: ceph
-#ceph_nfs_docker_imagename: ganesha
-#ceph_nfs_docker_image_tag: latest
+#ceph_docker_image: "ceph/ganesha"
```
ktdreyer: Is `ceph/ganesha` a thing? I don't see it @ https://hub.docker.com/r/ceph/ . It looks like the ganesha packages are installed in the "ceph/daemon" image. (Separate PR to change this?)

infrastructure-playbooks/purge-docker-cluster.yml
```diff
@@ -368,14 +368,14 @@
...
-      image: "{{ ceph_mon_docker_username }}/{{ ceph_mon_docker_imagename }}:{{ ceph_mon_docker_image_tag }}"
+      image: "{{ ceph_docker_image }}:{{ ceph_docker_image_tag }}"
       name: "{{ ansible_hostname }}"
       state: absent
     ignore_errors: true
 
   - name: remove restapi container
     docker:
-      image: "{{ ceph_restapi_docker_username }}/{{ ceph_restapi_docker_imagename }}:{{ ceph_restapi_docker_image_tag }}"
+      image: "{{ ceph_restapi_docker_image }}:{{ ceph_restapi_docker_image_tag }}"
```
andrewschoen: This should also be:

```
{{ ceph_docker_image }}:{{ ceph_docker_image_tag }}
```

roles/ceph-mds/tasks/docker/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: check if a cluster is already running
-  shell: "docker ps | grep -sq '{{ceph_mds_docker_username}}/{{ceph_mds_docker_imagename}}:{{ceph_mds_docker_image_tag}}'"
+  shell: "docker ps | grep -sq '{{ ceph_docker_image }}:{{ ceph_docker_image_tag }}'"
```
ktdreyer: We could use `docker ps -q -a --filter='ancestor={{ ceph_docker_image }}:{{ ceph_docker_image_tag }}'` to make this search more precise instead of text-matching with grep. (Maybe a separate PR?)

leseb: I remember having an issue using this syntax, maybe it was with Atomic I'm not sure.

infrastructure-playbooks/purge-docker-cluster.yml
leseb: Thanks!

roles/ceph-mds/tasks/docker/main.yml
ktdreyer: Should we have Atomic tests in our CI so we verify that?

leseb: We are actually testing on Atomic for Docker :)

leseb: Or maybe this was because of the syntax/vars we were using, let me try your suggestion, the CI will tell us :)

group_vars/nfss.yml.sample
leseb: I don't know, I just re-used what was present. Maybe @dang knows?

leseb:

PR #1305, opened 2017-02-10
bugfix: Add missing conditions for kraken release

guits:
76ddcbc introduced an issue by removing some needed conditions on tasks
that need to be run only on release >= kraken.


andrewschoen:
Thanks for this! I had noticed the regression the other day myself but hadn't had time to put a PR together yet.

PR #1307, opened 2017-02-14
rgw-standalone.yml will also now collect keys

andrewschoen:
This playbook is used to add an rgw node to an existing cluster when you
don't want to rerun the site.yml against your entire cluster again.

This is also used by the api/rgw/configure endpoint of ceph-installer.

### DIFF
--- a/infrastructure-playbooks/rgw-standalone.yml
+++ b/infrastructure-playbooks/rgw-standalone.yml
@@ -1,8 +1,12 @@
 ---
 # Run ceph-rgw role standalone
 
-# Need to load the facts from mons because ceph-common need them to generate the ceph.conf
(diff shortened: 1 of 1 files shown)


leseb:
We already have https://github.com/ceph/ceph-ansible/blob/master/infrastructure-playbooks/rgw-standalone.yml

andrewschoen:
@leseb I don't believe ``rgw-standalone.yml`` will work in the ceph-installer use case because we also need to fetch keys and can't depend on what's in the ``fetch_directory`` to be correct.

andrewschoen:
@leseb I've modified this so that ``rgw-standalone.yml`` also collects keys so that I can use it with ceph-installer and removed ``rgw-configure.yml``

Issue #1296, opened 2017-02-08
harmonize docker systemd unit file location


ibotty:
They are stored in /etc/systemd/system or /var/lib/ceph. Which one, do you prefer? I would write a patch.

leseb:
@ibotty where do you see systemd files on `/var/lib/ceph/`? AFAIR we only store the docker systemd unit files in `/etc/systemd/system`.

ibotty:
ceph-ansible$ git grep /var/lib/ceph/ceph-
roles/ceph-mds/tasks/docker/start_docker_mds.yml:    dest: /var/lib/ceph/ceph-mds@.service
roles/ceph-mds/tasks/docker/start_docker_mds.yml:    src: /var/lib/ceph/ceph-mds@.service
roles/ceph-rgw/tasks/docker/start_docker_rgw.yml:    dest: /var/lib/ceph/ceph-rgw@.service
roles/ceph-rgw/tasks/docker/start_docker_rgw.yml:    src: /var/lib/ceph/ceph-rgw@.service

leseb:
@ibotty I never paid attention to that, so please submit a PR to have everything in `/etc/systemd/system` just like mon and osd.

PR #1313, opened 2017-02-16
docker: homogenise systemd file location

leseb:
So unit files were stored in /var/lib/ceph some where in
/etc/systemd/system. Now they are all under /etc/systemd/system.

closes: #1296


leseb:
jenkins test ansible2.2-purge_dmcrypt

leseb:
"ansible2.2-docker_dedicated_journal" failure is expected as the scenario is not present yet.

PR #1323, opened 2017-02-21
Update ceph_keys.yml

WingkaiHo:
jewel version need to build the {{ cluster }}.client.admin.keyring exists key

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -8,8 +8,7 @@
   always_run: true
   when:
     - cephx
-    - ceph_release_num.{{ ceph_release }} > ceph_release_num.jewel
-
+    - ceph_release_num.{{ ceph_release }} >= ceph_release_num.jewel
 # NOTE (leseb): wait for mon discovery and quorum resolution
 # the admin key is not instantaneously created so we have to wait a bit
 - name: "wait for {{ cluster }}.client.admin.keyring exists"


leseb:
Jewel works well without this, what's the issue? Do you have any logs to share?

WingkaiHo:
wait for {{ cluster }}.client.admin.keyring exists

timeout

leseb:
Is the monitor started? The `ceph-create-keys` should be triggered by systemd, so we don't need to. Can I have a look at your systemd unit file for ceph-mon?

WingkaiHo:
[Unit]
Description=Ceph cluster monitor daemon

# According to:
#   http://www.freedesktop.org/wiki/Software/systemd/NetworkTarget
# these can be removed once ceph-mon will dynamically change network
# configuration.
After=network-online.target local-fs.target time-sync.target ceph-create-keys@%i.service
Wants=network-online.target local-fs.target time-sync.target ceph-create-keys@%i.service

PartOf=ceph-mon.target

[Service]
LimitNOFILE=1048576
LimitNPROC=1048576
EnvironmentFile=-/etc/sysconfig/ceph
Environment=CLUSTER=ceph
ExecStart=/usr/bin/ceph-mon -f --cluster ${CLUSTER} --id %i --setuser ceph --setgroup ceph
ExecReload=/bin/kill -HUP $MAINPID
PrivateDevices=yes
ProtectHome=true
ProtectSystem=full
PrivateTmp=true
TasksMax=infinity
Restart=on-failure
StartLimitInterval=30min
StartLimitBurst=3

[Install]
WantedBy=ceph-mon.target

WingkaiHo:
ok thank

PR #1315, opened 2017-02-17
adds a testing scenario for rolling updates

andrewschoen:
This currently tests a rolling update on a cluster using dmcrypt dedicated journals from jewel to kraken. I had to change to using ``yum_repository`` to manage the ``ceph_stable`` repo because the ceph-release package from jewel -> kraken has the same version number and yum does not know to update anything. Also, this ways gives us much more flexibility in using other mirrors.

PR #1324, opened 2017-02-21
group_vars: rm Infernalis comment

ktdreyer:
As of Infernalis, the Ceph daemons run as an unprivileged "ceph" UID, and this is by design.

Commit f19b765f7995dab4aca46e1885320dc440485ead altered the default civetweb port from 80 to 8080 with a comment in the commit log about "until this gets solved"

Remove the comment about permissions on Infernalis, because this is always going to be the case on the Ceph versions we support, and it is just confusing.

If users want to expose civetweb to s3 clients using privileged TCP ports, they can redirect traffic with iptables, or use a reverse proxy application like HAproxy.

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -293,7 +293,7 @@ dummy:
 #
 #radosgw_dns_name: your.subdomain.tld # subdomains used by radosgw. See http://ceph.com/docs/master/radosgw/config/#enabling-subdomain-s3-calls
 #radosgw_resolve_cname: false # enable for radosgw to resolve DNS CNAME based bucket names
-#radosgw_civetweb_port: 8080 # on Infernalis we get: "set_ports_option: cannot bind to 80: 13 (Permission denied)"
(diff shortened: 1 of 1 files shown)


leseb:
Oops sorry, can you rebase on top of https://github.com/ceph/ceph-ansible/pull/1325? Thanks!

andrewschoen:
This change should have been done in ``ceph-common/defaults/main.yml`` because it'll be overwritten next time somebody runs ``generate_group_vars_sample.sh``.  I've fixed this in https://github.com/ceph/ceph-ansible/pull/1315/commits/6cf842eb396d0a26882db9658cf7010ff84a7733

leseb:
Rahhh stupid me, usually I'm the guy saying "please update `ceph-common/defaults/main.yml` and run `generate_group_vars_sample.sh`"...

ktdreyer:
I am getting the sense that PRs are merging too quickly without enough review

PR #1325, opened 2017-02-21
Ipv6 rgw doc

leseb:


ktdreyer:
mind waiting a bit before merging? I am not sure what the `docker-common` change has to do with this?

leseb:
@ktdreyer it's a separate commit so it's not a problem right?

ktdreyer:
I like "one PR per logical change, one logical change per PR"

leseb:
Alright, will keep that in mind :)

PR #1340, opened 2017-03-02
 - add dependencies for redhat 7 upstream install

laurencegill:
add the repos needed for package dependencies as instructed in
 the ceph docs:
  http://docs.ceph.com/docs/master/start/quick-start-preflight/#rhel-centos

 the packages have been moved to the extras repo:
  http://pkgs.fedoraproject.org/cgit/rpms/python-flask.git/commit/?h=epel7

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -38,7 +38,8 @@
 
 - include: ./pre_requisites/prerequisite_redhat_install.yml
   when:
-    - ansible_os_family == 'RedHat'
+    - ansible_distribution == "RedHat"
+    - ansible_distribution_major_version >= 7
     - not ceph_rhcs_iso_install
   tags:
     - package-install
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -40,7 +40,7 @@
   when:
     - ansible_distribution == "RedHat"
     - ansible_distribution_major_version >= 7
-    - not ceph_rhcs_iso_install
+    - not ceph_rhcs
   tags:
     - package-install
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -38,8 +38,8 @@
 
 - include: ./pre_requisites/prerequisite_redhat_install.yml
   when:
-    ansible_os_family == 'RedHat' and
-    not ceph_rhcs_iso_install
+    - ansible_os_family == 'RedHat'
+    - not ceph_rhcs_iso_install
   tags:
     - package-install
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
@@ -47,8 +47,8 @@
 
 - include: ./installs/install_on_redhat.yml
   when:
-    ansible_os_family == 'RedHat' and
-    not ceph_rhcs_iso_install
+    - ansible_os_family == 'RedHat'
+    - not ceph_rhcs_iso_install
   tags:
     - package-install
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -36,6 +36,15 @@
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
   static: False
 
+- include: ./pre_requisites/prerequisite_redhat_install.yml
+  when:
+    ansible_os_family == 'RedHat' and
+    not ceph_rhcs_iso_install
+  tags:
+    - package-install
+  # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
+  static: False
+
 - include: ./installs/install_on_redhat.yml
   when:
     ansible_os_family == 'RedHat' and
--- a/roles/ceph-common/tasks/pre_requisites/prerequisite_redhat_install.yml
+++ b/roles/ceph-common/tasks/pre_requisites/prerequisite_redhat_install.yml
@@ -1,0 +1,36 @@
+---
+- name: determine if node is registered with subscription-manager.
+  command: subscription-manager identity
+  register: subscription
+  changed_when: false
+  always_run: true
+
+- name: check if the rhel 7 extras repo is already present
+  shell: yum --noplugins --cacheonly repolist | grep -sq rhel-7-server-extras-rpms
+  changed_when: false
+  failed_when: false
+  register: rh_extras_repo
+  always_run: true
+  when: mon_group_name in group_names
+
+- name: enable rhel 7 extras repository
+  command: subscription-manager repos --enable rhel-7-server-extras-rpms
+  changed_when: false
+  when:
+    - mon_group_name in group_names
+    - rh_extras_repo.rc != 0
+
+- name: check if the EPEL repo is already present
+  shell: yum --noplugins --cacheonly repolist | grep -sq ^epel/{{ ansible_architecture }}
+  changed_when: false
+  failed_when: false
+  register: rh_epel_repo
+  always_run: true
+  when: mon_group_name in group_names
+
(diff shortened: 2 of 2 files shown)


roles/ceph-common/tasks/main.yml
```diff
@@ -36,6 +36,15 @@
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
   static: False
 
+- include: ./pre_requisites/prerequisite_redhat_install.yml
+  when:
+    ansible_os_family == 'RedHat' and
```
leseb: Please use the following syntax:

```
when:
  - ansible_os_family == 'RedHat'
  - not ceph_rhcs_iso_install
```

roles/ceph-common/tasks/pre_requisites/prerequisite_redhat_install.yml
```diff
@@ -0,0 +1,36 @@
+---
+- name: determine if node is registered with subscription-manager.
```
leseb: please remove the `.` at the end of the line

roles/ceph-common/tasks/pre_requisites/prerequisite_redhat_install.yml
```diff
@@ -0,0 +1,36 @@
...
+  when: mon_group_name in group_names
+
+- name: enable rhel 7 extras repository
+  command: subscription-manager repos --enable rhel-7-server-extras-rpms
+  changed_when: false
+  when:
+    - mon_group_name in group_names
+    - rh_extras_repo.rc != 0
+
+- name: check if the EPEL repo is already present
```
leseb: please remove capital letters

roles/ceph-common/tasks/pre_requisites/prerequisite_redhat_install.yml
```diff
@@ -0,0 +1,36 @@
...
+  shell: yum --noplugins --cacheonly repolist | grep -sq ^epel/{{ ansible_architecture }}
+  changed_when: false
+  failed_when: false
+  register: rh_epel_repo
+  always_run: true
+  when: mon_group_name in group_names
+
+- name: enable EPEL repository
```
leseb: ditto

roles/ceph-common/tasks/pre_requisites/prerequisite_redhat_install.yml
```diff
@@ -0,0 +1,36 @@
...
+  shell: sudo yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
```
leseb: `sudo` is not needed here

leseb: I also wonder if we couldn't use an Ansible module for that. Like `yum`: http://docs.ansible.com/ansible/yum_module.html

laurencegill:
Agreed, I have made the changes (I wasn't sure whether to stick with the formatting of the other tasks)

Do we still need to use these lines with the yum module?

```
   yum:
     name: https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
     state: installed
-  changed_when: false
   when:
-    - rh_epel_repo.rc != 0
```

do we still need this task at all?

```
- name: check if the epel repo is already present
```

Excellent work by the way!

roles/ceph-common/tasks/pre_requisites/prerequisite_redhat_install.yml
```diff
@@ -0,0 +1,38 @@
...
+
+- name: enable rhel 7 extras repository
+  command: subscription-manager repos --enable rhel-7-server-extras-rpms
+  changed_when: false
+  when:
+    - mon_group_name in group_names
+    - rh_extras_repo.rc != 0
+
+- name: check if the epel repo is already present
+  shell: yum --noplugins --cacheonly repolist | grep -sq ^epel/{{ ansible_architecture }}
```
leseb: Please write this like this:

```
shell: |
  yum --noplugins --cacheonly repolist | grep -sq ^epel/{{ ansible_architecture }}
```

roles/ceph-common/tasks/pre_requisites/prerequisite_redhat_install.yml
```diff
@@ -0,0 +1,38 @@
...
+  changed_when: false
+  failed_when: false
+  register: rh_epel_repo
+  always_run: true
+  when: mon_group_name in group_names
+
+- name: enable epel repository
+  yum:
+    name: https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
```
zhsj: Why use the url directly, epel-release is a package that existed in the base repo.

laurencegill: Hi,

I don't see it in rhel7...

zhsj: sorry, it's not in the base repo, but in extra for centos and option for rhn
http://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F

To install from these place means the package can be verified by the existing gpg key.

laurencegill: I think this is what we are already doing, the package contains the keys.

I could be entirely missing your point though...

zhsj: I may misunderstand the words on the wiki. 

But could you use epel-release package name, rather than the url for CentOS. I can't determine if the epel-release package is available for RHEL.

Install epel-release from CentOS repo, so that we can benefit from the fastmirror plugin when the speed to dl.fedoraproject.org is not so good.

roles/ceph-common/tasks/pre_requisites/prerequisite_redhat_install.yml
```diff
@@ -0,0 +1,41 @@
...
+
+- name: check if the rhel 7 extras repo is already present
+  shell: yum --noplugins --cacheonly repolist | grep -sq rhel-7-server-extras-rpms
+  changed_when: false
+  failed_when: false
+  register: rh_extras_repo
+  always_run: true
+  when:
+    - (mon_group_name in group_names or client_group_name in group_names)
+
```
leseb: Please remove this extra line.

roles/ceph-common/tasks/main.yml
```diff
@@ -36,10 +36,20 @@
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
   static: False
 
+- include: ./pre_requisites/prerequisite_redhat_install.yml
+  when:
+    - ansible_distribution == "RedHat"
+    - ansible_distribution_major_version >= 7
+    - not ceph_rhcs_iso_install
```
leseb: Why not `ceph_rhcs` so we address both `ceph_rhcs_cdn_install` and `ceph_rhcs_iso_install`?

andrewschoen:
@laurencegill have you tried deploying on your rhel node just like we deploy to centos7 upstream? I think all you'd need to do is set ``ceph_stable: True`` and then configure everything else as normal. If so, what failures did you see?

I'm looking at the ``ceph-common`` and it seems like it might work if you add ``epel-release`` to ``redhat_package_dependencies`` in your group_vars. I'm not sure you need to do anything with subscription-manager if you're wanting to install upstream ceph.

laurencegill:
@andrewschoen the problem I found with rhel7 is when you try to install the ceph stable upstream packages you get package dependency errors during the install, due to these packages being moved into the extras repo, python-flash for example:
```
http://pkgs.fedoraproject.org/cgit/rpms/python-flask.git/commit/?h=epel7
```
this is why rhel extras is required for rhel7 install.

epel-release isn't available from the regular repo in rhel, but if you could add the link to the ```redhat_package_dependencies``` like in ```- name: enable epel repository``` this should work, although as the install fails without it should it just happen automatically?

leseb:

Issue #1338, opened 2017-03-02
docker: make dir permission work for both debian and redhat systems


leseb:
Here https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-rgw/tasks/docker/dirs_permissions.yml#L6-L7 we set the uid to 64045 which binds to ubuntu's ceph uid. On redhat, we have 167. So this should be configurable.

All the others daemons have the same problem.

We need a new var for this and default it to the current 64045.

leseb:
@guits can you have a look into this? Thanks!

PR #1345, opened 2017-03-03
ceph-docker: fix bootstrap directories permissions

guits:
Make bootstrap directories permissions work for both RedHat and Debian
os families.

Fix: #1338

### DIFF
--- a/roles/ceph-docker-common/defaults/main.yml
+++ b/roles/ceph-docker-common/defaults/main.yml
@@ -1,4 +1,9 @@
 ---
-
 ceph_docker_registry: docker.io
 ceph_docker_enable_centos_extra_repo: false
+
+# Set uid/gid to default '64045' for bootstrap directories.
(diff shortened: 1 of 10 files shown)


guits:
still WIP, after discussing with @leseb this PR needs to be fixed

roles/ceph-docker-common/defaults/main.yml
```diff
@@ -1,4 +1,9 @@
 ---
-
 ceph_docker_registry: docker.io
 ceph_docker_enable_centos_extra_repo: false
+
+# Set uid/gid to default '64045' for bootstrap directories.
```
leseb: + This value as to be set according to the base OS used by the container image, **not** the host.

guits: done

guits: Done

PR #1350, opened 2017-03-07
Makefile: handle "rc" version tags

ktdreyer:
When `git describe` returns an `rc` tag (eg `2.2.0rc1`), insert the RC number into the RPM's `Release` field.

This change will allow us to generate RPMs from the master branch with an NVR that is higher than v2.1.9, once we tag v2.2.0rc1 on that branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,19 @@
 # Try "make" (for SRPMS) or "make rpm"
 
 NAME = ceph-ansible
+
+# Set the RPM package NVR from "git describe".
+# Examples:
+#
+#  A "git describe" value of "v2.2.0rc1" would create an NVR
+#  "ceph-ansible-2.2.0-0.rc1.1.el7"
+#
+#  A "git describe" value of "v2.2.0rc1-1-gc465f85" would create an NVR
+#  "ceph-ansible-2.2.0-0.rc1.1.gc465f85.el7"
+#
+#  A "git describe" value of "v2.2.0" creates an NVR
+#  "ceph-ansible-2.2.0-1.el7"
+
 VERSION := $(shell git describe --tags --abbrev=0 --match 'v*' | sed 's/^v//')
 COMMIT := $(shell git rev-parse HEAD)
 SHORTCOMMIT := $(shell echo $(COMMIT) | cut -c1-7)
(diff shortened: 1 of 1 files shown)


ktdreyer:
CC @zhsj in case he wants to review, since he commented on the [ceph-ansible mailing list](http://lists.ceph.com/pipermail/ceph-ansible-ceph.com/2017-March/000041.html) about this

zhsj:
Hi @ktdreyer,
No comments for RPM NVR. But AFAIR, in debian packaging, we use `~` for upstream pre release.
https://www.debian.org/doc/debian-policy/footnotes.html#f37

So the example is:
If `git git describe --tags` returns `v2.1.0~rc-154-g95bece9`, then the upstream version is `2.1.0~rc+154`.
It make sure that `2.1.0~rc+154` < `2.1.0~rc+155` and < `2.1.0`

I'm not familar with RPM, you can just ignore what I said :)

But +1 for use rc version.

ktdreyer:
Thanks @zhsj . Looks like the [Fedora FPC is still evaluating guidelines for tilde versioning](https://pagure.io/packaging-committee/issue/398), and when that's complete and stable we could use it here in a future change.

ErwanAliasr1:
Why not splitting our versionning which is very predictable with a dash "-" and let a kind of free field after that ?

I mean sedding anything after the dash and use it after the "0." in the release ?

That would allow having any kind of versionning like rc/beta/pre whatever.

ktdreyer:
@ErwanAliasr1  Let's do that in a separate PR? Trying to balance simplicity vs flexibility here.

alfredodeza:
This really needs to move forward. We can't possibly keep discussing what the nicest thing for versioning is. If we need further discussion, lets do that, but without preventing @ktdreyer from getting packages out

ErwanAliasr1:
If that can't be discussed now then I don't get why I'm ask for my opinion. Asking for a review is being in a position to accept questions / remarks / other ideas. If we don't have time for that, then the peer reviewing is almost useless.

alfredodeza:
@ErwanAliasr1 your opinion(s) are still valuable :) We do have time to discuss and get good ideas going. I am proposing to go at it in the mailing list, for as long as it is needed, to reach some consensus.

What doesn't sound OK is to deviate current progress for packaging for what seems like a new discussion about using `-` or just a plain `rc`.

Nothing in this merged PR should be considered final, we still can revert/change/update if we want to add `-` (or anything else) later.

ErwanAliasr1:
Doing fast is not always the good way to do things but I'll not be the picky guy here.
My proposal was not a big change regarding the current code and didn't expected it to be received like that. Let's go this way and see later.

PR #1358, opened 2017-03-13
ceph-common: install nfs-ganesha FSALs on Debian

ktdreyer:
Prior to this change, ceph-ansible would install the main NFS Ganesha server daemon on Ubuntu, but it would skip the Ceph FSALs.

Running `apt-get install nfs-ganesha` will only install the main NFS Ganesha server. It does *not* pull in the RGW FSAL (`/usr/lib/x86_64-linux-gnu/ganesha/libfsalrgw.so`)

Running `apt-get install nfs-ganesha-fsal` will install the RGW FSAL as well as the main NFS Ganesha server package.

### DIFF
--- a/roles/ceph-common/tasks/installs/install_on_debian.yml
+++ b/roles/ceph-common/tasks/installs/install_on_debian.yml
@@ -53,7 +53,7 @@
 
 - name: install NFS gateway
   apt:
-    pkg: nfs-ganesha
+    pkg: nfs-ganesha-fsal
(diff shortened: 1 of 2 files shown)


leseb:
Signed-off-by?

ktdreyer:
I added `Signed-off-by: Ken Dreyer <kdreyer@redhat.com>`

(We're enforcing that for ceph-ansible now? Might be a good idea to announce it :)

leseb:
We don't explicitly enforce it but it has been a common practice :)

PR #1365, opened 2017-03-15
Make ceph-common aware of osd config fragments

czunker:
This removes the implicit order requirement when using OSD fragments.
When you use OSD fragments and ceph-osd role is not the last one,
the fragments get removed from ceph.conf by ceph-common.

It is not nice to have this code at two locations, but this is
necessary to prevent problems, when ceph-osd is the last role as
ceph-common gets executed before ceph-osd.

This could be prevented when ceph-common would be explicitly called
at the end of the playbook.

Signed-off-by: Christian Zunker <christian.zunker@codecentric.de>

### DIFF
--- a/roles/ceph-common/tasks/generate_ceph_conf.yml
+++ b/roles/ceph-common/tasks/generate_ceph_conf.yml
@@ -1,22 +1,33 @@
 ---
-- name: create ceph conf directory
+- name: create ceph conf directory and assemble directory
   file:
-    path: /etc/ceph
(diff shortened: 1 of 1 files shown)


D3N14L:
ceph-common does not respect `/etc/ceph/ceph.d/*` configuration snippets created by ceph-osd. Any chance we can get this merged any time soon? It would solve quiet some hicups for us. It would also make it possible to use config snippets in other roles...

PR #1369, opened 2017-03-16
activate LVM osds as well

wind0204:
The original phrase \\1 didn't work for me so I simply put just one backslash. so there is a discordance between the first regex_replace filter and the second; I hope the ceph-ansible maintainer will resolve this issue.

Disclaimer: I only tested with the task named 'activate osd(s) when device is a disk' but I added it to all 3 tasks without testing. I just don't have enough energy and time to test them all but I guess ceph-ansible team could have the ability to do it efficiently in a short time span.

Best Regards,
Nicholas Gim.

P.S. recently made a pull request for ceph lvm support ( https://github.com/ceph/ceph/pull/13994 )

Signed-off-by: Gunwoo Gim (a.k.a. Nicho1as) <wind8702@gmail.com>

### DIFF
--- a/roles/ceph-osd/tasks/activate_osds.yml
+++ b/roles/ceph-osd/tasks/activate_osds.yml
@@ -3,7 +3,7 @@
 # partition.
 
 - name: automatically activate osd disk(s) without partitions
-  command: ceph-disk activate "/dev/{{ item.key | regex_replace('^(\/dev\/cciss\/c[0-9]{1}d[0-9]{1})$', '\\1p') }}1"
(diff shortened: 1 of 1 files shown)


alfredodeza:
hey @wind0204 we are going to have a separate tool for deploying OSDs from LVM volumes, ceph-disk is not going to support this.

The documentation is at http://docs.ceph.com/ceph-lvm/master/index.html and we should have a release in the next few months. Let us know if you have any ideas/feedback.

Thank you!

PR #1378, opened 2017-03-16
ceph-mon: always call ceph-create-keys

andrewschoen:
After the jewel release the mon startup does not generate keys, but it's
still harmless to call ceph-create-keys with jewel because this task has
a 'creates' argument that will cause it not to run if the keys already
exist.

Removing this when condition also allows the downstream CI tests to
install kraken or luminous without resetting ceph_stable_release, which does not
pertain to rhcs.


ktdreyer:
LGTM once it passes Jenkins

andrewschoen:
@leseb any concerns with this one?

leseb:
No I don't have any real concerns, I just thought we passed that, but apparently the issue is back on Kraken.

PR #1383, opened 2017-03-19
Update main.yml

Xanky06:
The "/usr/share" folder is mount in read-only which stop the installation process in the vagrant/docker scenario. Using the /root solve the issue in my case


leseb:
No need to change the default behavior, simply edit `group_vars/osds.yml`.

PR #1384, opened 2017-03-20
Add two variables to allow setting the monitor hosts explicitly.

jap:
This allows one to run a playbook for the monitor role with ``--limit`` to a single host
without using fact caching.

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -258,6 +258,10 @@ ip_version: ipv4
 mon_use_fqdn: false # if set to true, the MON name used will be the fqdn in the ceph.conf
 monitor_address_block: false
 
+# Use the below variables to set these explicitly instead of relying on gathered facts.
+# mon_initial_members: [ mon1, mon2, mon3 ]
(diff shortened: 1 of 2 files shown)


roles/ceph-common/defaults/main.yml
```diff
@@ -258,6 +258,10 @@ ip_version: ipv4
 mon_use_fqdn: false # if set to true, the MON name used will be the fqdn in the ceph.conf
 monitor_address_block: false
 
+# Use the below variables to set these explicitly instead of relying on gathered facts.
+# mon_initial_members: [ mon1, mon2, mon3 ]
```
leseb: Not this way, you can use `ceph_conf_overrides` to declare them.
Something like:

```
ceph_conf_overrides:
  global:
    mon_initial_members: mon1, mon2, mon3
    mon_host: 10.0.0.1, 10.0.0.2, 10.0.0.3
```

See: https://github.com/ceph/ceph-ansible/blob/master/group_vars/all.yml.sample#L359-L374
So adding this is not necessary.

jap: Thanks, that works even better! Will drop the PR.

PR #1389, opened 2017-03-23
tests: enable testing of rhcs

andrewschoen:
This adds a new prefix, ``rhcs-``, that when added will instruct our testing scenarios to install Red Hat Ceph Storage instead of upstream ceph.

### DIFF
--- a/tests/functional/rhcs_setup.yml
+++ b/tests/functional/rhcs_setup.yml
@@ -13,6 +13,13 @@
         dest: "{{ change_dir }}/vagrant_variables.yml"
       when: change_dir is defined
 
+    - name: change ceph/ubuntu-xenial vagrant box name to rhel7
+      replace:
+        regexp: "ceph/ubuntu-xenial"
+        replace: "rhel7"
+        dest: "{{ change_dir }}/vagrant_variables.yml"
+      when: change_dir is defined
+
 - hosts: all
   gather_facts: true
   become: yes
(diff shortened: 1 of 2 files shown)


tests/functional/rhcs_setup.yml
```diff
@@ -0,0 +1,90 @@
...
+    - name: install nightly rhel7 repo
+      get_url:
+        url: "{{ rhel7_repo_url }}"
+        dest: /etc/yum.repos.d
+        owner: root
+        group: root
+      when: not is_atomic
+
+    - name: set MTU on eth0
+      command: "ifconfig eth0 mtu 1400 up"
```
leseb: why?

andrewschoen: The openstack instance we are testing this with starts the vms with a MTU of 1500 which makes yum fail, this was just an easy hack to fix that and get tests running.

I'm sure there is a better way, if you know of one I'd love to remove this from the playbook.

leseb: I don't know any, the only comment I could do is use `ip` instead of `ifconfig` but since the result is the same...

andrewschoen: Yeah, it is and I've tested this out already so I'd prefer not to change. I don't like this either though as it seems to be very environment specific, but we have little control over the environment we'll be running these in.

leseb: Yes, I'm fine with this :). Feel free to merge.

andrewschoen:
jenkins test ansible2.2-update_dmcrypt

Issue #1354, opened 2017-03-09
Setting ntp_service_enabled = False still installs ntp


gaudenz:
Even if the variable to not enabled the NTP service is set, the playbooks still install ntp through the XXX_package_dependencies variables. This makes it impossible to use a different NTP implementation (like openntpd or chrony) without also overriding the XXX_package_dependencies variable.

IMO installing the ntp package should be in it's own task and only be done when ntp_service_enabled is set to to True. I expect that in most environments where you manage all your hosts with Ansible you already have some kind of base playbook which takes care of installing and configuring an NTP daemon.

leseb:
@guits can you look into this please?

PR #1399, opened 2017-03-28
Common: Do not install ntp when ntp_service_enabled is false

guits:
ntp is still installed even if ntp_service_enabled is set to false.
That could be a problem if the time synchronization is managed by
something else than ceph-ansible or if you want to use different NTP
implementation as suggested in #1354.

Fixes: #1354


guits:

Issue #1406, opened 2017-04-02
Suggest: add a test to check if OS support osd_mkfs_type filesystem


harobed:
Suggest:

add in [roles/ceph-common/tasks/checks/check_system.yml](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/checks/check_system.yml) a test to check if OS support `osd_mkfs_type` filesystem.

Best regards,
Stéphane

PR #1407, opened 2017-04-02
Check if osd hosts support osd_mkfs_type filesystem

harobed:
Check if osd hosts support osd_mkfs_type filesystem #1406

### DIFF
--- a/roles/ceph-osd/tasks/check_system.yml
+++ b/roles/ceph-osd/tasks/check_system.yml
@@ -1,0 +1,3 @@
+---
+- name: fail on unsupported {{osd_mkfs_type}} filesystem (configured by osd_mkfs_type variable)
+  command: grep -q "[[:space:]]{{osd_mkfs_type}}$" /proc/filesystems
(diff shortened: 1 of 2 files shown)


harobed:
`xfs` isn't supported by Ubuntu "xenial" ?

What is purge cluster ?

roles/ceph-osd/tasks/check_system.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: fail on unsupported {{osd_mkfs_type}} filesystem (configured by osd_mkfs_type variable)
+  command: grep -q "[[:space:]]{{osd_mkfs_type}}$" /proc/filesystems
```
leseb: Please add to the task:

```
changed_when: false
failed_when: false
```

Can you also add a `fail` task to fail properly with a useful message?

harobed: Ok, I'll do that.

leseb: Can you do this ASAP? We want to tag stable-2.2 by the end of the week.
Thanks!

zhsj:
Hi, the xfs module is not loaded by default on Debian, but it has. So the check for /proc/filesystem is not accurate.

/proc/filesystem only shows the fs module used currently. But some fs kernel modules not loaded are also available to use.

leseb:

PR #1419, opened 2017-04-06
Allow NFS role to copy admin key

dang:
If cephx is enabled, then the admin key is necessary to add RGW users.
Add the ability to copy the admin key for the object NFS case.


roles/ceph-nfs/tasks/pre_requisite.yml
```diff
@@ -10,6 +10,20 @@
     - /var/lib/nfs/ganesha
     - /var/run/ganesha
 
+- name: copy rados gateway admin key
```
leseb: isn't this just the admin key instead of the rgw admin key?

leseb:
@dang bump

PR #1420, opened 2017-04-07
Added Sandisk FusionIO support and added comments for running bluestore

piwi3910:
added support for Sandisk FusionIO devices

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -271,7 +271,7 @@ dummy:
 #osd_mkfs_type: xfs
 #osd_mkfs_options_xfs: -f -i size=2048
 #osd_mount_options_xfs: noatime,largeio,inode64,swalloc
-#osd_objectstore: filestore # Set to bluestore if you are planning to use the bluestore scenario in osd.yml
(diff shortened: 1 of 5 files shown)


leseb:

group_vars/all.yml.sample
```diff
@@ -271,7 +271,7 @@ dummy:
 #osd_mkfs_type: xfs
```
leseb: Instead of editing this file you should edit `roles/ceph-osd/defaults/main.yml` and then run `generate_group_vars_sample.sh`. Same for `roles/ceph-common/defaults/main.yml`.

Please also note that the option `enable experimental unrecoverable data corrupting features: 'bluestore rocksdb'` won't stay forever so we should flag this by saying something like "only valid for Kraken, might go away for Luminous".

piwi3910:
can we commit the fusionIO support and just remove my other commit?
https://github.com/ceph/ceph-ansible/pull/1420/commits/380efeeac703667968defa192b42ff810d22d72a

leseb:
https://github.com/ceph/ceph-ansible/pull/1527

Issue #1013, opened 2016-10-06
persistent disabling of transparent hugepages


bengland2:
in [ceph-ansible kernel tuning yml](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/misc/system_tuning.yml) it disables transparent hugepages by echoing to sysfs but this is not a persistent setting across reboots. Note that disable_transparent_hugepage var is [True by default](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/defaults/main.yml#L327).  Yaml to disable THP on the live system is [here](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/misc/system_tuning.yml#L7).

See [this article](http://unix.stackexchange.com/questions/99154/disable-transparent-hugepages) about the problems with persisting this setting across reboots.   Since it's in sysfs under /sys/kernel/mm/transparent_hugepage/enabled and not /proc/sys/ , you can't add to sysctl.conf, right?    A **tuned** profile for Ceph would solve this for Fedora, Centos and RHEL but this would not work for Ubuntu would it?  It could also be implemented as a boot-time systemctl unit file (my favorite since it is portable across distros now and can be set early in boot cycle before Ceph runs).

Here is [kernel article](https://www.kernel.org/doc/Documentation/vm/transhuge.txt) that exhaustively discusses THP including pros and cons.

leseb:
I won't say no to a *tuned* profile for Red Hat based systems Ben.
I agree this Kernel tuning should survive reboots.

Do you mind sending a PR for this?

bengland2:
Perhaps this could be a 1-line add to an existing unit file installed by Ceph ansible, but we should only disable THP where it's needed, for OSD hosts that thrash memory.

bengland2:
best I can do so far is in **templates/ceph-osd.service.j2** add this:

    {% if disable_transparent_hugepage -%}
    ExecStartPre=-/usr/lib/ceph/ceph-disable-thp.sh
    {% endif -%}

where this script contains:

    #!/bin/bash
    echo never > /sys/kernel/mm/transparent_hugepage/enabled
    logger "disabled transparent huge pages"

ok?

leseb:
@bengland2 templates/ceph-osd.service.j2 is only for container based deployments so this won't work for traditionnal non-containerized deployments.

bengland2:
[...]

PR #1422, opened 2017-04-08
ceph-common: persist "disable_transparent_hugepage"

### LINKED ISSUE #1408: ceph-common: persist "disable_transparent_hugepage"
On systemd with [tmpfiles](https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html) possible execute simple operations on boot, like create/delete/write. So it's possible to set "persist" values on sysfs.

write `never` to `/sys/kernel/mm/transparent_hugepage/enabled`
```
w /sys/kernel/mm/transparent_hugepage/enabled - - - - never
```

If this useful for ceph-ansible I can PR.

k0ste:
Resolve #1408 and #1013.

First check tmpfiles.d directory for local administrator is present (or create it).
After file is created - systemd-tmpfiles is executed and if exit zero exit status - all okay and hugepage is tuned (can be covered in tests).

### DIFF
--- a/roles/ceph-common/tasks/misc/system_tuning.yml
+++ b/roles/ceph-common/tasks/misc/system_tuning.yml
@@ -4,11 +4,25 @@
   changed_when: false
   failed_when: false
 
+- name: create tmpfiles.d directory
+  file:
+    path: "/etc/tmpfiles.d"
+    state: "directory"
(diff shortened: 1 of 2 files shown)


leseb:
jenkins test this please

PR #1429, opened 2017-04-11
test: use vagrant inventory file

leseb:
Instead of writing a host file for each scenario, let's re use the file
generated by vagrant during the up sequence.


andrewschoen:
I'm -1 on this because I think placing an actual host file is what a user would do, so it's closer to a  real usage test. Also, I've added host vars to many of the inventory files to facilitate different testing scenarios like: https://github.com/ceph/ceph-ansible/blob/master/tests/functional/centos/7/cluster/hosts#L2

leseb:
@andrewschoen I was waiting for your input thanks, this was actually a test. I think vagrant does what a user would do but you have a point with this host file having options. It's just that maintaining this host file for each scenario is not ideal but I guess I can live with that :).

PR #1436, opened 2017-04-11
Backport osd fragments fix

mcv21:
This backports 3 fixes from master onto stable-2.1 They are 349b9ab3e72f188c50b5f84f3ce5ee2fddcfbf80 (which means that you don't lose the osd fragments from ceph.conf on nodes that are members of another group as well as osds) and the two commits from my PR https://github.com/ceph/ceph-ansible/pull/1431 which mean you don't pick up stray files from `/etc/ceph/ceph.d`

### DIFF
--- a/roles/ceph-common/tasks/generate_ceph_conf.yml
+++ b/roles/ceph-common/tasks/generate_ceph_conf.yml
@@ -25,6 +25,7 @@
   assemble:
     src: /etc/ceph/ceph.d/
     dest: /etc/ceph/{{ cluster }}.conf
+    regexp: "^(({{cluster}})|(osd)).conf$"
     owner: "ceph"
     group: "ceph"
     mode: "0644"
--- a/roles/ceph-common/tasks/generate_ceph_conf.yml
+++ b/roles/ceph-common/tasks/generate_ceph_conf.yml
@@ -1,22 +1,33 @@
 ---
-- name: create ceph conf directory
+- name: create ceph conf directory and assemble directory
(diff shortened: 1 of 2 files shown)


mcv21:
Picking one of these tests (jewel-ansible2.2-xenial_cluster), the error seems to be:

`ERROR: unknown environment 'jewel-ansible2.2-xenial_cluster'`

Does that mean the tests don't work on the Stable-2.1 branch?

leseb:
@andrewschoen @alfredodeza any ideas?

@mcv21 If you can wait one to two weeks, we should have finalized stable-2.2, just an FYI :). although I don't mind merging this once the issue with the CI is resolved.

alfredodeza:
we were discussing with @andrewschoen that backporting the new tests for stable-2.1 support is very difficult at the moment. I would prefer getting this to 2.2

andrewschoen:
I think the biggest issue is that to enable all the tests we have now for master in the stable-2.1 branch then we'd have to also backport task changes, which we are trying to avoid.  We've already shipped stable-2.1 downstream and if we bring in changes that hasn't been through QE then we risk regressions.

All that being said, I think it's best to wait until stable-2.2 which should be cut soon.

mcv21:
From our perspective, having a fixed 2.1 [CI tests notwithstanding] would beat having to move to 2.2 (given check-mode doesn't really work, it's always going to be a bit of a risk moving to a new ceph-ansible version).

PR #1439, opened 2017-04-12
tests: fix ceph_rhcs setup

andrewschoen:
When setting ``CEPH_STABLE = false`` as an environment variable it gets passed to --extra-vars as a string, which will always evaluate to ``true`` in the tasks. This PR avoids that by modifying the ``group_vars/all`` for the scenario being tested when we've used the ``rhcs`` prefix.

This also fixes a bug where the ``installs/install_on_redhat.yml`` task was always being included unless you were trying to do an iso install for rhcs.

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -38,8 +38,8 @@
 
 - include: ./installs/install_on_redhat.yml
   when:
-    ansible_os_family == 'RedHat' and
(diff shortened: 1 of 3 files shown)


andrewschoen:

andrewschoen:
jenkins test jewel-ansible2.2-docker_dedicated_journal

andrewschoen:
this needs backport to ``stable-2.2``

PR #1446, opened 2017-04-13
Backport "Make ceph-common aware off osd config fragments" to stable-2.1

D3N14L:
This is already in master, I just backported it to the stable-2.1 branch.
Original commit message below this line:

Make ceph-common aware off osd config fragments

This removes the implicit order requirement when using OSD fragments.
When you use OSD fragments and ceph-osd role is not the last one,
the fragments get removed from ceph.conf by ceph-common.

It is not nice to have this code at two locations, but this is
necessary to prevent problems, when ceph-osd is the last role as
ceph-common gets executed before ceph-osd.

This could be prevented when ceph-common would be explicitly called
at the end of the playbook.

Signed-off-by: Christian Zunker <christian.zunker@codecentric.de>

### DIFF
--- a/roles/ceph-common/tasks/generate_ceph_conf.yml
+++ b/roles/ceph-common/tasks/generate_ceph_conf.yml
@@ -1,22 +1,33 @@
 ---
-- name: create ceph conf directory
+- name: create ceph conf directory and assemble directory
   file:
-    path: /etc/ceph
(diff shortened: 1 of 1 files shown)


D3N14L:
Did not see that there already is a pull request for this: #1436

PR #1430, opened 2017-04-11
test: add mgr section to the host inventory file

leseb:
Without this, we don't test the mgr role so we need to add it.

Signed-off-by: Sébastien Han <seb@redhat.com>

PR #1447, opened 2017-04-13
Fixing mgr group references where it is used with with_items and/or when

D3N14L:
"with_items:" gets evaluated before "when:" to be able to skip specific
items. A few tasks using "groups[mgr_group_name]" in their "with_items"
tried to implement a check if the group has more then 0 elements with
"when". This breaks, if the mgr group is not defined:
```
TASK [ceph-mon : add mgr keys to config and keys paths]
************************
fatal: [mon0001]: FAILED! => {"failed": true, "msg": "'dict object' has
no attribute u'mgrs'"}
fatal: [mon0002]: FAILED! => {"failed": true, "msg": "'dict object' has
no attribute u'mgrs'"}
fatal: [mon0003]: FAILED! => {"failed": true, "msg": "'dict object' has
no attribute u'mgrs'"}
```
This commit fixes the "with_items" fields on those tasks and removes
unnecessary "when"s.

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -52,7 +52,7 @@
     - cephx
     - groups.get(mgr_group_name, []) | length > 0
     - inventory_hostname == groups[mon_group_name]|last
-  with_items: "{{ groups[mgr_group_name] }}"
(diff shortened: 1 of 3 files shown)


leseb:
In progress in https://github.com/ceph/ceph-ansible/pull/1430

PR #1452, opened 2017-04-17
ceph-common: install ceph-mgr for rhcs

andrewschoen:


andrewschoen:
@ktdreyer are these the right package names for rhcs?

PR #1455, opened 2017-04-18
Common: Restore check_socket

guits:
Since `5bec62b` force the user to have a configuration that makes the
PGs in 100% clean state. We need to have a better logging in the
`restart_*_daemon.sh` scripts. Indeed, if a daemon fails to start we
shouldn't not being deeply debugging why.

### DIFF
--- a/roles/ceph-common/tasks/checks/check_socket.yml
+++ b/roles/ceph-common/tasks/checks/check_socket.yml
@@ -1,0 +1,15 @@
+---
+# These checks are used to avoid running handlers at initial deployment.
+- name: check for a ceph socket
+  shell: "stat /var/run/ceph/*.asok > /dev/null 2>&1"
+  changed_when: false
+  failed_when: false
+  always_run: true
+  register: socket
+
+- name: check for a rados gateway socket
+  shell: "stat {{ rbd_client_admin_socket_path }}*.asok > /dev/null 2>&1"
+  changed_when: false
+  failed_when: false
+  always_run: true
+  register: socketrgw
--- a/roles/ceph-common/templates/restart_mon_daemon.sh.j2
+++ b/roles/ceph-common/templates/restart_mon_daemon.sh.j2
@@ -15,6 +15,9 @@ while [ $RETRIES -ne 0 ]; do
   let RETRIES=RETRIES-1
 done
 # If we reach this point, it means there is a problem with the quorum
+echo "Error with quorum."
+echo "cluster status:"
+ceph --cluster ${CLUSTER} -s
 exit 1
 }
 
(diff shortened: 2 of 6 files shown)


andrewschoen:
I ran this branch through our ceph-installer tests and it's still failing. https://jenkins.ceph.com/view/ceph-installer/job/ceph-installer-tests-ansible2.2-nightly_centos7/111/console

I think this is happening because with our ceph-installer tests we only deploy 1 mon and 1 osd and we will never pass the check in the ``restart_osd_daemon.sh`` script.  From what I can see the osd daemon did start just fine. What was the reason we added that check? 

Another issue here is that ceph-installer is setup to allow for the building of a cluster one node at a time, so I think we'd hit this issue regardless of if we planned to test a larger cluster.

roles/ceph-common/templates/restart_mon_daemon.sh.j2
```diff
@@ -15,6 +15,9 @@ while [ $RETRIES -ne 0 ]; do
   let RETRIES=RETRIES-1
 done
 # If we reach this point, it means there is a problem with the quorum
+echo "Error with quorum."
+echo "cluster status:"
+ceph -s
```
leseb: `--cluster {{ cluster }}`

guits: done

leseb:
@andrewschoen can you try again?

roles/ceph-common/tasks/checks/check_socket.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: check for a ceph socket
```
andrewschoen: I'm thinking we should leave some comments here on why we want to do this.

leseb: +1

andrewschoen:
We'll need to get this backported to stable-2.2 as well.

guits:

leseb:
@andrewschoen can we try to address https://github.com/ceph/ceph-ansible/issues/1457 as well here?

guits:
@leseb I pushed a new commit regarding #1457

andrewschoen:
@guits would you mind removing the pg configuration from one of our testing scenarios so we could catch this issue if it comes back?

Perhaps the journal_collocation test: https://github.com/ceph/ceph-ansible/blob/master/tests/functional/centos/7/journal-collocation/group_vars/all#L15

guits:
jenkins test kraken-ansible2.2-docker_dmcrypt_journal_collocation

guits:
@andrewschoen : done!

andrewschoen:
@guits thanks for doing that. Would you mind getting this backported to the stable-2.2 branch now?

PR #1456, opened 2017-04-19
Apparmor on Ubuntu Xenial will not permit containers to mount devices.

netservers:
The same issue is discussed here: https://github.com/moby/moby/issues/9950
This patch resolves the issue by adding the "--security-opt apparmor:unconfined" option in ceph-osd-run.sh.j2, but only for Debian systems.

### DIFF
--- a/roles/ceph-osd/templates/ceph-osd-run.sh.j2
+++ b/roles/ceph-osd/templates/ceph-osd-run.sh.j2
@@ -24,6 +24,9 @@ fi
   --rm \
   --net=host \
   --cap-add SYS_ADMIN \
+  {% if ansible_distribution == 'Ubuntu' -%}
+  --security-opt apparmor:unconfined \
+  {% endif -%}
   --pid=host \
   {% if not osd_containerized_deployment_with_kv -%}
   -v /var/lib/ceph:/var/lib/ceph \


roles/ceph-osd/templates/ceph-osd-run.sh.j2
```diff
@@ -24,6 +24,9 @@ fi
   --rm \
   --net=host \
   --cap-add SYS_ADMIN \
+  {% if ansible_os_family == "Debian" -%}
```
leseb: why not "Ubuntu" instead? `ansible_distribution == 'Ubuntu'`? I don't think Debian uses Apparmor since it's an Ubuntu thing.

leseb:
@netservers looks good, but please squash your commits into one.

netservers:
@leseb I've not done a commit squash before, and I don't see any option in Github to do it. Do I need write access to your repo to do that?

leseb:
You have to do this in Git, see my comment here: https://github.com/ceph/ceph-docker/pull/580#issuecomment-290356545

You will have to adapt the commit ID, if you don't feel comfortable doing this then please create a new branch and push a new PR with just a single comm :).

netservers:
OK. How does that look? It seems pretty squashed to me.

leseb:
@netservers yes looks good now, thanks!

PR #1458, opened 2017-04-19
rgw: set tuning parameters

alimaredia:
Add more rgw ceph config variables and create
pools for rgw.

Suggestions for parameters in comments

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -293,7 +293,7 @@ dummy:
 #radosgw_resolve_cname: false # enable for radosgw to resolve DNS CNAME based bucket names
 #radosgw_civetweb_port: 8080
 #radosgw_civetweb_bind_ip: "{{ ansible_default_ipv4.address }}" # when using ipv6 enclose with brackets: "[{{ ansible_default_ipv6.address }}]"
-#radosgw_civetweb_num_threads: 50
+#radosgw_civetweb_num_threads: 100
 # For additional civetweb configuration options available such as SSL, logging,
 # keepalive, and timeout settings, please see the civetweb docs at
 # https://github.com/civetweb/civetweb/blob/master/docs/UserManual.md
--- a/roles/ceph-rgw/defaults/main.yml
+++ b/roles/ceph-rgw/defaults/main.yml
@@ -21,6 +21,32 @@ cephx: true
 rgw_pull_port: "{{ radosgw_civetweb_port }}"
 rgw_pull_proto: "http"
 
+########
+#TUNING#
+########
+
+# To support buckets with a very large number of objects it's 
+# important to split them into shards. We suggest about 100K
+# objects per shard as a conservative maximum.
+#rgw_override_bucket_index_max_shards: 16
+ 
+# Consider setting a quota on buckets so that exceeding this
+# limit will require admin intervention.
+#rgw_bucket_default_quota_max_objects: 1638400 # i.e., 100K * 16
+
+# This dictionary will create pools with the given number of pgs. 
+# This is important because they would be created with the default
+# of 8. 
+# New pools and their corresponding pg_nums can be created
+# by adding to the create_pools dictionary (see foo).
+#create_pools:
+#  defaults.rgw.buckets.data:
+#    pg_num: 16
+#  defaults.rgw.buckets.index:
+#    pg_num: 32
+#  foo:
+#    pg_num: 4
+
 ##########
 # DOCKER #
 ##########
--- a/group_vars/rgws.yml.sample
+++ b/group_vars/rgws.yml.sample
@@ -29,6 +29,32 @@ dummy:
 #rgw_pull_port: "{{ radosgw_civetweb_port }}"
 #rgw_pull_proto: "http"
 
+########
+#TUNING#
+########
+
+# To support buckets with a very large number of objects it's 
+# important to split them into shards. We suggest about 100K
+# objects per shard as a conservative maximum.
+#rgw_override_bucket_index_max_shards: 16
+ 
+# Consider setting a quota on buckets so that exceeding this
+# limit will require admin intervention.
+#rgw_bucket_default_quota_max_objects: 1638400 # i.e., 100K * 16
+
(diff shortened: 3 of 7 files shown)


group_vars/all.yml.sample
```diff
@@ -326,6 +326,8 @@ dummy:
 #radosgw_dns_s3website_name: your.subdomain.tld # subdomain used by radosgw for website bucket hosting.
 # Rados Gateway options
 #email_address: foo@bar.com
+#TODO: PROVIDE SUGGESTIONS FOR DIFFERENT CONFIGURATIONS
```
leseb: Do not edit this file directly, if it's a rgw option then this should go into `roles/ceph-rgw/defaults/main.yml`, if it's global this goes into: roles/ceph-common/defaults/main.yml`. Then run `generate_group_vars_sample.sh`.

alimaredia:
@alfredodeza this is my first foray into ceph-ansible. After reading the docs I think I got the syntax (mostly) right. ATM we still need to finalize comments and parameters.

roles/ceph-mon/tasks/set_rgw_pool_pg_num.yml
```diff
@@ -0,0 +1,9 @@
+# Uncomment below lines to create pools for the rgw
```
leseb: These are variables in a dict so they should go into: `roles/ceph-mon/defaults/main.yml`

roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -56,6 +56,8 @@
 
 - include: set_osd_pool_default_pg_num.yml
 
+- include: set_rgw_pool_pg_num.yml
```
leseb: See my last comment, you can remove this file.

roles/ceph-mon/tasks/rgw_pool_pgs.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: create rgw pools if rgw_pool is defined
+  shell: |
```
leseb: the command module should be enough here.

roles/ceph-mon/tasks/set_rgw_pool_pg_num.yml
```diff
@@ -0,0 +1,9 @@
+# Uncomment below lines to create pools for the rgw 
+# TODO: recommend tuning pg_nums for different configurations
+#rgw_pool:
+#  rgw_pool_name1:
```
alimaredia: @cbodley are there any default values we can add to this, so the entire file isn't commented out?

roles/ceph-mon/tasks/rgw_pool_pgs.yml
```diff
@@ -0,0 +1,7 @@
...
+    ceph --connect-timeout 5 --cluster {{ cluster }} osd pool {{ item.key }} {{ item.value.pg_num }}
+  changed_when: false
+  failed_when: false
+  with_dict: "{{ rgw_pool }}"
```
cbodley: this task looks useful outside of rgw, maybe we should call it something more generic like `create_pool_pgs.yml`, and use `create_pools` for the dict name?

roles/ceph-mon/tasks/set_rgw_pool_pg_num.yml
```diff
@@ -0,0 +1,9 @@
...
+#    pg_num: y1
+#  rgw_pool_name2:
+#    pg_num: y2
+#  rgw_pool_name3:
+#    pg_num: y3
```
cbodley: does this stuff not belong under `group_vars` for users to customize?

roles/ceph-rgw/defaults/main.yml
```diff
@@ -21,6 +21,28 @@ cephx: true
...
 
+########
+#TUNING#
+########
+# The below values were determined to be safe defaults to 
+# prevent users from putting too many objects into their
+# buckets which causes performance and stability issues.
+#ceph_conf_overrides:
+#    rgw:
+#      radosgw_civetweb_num_threads: 50
```
alimaredia: @mattbenjamin is the suggested number for this supposed to be lower? This is what it defaults to now?

roles/ceph-rgw/defaults/main.yml
```diff
@@ -21,6 +21,28 @@ cephx: true
 rgw_pull_port: "{{ radosgw_civetweb_port }}"
 rgw_pull_proto: "http"
 
+########
+#TUNING#
+########
+# The below values were determined to be safe defaults to 
+# prevent users from putting too many objects into their
+# buckets which causes performance and stability issues.
+#ceph_conf_overrides:
```
alimaredia: @leseb is this the appropriate location and method to set rgw ceph.conf values?

roles/ceph-rgw/defaults/main.yml
```diff
@@ -21,6 +21,28 @@ cephx: true
 rgw_pull_port: "{{ radosgw_civetweb_port }}"
 rgw_pull_proto: "http"
 
+########
+#TUNING#
+########
```
alimaredia: @cbodley here's my first crack at the english ;)

roles/ceph-rgw/defaults/main.yml
```diff
@@ -21,6 +21,28 @@ cephx: true
...
+# The below values were determined to be safe defaults to 
+# prevent users from putting too many objects into their
+# buckets which causes performance and stability issues.
+#ceph_conf_overrides:
+#    rgw:
+#      radosgw_civetweb_num_threads: 100
```
alimaredia: TODO: double check to make sure this conf variable is necessary to override.

roles/ceph-rgw/defaults/main.yml
leseb: No keep it here :). However, this won't work as expected I'm afraid since we have one rgw section per rgw. The issue reported into this PR: https://github.com/ceph/ceph-ansible/pull/1018 should be fixed before we can merge this.

GregMeno:

PR #1459, opened 2017-04-20
mon: update mgr key capabilities

jcsp:
This is to allow ceph-mgr daemons to remote control
osd and mds daemons with MCommand messages.

Fixes: http://tracker.ceph.com/issues/19713
Signed-off-by: John Spray <john.spray@redhat.com>


jcsp:
@leseb thanks for the review, I've updated it.

leseb:
@jcsp Thanks, quick question I see there is no bootstrap key for the mgr, will it make sense to have one?

PR #1461, opened 2017-04-20
remove osd directory scenario

GregMeno:
Signed-off-by: Gregory Meno <gmeno@redhat.com>


alfredodeza:
This looks OK to me. @andrewschoen ?

andrewschoen:
+1

alfredodeza:
I believe the explanation is that directories do not provide anything for clusters unless it is playing with the idea of an OSD without having an actual device.

Proof-of-concept clusters or actual production clusters will never want to use this. We also do not test it anywhere for this same reason.

There might be a few more?

GregMeno:
@leseb I am proposing that this is just extra rope to hang us with, I don't want to support it. I did a quick poll in LA and no-one said keep it.

GregMeno:

GregMeno:

leseb:
Thanks guys, sounds good now :)

PR #1464, opened 2017-04-21
purge-cluster: remove unneeded tasks

andrewschoen:
This removes tasks from the end of the purge-cluster playbook that are not needed anymore. This was noticed because in rhcs we install ``ceph-common`` on clients nodes and that package does not include ``ceph.target`` so the playbook would fail.  Removing the packages should stop that service anyway if it is running, eventually we'll want to not install ``ceph-base`` upstream on client nodes either as it's not meant to be installed standalone.

There is also some cleanup on sysv and upstart tasks.

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -423,11 +423,6 @@
 
   tasks:
 
-  - name: check for anything running ceph
-    shell: "ps awux | grep -- /usr/bin/[c]eph-"
-    register: check_for_running_ceph
-    failed_when: check_for_running_ceph.rc == 0
-
   - name: purge ceph packages with yum
     yum:
       name: "{{ item }}"
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -481,19 +481,6 @@
      path: /var/log/ceph
      state: absent
 
-  - name: remove from sysv
-    shell: "update-rc.d -f ceph remove"
-    when: ansible_service_mgr == 'sysvinit'
-
-  - name: remove upstart and sysv files
-    shell: "find /etc -name '*ceph*' -delete"
-    when: ansible_service_mgr == 'upstart'
-
-  - name: remove upstart and apt logs and cache
-    shell: "find /var -name '*ceph*' -delete"
-    failed_when: false
-    when: ansible_distribution == 'Ubuntu'
-
   - name: request data removal
     local_action: shell echo requesting data removal
     become: false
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/purge-cluster.yml
```diff
@@ -423,18 +423,6 @@
...
   tasks:
 
-  - name: stop ceph.target with systemd
-    service:
-      name: ceph.target
-      state: stopped
-      enabled: no
-    when: ansible_service_mgr == 'systemd'
-
-  - name: check for anything running ceph
```
leseb: Why? I still think it's a valuable task, we've seen many issues when systemd resources failed to stop daemons.

andrewschoen: Sure, we can keep this and move it after the package removals. @ktdreyer ensures me that removing these packages should stop all the daemons, and if not it's a packaging bug that needs to be addressed.

andrewschoen:
jenkins test jewel-ansible2.2-purge_cluster

infrastructure-playbooks/purge-cluster.yml
leseb: Cool, let's keep it then :)

ktdreyer: If you see daemons still running after the package manager has removed/purged the packages, please file a ticket!

infrastructure-playbooks/purge-cluster.yml
```diff
@@ -531,6 +506,11 @@
       state: absent
     when: ansible_os_family == 'RedHat'
 
+  - name: check for anything running ceph
+    shell: "ps awux | grep -- /usr/bin/[c]eph-"
```
leseb: I tend to believe simply using `ps -u ceph -U ceph` should be more than enough and less error prone than this `grep`. Another benefit would be to use the `command` module.

andrewschoen: ok, I made that change

andrewschoen:

andrewschoen:
jenkins test kraken-ansible2.2-docker_cluster

andrewschoen:
jenkins test kraken-ansible2.2-xenial_cluster

andrewschoen:

andrewschoen:
@ktdreyer thoughts on backporting this to ``stable-2.2``? We do not have a BZ for it currently but I imagine we will soon.

ktdreyer:
sure, backport it

andrewschoen:
We seem to be having issues pulling from docker hub, perhaps a side effect of this scheduled maintence: https://status.docker.com/pages/maintenance/533c6539221ae15e3f000031/58fe66db1e525b480e00008f

Regardless, this PR does not introduce any changes to our docker tasks so I think it's safe to merge.

Issue #1445, opened 2017-04-13
ceph-mon task currently chmods all the files of /etc/ceph/*.keyring


wind0204:
I mean, do ceph-mons really have to chmod all the files of /etc/ceph/*.keyring?

13:37:55 Nicho1as | Hi, ceph-ansible playbook's ceph-mon role chmods all the files of /etc/ceph/*.keyring but I have to keep /etc/ceph/ceph.client.libvirt.keyring as 0640 not 0600... what would you guys suggest as a preferable solution? I mean do ceph-mons really have to chmod all the files of /etc/ceph/*.keyring?
13:38:29          | it's written in ceph-ansible/roles/ceph-mon/tasks/ceph_keys.yml
13:38:56          | leseb: I'd like to hear your opinion..
16:27:27          | for the time being I just put a task right after that which chmods the file I need to keep 0640..
16:34:29      --> | *****04 (~*****04@s******.******.***) has joined #ceph-ansible
16:44:40      --> | ******ne-***** (~******ne@1**.***.**.**) has joined #ceph-ansible
16:55:54    leseb | Nicho1as can you open an issue for this please? on github :) we will address that
16:56:27 Nicho1as | Sure, I will after dinner
16:57:13          | Oh, I could do it right now ... just copying the irc log :-q hehe

guits:
PR #1471 was closed due to inactivity,  closing this issue as well.
Feel free to re-open.

PR #1471, opened 2017-04-25
Changes permission of libvirt keyring

nilamdyuti:
Fixes: #1445

Signed-off-by: Nilamdyuti Goswami <ngoswami@redhat.com>

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -108,14 +108,11 @@
   when: cephx
 
 - name: set ceph libvirt keyring permission
-  file:
-    path: "/etc/ceph/{{ cluster }}.client.libvirt.keyring"
-    owner: "ceph"
-    group: "ceph"
-    mode: "0640"
-    command: chown {{ owner }}:{{ group }} {{ path }} ; chmod {{ mode }} {{ path }}
-    register: change_perm_result
-    changed_when: change_perm_result.stdout
+  shell: chown ceph:ceph *.libvirt.keyring ; chmod 0640 *.libvirt.keyring
+  args:
+    chdir: /etc/ceph/
+  changed_when: false
+  always_run: true
   when:
     - is_libvirt_keyring_exists == 0
     - cephx
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -83,7 +83,7 @@
     - "{{ inventory_hostname == groups[mon_group_name] | last }}"
 
 - name: find ceph keys
-  shell: ls -1 /etc/ceph/*.keyring
+  shell: find /etc/ceph/ -name "*.keyring" ! -name "*.libvirt.keyring"
   changed_when: false
   register: ceph_keys
   always_run: true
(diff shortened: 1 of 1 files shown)


roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -99,13 +99,32 @@
     - "{{ ceph_keys.get('stdout_lines') | default([]) }}"
   when: cephx
 
+- name: check if libvirt keyring exists
+  command: find /etc/ceph -name "*.libvirt.keyring"
```
leseb: will the `*` work with the `command` module?

roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -99,13 +99,32 @@
...
+  always_run: true
+  run_once: true
+  when: cephx
+
+- name: set ceph libvirt keyring permission
+  file:
+    path: "/etc/ceph/{{ cluster }}.client.libvirt.keyring"
+    owner: "ceph"
+    group: "ceph"
+    mode: "0640"
```
leseb: Don't you want these owner/group/mode to be configurable?

roles/ceph-mon/tasks/ceph_keys.yml
nilamdyuti: @leseb Yes, you are right. Wildcard ideally wouldn't work with command module. I have to use the shell module for the command, take the output in a register variable, then use command module. Then it should work.

roles/ceph-mon/tasks/ceph_keys.yml
nilamdyuti: @leseb Sorry, I missed that. I will make the changes.

nilamdyuti:
@leseb Made the changes. Kindly review.

nilamdyuti:
@leseb Are the changes acceptable or modification is required?

leseb:
@nilamdyuti I don't see anything.

nilamdyuti:
@leseb Sorry, my bad. I had amended the commit. Maybe that's why you don't see the diff. The new changes can be seen in between line 102 to line 122. Old unchanged changes are in line 86 and 130.

nilamdyuti:
@leseb Were you able to see the changes?

leseb:
@nilamdyuti nop, can you?

nilamdyuti:
@leseb That's strange. Yes, I can. As I said before, the changes can be seen from line 102 to line 122.

nilamdyuti:
@leseb I can see the changes in both `unified` and `split` view under `Files changed` tab.

roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -99,13 +99,35 @@
...
+    command: chown {{ owner }}:{{ group }} {{ path }} ; chmod {{ mode }} {{ path }}
```
leseb: There is something wrong with your patch here, you included the 'command' module in the file module.
Also, where are those variables declared? owner? group? path? mode?

nilamdyuti: @leseb My bad. Making the changes.

nilamdyuti:
@leseb Committed the changes. Kindly review.

roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -99,13 +99,32 @@
...
+- name: check if libvirt keyring exists
+  command: find /etc/ceph -name "{{ cluster }}.client.libvirt.keyring"
+  changed_when: false
+  register: is_libvirt_keyring_exists
+  always_run: true
+  run_once: true
+  when: cephx
+
+- name: set ceph libvirt keyring permission
+  shell: chown ceph:ceph *.libvirt.keyring ; chmod 0640 *.libvirt.keyring
```
leseb: Previously you were using the `file` module which is exactly what you should be doing.
My last comment was the owner/mode might need to be configurable with variables so this is not hardcoded.

nilamdyuti: @leseb Okay. Making the changes.

nilamdyuti: @leseb In task `set keys permissions` of this file `roles/ceph-mon/tasks/ceph_keys.yml`  owner/group/mode are not configurable with variables as well. They are simply used the way they are used in a `file` module. That's what I had done in my first commit for the task `set ceph libvirt keyring permission` that I had added.

Do owner/group/mode actually need to be configurable with variables?

nilamdyuti: @leseb Added the `file` module again. Kindly check.

nilamdyuti: @leseb Are the changes acceptable or modification is required?

nilamdyuti:
@leseb Can you please check if the changes are acceptable?

leseb:
@nilamdyuti I don't think this is the right approach to fix this. I don't really want to end up in a situation where everyone adds new code for a new key. Let me think about this.

guits:
This PR has been opened for more than a year, if you think it's still worth to pursue feel free to re-open.

PR #1476, opened 2017-04-27
improve shrink-osd.yml can shrink osd when disk damage

WingkaiHo:
When the osd's disk or partition damage,  we could not use the ceph-disk deactivate and ceph-disk destroy to shrink osd.  It must use the ceph command to shrink osd. 

You can use the sgdisk --zap-all -- /dev/device_name to simulate partition damage.
This osd could not be shrink by this playbook, it will throw error "Cannot find any match device!!" at task deactivating osd(s)


WingkaiHo:
thanks

PR #1483, opened 2017-04-28
tests: collocate a mgr daemon on a mon node

andrewschoen:
So we can find potential issues with a mgr collocated on a mon node.

### DIFF
--- a/tests/functional/centos/7/cluster/hosts
+++ b/tests/functional/centos/7/cluster/hosts
@@ -17,3 +17,4 @@ client0
 
 [mgrs]
 mgr0
+mon0


leseb:
Not sure how relevant this is since by default a monitor will run a ceph-mgr.
This is how the ceph-mon unit file works, it wants to have ceph-mgr. During its initialization ceph-mgr will auto create a key etc...

guits:
@andrewschoen @leseb from what I have been facing these last few days, the `ceph-mgr` reference in `ceph-mon@.service` unit file you mention here is kind of something optional just to ensure `ceph-mgr` will be started first in case where it is precisely collocated.
`Wants=` doesn't work like `Requires=`, if the the unit listed in `Wants=` doesn't exist, it won't prevent the 'parent unit' from starting.

IMO this PR should be reopen.

guits:
Actually, I would add that each scenario testing in luminous without an explicit declaration of an mgr instance will deploy 'well' in the sense the job is `SUCCESS` but results with a cluster in `HEALTH_ERR` state since there is no mgr instance.

PR #1487, opened 2017-05-01
tests: actually test jewel when we say we are

andrewschoen:
Signed-off-by: Andrew Schoen <aschoen@redhat.com>

PR #1489, opened 2017-05-02
Backport: "tests: actually test jewel when we say we are #1487"

andrewschoen:
backport of #1487 

I did not backport 8472472194521a468b58bac76954cb86cbcdaece as it looks like we are missing other documentation commits and it was conflicting.


andrewschoen:
The failing tests all seem to be caused by the same issues we saw in the master branch with #1487

PR #1490, opened 2017-05-03
tests: allow for insecure docker registries when testing rhcs

andrewschoen:
Signed-off-by: Andrew Schoen <aschoen@redhat.com>

PR #1491, opened 2017-05-03
backport: "tests: allow for insecure docker registries when testing rhcs #1490"

andrewschoen:
backport of #1490


andrewschoen:
No need to wait for tests to finish as these only work with rhcs testing

PR #1501, opened 2017-05-05
tests: use the xenial container for jewel instead of centos7

andrewschoen:


andrewschoen:
I seem to be having better luck with the jewel xenial container than the centos7 one. We'll see if the CI agrees.

PR #1505, opened 2017-05-10
remove the install-ansible.sh script

andrewschoen:
I'm unsure why this is needed, users should be able to install ansible
on their own in the method they prefer.


andrewschoen:
I don't see a reason to wait for CI on this one.

ktdreyer:
`ceph-aio-no-vagrant.sh` referred to this `install-ansible.sh` script.

If we're going to clean up, let's delete `ceph-aio-no-vagrant.sh` as well?

leseb:
I thought we agreed on having more peers reviews before merging our own patch?
What was urgent about this one?

I still think `ceph-aio-no-vagrant.sh` can be beneficial and so is `install-ansible.sh`...

ktdreyer:
I am all for waiting for peer review. It looks like you merged the corresponding backport, though?

andrewschoen:
@leseb installing EPEL anywhere was raising flags in the downstream release. Why does ceph-ansible need to be the repo to store all these scripts?

leseb:
@andrewschoen it's just that there is no other places at the moment :).

@ktdreyer Yes I did, I was actually about to say something and then realized you guys might have a good reason. It's just that your comment made me think we should have had more peer review :).

Let me modify `ceph-aio-no-vagrant.sh` so we can keep it and not fail because `install-ansible.sh` does not exist anymore.

leseb:
@andrewschoen @ktdreyer https://github.com/ceph/ceph-ansible/pull/1511

PR #1508, opened 2017-05-10
rgw: move default bucket quota conf vars to global

alimaredia:
"rgw override bucket index max shards" and
"rgw bucket default quota max objects" were in the
client section of the ceph.conf and not being
applied, this commit moves them to global

Resolves: bz#1391500


andrewschoen:
jenkins test kraken-ansible2.2-update_cluster

PR #1511, opened 2017-05-10
ceph-aio-no-vagrant.sh: check if ansible is installed

leseb:
Since we don't provide install-ansible.sh anymore we just check if it
exists. If not we fail and ask for installation.

### DIFF
--- a/ceph-aio-no-vagrant.sh
+++ b/ceph-aio-no-vagrant.sh
@@ -18,7 +18,7 @@ INSTALL_RGW=true
 # FUNCTIONS
 
 show_help() {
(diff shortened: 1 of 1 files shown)


ktdreyer:
It looks like there are more things going on in this commit?

leseb:
@ktdreyer yes some shell lint, want me to separate the commits?

ktdreyer:
It's ok to leave this PR as is to keep it moving.

I think the future it would be great to break these out so it's easier to review and bisect.

leseb:
@ktdreyer sure, thanks!

Issue #1514, opened 2017-05-11
docker: mgr jewel Error EINVAL: bad entity name


leseb:
CI error:

```
TASK [ceph-mon : create ceph mgr keyring(s) when mon is containerized] *********
task path: /home/jenkins-build/build/workspace/ceph-docker-prs-ceph_ansible-jewel-centos7_cluster/ceph-ansible/roles/ceph-mon/tasks/docker/main.yml:93
failed: [mon2] (item=mgr0) => {
    "changed": false, 
    "cmd": [
        "docker", 
        "exec", 
        "ceph-mon-ceph-mon2", 
        "ceph", 
        "--cluster", 
        "test", 
        "auth", 
        "get-or-create", 
        "mgr.ceph-mgr0", 
        "mon", 
        "allow profile mgr", 
        "osd", 
        "allow *", 
        "mds", 
        "allow *", 
        "-o", 
        "/etc/ceph/test.mgr.ceph-mgr0.keyring"
    ], 
    "delta": "0:00:01.201008", 
    "end": "2017-05-11 08:47:59.175259", 
    "failed": true, 
    "item": "mgr0", 
    "rc": 22, 
    "start": "2017-05-11 08:47:57.974251", 
    "warnings": []
}

STDERR:

Error EINVAL: bad entity name
```

This is due to the fact that the 'mgr' entity name does not exist in Jewel.

PR #1499, opened 2017-05-04
Common: Do not install ceph-mgr packages on jewel

guits:
ceph-mgr tasks have to be skipped on jewel.

Fix: #1494

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #1515, opened 2017-05-11
mon: do not create mgr key on jewel

leseb:
The CI on Docker is reporting the following error:

STDERR:
Error EINVAL: bad entity name

This is due to the fact that this auth entity name does not exist on
Jewel so we should not create that key when running Jewel containers.

Fixes: https://github.com/ceph/ceph-ansible/issues/1514

### DIFF
--- a/roles/ceph-mon/tasks/docker/main.yml
+++ b/roles/ceph-mon/tasks/docker/main.yml
@@ -125,4 +125,5 @@
       - item.stat.exists == true
   when:
     - inventory_hostname == groups[mon_group_name]|last
-    - ceph_release_num.{{ ceph_stable_release }} > ceph_release_num.jewel
+    - not ceph_docker_image_tag.find('jewel') != -1
+    - ceph_docker_image != 'rhceph'


andrewschoen:
We must have missed this in #1499. @guits used ``ceph_stable_release`` to skip docker tasks that don't work with jewel i.e. https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/docker/main.yml#L128

Maybe checking the tag is a better approach though as ``ceph_stable_release`` doesn't really pertain to containers.

leseb:
I don't think https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/docker/main.yml#L128 is properly evaluated since `ceph_release_num` is only declared in ceph-common, role that is skipped during a containerized deployment.

Am I missing something?

leseb:
This patch should probably remove https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/docker/main.yml#L128 as well...

andrewschoen:
@leseb, yeah that's why we were using ``ceph_stable_release`` instead of ``ceph_release_num``.

leseb:
@andrewschoen `ceph_stable_release` is also part of ceph-common, I'm not sure how this is solving the issue. Or perhaps the CI puts it as an ENV var?

Anyway, what do you think about this patch?

andrewschoen:
@leseb yeah, it works because the CI is passing it in through extra-vars https://github.com/ceph/ceph-ansible/blob/master/tox.ini#L95

The patch works here but I'm concerned what happens when the tag is 'latest', or more concerning is the downstream container which will not mention jewel at all but will be based on jewel. We need some consistent way to know what ceph release the container is based on.

Also, I think you're right that this patch should maybe change https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/docker/main.yml#L128 as well.

leseb:
@andrewschoen right, what about this approach? It should cover both upstream and downstream. However this does not address the case where the tag is 'latest'.

andrewschoen:
@leseb maybe we don't care about ``latest`` right now? We know upstream ``latest`` will always be > jewel. I don't believe downstream will ever use the ``latest`` tag.

roles/ceph-mon/tasks/docker/main.yml
```diff
@@ -125,4 +125,4 @@
       - item.stat.exists == true
   when:
     - inventory_hostname == groups[mon_group_name]|last
-    - ceph_release_num.{{ ceph_stable_release }} > ceph_release_num.jewel
+    - not ceph_docker_image_tag.find('jewel') != -1 or not ceph_docker_image_tag.find('ceph-2') != -1
```
leseb: Yes, images name standardization is pretty bad at the moment. It'll be nice to have something a bit more solid. At least, something we can rely on without having to go back in a couple of months to update it...

leseb:
@andrewschoen yes I think we don't care about latest for now, just pushed a new version.

roles/ceph-mon/tasks/docker/main.yml
```diff
@@ -125,4 +125,4 @@
       - item.stat.exists == true
   when:
     - inventory_hostname == groups[mon_group_name]|last
-    - ceph_release_num.{{ ceph_stable_release }} > ceph_release_num.jewel
+    - not ceph_docker_image_tag.find('jewel') != -1 or ceph_docker_image_name != rhceph
```
leseb: good catch

roles/ceph-mon/tasks/docker/main.yml
```diff
@@ -125,4 +125,4 @@
       - item.stat.exists == true
   when:
     - inventory_hostname == groups[mon_group_name]|last
-    - ceph_release_num.{{ ceph_stable_release }} > ceph_release_num.jewel
+    - not ceph_docker_image_tag.find('jewel') != -1 or ceph_docker_image_name != 'rhceph'
```
andrewschoen: s/ceph_docker_image_name/ceph_docker_image

andrewschoen:
jenkins test this please

leseb:
jenkins test this please

leseb:
jenkins test jewel-ansible2.2-update_cluster

leseb:

andrewschoen:
@leseb I don't think that ``ceph_docker_image_tag.find('jewel') != 1`` is doing what we expect here? I couldn't find docs on that method at all. Perhaps we need to use ``search`` or ``match``?

https://docs.ansible.com/ansible/playbooks_tests.html#testing-strings

leseb:
@andrewschoen the command is correct, it's just that I need a `and` instead of a `or`.

djgalloway:
I'm still seeing this using master branches of ceph-docker and ceph-ansible.

https://paste.fedoraproject.org/paste/MwD0750xAKzAnXi7YpmmZF5M1UNdIGYhyRLivL9gydE=

andrewschoen:
@djgalloway if you're using the ceph-docker CI process we were still depending on the 'latest' tag there which this PR doesn't support.

I'm addressing this here https://github.com/ceph/ceph-docker/pull/641/commits/67cf31047aa0914795b81f366517732c3c66db19

PR #1516, opened 2017-05-12
generate_group_vars_sample: merge role variables

leseb:
In some case, ceph_docker_registry does not match with a defined host
group in the playbook so we make this variable widely available.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1447179

### DIFF
--- a/generate_group_vars_sample.sh
+++ b/generate_group_vars_sample.sh
@@ -2,20 +2,20 @@
 
 set -euo pipefail
 
+#############
+# VARIABLES #
+#############
+
 basedir=$(dirname "$0")
+no_header="ceph-docker-common" # pipe separated list of roles with no header, MUST end with '$', e.g: 'foo$|bar$'
+merge_in_all="ceph-common$|ceph-docker-common$" # pipe separated list of roles you want to merge in all.yml.sample, MUST end with '$', e.g: 'foo$|bar$'
 
-for role in "$basedir"/roles/ceph-*; do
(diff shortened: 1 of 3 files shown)


andrewschoen:
I'm not sure that this is the correct path here. I don't like duplicating variables from ``ceph-docker-common`` into ``ceph-common`` just because ``generate_group_vars.sh``  can't put vars from both roles into ``group_vars/all.yml.sample``.

Moving forward we'll have to continue this duplication for every var we put into ``ceph-docker-common``.

leseb:
@andrewschoen I'm not still not sure why we need to do that btw.
The meta declaration in the role should point to `docker-common.yml` once the role is started right?

andrewschoen:
@leseb the meta declaration just includes the role, it does nothing with the group_vars. For ansible to use files defined in ``group_vars/`` the name of the file *must* match a host group. Like ``mons``, ``osds``, etc. There is no group ``docker-common`` so the variables defined there are never used, as proved by https://bugzilla.redhat.com/show_bug.cgi?id=1447179 and the ansible documentation.

leseb:
@andrewschoen oh yes, thanks for the refresh :). I'll work on a different approach. Stay tuned.

leseb:

GregMeno:
@leseb so it seems like the intent here is to move the ceph_docker_registry in the all.yml.sample right?

leseb:
@GregMeno correct

PR #1518, opened 2017-05-12
ceph-common: improve error message on restart osd daemon script

alfredodeza:

### DIFF
--- a/roles/ceph-common/templates/restart_osd_daemon.sh.j2
+++ b/roles/ceph-common/templates/restart_osd_daemon.sh.j2
@@ -13,7 +13,10 @@ check_pgs() {
     let RETRIES=RETRIES-1
   done
   # PGs not clean, exiting with return code 1
-  echo "Error with PGs, check config"
+  echo "Error while running 'ceph $CEPH_CLI -s', PGs were not reported as active+clean"
+  echo "It is possible that the cluster has less OSDs than the replica configuration"
+  echo "Will refuse to continue"
+  ceph $CEPH_CLI -s
   exit 1
 }
 


roles/ceph-common/templates/restart_osd_daemon.sh.j2
```diff
@@ -13,7 +13,11 @@ check_pgs() {
     let RETRIES=RETRIES-1
   done
   # PGs not clean, exiting with return code 1
-  echo "Error with PGs, check config"
+  echo "Error while running `$CEPH_CLI -s`, PGs were not reported as active+clean"
```
leseb: Is `$CEPH_CLI -s` really important since we show Ceph's status after?

roles/ceph-common/templates/restart_osd_daemon.sh.j2
```diff
@@ -13,7 +13,11 @@ check_pgs() {
...
+  echo "It is possible that the cluster has less OSDs than the replica configuration"
+  echo "Will refuse to continue"
+  # Get `ceph -s` output for added information
```
leseb: should be `ceph --cluster <name> -s`.

roles/ceph-common/templates/restart_osd_daemon.sh.j2
alfredodeza: actually, it should be `ceph $CEPH_CLI -s`. You are right that we show the status after, but we don't say what exactly we were running which I think it is important.

roles/ceph-common/templates/restart_osd_daemon.sh.j2
alfredodeza: This is fixed

leseb: Actually not sure if this comment is needed

roles/ceph-common/templates/restart_osd_daemon.sh.j2
leseb: Sure, just make sure the code does not executed in the `echo` :)

roles/ceph-common/templates/restart_osd_daemon.sh.j2
```diff
@@ -13,7 +13,11 @@ check_pgs() {
     let RETRIES=RETRIES-1
   done
   # PGs not clean, exiting with return code 1
-  echo "Error with PGs, check config"
+  echo "Error while running `ceph $CEPH_CLI -s`, PGs were not reported as active+clean"
```
leseb: use single quotes instead of backquote, otherwise the code command will be executed inside the `echo`.

roles/ceph-common/templates/restart_osd_daemon.sh.j2
```diff
@@ -13,7 +13,11 @@ check_pgs() {
...
+  echo "It is possible that the cluster has less OSDs than the replica configuration"
+  echo "Will refuse to continue"
+  # Get `ceph --cluster <name> -s` output for added information
```
leseb: Is this really needed?

roles/ceph-common/templates/restart_osd_daemon.sh.j2
alfredodeza: Ah, my RestructuredText habits kicking in here. FIxed

roles/ceph-common/templates/restart_osd_daemon.sh.j2
alfredodeza: Guess not. Removed.

guits:

PR #1552, opened 2017-05-19
common: explicitly set rhel os version support

leseb:
Clarify in the error message that only RHEL version >= 7.3 are
supported.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1452431


leseb:
@ktdreyer @andrewschoen doing the backport now

PR #1554, opened 2017-05-19
Add an option for setting rbd_default_features in ceph.conf

ericlake:
Having exclusive-lock, object-map, fast-diff, and deep-flatten enabled was
preventing mapping on CentOS 7.


leseb:
Please use config override to add a new config option into your `ceph.conf`, see https://github.com/ceph/ceph-ansible/blob/master/group_vars/all.yml.sample#L350-L365

PR #1555, opened 2017-05-22
rolling-update: set/unset flags on the right container

leseb:
Problem: we are delegating the set/unset flag to a monitor node but we
try to call an osd container

Solution: use the right container name.


leseb:
jenkins test this please

leseb:
CI failures unrelated.

PR #1557, opened 2017-05-22
common: fix installation condition

leseb:
Problem: we could end up in situation where we would install a package
on a machine that does not have the right repo enabled. Because the
condition was set to OR we weren't pinning a particular host but just a
condition. Let's say someone sets 'ceph_origin == "distro"', this would
try to install OSD packages on Monitors.

Solution: use a AND condition to first pin to the group_name (which
identifies a set of hosts) AND then after this one of the installation
condition.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1453119

### DIFF
--- a/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
+++ b/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
@@ -29,6 +29,6 @@
     - ceph_rhcs
     - not ceph_rhcs_cdn_install
     - not ceph_rhcs_iso_install
-    - ceph_origin == "upstream"
+    - ceph_origin != 'distro'
   tags:
     - package-install
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -74,8 +74,6 @@
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
   when:
     - mon_group_name in group_names
-      or ceph_origin == "distro"
-      or ceph_custom
 
 - name: install distro or red hat storage ceph osd
   package:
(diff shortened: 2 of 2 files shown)


andrewschoen:
Actually, I'm not sure the condition ``- (ceph_origin == "distro" or ceph_custom)`` is needed at all. ``ceph_origin: distro`` is set when we want to avoid setting up the CDN or an ISO, but those tasks have different checks now: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/main.yml#L20-L35

We just need ``ceph_origin: distro`` set so that the mandatory checks pass here: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/checks/check_mandatory_vars.yml#L32

Maybe that line in check_mandatory_vars.yml should be ``- ceph_origin != "distro"`` as that's the other valid install option for rhcs.

leseb:
jenkins test this please

leseb:
`ceph_origin: distro` is used when we want to install the Ceph version that comes with the default repository of the distribution.

 @andrewschoen that's right, I also noticed @zhsj had an attempt to fix this too in https://github.com/ceph/ceph-ansible/pull/1283. I guess I didn't quite understand the real purpose of the commit. @zhsj I'm adding you as a Co-Authored-By for this commit :).

I'm fixing this properly now.

leseb:
@andrewschoen I'm just wondering what's the real usage of this check: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/checks/check_mandatory_vars.yml#L32

Is this line really needed?

andrewschoen:
@leseb I think https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/checks/check_mandatory_vars.yml#L32 needs to stay here, but perhaps changed to ``- ceph_origin != "distro"``. It's important because that is a valid way of deploying RHCS, with preprepared repos present on the nodes.

leseb:
@andrewschoen final review please

PR #1563, opened 2017-05-23
Use host-passthrough for libvirt vCPUs

zmc:


zmc:
Without this, ceph-docker unit tests won't work on smithi machines (and maybe more)

leseb:
Thanks @zmc

PR #1564, opened 2017-05-23
Fixing partition detection regex for FusionIO devices.

arodd:
This resolves a bug identifying partitions for FusionIO devices.


leseb:
CI issues are unrelated, moving on with this one.

PR #1570, opened 2017-05-29
DNM: docker: bindmount /var/log/ceph

leseb:
Currently, we do not export /var/log/ceph to the containers and we
remove them once they are stopped so we have no way of getting the logs
after e.g. a osd daemon crashes. The logs are just lost and we have no
idea why the daemon in the container stopped.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1455509


leseb:
As per my last comment here: https://bugzilla.redhat.com/show_bug.cgi?id=1455509#c4 I don't think that's a good idea anymore.

PR #1573, opened 2017-05-30
keep port as part of the control path

duritong:
On a Vagrant setup, every host is 127.0.0.1 and every user is vagrant.
To not have ansible do the same stuff on the same machine (the one
we connected first), we should keep the port as part of the
control path.

Also this helps in setups where different hosts might be hidden
behind the same IP, but on different ports (e.g. using DNAT).

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -10,4 +10,4 @@ log_path = /var/log/ansible.log
 
 [ssh_connection]
 # see: https://github.com/ansible/ansible/issues/11536
-control_path = %(directory)s/%%h-%%r
+control_path = %(directory)s/%%h-%%r-%%p


leseb:
@duritong lgtm but can you elaborate on the "To not have ansible do the same stuff on the same machine" please?

duritong:
@leseb well, ssh will reuse an existing control socket based on the calculated path. And hence if the only differentiation between connecting to a host is the port (as it is with a local Vagrant setup where everything goes to 127.0.0.1 but differentiated by port), it will mean that ansible's commands will get executed on another host intended, as ssh is reusing the already existing control socket.

As an example, with a 2 node setup, commands intended for the ODS (like preparing the disks) will get executed on the mon host (and obviously failing), as the first connection was made to the mon host, so the socket behind the path `.ansible/127.0.0.1-root` is a connection to the mon host, which will be reused by ssh when connecting to the OSD host.

Or in other words: the local Vagrant use case of that repository is completely broken without that patch.

guits:
jenkins test jewel-ansible2.2-xenial_cluster

guits:

leseb:
Thanks @duritong

PR #1576, opened 2017-05-31
tests: set -e DEBUG=verbose on all docker testing scenarios

andrewschoen:

### DIFF
--- a/tests/functional/centos/7/docker-cluster-dedicated-journal/group_vars/all
+++ b/tests/functional/centos/7/docker-cluster-dedicated-journal/group_vars/all
@@ -21,5 +21,6 @@ devices:
   - /dev/sda
 raw_journal_devices:
   - /dev/sdb
-ceph_osd_docker_prepare_env: -e CLUSTER={{ cluster }} -e OSD_JOURNAL_SIZE={{ journal_size }} -e OSD_FORCE_ZAP=1 -e OSD_JOURNAL={{ raw_journal_devices[0] }}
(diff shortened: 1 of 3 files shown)


ktdreyer:
@leseb how can we reduce the impact of debug logging so that `DEBUG=verbose` does not cause the containers to behave differently?

andrewschoen:
@leseb I'm confused, enabling debug mode here changes the way the container operates? I was hoping it'd allow us to have more insight on why things fail on the PRs without having to modify the branch to enable debug and do a manual run of the test.

I ran this branch against the ``purge_docker_cluster`` test and got the following: https://jenkins.ceph.com/view/ceph-ansible/job/ceph-ansible-scenario/96/console

leseb:
`verbose` activates `set -x` which enables tracing which introduces latency in the code execution. Thus this might hide potential raise conditions.

@andrewschoen I understand the intent and I agree this would be useful. With that I think we are getting back to the discussion about adding "jenkins test <scenario> debug", which ended with "boot yourself a slave and try to reproduce" :).

ktdreyer:
Could we have Jenkins re-run with "-x" if the test fails?

andrewschoen:
@ktdreyer not easily, as you can see from the diff it takes modifying config options in group_vars to include ``-e DEBUG=verbose`` when you deploy the container.

What @leseb is asking for is a way from a pr to say "jenkins debug $scenario" and then that slave that ran the test would stay around for manual debugging. This is non trivial because currently mita has no way to spin up a 'single-use' type of slave. There is also difficulty in knowing when a node that's used in this way should be cleaned up.

leseb:
@andrewschoen can you have a condition in tox to add more options to `ceph_osd_docker_prepare_env` if a special flag is passed to the `jenkins test` line? And then add `-e DEBUG=verbose`?

PR #1582, opened 2017-06-02
purge-docker-cluster: include ceph_docker_registry

andrewschoen:
We need to include ceph_docker_registry when removing containers/images
because if we don't it will assume docker.io which is not always where
the image originated from, causing the playbook to fail.

### DIFF
--- a/infrastructure-playbooks/purge-docker-cluster.yml
+++ b/infrastructure-playbooks/purge-docker-cluster.yml
@@ -43,6 +43,11 @@
          on the command line when invoking the playbook"
     when: remove_packages != 'yes'
 
+  - name: set ceph_docker_registry value if not set
+    set_fact:
+      ceph_docker_registry: "docker.io"
+    when: ceph_docker_registry is not defined
+
 
 - name: purge ceph mds cluster
 
@@ -62,7 +67,7 @@
 
   - name: remove ceph mds container
     docker:
-      image: "{{ ceph_docker_image }}:{{ ceph_docker_image_tag }}"
+      image: "{{ ceph_docker_registry }}/{{ ceph_docker_image }}:{{ ceph_docker_image_tag }}"
       name: "ceph-mds-{{ ansible_hostname }}"
       state: absent
     ignore_errors: true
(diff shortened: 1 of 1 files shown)


andrewschoen:
jenkins test kraken-ansible2.2-update_dmcrypt

andrewschoen:
This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1455187

ktdreyer:
Can we simplify this by leaving out the "docker.io" default?

It works from the docker CLI:
```
$ docker rmi centos:7
Untagged: centos:7
Untagged: docker.io/centos@sha256:4eda692c08e0a065ae91d74e82fff4af3da307b4341ad61fa61771cc4659af60
Deleted: sha256:a8493f5f50ffda70c2eeb2d09090debf7d39c8ffcd63b43ff81b111ece6f28bf
Deleted: sha256:36018b5e978717a047892794aebab513ba6856dbe1bdfeb478ca1219df2c7e9c
```

andrewschoen:
> Can we simplify this by leaving out the "docker.io" default?

No, because in most cases if people are using docker.io they will not have set ``ceph_docker_registry`` in their ``group_vars/`` and depend on the default value in ``roles/ceph-docker-common/defaults/main.yml`` and the ``purge-docker-cluster.yml`` playbook does not load any role defaults. This means that if I didn't include the default ``ceph_docker_registry`` would be undefined and the playbook would fail.

andrewschoen:
@ktdreyer I could try with one task at the top of the playbook that sets it to docker.io if it's not defined.

andrewschoen:
jenkins test jewel-ansible2.2-purge_docker_cluster

andrewschoen:
jenkins test jewel-ansible2.2-update_dmcrypt

andrewschoen:

andrewschoen:
jenkins test kraken-ansible2.2-journal_collocation

andrewschoen:
jenkins test kraken-ansible2.2-purge_dmcrypt

andrewschoen:
jenkins test kraken-ansible2.2-update_dmcrypt

PR #1586, opened 2017-06-05
rpm: do not package iscsi files

ktdreyer:
Currently we cannot install the ceph-iscsi-ansible RPM on a node where the ceph-ansible RPM is already installed.

ceph-iscsi-ansible should install on top of the ceph-ansible environment without issues.


ktdreyer:
I've tested the RPMs from Shaman and verified that this fixes the issue reported at https://bugzilla.redhat.com/1454945

ktdreyer:
We don't currently do much with the RPMs upstream, so I've skipped backporting. We'll catch up when we're at ceph-ansible v2.3.0.

PR #1589, opened 2017-06-06
remove ceph-iscsi-gw play from site.yml.sample

andrewschoen:
We ship ceph-iscsi-gw in a separate repo downstream and do not package
it with ceph-ansible. Including the play for ceph-iscsi-gw in
site.yml.sample makes the playbook fail when using the downstream
packages.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1454945

### DIFF
--- a/site.yml.sample
+++ b/site.yml.sample
@@ -11,7 +11,6 @@
   - restapis
   - rbdmirrors
   - clients
-  - iscsigws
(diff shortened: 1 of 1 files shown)


ktdreyer:
Having this in the tree at all is a problem, until it is really functional :(

PR #1592, opened 2017-06-06
backport: DNM: tox: add ansible2.3

ktdreyer:
Not sure if this is the right way to add this... But I really want us to be testing on both ansible2.2 and ansible2.3.


ktdreyer:
I think this means I need to add Jenkins jobs for ansible2.3 as well? (And they'll fail on stable-2.2 until I've backported this?)

andrewschoen:
> I think this means I need to add Jenkins jobs for ansible2.3 as well? (And they'll fail on stable-2.2 until I've backported this?)

Yeah, you'll need to update https://github.com/ceph/ceph-build/blob/master/ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml#L6. Perhaps it's time to add an ``ansible_version`` factor in that JJB so we can add ``ansible2.2`` and ``ansible2.3`` easily.

ktdreyer:
Thanks. I will merge+backport this PR before making the ceph-build.git change.

guits:
jenkins test kraken-ansible2.2-xenial_cluster

leseb:
@ktdreyer status of this PR?

ktdreyer:
This is uneeded now

PR #1633, opened 2017-06-29
ansible: fail if user selects OSD auto detection and raw devices are …

fullerdj:
…mounted

Signed-off-by: Douglas Fuller <dfuller@redhat.com>

### DIFF
--- a/roles/ceph-osd/tasks/check_devices_auto.yml
+++ b/roles/ceph-osd/tasks/check_devices_auto.yml
@@ -10,6 +10,27 @@
     - ansible_devices is defined
     - item.value.removable == "0"
 
+- name: check if any of the raw partitions are mounted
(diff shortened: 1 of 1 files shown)


roles/ceph-osd/tasks/check_devices_auto.yml
```diff
@@ -10,6 +10,27 @@
     - ansible_devices is defined
     - item.value.removable == "0"
 
+- name: check if any of the raw partitions are mounted
+  shell: "mount |grep '^/dev/{{ item.key }} '"
```
leseb: `s/mount |grep/mount | grep -sq/`?

fullerdj: Sure.

PR #1644, opened 2017-07-03
osd: validate devices variable input

fullerdj:
Fail with a sane message if the devices variable is not an array
during manual device assignment.

Signed-off-by: Douglas Fuller <dfuller@redhat.com>

### DIFF
--- a/roles/ceph-osd/tasks/check_mandatory_vars.yml
+++ b/roles/ceph-osd/tasks/check_mandatory_vars.yml
@@ -55,6 +55,15 @@
     - not osd_auto_discovery
     - devices|length == 0
 
+- name: make sure the devices variable is a list
+  fail:
+    msg: "devices: must be a list, not a string, i.e. [ \"/dev/sda\" ]"
+  when:
+    - osd_group_name is defined
+    - osd_group_name in group_names
+    - not osd_auto_discovery
+    - devices is string
+
 - name: verify journal devices have been provided
   fail:
     msg: "please provide devices and raw journal devices to your osd scenario"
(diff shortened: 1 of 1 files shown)


roles/ceph-osd/tasks/check_devices_static.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- name: make sure the devices variable is a list
+  fail:
+    msg: "devices: must be a list, not a string, i.e. [ \"/dev/sda\" ]"
+  when: devices[0] | length == 1
```
andrewschoen: I think a safer check here might be: ``when: devices is string``.

Could you try that out and see if it works?

leseb: +1 for `when: devices is string`, it works for me.

leseb:
You should also have the same check for `raw_journal_devices`, see https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/check_mandatory_vars.yml#L71

roles/ceph-osd/tasks/check_mandatory_vars.yml
```diff
@@ -66,6 +73,18 @@
...
       or devices|length == 0
 
+- name: make sure the raw_journal_devices variable is a list
+  fail:
+    msg: "raw_journal_devices: must be a list, not a string, i.e. [ \"/dev/sda\" ]"
+  when:
+    - osd_group_name is defined
+    - osd_group_name in group_names
+    - not containerized_deployment
+    - raw_multi_journal
```
leseb: just to make sure the `or` doesn't get messed up, let's move `raw_journal_devices is string` after `raw_multi_journal`.

roles/ceph-osd/tasks/check_mandatory_vars.yml
```diff
@@ -55,6 +55,13 @@
     - not osd_auto_discovery
     - devices|length == 0
 
+- name: make sure the devices variable is a list
+  fail:
+    msg: "devices: must be a list, not a string, i.e. [ \"/dev/sda\" ]"
+  when:
+    - not osd_auto_discovery
```
leseb: This might not be needed but let's add this for consistency:

```
- osd_group_name is defined
- osd_group_name in group_names
```

Issue #1380, opened 2017-03-16
ceph-ansible: "Timeout when waiting for file /etc/ceph/ceph.client.admin.keyring"


gator1:
Hi,

I set up seven KVM centos CentOS-7-x86_64-Minimal-1611 VMs.
They use bridged network on the host with static ip address.

I set up a user cephuser in wheel group but ceph-ansible doesn't like that user.

Now instead I use root. I tried to set up three monitors (mon-node1--3) and three osds (osd-node1-3)

I use this all file to run ansible-playbook site.yml -u root with site.yml copied from sample unchanged. 

I checked mon-node1 /etc/ceph and there is no keyring file. 

Do I need certain utility for this to work? ssh-keygen works. 

Please help. 

[all.txt](https://github.com/ceph/ceph-ansible/files/849284/all.txt)

](url)

TASK [ceph-mon : wait for ceph.client.admin.keyring exists] ********************
fatal: [mon-node1]: FAILED! => {"changed": false, "elapsed": 300, "failed": true, "msg": "Timeout when waiting for file /etc/ceph/ceph.client.admin.keyring"}
fatal: [mon-node3]: FAILED! => {"changed": false, "elapsed": 300, "failed": true, "msg": "Timeout when waiting for file /etc/ceph/ceph.client.admin.keyring"}
fatal: [mon-node2]: FAILED! => {"changed": false, "elapsed": 300, "failed": true, "msg": "Timeout when waiting for file /etc/ceph/ceph.client.admin.keyring"}
	to retry, use: --limit @/root/ceph-ansible/site.retry

PLAY RECAP *********************************************************************
mon-node1                  : ok=41   changed=1    unreachable=0    failed=1   
mon-node2                  : ok=39   changed=1    unreachable=0    failed=1   
mon-node3                  : ok=39   changed=1    unreachable=0    failed=1   
osd-node1                  : ok=2    changed=0    unreachable=0    failed=0   
osd-node2                  : ok=2    changed=0    unreachable=0    failed=0   
osd-node3                  : ok=2    changed=0    unreachable=0    failed=0

gator1:
monitor kind of started but ceph needs the keys on monitors. 

There must be some simple issues. From the deplot node (called mgmt) I can ssh root@mon-node1 without password. Does that mean I did everything regarding to the key?


[root@mon-node1 ceph]# ceph -s
[...]

PR #1654, opened 2017-07-05
Allow ceph-mon systemd overrides to be specified

andymcc:
ceph-mon can fail to start under certain circumstances (for example,
when running in a container) because the default systemd service
configuration causes namespace issues.

To work around this we can override the system service settings by
placing an overrides file in the ceph-mon@.service.d directory. This can
be generic so as to allow any potential changes required to the ceph-mon
service file.

The overrides file is only setup when the "ceph_mon_systemd_overrides"
config_template override variable is specified.

### DIFF
--- a/roles/ceph-mds/defaults/main.yml
+++ b/roles/ceph-mds/defaults/main.yml
@@ -19,3 +19,14 @@ copy_admin_key: false
 
 ceph_mds_docker_extra_env: -e CLUSTER={{ cluster }} -e MDS_NAME={{ ansible_hostname }}
 ceph_config_keys: [] # DON'T TOUCH ME
+
+
+###########
+# SYSTEMD #
+###########
+# ceph_mds_systemd_overrides will override the systemd settings
+# for the ceph-mds services.
+# For example,to set "PrivateDevices=false" you can specify:
+#ceph_mds_systemd_overrides:
+#  Service:
+#    PrivateDevices: False
--- a/roles/ceph-mgr/defaults/main.yml
+++ b/roles/ceph-mgr/defaults/main.yml
@@ -5,3 +5,14 @@
 
 ceph_mgr_docker_extra_env: -e CLUSTER={{ cluster }} -e MGR_NAME={{ ansible_hostname }}
 ceph_config_keys: [] # DON'T TOUCH ME
+
+
+###########
+# SYSTEMD #
+###########
+# ceph_mgr_systemd_overrides will override the systemd settings
+# for the ceph-mgr services.
+# For example,to set "PrivateDevices=false" you can specify:
+#ceph_mgr_systemd_overrides:
+#  Service:
+#    PrivateDevices: False
--- a/roles/ceph-mon/defaults/main.yml
+++ b/roles/ceph-mon/defaults/main.yml
@@ -116,3 +116,14 @@ ceph_mon_docker_extra_env: -e CLUSTER={{ cluster }} -e FSID={{ fsid }} -e MON_NA
 mon_docker_privileged: false
 mon_docker_net_host: true
 ceph_config_keys: [] # DON'T TOUCH ME
+
+
+###########
+# SYSTEMD #
+###########
+# ceph_mon_systemd_overrides will override the systemd settings
+# for the ceph-mon services.
+# For example,to set "PrivateDevices=false" you can specify:
+#ceph_mon_systemd_overrides:
+#  Service:
+#    PrivateDevices: False
--- a/roles/ceph-osd/defaults/main.yml
+++ b/roles/ceph-osd/defaults/main.yml
@@ -238,3 +238,14 @@ ceph_osd_docker_prepare_env: -e CLUSTER={{ cluster }} -e OSD_JOURNAL_SIZE={{ jou
 #
 ceph_osd_docker_extra_env: -e CLUSTER={{ cluster }} -e OSD_JOURNAL_SIZE={{ journal_size }}
 ceph_osd_docker_run_script_path: "/usr/share" # script called by systemd to run the docker command
+
+
+###########
+# SYSTEMD #
+###########
+# ceph_osd_systemd_overrides will override the systemd settings
+# for the ceph-osd services.
+# For example,to set "PrivateDevices=false" you can specify:
+#ceph_osd_systemd_overrides:
+#  Service:
+#    PrivateDevices: False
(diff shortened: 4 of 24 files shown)


leseb:
Don't we have the same problem for other daemons in container?

andymcc:
I imagine you would, yes! Happy to apply to each role in that case, do you think the approach is decent enough, or needs more work?

leseb:
@andymcc I'm not sure, as I don't really understand what the problem is :).
Can you show me an error example?

Thanks!

andymcc:
Sure! So the issue is very similar to: https://github.com/ceph/ceph-ansible/issues/1380

As an example, without specifying "PrivateDevices=false" in the [Service] init file (on CentOS7 at least, the Ubuntu systemd files don't have that set to true) you will get the service failing as follows - because PrivateDevices is defaulted to true in the init file from the package:

~ systemctl status ceph-mon@aio1-ceph-mon-container-04c839c6.service
● ceph-mon@aio1-ceph-mon-container-04c839c6.service - Ceph cluster monitor daemon
   Loaded: loaded (/usr/lib/systemd/system/ceph-mon@.service; enabled; vendor preset: disabled)
   Active: activating (auto-restart) (Result: exit-code) since Wed 2017-07-05 14:52:25 UTC; 5s ago
  Process: 7833 ExecStart=/usr/bin/ceph-mon -f --cluster ${CLUSTER} --id %i --setuser ceph --setgroup ceph (code=exited, status=226/NAMESPACE)
 Main PID: 7833 (code=exited, status=226/NAMESPACE)

Once we add the systemd override file inside /etc/systemd/system/ceph-mon@.service.d/*.conf
with PrivateDevices=false - it overrides the value in the actual systemd script for the service and so it starts normally:

~ systemctl status ceph-mon@aio1-ceph-mon-container-04c839c6.service
● ceph-mon@aio1-ceph-mon-container-04c839c6.service - Ceph cluster monitor daemon
   Loaded: loaded (/usr/lib/systemd/system/ceph-mon@.service; enabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/ceph-mon@.service.d
           └─test.conf
   Active: active (running) since Wed 2017-07-05 14:55:18 UTC; 1s ago
 Main PID: 7870 (ceph-mon)
   CGroup: /lxc/aio1_ceph-mon_container-04c839c6/system.slice/system-ceph\x2dmon.slice/ceph-mon@aio1-ceph-mon-container-04c839c6.service
           └─7870 /usr/bin/ceph-mon -f --cluster ceph --id aio1-ceph-mon-container-04c839c6 --setuser ceph --setgroup ceph

Jul 05 14:55:18 aio1-ceph-mon-container-04c839c6 systemd[1]: Started Ceph cluster monitor daemon.
Jul 05 14:55:18 aio1-ceph-mon-container-04c839c6 systemd[1]: Starting Ceph cluster monitor daemon...
Jul 05 14:55:18 aio1-ceph-mon-container-04c839c6 ceph-mon[7870]: starting mon.aio1-ceph-mon-container-04c839c6 rank 0 at 172.29.238.125:6789/0 mon_data /var/lib/ceph/mon/ceph-aio1-cep...71435da7042
Hint: Some lines were ellipsized, use -l to show in full.

=================

TL;DR I think making it generic such that you can override the service's systemd settings if you want (and in a generic way) would mean that this solves more than just 1 use case.

I haven't deployed osds or other services inside a container but I'm guessing it would impact the other services too. The example in the defaults/main.yml is what I would use to resolve the ceph-mon inside the container issue that we're seeing.

leseb:
jenkins test this please

leseb:
@andymcc sorry for late reply, I understand the issue, it looks like you are running your deployment on LXC containers. It's clearer now. Would you mind working on the support for others daemons too?

Thanks!

andymcc:
Will fix up for the other daemons and resubmit!

leseb:
@andymcc sorry I missed your last update, looking into this.

roles/ceph-mds/defaults/main.yml
```diff
@@ -19,3 +19,13 @@ copy_admin_key: false
 
 ceph_mds_docker_extra_env: -e CLUSTER={{ cluster }} -e MDS_NAME={{ ansible_hostname }}
 ceph_config_keys: [] # DON'T TOUCH ME
+
```
leseb: Please add one more line, just like you did below.

roles/ceph-mgr/defaults/main.yml
```diff
@@ -5,3 +5,13 @@
 
 ceph_mgr_docker_extra_env: -e CLUSTER={{ cluster }} -e MGR_NAME={{ ansible_hostname }}
 ceph_config_keys: [] # DON'T TOUCH ME
+
```
leseb: ditto

roles/ceph-mon/defaults/main.yml
```diff
@@ -120,3 +120,13 @@ ceph_mon_docker_extra_env: -e CLUSTER={{ cluster }} -e FSID={{ fsid }} -e MON_NA
 mon_docker_privileged: false
 mon_docker_net_host: true
 ceph_config_keys: [] # DON'T TOUCH ME
+
```
leseb: ditto

roles/ceph-osd/defaults/main.yml
```diff
@@ -227,3 +227,13 @@ ceph_osd_docker_prepare_env: -e CLUSTER={{ cluster }} -e OSD_JOURNAL_SIZE={{ jou
 #
 ceph_osd_docker_extra_env: -e CLUSTER={{ cluster }} -e OSD_JOURNAL_SIZE={{ journal_size }}
 ceph_osd_docker_run_script_path: "/usr/share" # script called by systemd to run the docker command
+
```
leseb: ditto

roles/ceph-rgw/defaults/main.yml
```diff
@@ -51,3 +51,13 @@ ceph_rgw_civetweb_port: "{{ radosgw_civetweb_port }}"
 ceph_rgw_docker_extra_env: -e CLUSTER={{ cluster }} -e RGW_CIVETWEB_PORT={{ ceph_rgw_civetweb_port }}
 ceph_config_keys: [] # DON'T TOUCH ME
 rgw_config_keys: "/" # DON'T TOUCH ME
+
```
leseb: ditto

andymcc: fixing now - sorry i got sidetracked and forgot about this :)

andymcc:
Should be fixed now - and rebased on the change to move out the starting of the ceph-osd services.

leseb:
jenkins test this please

leseb:
CI failures due to https://github.com/ceph/ceph-docker/pull/754. Safe to merge.

Issue #1688, opened 2017-07-17
Keys created by "create openstack key(s)" task are not readable by openstack users nova, cinder, etc.


fultonj:
Keys created by [ceph-ansible/roles/ceph-mon/tasks/openstack_config.yml](https://github.com/ceph/ceph-ansible/blob/adf752f2419e91fc5da269e052432b4d0d8e76dc/roles/ceph-mon/tasks/openstack_config.yml#L11-L17) are not readable by the OpenStack users who need to use them when using ceph-ansible to configure the keys for OpenStack clients that will run on the same servers. 

For example, the cinder log gets the following error: 

`2017-07-17 17:00:11.086 46300 ERROR cinder.utils [req-6601e808-6e19-4631-aa9b-43bee8735640 - - - - -] Volume driver RBDDriver not initialized`

but `chmod o+r ceph.client.openstack.keyring` resolves the issue and restarting cinder after runing just that command shows: 

`2017-07-17 17:41:07.862 71387 INFO cinder.volume.manager [req-ec35b612-970f-4c6e-aca0-270c6d5a9be0 - - - - -] Starting volume driver RBDDriver (1.2.0)`

PR #1689, opened 2017-07-18
Allow user to define ACLs for OpenStack keys

fultonj:
The openstack_keys structure now supports a key called acls whose value is a list of strings that one could pass to setfacl. The ansible ACL module applies the ACLs to all openstack keys with this property.

Fixes: #1688


fultonj:
problem @guits @gfidente this change, tested on tripleo, works for the clients, e.g. nova [1], but not on the monitor/controller nodes [2]. I realize I dropped this change into the client config, not the mon config, but because in tripleo the controller node is a client in that it runs cinder and glance by the default, shouldn't the client code have changed the controller node (which happens to be the monitor node)? 

Do we change this in how TripleO calls ceph-ansible or in ceph-ansible itself? 

[1] 
```
[root@overcloud-novacompute-0 ~]# getfacl /etc/ceph/ceph.client.openstack.keyring
getfacl: Removing leading '/' from absolute path names
# file: etc/ceph/ceph.client.openstack.keyring
# owner: root
# group: root
user::rw-
user:glance:r--
user:nova:r--
user:cinder:r--
user:gnocchi:r--
group::---
mask::r--
other::---

[root@overcloud-novacompute-0 ~]# 
```

[2] 
```
[root@overcloud-controller-0 ~]# getfacl /etc/ceph/ceph.client.openstack.keyring
getfacl: Removing leading '/' from absolute path names
# file: etc/ceph/ceph.client.openstack.keyring
# owner: root
# group: root
user::rw-
group::---
other::---

[root@overcloud-controller-0 ~]# 
```

gfidente:
I think we just need to make the same changes into roles/ceph-mon/tasks/openstack_config.yml

roles/ceph-client/tasks/create_users_keys.yml
```diff
@@ -1,14 +1,44 @@
...
+
+- name: add key(s) to ceph
+  command: "ceph --cluster {{ cluster }} auth import -i /etc/ceph/{{ cluster }}.{{ item.0.name }}.keyring"
+  changed_when: false
+  with_together:
+    - "{{ keys }}"
+    - "{{ keys_exist.results }}"
+  when: item.1.rc != 0
+
+- name: setfacl for openstack keys
```
gfidente: this can probably go into ceph-common and be consumed by both

fultonj:
@gfidente I was able to get this working as is by adding `- OS::TripleO::Services::CephClient` to roles_data.yml under Controller. It makes sense to me because the controller hosts the ceph clients for glance and cinder by default. 

```
(undercloud) [stack@undercloud upgrade]$ ansible clients -b -m shell -a "hostname; getfacl /etc/ceph/ceph.client.openstack.keyring"
192.168.24.18 | SUCCESS | rc=0 >>
overcloud-novacompute-0
# file: etc/ceph/ceph.client.openstack.keyring
# owner: root
# group: root
user::rw-
user:glance:r--
user:nova:r--
user:cinder:r--
user:gnocchi:r--
group::---
mask::r--
other::---getfacl: Removing leading '/' from absolute path names

192.168.24.9 | SUCCESS | rc=0 >>
overcloud-controller-0
# file: etc/ceph/ceph.client.openstack.keyring
# owner: root
# group: root
user::rw-
user:glance:r--
user:nova:r--
user:cinder:r--
user:gnocchi:r--
group::---
mask::r--
other::---getfacl: Removing leading '/' from absolute path names

(undercloud) [stack@undercloud upgrade]$ 
```

gfidente:
@guits so I think @fultonj is suggesting to set the same ip as inventory host for multiple roles to have the code only in the client task, I am waiting for directions, it it sounds valid approach in ceph-ansible we can make changes in tripleo to behave appropriately

fultonj:
Actually we can't make this change in TripleO to tht/roles_data.yaml because it breaks puppet-ceph [1] and we want to have both working ceph-ansible (containers) and puppet-ceph (non-containerized ceph) in Pike. Thus, I will update this PR. 

[1] http://paste.openstack.org/show/615760/

guits:
needs to be properly rebased on master

fultonj:
see https://github.com/ceph/ceph-ansible/pull/1696

PR #1699, opened 2017-07-20
Make ceph-fetch-keys depend on ceph-common (for cluster_uuid)

### LINKED ISSUE #1698: ceph-fetch-keys missing dependency on ceph-common
Hi,

A number of the infrastructure playbooks include the ceph-fetch-keys role; this role, however, will fail (complaining about missing `cluster_uuid.stdout`) in the common case where fsid is set to `"{{ cluster_uuid.stdout }}"`; it needs to depend upon the `ceph-common` role, where that variable is set (having extracted the fsid from the running cluster via the "is ceph running already?" task).

I'll open a PR shortly with the (pretty simple) patch, but it seemed worth opening an issue separately, in case I'm missing something obvious.

Example playbook to repro:
```yaml
---
- hosts: mons
  become: True
  roles:
     - ceph-fetch-keys
```

Output from running this:
mv3@deskpro108655:~/software/ceph-ansible$ ansible-playbook --ask-vault-pass -i inventory.py infrastructure-playbooks/keysbug.yml 
Vault password: 

PLAY [mons] ********************************************************************

TASK [Gathering Facts] *********************************************************
ok: [sto-1-1]
ok: [sto-3-1]
ok: [sto-2-1]

TASK [ceph-fetch-keys : find ceph keys] ****************************************
[DEPRECATION WARNING]: always_run is deprecated. Use check_mode = no instead..

This feature will be removed in version 2.4. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: always_run is deprecated. Use check_mode = no instead..

This feature will be removed in version 2.4. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: always_run is deprecated. Use check_mode = no instead..

This feature will be removed in version 2.4. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
ok: [sto-1-1]
ok: [sto-3-1]
ok: [sto-2-1]

TASK [ceph-fetch-keys : set keys permissions] **********************************
changed: [sto-1-1] => (item=/etc/ceph/ceph.client.admin.keyring)
changed: [sto-1-1] => (item=/etc/ceph/ceph.client.irods-test.keyring)
changed: [sto-1-1] => (item=/etc/ceph/ceph.client.openstack-delta2.keyring)
changed: [sto-1-1] => (item=/etc/ceph/ceph.client.openstack-delta.keyring)
changed: [sto-1-1] => (item=/etc/ceph/ceph.client.openstack-demo.keyring)
changed: [sto-1-1] => (item=/etc/ceph/ceph.client.openstack-epsilon.keyring)
changed: [sto-1-1] => (item=/etc/ceph/ceph.client.openstack.keyring)
changed: [sto-1-1] => (item=/etc/ceph/ceph.openstack.client.keyring)
changed: [sto-2-1] => (item=/etc/ceph/ceph.client.admin.keyring)
changed: [sto-2-1] => (item=/etc/ceph/ceph.client.irods-test.keyring)
changed: [sto-2-1] => (item=/etc/ceph/ceph.client.openstack-delta2.keyring)
changed: [sto-2-1] => (item=/etc/ceph/ceph.client.openstack-delta.keyring)
changed: [sto-2-1] => (item=/etc/ceph/ceph.client.openstack-demo.keyring)
changed: [sto-2-1] => (item=/etc/ceph/ceph.client.openstack-epsilon.keyring)
changed: [sto-2-1] => (item=/etc/ceph/ceph.client.openstack.keyring)
changed: [sto-2-1] => (item=/etc/ceph/ceph.openstack.client.keyring)
changed: [sto-3-1] => (item=/etc/ceph/ceph.client.admin.keyring)
changed: [sto-3-1] => (item=/etc/ceph/ceph.client.irods-test.keyring)
changed: [sto-3-1] => (item=/etc/ceph/ceph.client.openstack-delta2.keyring)
changed: [sto-3-1] => (item=/etc/ceph/ceph.client.openstack-delta.keyring)
changed: [sto-3-1] => (item=/etc/ceph/ceph.client.openstack-demo.keyring)
changed: [sto-3-1] => (item=/etc/ceph/ceph.client.openstack-epsilon.keyring)
changed: [sto-3-1] => (item=/etc/ceph/ceph.client.openstack.keyring)
changed: [sto-3-1] => (item=/etc/ceph/ceph.openstack.client.keyring)

TASK [ceph-fetch-keys : copy keys to the ansible server] ***********************
fatal: [sto-1-1]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: {{ cluster_uuid.stdout }}: 'cluster_uuid' is undefined\n\nThe error appears to have been in '/nfs/users/nfs_m/mv3/software/ceph-ansible/roles/ceph-fetch-keys/tasks/main.yml': line 17, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: copy keys to the ansible server\n  ^ here\n"}

NO MORE HOSTS LEFT *************************************************************
        to retry, use: --limit @/nfs/users/nfs_m/mv3/software/ceph-ansible/infrastructure-playbooks/keysbug.retry

PLAY RECAP *********************************************************************
sto-1-1                    : ok=3    changed=1    unreachable=0    failed=1   
sto-2-1                    : ok=3    changed=1    unreachable=0    failed=0   
sto-3-1                    : ok=3    changed=1    unreachable=0    failed=0

mcv21:
A typical deployment of ceph-ansible has fsid set to "{{
cluster_uuid.stdout }}". The ceph-fetch-keys role needs fsid defined
(it stores keys in "{{ fetch_directory }}/{{ fsid }}/{{ item }}"), so
should depend upon this role.

closes #1698

### DIFF
--- a/roles/ceph-fetch-keys/meta/main.yml
+++ b/roles/ceph-fetch-keys/meta/main.yml
@@ -13,4 +13,6 @@ galaxy_info:
         - 7
   categories:
     - system
-dependencies: []
+dependencies:
+  - { role: ceph.ceph-common, when: not containerized_deployment }
+  - { role: ceph.ceph-docker-common, when: containerized_deployment }


guits:
I'm not sure whether we want to play all the task from `ceph-common`|`ceph-docker-common` when running `ceph-fetch-keys`

@andrewschoen @leseb ?

guits:
Perhaps adding a task to get the current `fsid` from `ceph.conf` and overriding the variable `fsid` to the retrieved value could be enough?

mcv21:
ceph-common has a bunch of code to work out what `fsid` should be; it would be a shame to duplicate it. If ceph-common is felt to be too heavy-weight, I guess you'd have to split the fsid-generating bits out of it into another role and have both ceph-common and ceph-fetch-keys depend on that.

guits:
@mcv21 I would say I'm -1 on this.

the easiest way is to set manually the `fsid` variable in `group_vars/` or just passing it with `--extra-vars`. I don't see the benefits running all tasks in `ceph-common` only to set one variable while you could simply define it manually.
In the worse case, we can add a task to retrieve it either via `ceph.conf` or maybe with `ceph fsid` command

mcv21:
A couple of thoughts:

Firstly, two of the three existing infrastructure-playbooks already end up pulling in ceph-common later in their execution (and the third is the taking-over-a-cluster one which is a special case) - 
osd-configure later uses ceph-osd (which depends on ceph-common)
rgw-standalone later uses ceph-rgw (which depends on ceph-common)

So adding this dependency isn't going to suddenly make those playbooks much heavier. I'd also argue that most uses of ceph-fetch-keys are going to be followed by running one or more of the other ceph-foo roles, most of which depend on ceph-common already.

Secondly, it seems to me unhelpful to require different handling of `fsid` for this role than for the other bits of ceph-ansible; setting it as cluster_uuid.stdout is going to be a common deployment option, so making this role require special-casing violates the principal of least surprise, and makes it less useful for administrators.

Thirdly, even if you disregard both of the above, the correct thing is surely to separate out the code from ceph-common that arranges for fsid to be set into another role and have ceph-common and ceph-fetch-keys depend on that, so there's still only one consistent mechanism?

andrewschoen:
I agree with @guits here.

@mcv21 if you add ``ceph-common`` as a dependency of ``ceph-fetch-keys`` and then run another role like ``ceph-osd`` as in the case of ``osd-configure.yml`` then you'd end up running the ``ceph-common`` role twice which you definitely don't want.

mcv21:
OOI, why do you say you definitely don't want to run ceph-common more than once? Isn't the point that it should be idempotent?

When we run ceph-ansible on our cluster, the ceph-common roles gets run many times (once for the mons, once for the osds, once for the rgs - meaning that hosts which are osds+mons or osds+rgws have it run against them several times).

guits:
the fact that the two infrastructure-playbook you mentioned depend on ceph-common is normal since they aim to deploy osd(s) or rgw(s), so they need to run ceph-common, I would even say this is mandatory.
In your case you would want to re-run something which is absolutely not necessary.

andrewschoen:
> OOI, why do you say you definitely don't want to run ceph-common more than once? Isn't the point that it should be idempotent?

On larger clusters this can extend the playbook execution by quite a bit. Yes, the ``ceph-common`` role is idempotent and currently if you have a single node host two different daemons if will run multiple times. I don't think that's a good behavior either. We've actually had requests to remove the other roles dependency on ``ceph-common`` entirely and have it put explicitly in our example playbooks. This would allow for some flexibility for people that host multiple daemons on a single node.

The ``fsid`` handling tasks could be put into another role, yes.

PR #1703, opened 2017-07-24
fix: add partprobe after ceph-disk prepare

yanyixing:
In my scenario, I used particular disk partitions for ceph devices, and jraw_journal_devices.
And ceph-disk activate always error, 
mkjournal error creating journal on /var/lib/ceph/tmp/mnt.n6qlKY/journal:  Permission denied
If I run partprobe before ceph-disk activate, it works well

### DIFF
--- a/roles/ceph-osd/tasks/scenarios/raw_multi_journal.yml
+++ b/roles/ceph-osd/tasks/scenarios/raw_multi_journal.yml
@@ -40,4 +40,7 @@
     - osd_objectstore == 'bluestore'
     - not osd_auto_discovery
 
+- name: partprobe
+  command: partprobe
+  
 - include: ../activate_osds.yml


leseb:
I guess the problem is that your partitions have wrong permissions, when you call partprobe, a udev event is fired which triggers the ceph udev rules that apply the right permissions on the device.
Can't you just get the right permissions from the get go?

yanyixing:
My partitions are as blew
```
brw-rw---- 1 root disk 8, 16 Jul 24 22:26 /dev/sdb
brw-rw---- 1 root disk 8, 17 Jul 24 22:26 /dev/sdb1
brw-rw---- 1 root disk 8, 18 Jul 24 22:26 /dev/sdb2

ceph-disk prepare /dev/sdb2 /dev/sdb1
ceph-disk activate /dev/sdb2 

the second command error by the Permission denied problem

what right permission should be ?

leseb:
The devices should be owned by `ceph:ceph`.

yanyixing:
when I partitioned this disk , ceph is not installed , so there is no ceph user

yanyixing:
@leseb any suggestions?

leseb:
@yanyixing you can then use `64045` for Ubuntu and `167` for CentOS.

leseb:
My advice should work :).
Feel free to re-open if you need anything.

Issue #1712, opened 2017-07-25
ceph-ansible broken: Error initializing cluster client


tmuthamizhan:
looks like ceph-ansible is broken and failing with error "Error initializing cluster client: Error('error calling conf_read_file: error code 22',)"

latest logs: http://pulpito.front.sepia.ceph.com/teuthology-2017-07-25_05:15:47-ceph-ansible-kraken-distro-basic-vps/

leseb:
The problem is that you used ceph_dev on a kraken branch, then because of this https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/release.yml#L2-L4 ceph_release was set to kraken because the current default of ceph_stable_release is kraken. However, because of ceph_dev this https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/release.yml#L6-L12 overrode ceph_release.
Fix here: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/release.yml#L6-L12

From now on please, always set on which Ceph release you are working on, even with ceph_dev.

vasukulkarni:
I am reopening to get different solution to this, It is hard to handle this in teuthology, since a ceph_dev_branch can point to something like "wip-testing-2" , at that stage without installing its not possible to detect what is ceph_stable_release, since ceph-ansible handles the install its knows wheter its 10.x, 11.x or 12.x for various release flag.

andrewschoen:
@vasukulkarni I disagree, I think making ``ceph_stable_release`` mandatory is a workable solution. The problem is that we sometimes need to make decisions in ceph-ansible before install for differing versions of ceph, this is why it's nice to know what version you're expecting to install before we actually install the packages.

I recently was testing the ``wip-volume`` branch of ceph and I knew it was based of luminous so I just set ``ceph_stable_release: luminous`` as well. Why couldn't teuthology or the user initiating the teuthology test do the same?

vasukulkarni:
@andrewschoen Its just not only me running the tests, there could be other users who are not aware what to set in overrides when running tests, since many things are handled inside the task.  Something thats gets set automatically inside ceph-ansible is a good solution. Is it not possible to handle this in ceph-ansible now?

tmuthamizhan:
@vasukulkarni, we can very well set it along with ceph_dev_branch in
ceph-ansible task.

leseb:
[...]

PR #1713, opened 2017-07-26
common: make ceph_stable_release mandatory

leseb:
Fix https://github.com/ceph/ceph-ansible/issues/1712

### DIFF
--- a/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
+++ b/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
@@ -22,6 +22,24 @@
   tags:
     - package-install
 
+- name: make sure ceph_stable_release is set
+  fail:
+    msg: "make sure ceph_stable_release is set to a release name (e.g: luminous), http://docs.ceph.com/docs/master/release-notes/"
+  when:
+    - ceph_stable_release == 'dummy'
+    - not ceph_rhcs
+  tags:
+    - package-install
+
+- name: make sure ceph_stable_release is correct
+  fail:
+    msg: "wrong release name, see http://docs.ceph.com/docs/master/release-notes/"
+  when:
+    - ceph_stable_release not in ceph_release_num
+    - not ceph_rhcs
+  tags:
+    - package-install
+
 - name: verify that a method was chosen for red hat storage
   fail:
     msg: "choose between ceph_rhcs_cdn_install and ceph_rhcs_iso_install"
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -105,7 +105,7 @@ dummy:
 #ceph_stable: false # use ceph stable branch
 #ceph_mirror: http://download.ceph.com
 #ceph_stable_key: https://download.ceph.com/keys/release.asc
-#ceph_stable_release: kraken # ceph stable release
+#ceph_stable_release: dummy
 #ceph_stable_repo: "{{ ceph_mirror }}/debian-{{ ceph_stable_release }}"
 
 ######################################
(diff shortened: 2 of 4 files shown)


leseb:
jenkins test luminous-ansible2.2-purge_cluster

leseb:
jenkins test luminous-ansible2.2-purge_dmcrypt

leseb:

leseb:

leseb:
jenkins test luminous-ansible2.2-bluestore_journal_collocation

roles/ceph-common/tasks/checks/check_mandatory_vars.yml
```diff
@@ -22,6 +22,14 @@
   tags:
     - package-install
 
+- name: make sure ceph_stable_release is set
+  fail:
+    msg: "make sure ceph_stable_release is set to a release name, http://docs.ceph.com/docs/jewel/releases/"
+  when:
+    - ceph_stable_release == 'dummy'
```
alfredodeza: this looks naive enough that a typo in a release would go through.  Having an up to date set of current releases may not be something you want to maintain though.

What is the difference between this and just making sure it is not undefined?

leseb: Is it better this way?

leseb: It's easier for the check to have it defined and also for the users so they know the variable exists and where to find/edit it.

roles/ceph-common/tasks/checks/check_mandatory_vars.yml
```diff
@@ -22,6 +22,21 @@
...
+  tags:
+    - package-install
+
+- name: make sure ceph_stable_release is correct
+  fail:
+    msg: "wrong release name, see http://docs.ceph.com/docs/jewel/releases/"
```
alfredodeza: you may want to use 'master' here no?

roles/ceph-common/tasks/checks/check_mandatory_vars.yml
alfredodeza: I don't think you want to use a 'jewel' url.

roles/ceph-common/tasks/checks/check_mandatory_vars.yml
```diff
@@ -22,6 +22,21 @@
   tags:
     - package-install
 
+- name: make sure ceph_stable_release is set
+  fail:
+    msg: "make sure ceph_stable_release is set to a release name, http://docs.ceph.com/docs/jewel/releases/"
```
andrewschoen: Maybe we should just list which releases we support here instead of linking to the docs? Which I think is jewel, kraken and now luminous.

Those docs don't mention kraken or luminous anyway.

roles/ceph-common/tasks/checks/check_mandatory_vars.yml
andrewschoen: If you're wanting to test the ``master`` branch that would be set like ``ceph_dev_branch: master``. The ``ceph_stable_release`` var would still need to be set to the codename of the release that master is currently at.

roles/ceph-common/tasks/checks/check_mandatory_vars.yml
leseb: Yes but I was trying to avoid re-editing the code for every new release...
I was hoping the official doc could be the source of truth... It looks like it's not the case

roles/ceph-common/tasks/checks/check_mandatory_vars.yml
leseb: Yes.

leseb:
jenkins test luminous-ansible2.2-bluestore_cluster

vasukulkarni:
@leseb it seems redundant to me to add ceph_stable_release as well to config, why can't this be auto detected and handled based on ceph-dev?

tmuthamizhan:
@vasukulkarni +1

Issue #1142, opened 2016-12-02
change systemd activation and units


leseb:
see: https://github.com/ceph/ceph/pull/12241

it seems that we should not be activating the resources ceph-osd@$id but only keep it ephemeral til reboot.

PR #1715, opened 2017-07-26
osd: do not enable osd@id unit file

leseb:
ceph-disk is responsable for enabling the unit file if needed. Actually
since https://github.com/ceph/ceph/pull/12241 it seems that it's not
even needed. On an event of a restart, udev rules will be trigger and
they will ceph-disk activate the device too so the 'enabled' is not
needed.

Closes: https://github.com/ceph/ceph-ansible/issues/1142

### DIFF
--- a/roles/ceph-osd/tasks/activate_osds.yml
+++ b/roles/ceph-osd/tasks/activate_osds.yml
@@ -95,6 +95,5 @@
   service:
     name: ceph-osd@{{ item }}
     state: started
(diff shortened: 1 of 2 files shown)


alfredodeza:
@leseb you need to remove the test that makes sure it is enabled too:

https://github.com/ceph/ceph-ansible/blob/master/tests/functional/tests/osd/test_osds.py#L25

PR #1718, opened 2017-07-27
change the partition's ownership

yanyixing:
if osd using a partition, and the disk has partitioned manually,  the partition's ownership should be changed to ceph:ceph. 
Otherwise ceph-disk activate will be failed for something permission denied error


alfredodeza:
This sounds like a ceph-disk issue though. ceph-disk already sets permissions always when dealing with devices

yanyixing:
if ceph-disk create the partition, it will invoking the partprobe command. In this scenario, partitions are created manaually , so I change the ownership.

PR #1735, opened 2017-08-02
common: override and autodetect ceph_stable_release

leseb:
For ceph_dev and rhcs installation we need to detect the release since
we do not declare it explicitly. Keeping the default ceph_stable_release
could lead to several things going wrong and some have already been
reported.

Fixes: https://github.com/ceph/ceph-ansible/issues/1712 and https://bugzilla.redhat.com/show_bug.cgi?id=1476210

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -74,6 +74,12 @@
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
   static: False
 
+# override ceph_stable_release for ceph_dev and rhcs installations since ceph_stable_release is not mandatory
+- include: ./release-rhcs.yml
+  when: (ceph_rhcs or ceph_dev)
+  tags:
+    - always
+
 - include: ./misc/ntp_redhat.yml
   when:
     - ansible_os_family == 'RedHat'
--- a/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
+++ b/roles/ceph-common/tasks/checks/check_mandatory_vars.yml
@@ -27,7 +27,8 @@
     msg: "make sure ceph_stable_release is set to a release name (e.g: luminous), http://docs.ceph.com/docs/master/release-notes/"
   when:
     - ceph_stable_release == 'dummy'
-    - not ceph_rhcs
+    - ceph_origin == 'upstream'
+    - ceph_stable
   tags:
     - package-install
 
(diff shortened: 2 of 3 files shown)


roles/ceph-common/tasks/main.yml
```diff
@@ -79,6 +79,12 @@
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
   static: False
 
+# override ceph_stable_release for ceph_dev and rhcs installations since ceph_stable_release is not mandatory
+- include: ./release-rhcs.yml
+  when: (rhcs or ceph_dev)
```
andrewschoen: I think this needs to be ``ceph_rhcs``.

leseb: Done, thanks!

leseb:
@andrewschoen about the `site.yml.sample`, this is tough to handle. IMHO this is the responsibility of the deployer to configure the `site.yml` according to its need. I'm not even sure if this condition works as expected right now. Is this a real blocker for this PR?

andrewschoen:
@leseb the condition does work as expected now, otherwise none of our luminous or kraken CI tests would be working.

I suppose since this isn't removing the code that makes setting ``ceph_stable_release`` mandatory it'd still work. This just means that users of RHCS 3.0 will need to set ``ceph_stable_release: luminous`` or otherwise modify ``site.yml.sample`` which are things they don't currently do.

I do agree that ``ceph_stable_release`` is a bit clunky to use. Eventually I think it'd be nice to see it go away and replaced with config like:

```
install_method: upstream
install_version: jewel
```

or

```
install_method: rhcs
install_version: 3
```

leseb:
@andrewschoen I agree we need to simplify the installation process or at least clarify/ease the user experience.

I disagree, this is the purpose of this patch to detect which Ceph version has been deployed, so RHCS users will benefit from that too. Am I missing something from your last comment?

andrewschoen:
@leseb no, I think this is fine for RHCS users. I was just pointing out that they'll need to manually set ``ceph_stable_release: luminous`` when deploying rhcs 3.0, which is a new step.

Initially I thought this PR was about removing the mandatory requirement of setting``ceph_stable_release`` but it's not.

leseb:
@andrewschoen it's not for the community installation, so basically:

*  `ceph_stable_release` is mandatory for the upstream install
*  `ceph_stable_release` is **not** mandatory for rhcs and ceph_dev as it's auto-discovered after package installation

andrewschoen:
@leseb Ok, I think I understand now. We'll just leave it up to users to modify ``site.yml.sample`` if they wish to deploy mgr nodes when using RHCS.

vasukulkarni:
👍

leseb:
jenkins test this please

leseb:
jenkins test this please

Issue #1737, opened 2017-08-02
ceph-ansible: failure in ceph-mon playbook


tmuthamizhan:
ceph branch: jewel

ceph-ansible fails at "generate monitor initial keyring" in ceph-mon playbook with the following error "The conditional check 'not containerized_deployment' failed. The error was: error while evaluating conditional (not containerized_deployment): 'containerized_deployment' is undefined"

latest logs can be found at - http://qa-proxy.ceph.com/teuthology/teuthology-2017-08-02_04:15:01-ceph-ansible-jewel-distro-basic-vps/1471779/teuthology.log

leseb:
What's the content of `cephlab.yml`? I suspect you need to update it to reflect our recent changes with the introduction of the `ceph-defaults` role, see: https://github.com/ceph/ceph-ansible/blob/master/site.yml.sample

tmuthamizhan:
@leseb, can you please merge the fix so we can proceed with testing?

guits:
@tmuthamizhan, unless I missed something, I think there is a misunderstanding here? There is no fix to push, you just need to include the `ceph-defaults` role in your `cephlab.yml`, indeed, we merged some days ago in master some commits which introduced a new role `ceph-defaults` that needs to be called before playing `ceph-common`, this role handles all default variables and sets all many facts that which are all consumed by `ceph-common`.

tmuthamizhan:
@guits, where can i find cephlab.yml?
I believe this is a ceph-ansible change and thought you may simply add the new role ceph-defaults before ceph-common.

please correct me if am wrong

andrewschoen:
@tmuthamizhan you need to explicitly define ``ceph-defaults`` and ``ceph-common`` roles now. It looks like teuthology does that here https://github.com/ceph/teuthology/blob/master/teuthology/task/ceph_ansible.py#L32

Our ``site.yml.sample`` playbook also does this now: https://github.com/ceph/ceph-ansible/blob/master/site.yml.sample#L44

tmuthamizhan:
@andrewschoen , thank you! it works fine and am stuck with a new issue - https://github.com/ceph/ceph-ansible/issues/1752

PR #1740, opened 2017-08-03
Add ceph-defaults to galaxy role breakout script

Logan2211:
The `ceph/ansible-ceph-defaults` repo will need to be created.

This is needed because downstream projects that consume ceph-ansible, such as openstack-ansible, are unable to do so due to #1737 until there is a ceph-defaults breakout role created.


leseb:
@Logan2211 good catch thanks!

Logan2211:
@leseb don't forget https://github.com/ceph/ansible-ceph-defaults needs to be created and populated now ;) eager to get this done so we can get https://review.openstack.org/#/c/490476/ moving forward

leseb:
@Logan2211 done! https://github.com/ceph/ansible-ceph-defaults
And it's in the galaxy too!

Logan2211:
Thank you!

PR #1743, opened 2017-08-03
generate_group_vars_sample: rhcs.yml.sample

leseb:
Add a pre-configured bare minimum rhcs.yml.sample that can save
configuration steps.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1478071


ktdreyer:
The "generate group vars" thing is getting more complicated, but sure.

Good call on needs-docs.

ktdreyer:
I must be missing the point of the "needs-docs" label

leseb:
@ktdreyer the idea is once we branch stable-2.3 and start doing QE we will review all the PR labelled like this and discuss with the doc team accordingly.

ktdreyer:
@leseb why not write the docs as part of the PR?

leseb:
We have some portion of it in the commits, that's why we will revisit it with the doc team.

ktdreyer:
I should clarify - why not update https://github.com/ceph/ceph-ansible/tree/master/docs in each PR?

leseb:
Because I forgot and time is often limited... But yes I agree, this is not optimal.

PR #1746, opened 2017-08-04
defaults: fail if can not write ansible logs

leseb:

### DIFF
--- a/roles/ceph-defaults/tasks/ansible_logs.yml
+++ b/roles/ceph-defaults/tasks/ansible_logs.yml
@@ -1,0 +1,23 @@
+---
+- name: test if we have a local ansible.cfg file
+  local_action: stat path=ansible.cfg
+  register: ansible_cfg
+
+- name: lookup the value of log_path in ansible.cfg
+  local_action: shell awk -F "=" '/log_path/ {print $2}' ansible.cfg
+  register: ansible_log_location
+  when: ansible_cfg.stat.exists == True
+
+- name: test if ansible_cfg.stdout can be written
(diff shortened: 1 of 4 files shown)


leseb:
The entire fails because the patch works :p, we now need to decide if we make this optional or not...

guits:
why would we want to make this test mandatory?

leseb:
@guits this still needs to be discussed but we have been having many users reporting issues and not able to provide enough logs. So this might be useful. This is still DNM for me as we need to discuss this.

roles/ceph-defaults/tasks/ansible_logs.yml
```diff
@@ -0,0 +1,26 @@
...
+  changed_when: false
+  failed_when: false
+  register: mktemp
+  when: ansible_cfg.stat.exists == True
+
+- name: fail if log_path directory from ansible.cfg is not writable
+  fail:
+    msg: "make sure log_path directory from ansible.cfg is writable by the user running ansible"
+  when:
+    - mktemp.stdout != '0'
```
guits: shouldn't it be `mktemp.rc != 0` ?

roles/ceph-defaults/tasks/ansible_logs.yml
```diff
@@ -0,0 +1,26 @@
...
+    - ansible_cfg.stat.exists == True
```
guits: shouldn't we stat for `{{ ansible_log_location }}` and check for `stat_ansible_log_location.stat.exists == False` instead ?

guits:
jenkins test pipeline

guits:
jenkins test pipeline

guits:
jenkins test this please

Issue #1755, opened 2017-08-09
Allow user to specify the mode of the openstack keys to be created


fultonj:
The ACL feature for OpenStack [1] does not work for OpenStack services that run in Kolla containers (whose docker images also have a different UID). Please allow the user to use the same interface to specify a file mode too as found in puppet-ceph [2]. This is not a request to hard code a permission but to allow a user to pass a permission that they choose as a parmeter that ceph-ansible will then set. 

[1] https://github.com/ceph/ceph-ansible/commit/73633f05a092190df8efa3aa5745aa1e0c4def19 
[2] https://github.com/openstack/puppet-ceph/blob/28e8f4525f4448a9f585f7ff4212fa9df58f4464/examples/common.yaml#L34

PR #1759, opened 2017-08-10
Set the permissions mode on all of the OpenStack keys

fultonj:
The original fix to issue #1755 only set the permissions on
the monitors to which the key was copied, but not the original
monitor where the key was created. Thus, we use a separate task
to set the permission of the key.


andrewschoen:

Issue #1751, opened 2017-08-08
what branch to use to deploy a stable cluster ?


acomisario:
Sorry for this question, but i see so many branches, that i need to ask this.
what branch should i use ? master ? always ?

best.

dvusboy:
Same boat here. I think there are `stable-*` with the version number referring to `ceph-ansible` releases. I don't know about the `ansible-*`. If I'm stuck with Ansible 1.9, can I use the branch `ansible-1.9`? Is it getting backported bug-fixes?

Many thanks for the great work though.

xahare:
some notes on the different branches would be great! and, maybe a faq entry.

having tried a few, seems for jewel, stable-2.2 is good. for luminus, tag v2.3.0rc5 works.

master also worked, but had to change a couple osd variables. id like to know how stable master is supposed to be. most of the combinations i tried failed in various ways. most tests were with ubuntu xenial and ansible 2.3.1.0.

fghaas:
_cough_ #1732

xahare:
i dont remmeber the error, but luminous didnt work for me on 2.2 either. now im using 3.0rcX on a test luminous cluster and its doing well. a recent commit made it clear thats the branch they want to work on.

leseb:
You have to use `stable-*` branches. We only support Ansible starting from 2.2, the latest tag 3.0rc series test on Ansible 2.3.1. We will continue to do so. The next major release (incoming) will likely only support Ansible starting from 2.3.1.

Thanks and sorry for the confusion, we will try to clarify all of that in the README soon.

fghaas:
@leseb I suggest you keep this issue open, then close it with the commit that actually updates the `README`.

leseb:
@fghaas sounds good to me.

Issue #1763, opened 2017-08-15
site.yml.sample should be the main site.yml instead of sample file


vasukulkarni:
Since site.yml.sample actually has various roles that are run in proper order for all hosts we shouldn't let user create their own during a new deployment, the file should be renamed 'site.yml' and should be used as is like other infrastructure playbooks (eg: purge-cluster)

andrewschoen:
Actually, it's expected for a user to change ``site.yml.sample`` to fit the needs of their own cluster. However, we've tried to design it so it works for most common use cases.

leseb:
Closing this with @andrewschoen's answer.

PR #1772, opened 2017-08-17
documentation update for osd scenarios and basic installation/usage

andrewschoen:
This does the following:

- removes README.md and adds README.rst that directs users to the hosted docs

- provides documentation for the ``collocated`` and ``non-collocated`` osd scenarios

- provides basic usage and installation of ``ceph-ansible`` documentation

- defines which ``stable-*`` branches are supported and what ceph and ansible versions they work for

Closes: https://github.com/ceph/ceph-ansible/issues/1751


docs/source/dev/index.rst
```diff
@@ -0,0 +1,85 @@
...
+Submit a patch
+--------------
+
+To start contributing just do::
+
+    $ git checkout -b my-working-branch
+    $ # do your changes #
+    $ git add -p
+
+If your change impacts a variable file in a role such as `roles/ceph-common/defaults/main.yml`, you need to generate a `group_vars` file::
```
alfredodeza: these need double back ticks

docs/source/dev/index.rst
```diff
@@ -0,0 +1,85 @@
...
+    jenkins test $scenario_name
+
+For example::
+
+    jenkins test luminous-ansible2.3-journal_collocation
+
+Backporting changes
+-------------------
+
+If a change should be backported to a `stable-*` Git branch:
```
alfredodeza: double backticks here too

docs/source/dev/index.rst
```diff
@@ -0,0 +1,85 @@
...
+
+- Mark your PR with the GitHub label "Backport" so we don't lose track of it.
+- Fetch the latest updates into your clone: `git fetch`
```
alfredodeza: in this list, there are a bunch of missing double backticks

docs/source/index.rst
```diff
@@ -8,30 +8,192 @@ ceph-ansible
...
+========
+
+The following branches should be used depending on your requirements. The ``stable-*``
+branches have been QE tested and sometimes recieve backport fixes throughout their lifecycle.
+The ``master`` branch should be considered experimental and used with caution.
+
+- ``stable-2.1`` Support for ceph version ``jewel``. This branch supports ansible versions
+  ``2.1`` and ``2.2.1``.
+
+- ``stable-2.2`` Support for ceph versions ``jewel`` and ``kraken``. This branch supports ansible versions
```
vasukulkarni: Does this mean one should be only using *latest* stable-* version?  and it wolud end up supporting all prevoius stable ceph releases?

docs/source/index.rst
```diff
@@ -8,30 +8,192 @@ ceph-ansible
...
+
+.. note::
+
+    For more information on ansible inventories please refer to the ansible documentation: http://docs.ansible.com/ansible/latest/intro_inventory.html
+
+Playbook
+--------
+
+You must have a playbook to pass to the ``ansible-playbook`` command when deploying your cluster. There is a sample playbook at the root of the ``ceph-ansible``
+project called ``site.yml.sample``. This playbook should work fine for most usages, but it does include by default every daemon group which might not be
```
vasukulkarni: This part is where i disagree, I think it should be used as is to reduce any user/admin errors or if user runs his own custom playbook it could be considered unsupported or something?

docs/source/index.rst
```diff
@@ -8,30 +8,192 @@ ceph-ansible
...
+appropriate for your cluster setup. Perform the following steps to prepare your playbook:
+
+- Rename the sample playbook: ``mv site.yml.sample site.yml``
```
vasukulkarni: If we rename probably this step will go away?

docs/source/osds/scenarios.rst
```diff
@@ -1,6 +1,130 @@
...
+----------
+This OSD scenario uses ``ceph-disk`` to create OSDs with collocated journals
+from raw devices.
+
+Use ``osd_scenario: collocated`` to enable this scenario. This scenario also
+has the following required configuration options:
+
+- ``devices``
+
+This scenario supports encrypting your OSDs by setting ``dmcrypt: True``.
```
vasukulkarni: is the osd_scenario a list for multiple options?

vasukulkarni:
Overall looks great, thanks @andrewschoen

docs/source/index.rst
andrewschoen: It really doesn't matter which one you use, it just depends on your requirements. Some might require ansible 2.1 and only deploy jewel so they might use ``stable-2.1``. Others might chose to use ``master`` to get quick fixes and new features. I've just tried to list here what each branch supports because this question has come up often.

docs/source/index.rst
andrewschoen: Let's discuss this in https://github.com/ceph/ceph-ansible/issues/1763

For now I'd like to just document the current state of things.

docs/source/osds/scenarios.rst
andrewschoen: No, ``osd_scenario`` is just a string that represents a single osd scenario.  The ``dmcrypt`` option is used with ``osd_scenario``.  I'll add an example to make this more clear. It'll be something like:

```
osd_scenario: "collocated"
dmcrypt: true
```

This keeps us from having so many osd scenarios like ``collocated`` and ``collocated_dmcrypt``, etc.

docs/source/index.rst
alfredodeza: @vasukulkarni lets not make that an issue for this PR which addresses docs at its current state. I don't see a problem in changing this once that behavior changes

vasukulkarni: Sure not blocking this, will continue the discussion at https://github.com/ceph/ceph-ansible/issues/1763

docs/source/index.rst
vasukulkarni: @andrewschoen  Sure

PR #1773, opened 2017-08-17
Update ceph_rgw_docker_extra_env to add bind ip

colonwq:
This patch adds passing the RGW_CIVETWEB_IP to the docker
container. This IP defaults to the value of radosgw_civetweb_bind_ip.
radosgw_civetweb_bind_ip default to ipv4.default

Without this value, the RGW containter will bind to 0.0.0.0

### DIFF
--- a/roles/ceph-rgw/defaults/main.yml
+++ b/roles/ceph-rgw/defaults/main.yml
@@ -48,6 +48,7 @@ rgw_pull_proto: "http"
 ##########
 
 ceph_rgw_civetweb_port: "{{ radosgw_civetweb_port }}"
-ceph_rgw_docker_extra_env: -e CLUSTER={{ cluster }} -e RGW_CIVETWEB_PORT={{ ceph_rgw_civetweb_port }}
+ceph_rgw_civetweb_ip: "{{ radosgw_civetweb_bind_ip }}"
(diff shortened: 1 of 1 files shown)


gfidente:
Maybe you could pay an update to group_vars/rgws.yml.sample too but it's only a sample file

roles/ceph-rgw/defaults/main.yml
```diff
@@ -48,6 +48,7 @@ rgw_pull_proto: "http"
 ##########
 
 ceph_rgw_civetweb_port: "{{ radosgw_civetweb_port }}"
-ceph_rgw_docker_extra_env: -e CLUSTER={{ cluster }} -e RGW_CIVETWEB_PORT={{ ceph_rgw_civetweb_port }}
+ceph_rgw_civetweb_ip: "{{ radosgw_civetweb_bind_ip }}"
+ceph_rgw_docker_extra_env: -e CLUSTER={{ cluster }} -e RGW_CIVETWEB_PORT={{ ceph_rgw_civetweb_port }} -e RGW_CIVETWEB_IP {{ ceph_rgw_civetweb_ip }}
```
leseb: This should be `RGW_CIVETWEB_IP={{ radosgw_civetweb_bind_ip }}`, you forgot the `=`. There is no need to add an extra variable.

colonwq: Couldn't the same be said about ceph_rgw_civetweb_port?

leseb:
@colonwq up

leseb:
As I can not take over, I'm closing this and opened https://github.com/ceph/ceph-ansible/pull/1839, @colonwq I kept your commit so you still get the credits ;)

Issue #1770, opened 2017-08-17
purge-cluster.yml and bluestore


bengland2:
I seem to be having to clear out all partitions on my dedicated_devices using this task in order to get clean reinstall of bluestore config.  Looking at code in purge-cluster.yml I couldn't figure out where it went wrong but suspect that it's not detecting block.db or block.wal partitions automatically.  Specifically the task "get ceph journal partitions" may be to blame.  Anyone else seeing this?  Here's my workaround:
```
  - name: clear journal and db partitions from SSD
    when: ansible_hostname in groups["osds"]
    shell: "for ssd in {{dedicated_devices}} ; do parted -s /dev/xvdb mktable gpt ; done"
```
Seems like the workaround is a lot simpler, but it might not pick up everything that you could find by searching for ceph-related partition UUIDs/labels.

leseb:
Nice catch @bengland2, I just pushed: https://github.com/ceph/ceph-ansible/pull/1778, do you mind trying it out?
Thanks!

PR #1778, opened 2017-08-21
purge: add ability to purge bluestore osd

leseb:
We now purge block db and/or wal partitions if we find any.

Closes: https://github.com/ceph/ceph-ansible/issues/1770


leseb:
@bengland2 thanks for testing this out :)

leseb:
Merging this as CI failure are fixed, but the docker images are being built.

Issue #1733, opened 2017-07-31
OSD autodiscovery breaks


fghaas:
This is from the `master` branch, commit f9eeedfbeef321f606300a471865642b482a5b9b:

```
TASK [ceph-osd : fail if any of the raw partitions are mounted] *******************************************************
fatal: [daisy]: FAILED! => {"failed": true, "msg": "The conditional check 'item.rc == 0' failed. The error was: error while evaluating conditional (item.rc == 0): 'dict object' has no attribute 'rc'\n\nThe error appears to have been in '/home/training/ceph-ansible/roles/ceph-osd/tasks/check_devices_auto.yml': line 28, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: fail if any of the rawpartitions are mounted\n  ^ here\n"}
fatal: [eric]: FAILED! => {"failed": true, "msg": "The conditional check 'item.rc == 0' failed. The error was: error while evaluating conditional (item.rc == 0): 'dict object' has no attribute 'rc'\n\nThe error appears to have been in '/home/training/ceph-ansible/roles/ceph-osd/tasks/check_devices_auto.yml': line 28, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: fail if any of the raw partitions are mounted\n  ^ here\n"}
fatal: [frank]: FAILED! => {"failed": true, "msg": "The conditional check 'item.rc == 0' failed. The error was: error while evaluating conditional (item.rc == 0): 'dict object' has no attribute 'rc'\n\nThe error appears to have been in '/home/training/ceph-ansible/roles/ceph-osd/tasks/check_devices_auto.yml': line 28, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: fail if any of the raw partitions are mounted\n  ^ here\n"}
```

This obviously breaks only with `osd_auto_discovery: true` in `group_vars/osds.yml`, if you enumerate the `devices` instead then the error doesn't trigger.

mistur:
I have the same behavior 

```TASK [ceph-osd : fail if any of the raw partitions are mounted] *********************************************************************************************************************************************************************************************************************************************************
task path: /root/git/ceph-ansible-luminous/roles/ceph-osd/tasks/check_devices_auto.yml:28
fatal: [cluster007.cluster]: FAILED! => {
[...]

PR #1786, opened 2017-08-22
mon, osd: fix skipped condition

leseb:
To be properly evaluated the "skipped" conditions must always have the
first place on the list of condition, otherwise the other conditions are
evaluated before and make the task fail.


fghaas:
@leseb You probably want to add `Fixes: #1733` to your commit message here.

leseb:
@fghaas right, thanks!

Issue #1787, opened 2017-08-22
purge-cluster.yml does not support osd_scenario=lvm


jharriga:
I have a ceph cluster deployed with osd_scenario=lvm
I tried running purge-cluster and no go... it needs 'devices' defined.
    fatal: [gprfs041.sbu.lab.eng.bos.redhat.com]: FAILED! => {"failed": true, "msg": "The conditional
    check 'devices|length == 0' failed. The error was: error while evaluating conditional (devices|length
    == 0): 'devices' is undefined\n

Since I am running with osd_scenario=lvm I have not defined that var.

FYI, the vars in 'group_vars/osds' look like this

osd_scenario: lvm
lvm_volumes:
  cached-lv0: /dev/nvme0n1p5
  cached-lv1: /dev/nvme0n1p6
  cached-lv2: /dev/nvme0n1p7
  noncached-lv0: /dev/nvme0n1p8
  noncached-lv1: /dev/nvme0n1p9
  noncached-lv2: /dev/nvme0n1p10

andrewschoen:
Thanks for reporting this. We currently have not build support into ``purge-cluster.yml`` for the ``lvm`` osd scenario. @alfredodeza and I were just discussing a ``ceph-volume lvm zap`` command to assist with this process, but we're not sure when it'll land in ceph. In the meantime I could try to add some tasks to ``purge-cluster.yml`` to handle this outside of ``ceph-volume``.

jharriga:
I am trying to do some performance testing with lvm osd scenario which will require tearing down and rebuilding the cluster with various lvm configurations.
I could really use a  purge ability. Thanks.

PR #1797, opened 2017-08-23
adds purge support for the lvm_osds osd scenario

andrewschoen:
To add this support, restructuring of ``lvm_volumes`` was necessary. The new format provides more flexibility in regards to how you define your lvm-based OSDs.

This also adds a ``purge_lvm_osds`` testing scenario to ensure it works correctly.

Fixes: https://github.com/ceph/ceph-ansible/issues/1787

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -194,12 +194,17 @@
 
   tasks:
 
+  - name: set devices if osd scenario is lvm
+    set_fact:
+      devices: []
+    when: osd_scenario == "lvm"
+
   - name: check for a device list
     fail:
       msg: "OSD automatic discovery was detected, purge cluster does not support this scenario. If you want to purge the cluster, manually provide the list of devices in group_vars/{{ osd_group_name }} using the devices variable."
     when:
       - devices|length == 0
-      - osd_auto_discovery
+      - osd_auto_discovery|default(false)
 
   - name: get osd numbers
     shell: "if [ -d /var/lib/ceph/osd ] ; then ls /var/lib/ceph/osd | sed 's/.*-//' ; fi"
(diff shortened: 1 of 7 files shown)


infrastructure-playbooks/purge-cluster.yml
```diff
@@ -321,6 +326,22 @@
       - ceph_disk_present.rc == 0
       - ceph_data_partlabels.rc == 0
 
+  - name: remove osd logical volumes
```
alfredodeza: can you add a comment here so that we know this needs to go away once `ceph-volume lvm zap` becomes available?

andrewschoen:
@jharriga would you mind taking a look at this please?

infrastructure-playbooks/purge-cluster.yml
andrewschoen: sure

roles/ceph-osd/tasks/check_mandatory_vars.yml
```diff
@@ -77,15 +77,15 @@
     - not osd_auto_discovery
     - lvm_volumes|length == 0
 
-- name: make sure the lvm_volumes variable is a dictionary
+- name: make sure the lvm_volumes variable is a list
   fail:
     msg: "lvm_volumes: must be a dictionary"
```
alfredodeza: this needs to be updated

andrewschoen:
jenkins test luminous-ansible2.3-update_docker_cluster

andrewschoen:
The failure in ``luminous-ansible2.3-update_docker_cluster`` is unrelated to this PR.

jharriga:
I just ran this against my cluster and it completed the purge with no errors.
Thank you for such quick response!

andrewschoen:
@jharriga Yes, when purging it does ``lvremove`` the lvs defined in ``lvm_volumes``, it also zaps any disks used for journals. In my testing I deployed a cluster, purged it, setup the lvs again and redeployed and everything worked fine. I guess I'm not sure what other behavior should be taken here. What would you expect to happen?

alfredodeza:
This is also how purge works for regular disks, the partitions get destroyed. The one difference, and where we divert (although temporarily) is that there is no facility to create/re-create lv's, unlike ceph-disk which will create them again

jharriga:
Just wanted to check that the difference of not currently create/re-create is temporary.
FYI, I recreated the lv's, ran ceph-ansible and the cluster is back.
Thanks.

PR #1799, opened 2017-08-23
ceph-osd: ceph-volume requires --data to be in vg/lv format

andrewschoen:


alfredodeza:
Merging this so that we can verify: https://github.com/ceph/ceph/pull/17207

Kind of chicken-egg issue here, because we need this in master since ceph-volume is changing.

alfredodeza:
Already ran correctly at https://2.jenkins.ceph.com/job/ceph-ansible-scenario/3/console even though the checks here are still going.

PR #1800, opened 2017-08-23
ceph-osd: Fix osd start sequence

b-ranto:
The script can fail to get the osd id because the osds are activated by
udev and it can take a while for them to activate. This commit fixes
that by trying to get all the osds per node in a loop.

This commit also makes the osd services enabled so that they are
available after reboot.

### DIFF
--- a/roles/ceph-osd/tasks/start_osds.yml
+++ b/roles/ceph-osd/tasks/start_osds.yml
@@ -9,6 +9,8 @@
   failed_when: false
   always_run: true
   register: osd_id
+  until: osd_id.stdout_lines|length == devices|unique|length
+  retries: 10
 
 - name: ensure systemd service override directory exists
   file:
@@ -32,5 +34,6 @@
   service:
     name: ceph-osd@{{ item }}
     state: started
+    enabled: true
   with_items: "{{ (osd_id|default({})).stdout_lines|default([]) }}"
   changed_when: false


b-ranto:
FYI: This might not show up as an issue at the moment because the udev scripts (well, ceph-disk trigger/activate) actually start the services by default but that should change in the future so that osds could obey the init system settings -- currently, the osds are started on boot even if they are disabled.

roles/ceph-osd/tasks/start_osds.yml
```diff
@@ -32,5 +34,6 @@
   service:
     name: ceph-osd@{{ item }}
     state: started
+    enabled: true
```
leseb: 👎 let's check with @alfredodeza but we removed that specifically because this is being done by ceph-disk already and we shouldn't change the behavior.

b-ranto: You are calling `ceph-disk prepare` only. That won't enable the service, it will just prepare the partitions. Once the partitions are prepared, udev detects them and takes over. It will call `ceph-disk trigger` and that step used to enable the service but that is what I am trying to change at the moment -- we use the same udev activation process on each boot as well and we definitely should not enable the service on each boot. Otherwise, we will disrespect the init system settings for enabled/disabled services.

I have a ceph PR for the ceph part of the change here:
https://github.com/ceph/ceph/pull/17209

In any case, I don't see how this line would hurt something even if it enabled the service ~twice? Maybe, I am missing something?

You can look at https://bugzilla.redhat.com/show_bug.cgi?id=1482201 for more details on why I am doing this.

leseb: That's not true, we also call `ceph-disk activate`. Anyway, I understand the issue, so if you believe that's the right approach then I don't have any objection to this patch.

leseb: One more thing, is it only true for osds? Don't we need the same for other daemons?
Thanks!

b-ranto: You are right, I missed that they are activated in a separate file. The other daemons do have that line in them (enabled: true, at least mon and mgr that I checked) -- the installation would not work without it, nothing would start after reboot. It is only the osds that were treated specially because of the way they are usually activated.

roles/ceph-osd/tasks/start_osds.yml
```diff
@@ -9,6 +9,8 @@
   failed_when: false
   always_run: true
   register: osd_id
+  until: osd_id.stdout_lines|length == devices|length
```
b-ranto: Hmm, I noticed you use `devices|unique` in a couple of places. Should I change this to `devices|unique|length`?

roles/ceph-osd/tasks/start_osds.yml
leseb: Understood.

roles/ceph-osd/tasks/start_osds.yml
leseb: Yes please.

PR #1804, opened 2017-08-24
Do not install additional dependencies when ceph_origin is distro

gfidente:
When Ceph is preinstalled or provided by the distribution repositories,
we should not install additional dependencies because they aren't needed or
should be pulled in by the package manager.

This change adds a when: ceph_origin != 'distro' condition to the tasks
which explicitly install dependencies.

### DIFF
--- a/roles/ceph-common/tasks/installs/install_on_debian.yml
+++ b/roles/ceph-common/tasks/installs/install_on_debian.yml
@@ -6,6 +6,7 @@
     update_cache: yes
     cache_valid_time: 3600
   with_items: "{{ debian_package_dependencies }}"
+  when: ceph_origin != 'distro'
 
 - name: configure ceph apt repository
   include: debian_ceph_repository.yml
--- a/roles/ceph-common/tasks/installs/install_on_redhat.yml
+++ b/roles/ceph-common/tasks/installs/install_on_redhat.yml
@@ -4,14 +4,18 @@
     name: "{{ item }}"
     state: present
   with_items: "{{ redhat_package_dependencies }}"
-  when: ansible_distribution == "RedHat"
+  when:
+    - ansible_distribution == "RedHat"
+    - ceph_origin != 'distro'
 
 - name: install centos dependencies
   package:
     name: "{{ item }}"
     state: present
   with_items: "{{ centos_package_dependencies }}"
-  when: ansible_distribution == "CentOS"
+  when:
+    - ansible_distribution == "CentOS"
+    - ceph_origin != 'distro'
 
 - name: configure ceph yum repository
   include: redhat_ceph_repository.yml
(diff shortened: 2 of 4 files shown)


gfidente:
Sorry guys I need to rebase the local branch, will do

gfidente:
@fultonj just so github notifies you of this

leseb:
jenkins test luminous-ansible2.3-bluestore_cluster

leseb:
jenkins test luminous-ansible2.3-dmcrypt_journal

leseb:
jenkins test luminous-ansible2.3-docker_cluster

leseb:
jenkins test luminous-ansible2.3-purge_dmcrypt

leseb:

leseb:
jenkins test luminous-ansible2.3-update_dmcrypt

gfidente:
We do use skip tags when possible; the scenario this change aims to fix is the one where you want to use the packages provided by the distro but they aren't installed yet ... on the nodes you have the repos configured already but you don't want the additional dependencies installed because everything (and only what is) needed for Ceph to work is pulled in as dependency by the Ceph packages.

I believe filter execution based on ceph_origin makes sense but if there are alternatives I am fine trying those.

leseb:
@gfidente these dependencies are mostly for Ansible not Ceph. If you don't want to have anything installed, why don't you use the `skip-tag`? If you do you will go through this task. You can also declare an empty array for `redhat_package_dependencies`?

gfidente:
As per my previous comment, we do use skip-tags when possible but in some scenarios want the Ceph packages to be installed, just not the ansible dependencies which aren't in fact needed by Ceph.

I don't see what is the problem with this change given it eventually fixes also a bug where ceph-ansible is installing the EPEL repo when ceph_origin is distro.

andrewschoen:
@gfidente could you try defining the following in your ``group_vars/all.yml`` and see if it solves your problem?

```
debian_package_dependencies: []
redhat_package_dependencies: []
centos_package_dependencies: []
```

This should make those tasks noop for you, while still allowing others to install these dependancies if they want with ``ceph-origin: distro``.

leseb:
@gfidente your fix is only valid for `centos_package_dependencies`. You're right we shouldn't install epel if ceph_origin is set to distro. Please rework your patch for this.

gfidente:
@leseb that might be an option, but the other packages are also unneeded

@andrewschoen I am not looking for a workaround (which we applied already), I am trying to fix what I think is an issue in ceph-ansible

leseb:
@gfidente why are you saying other packages are unneeded? Depending on the OS these are needed by Ansible:

* `libselinux-python`
* `python-setuptools`
* `python-pycurl`

gfidente:
I think these might be needed by Ansible on the node running the playbook, but are they on the nodes hosting Ceph? Testing in TripleO tells us either they *are* needed and installed as dependency by the Ceph packages or they are not because deployment succeeds when the package_dependencies array is empty.

leseb:
@gfidente you should have started by that I guess :). Are these dependancies packages already installed on the machines you're using? Maybe we could try to remove these tasks in this PR, just to see if the CI passed. Thanks!

Still, the only valid fix I see is for `centos_package_dependencies`.

leseb:
Closing this due to inactivity, feel free to re-open if you have anymore concerns.

PR #1805, opened 2017-08-24
Symlink infrastructure playbooks in the module root from .spec file

gfidente:
Given the infrastructure playbooks need to be executed from the
module root directory, we change the .spec file so that these are
linked in the module root when building the RPM.

### DIFF
--- a/ceph-ansible.spec.in
+++ b/ceph-ansible.spec.in
@@ -36,6 +36,12 @@ for f in ansible.cfg *.yml *.sample group_vars roles library plugins infrastruct
   cp -a $f %{buildroot}%{_datarootdir}/ceph-ansible
 done
 
+# Symlink infrastructure playbooks which need to be executed
+# from the module root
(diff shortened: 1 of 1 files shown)


gfidente:
@fultonj just so github notifies you of this

leseb:
@ktdreyer, what do you think of doing this? Personnally I'm not a huge fan of the approach but it'll obviously help some deployments.

gfidente:
I would be fine adding more ENV vars to the ansible-playbook execution if that helps, see [1]; those we're using apprently aren't sufficient.

1. https://review.openstack.org/#/c/495900/3/workbooks/ceph-ansible.yaml

leseb:
@gfidente I have an idea

leseb:
@gfidente let's close this in favor of https://github.com/ceph/ceph-ansible/pull/1808

gfidente:
@leseb ack

PR #1816, opened 2017-08-27
support Xen virtual disk device (xvd) used by Amazon EC2

bengland2:
I tested this fix on my Amazon EC2 cluster and it worked.   I assume it would be same for any Xen VM cluster.  If it works in the CI, then I think we're good.  

I would like in the future to see these regular expressions separated out into variables so that changes like this only need to be made in one place, but I went for the least risky change given all the activity here.  I may do a separate PR for this if you like that idea.

Issue #1520, opened 2017-05-14
purge-cluster.yml fails due to bug in ansible service module?


bengland2:
infrastructure-playbooks/purge-cluster.yml task below fails when Ceph is not installed:

      - name: stop ceph mdss with systemd
        service:
          name: ceph-mds@{{ ansible_hostname }}
          state: stopped
          enabled: no
        when: ansible_service_mgr == 'systemd'

The error is:

    TASK [stop ceph mdss with systemd] *********************************************
    fatal: [ip-172-31-27-179]: FAILED! => {"changed": false, "failed": true, "msg":  
     "Could not find the requested service ceph-mds@ip-172-31-27-179: cannot disable" }

versions:
ansible-2.2.1.0-1.el7.noarch
ceph-ansible-2.1.9-1.el7scon.noarch

To me this seems like a bug in the Ansible service module - if the service cannot be found, it should be safe to assume that the service is not running, right?    So in the Ansible tradition of idempotency, the service module should return success.  The service *is* disabled in the sense that it will not start when the system boots, since it is not present on the system.  This allows for purge-cluster to run successfully on a system where ceph is not present.  Do people agree with this?

But reality rears its ugly head -- if we can't get ansible to fix this (and they won't right away), then we have to start testing to see if the service is installed and add a when clause to above task to check before asking the service module to disable it, right?  :-(  

Why do I keep worrying about purge-cluster.yml?  After all, ansible playbooks are supposed to be idempotent.  I'm thinking about day-2 operations.  What if you want to shrink a cluster or change the nodes that a cluster lives on?  Then you'll need a playbook similar to this one right?  Also for testing purposes it's nice to be able to strip ceph off a cluster before running ceph-ansible, without redeploying the entire set of nodes.   

Perhaps with containerized Ceph services this problem goes away, trying that now...

leseb:
Why would run `purge-cluster.yml` on a node where Ceph is not installed?

bengland2:
as part of another script that puts ceph onto a cluster  and wants to make sure that all disks are wiped and any leftover ceph pieces are gone.  For example, previous ceph cluster may not have had an MDS.

guits:
My thought is that even if what you describe here was a real ansible issue, I'm not sure we should manage it at the ceph-ansible level.

guits:
This playbook is not supposed to fit this usecase.
Closing this issue

PR #1817, opened 2017-08-27
use when_failed to make stop of non-existent service ok

bengland2:
This patch makes purge-cluster.yml service shutdowns idempotent.   Thus it possible to run purge-cluster.yml if ceph-ansible site.yml has not yet run or has run but only partially completed.  We use failed_when: false in systemd and sysvinit shutdown tasks - this was already being done for Ubuntu pre-systemd releases.

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -66,10 +66,12 @@
       name: ceph-mds@{{ ansible_hostname }}
       state: stopped
       enabled: no
+    failed_when: false
     when: ansible_service_mgr == 'systemd'
 
   - name: stop ceph mdss
     shell: "service ceph status mds ; if [ $? == 0 ] ; then service ceph stop mds ; else echo ; fi"
+    failed_when: false
     when: ansible_service_mgr == 'sysvinit'
 
   - name: stop ceph mdss on ubuntu
@@ -96,10 +98,12 @@
       name: ceph-radosgw@rgw.{{ ansible_hostname }}
       state: stopped
       enabled: no
+    failed_when: false
     when: ansible_service_mgr == 'systemd'
 
   - name: stop ceph rgws
     shell: "service ceph-radosgw status ; if [ $? == 0 ] ; then service ceph-radosgw stop ; else echo ; fi"
+    failed_when: false
     when: ansible_service_mgr == 'sysvinit'
 
   - name: stop ceph rgws on ubuntu
@@ -126,6 +130,7 @@
     service:
       name: ceph-rbd-mirror@admin.service
       state: stopped
+    failed_when: false
     when: ansible_service_mgr == 'systemd'
 
   - name: stop ceph rbd mirror on ubuntu
@@ -152,10 +157,12 @@
     service:
       name: nfs-ganesha
       state: stopped
+    failed_when: false
     when: ansible_service_mgr == 'systemd'
 
   - name: stop ceph nfss
     shell: "service nfs-ganesha status ; if [ $? == 0 ] ; then service nfs-ganesha stop ; else echo ; fi"
+    failed_when: false
     when: ansible_service_mgr == 'sysvinit'
 
   - name: stop ceph nfss on ubuntu
@@ -417,10 +424,12 @@
       name: ceph-mon@{{ ansible_hostname }}
       state: stopped
       enabled: no
+    failed_when: false
     when: ansible_service_mgr == 'systemd'
 
   - name: stop ceph mons
     shell: "service ceph status mon ; if [ $? == 0 ] ; then service ceph stop mon ; else echo ; fi"
+    failed_when: false
     when: ansible_service_mgr == 'sysvinit'
 
   - name: stop ceph mons on ubuntu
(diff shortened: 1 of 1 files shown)


leseb:
AFAIK the service is idempotent already, so using `failed_when` does not change anything. See:

```
[stack@leseb-tarox ~]$ cat failed_when.yml
---
- hosts: localhost
  connection: local

  tasks:

    - name: stop ceph mdss with systemd
      service:
        name: ceph-mds@{{ ansible_hostname }}
        state: stopped
        enabled: no
      failed_when: false
[stack@leseb-tarox ~]$ ansible --version
ansible 2.3.1.0
  config file =
  configured module search path = Default w/o overrides
  python version = 2.7.12 (default, Aug  9 2016, 15:48:18) [GCC 6.1.1 20160621 (Red Hat 6.1.1-3)]
[stack@leseb-tarox ~]$ sudo systemctl status ceph-mds@leseb-tarox
● ceph-mds@leseb-tarox.service - Ceph metadata server daemon
   Loaded: loaded (/usr/lib/systemd/system/ceph-mds@.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
[stack@leseb-tarox ~]$ vim failed_when.yml
[stack@leseb-tarox ~]$ ansible-playbook failed_when.yml
 [WARNING]: Host file not found: /etc/ansible/hosts

 [WARNING]: provided hosts list is empty, only localhost is available


PLAY [localhost] ***********************************************************************************************************************************************************************************************

TASK [Gathering Facts] *****************************************************************************************************************************************************************************************
ok: [localhost]

TASK [stop ceph mdss with systemd] *****************************************************************************************************************************************************************************
ok: [localhost]

PLAY RECAP *****************************************************************************************************************************************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0
```

bengland2:
Seb,  you won't see the problem until the software has been uninstalled and there is no unit file.
Then comment out the **failed_when: false** line above and you'll see the problem.  See [issue 1520](https://github.com/ceph/ceph-ansible/issues/1520).  It *should* be idempotent already, that's a bug in ansible service module IMO.  If you ask ansible to stop a non-existent service, that should be a noop with success status, but it isn't.

leseb:
Right my bad. I think we might want to push this over to the Ansible project. In the meantime, let's move ahead with this. Thanks Ben!

leseb:
Actually, the only problem I see here is that the play might continue if the service stop fails and we don't really want that...
One way to handle that might be to use the command module and look up for a `No such file or directory` and then only fail if stderr does not contain it.

bengland2:
Is there any simpler way I could have done this?  I tried to do all this within the stop task by registering the result in a variable and testing the stdout_lines for 'service could not be found' but it gave me 

    fatal: [ip-172-31-12-170]: FAILED! => {"failed": true, "msg": "The conditional check ''service could not 
    be found' not in ceph_mds_status.stdout_lines | join | lower' failed. The error was: error while 
    evaluating conditional ('service could not be found' not in ceph_mds_status.stdout_lines | join | lower): 
    'dict object' has no attribute 'stdout_lines'"}

bengland2:
it appears this passed according to [Sebastien's comment here](https://github.com/ceph/ceph-ansible/pull/1816#issuecomment-325301914).

leseb:
@bengland2 what about osds? What if the osd_id list is empty?

bengland2:
If the OSD ID list is empty, then the 'get osd numbers' task registered an empty list, which means that all the OSD filesystems were already removed.  I don't think this can happen unless the OSD daemons were shut down first.  Since the OSD ids list is empty, then there will be no attempt to stop OSD daemons.   This means the ansible problem with stopping a non-existent service is not encountered.

leseb:
Well what you're testing is when ceph-ansible never ran so there might be no ceph-osd unit files at all so I'm expecting a failure. This task: https://github.com/bengland2/ceph-ansible/blob/c3988cca4b8d5b514b8f4433567fd3165b95cad1/infrastructure-playbooks/purge-cluster.yml#L247-L253 will never get skipped as there is no condition for osd_id.

bengland2:
Apparently since the with_items clause points to an empty list, it never gets executed in that case either.  I tried this with ansible-2.3.1.0-3.el7.noarch and it did not fail.  In any case, what Ceph OSD number should I use to implement the status check if the osd_ids list is empty?

bengland2:
I came up with this patch that will address your concern, I think.  Is this what you want before I commit it to this branch?

```
+  - name: see if ceph-osd.target is installed
+    shell: "systemctl status ceph-osd.target"
+    register: status_ceph_osd_target
+    failed_when: false
+
   - name: stop ceph-osd with systemd
     service:
       name: ceph-osd@{{item}}
       state: stopped
       enabled: no
     with_items: "{{ osd_ids.stdout_lines }}"
+    failed_when: "'unit ceph-osd.target could not be found' not in status_ceph_osd_target.stdout_lines | join | lower"
     when: ansible_service_mgr == 'systemd'
```

bengland2:
do not merge yet. found a problem.

leseb:
@bengland2 what's the status of this? Thanks!

guits:
@bengland2 any news for this?

bengland2:
can't work on this right now, sorry, it's up to you if you want to use it.  I'll leave development to the development team.

Issue #1775, opened 2017-08-18
CentOS and RHEL 7.3: Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!


vasukulkarni:
This seems to be ceph-ansible issue and fails for all CentOS 7.3 

"""
2017-08-18T19:07:25.922 INFO:teuthology.orchestra.run.vpm073.stdout:TASK [ceph-common : template ceph_conf_overrides] ******************************
2017-08-18T19:07:25.937 INFO:teuthology.orchestra.run.vpm073.stdout:task path: /home/ubuntu/ceph-ansible/roles/ceph-common/tasks/generate_ceph_conf.yml:13
2017-08-18T19:07:25.953 INFO:teuthology.orchestra.run.vpm073.stdout:fatal: [vpm039.front.sepia.ceph.com -> localhost]: FAILED! => {
2017-08-18T19:07:25.955 INFO:teuthology.orchestra.run.vpm073.stdout:    "changed": false,
2017-08-18T19:07:25.957 INFO:teuthology.orchestra.run.vpm073.stdout:    "checksum": "bda64e7c87f0fa320cb7c1cab9f8433e625d527c",
2017-08-18T19:07:25.959 INFO:teuthology.orchestra.run.vpm073.stdout:    "failed": true
2017-08-18T19:07:25.962 INFO:teuthology.orchestra.run.vpm073.stdout:}
2017-08-18T19:07:25.964 INFO:teuthology.orchestra.run.vpm073.stdout:
2017-08-18T19:07:25.966 INFO:teuthology.orchestra.run.vpm073.stdout:MSG:
2017-08-18T19:07:25.998 INFO:teuthology.orchestra.run.vpm073.stdout:
2017-08-18T19:07:26.000 INFO:teuthology.orchestra.run.vpm073.stdout:Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!
2017-08-18T19:07:26.009 INFO:teuthology.orchestra.run.vpm073.stdout:
2017-08-18T19:07:26.013 INFO:teuthology.orchestra.run.vpm073.stdout:NO MORE HOSTS LEFT *************************************************************
2017-08-18T19:07:26.015 INFO:teuthology.orchestra.run.vpm073.stdout:
2017-08-18T19:07:26.018 INFO:teuthology.orchestra.run.vpm073.stdout:PLAY RECAP *********************************************************************
[...]

PR #1821, opened 2017-08-28
tests: always use sitepackages=True

andrewschoen:
This is mostly important in rhcs testing so that our tests can use
packages installed on the distro.


andrewschoen:
This fixes our CI tests for issue #1775

PR #1824, opened 2017-08-28
dont use devices var anymore, works for osd_auto_discover

bengland2:
this enhancement allows purge-cluster.yml to handle sites with **osd_auto_discovery: true** .  Rather than use **devices** var to zap OSD data partitions, we find these partitions in the same way we find other types of partitions.  It was tested both with both **osd_scenario: non-collocated + osd_auto_discovery: false**  , and also with **osd_scenario: collocated + osd_auto_discovery: true** .

Since the **devices** var is not used, we don't need to clear it for **osd_scenario: lvm** either.

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -194,18 +194,6 @@
 
   tasks:
 
-  - name: set devices if osd scenario is lvm
-    set_fact:
-      devices: []
-    when: osd_scenario == "lvm"
-
-  - name: check for a device list
-    fail:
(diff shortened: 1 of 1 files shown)


bengland2:
why does the docs test fail?  It's a java exception in the log, never ran ceph-ansible, so I can't figure out what I could have possibly done that could have caused this.    Same with luminous-ansible2.3-journal_collocation_auto.

andrewschoen:
@bengland2 java exceptions like that are usually transient issues with jenkins. In those cases we typically just try to rerun.

andrewschoen:
jenkins test docs

PR #1838, opened 2017-08-31
Rgw units

leseb:


leseb:
Validated locally and by most of the previous CI runs.
Ci is struggling and I need to move forward with this.

PR #1839, opened 2017-08-31
Update ceph_rgw_docker_extra_env to add bind ip

leseb:
This patch adds passing the RGW_CIVETWEB_IP to the docker
container. This IP defaults to the value of radosgw_civetweb_bind_ip.
radosgw_civetweb_bind_ip default to ipv4.default

Without this value, the RGW containter will bind to 0.0.0.0


leseb:
luminous-ansible2.3-docker_cluster successfully passed, it's all that matters.

PR #1841, opened 2017-08-31
lvm-osds: test with a partition and an lv as journals

andrewschoen:
This adds support to the lvm_journals scenario so that you can use an lv for a journal. It also modifies the test so that it uses a partition for the other journal instead of a raw device, as ceph-volume is changing to not allow raw devices for journals.


andrewschoen:
I ran a successful test here using the wip branch from ceph for the ceph-volume changes: https://2.jenkins.ceph.com/job/ceph-ansible-scenario/12/console

This is the corresponding PR for ceph.git https://github.com/ceph/ceph/pull/17403

andrewschoen:
The CI failures here are unrelated to this PR. I'm going to merge so we can get ceph/ceph#17403 tested.

PR #1846, opened 2017-09-01
client: support for external cluster, generate admin key

leseb:
We need the ability to generate a client.admin key based on the external
cluster.


leseb:
Not needed.

Issue #1848, opened 2017-09-01
client configuration should not require ceph.client.admin.keyring


fultonj:
When using ceph-ansible's ability to configure a ceph client to talk to an external ceph cluster the task fails with the following message: 

2017-09-01 20:44:46,615 p=1728 u=mistral |  fatal: [192.168.24.10]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to find '/tmp/file-mistral-actiongUI08a/bddc7e56-8f09-11e7-a223-003da9b9b610/etc/ceph/ceph.client.admin.keyring' in expected paths."}

This task should be skipped in that scenario because it's not necessary for the client to have the admin key.

fultonj:
In addition, if the client admin key isn't copied, it shouldn't be used to try to `ceph auth import` the created client key and it is not necessary to verify the key either.

PR #1850, opened 2017-09-02
Add option to create client keyring file but not import it

fultonj:
Add new boolean parameter for client config create_key_file_only
with a default of false. When create_key_file_only is true, the
client tasks to connect to the external ceph cluster to verify
the key `ceph auth import` the key are skipped.

Fixes: #1848

### DIFF
--- a/roles/ceph-client/tasks/create_users_keys.yml
+++ b/roles/ceph-client/tasks/create_users_keys.yml
@@ -4,7 +4,9 @@
   with_items: "{{ pools }}"
   changed_when: false
   failed_when: false
-  when: pools | length > 0
+  when: 
+    - pools | length > 0
+    - copy_admin_key
 
 - name: create key(s)
   shell: "ceph-authtool -C /etc/ceph/{{ cluster }}.{{ item.name }}.keyring --name {{ item.name }} --add-key {{ item.key }} --cap mon \"{{ item.mon_cap|default('') }}\" --cap osd \"{{ item.osd_cap|default('') }}\" --cap mds \"{{ item.mds_cap|default('') }}\""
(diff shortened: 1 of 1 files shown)


roles/ceph-client/tasks/create_users_keys.yml
```diff
@@ -4,7 +4,9 @@
   with_items: "{{ pools }}"
   changed_when: false
   failed_when: false
-  when: pools | length > 0
+  when: 
+    - pools | length > 0
+    - not create_key_file_only
```
leseb: why not simply adding `- copy_admin_key`?

fultonj: The "create pools" and "create keys" tasks fail if the admin key is not copied. My goal is to have them not run in this scenario. Using one variable `copy admin key`, and not introducing another, as a condition of running this task, works for me. I'll update the PR.

leseb: Sure, please notify me when it's done. Thanks!

roles/ceph-client/tasks/create_users_keys.yml
```diff
@@ -22,14 +24,19 @@
   failed_when: false
   with_items: "{{ keys }}"
   register: keys_exist
+  when:
+    - copy_admin_key
 
 - name: add key(s) to ceph
   command: "ceph --cluster {{ cluster }} auth import -i /etc/ceph/{{ cluster }}.{{ item.0.name }}.keyring"
   changed_when: false
+  ignore_errors: true
```
leseb: I think we can remove this now right?

fultonj: If we remove ignore_errors:true, then item1.rc [1] will not be set because the previous task didn't run and then, according to my testing, the play will fail. If it's important to fail the play when the "add key(s) to ceph" task fails and not have ignore_errors:true, then one alternative is to have keys_exist.results set when not copy_admin_key. Should I add a task for that between "check if key(s) already exist(s)" and "add key(s) to ceph"? 

[1] https://github.com/fultonj/ceph-ansible/blob/b996c1f34e7c7e871b87e2295402ed667798e82b/roles/ceph-client/tasks/create_users_keys.yml#L39

leseb: You need to do the following in your `when` condition:

```
- not item.1.get("skipped")
```

Finally change `"{{ keys_exist.results }}"` to `"{{ keys_exist.results | default([])}}"`

Please try, thanks!

leseb:
jenkins test this please

PR #1855, opened 2017-09-05
Dump radosgw_keystone_config into [client.rgw.*] section

gfidente:
To configure radosgw it is necessary to provide some config options
related to the keystone deployment.

This change adds support for dumping into ceph.conf, in the
[client.rgw.*] section, arbitrary key=value settings using the
radosgw_keystone_config variable.


gfidente:
@guits @leseb @fultonj I like that this is a simple abstract structure which can be used to paste per-instance data and does not conflict with _overrides which remains functional for those who prefer it

leseb:
@gfidente what's the status of this?

PR #1859, opened 2017-09-06
container: introduce resource limitation for containers

leseb:
This can be controlled via 2 options:

* ceph_$DAEMON_docker_memory_limit
* ceph_$DAEMON_docker_cpu_limit

OSD and MDS have 5GB for memory and 1 CPU by default.
MON, RGW, RBD MIRROR, MGR have 2GB for memory and 1 CPU by default.

### DIFF
--- a/group_vars/mdss.yml.sample
+++ b/group_vars/mdss.yml.sample
@@ -25,6 +25,13 @@ dummy:
 # DOCKER #
 ##########
 
+# Resource limitation
+# For the whole list of limits you can apply see: docs.docker.com/engine/admin/resource_constraints
+# Default values are based from: https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/2/html/red_hat_ceph_storage_hardware_guide/minimum_recommendations
+# These options can be passed using the 'ceph_mds_docker_extra_env' variable.
+#ceph_mds_docker_memory_limit: 1g
+#ceph_mds_docker_cpu_limit: 1
+
 #ceph_mds_docker_extra_env: -e CLUSTER={{ cluster }} -e MDS_NAME={{ ansible_hostname }}
 #ceph_config_keys: [] # DON'T TOUCH ME
 
--- a/group_vars/mgrs.yml.sample
+++ b/group_vars/mgrs.yml.sample
@@ -11,6 +11,13 @@ dummy:
 # DOCKER #
 ##########
 
+# Resource limitation
+# For the whole list of limits you can apply see: docs.docker.com/engine/admin/resource_constraints
+# Default values are based from: https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/2/html/red_hat_ceph_storage_hardware_guide/minimum_recommendations
+# These options can be passed using the 'ceph_mgr_docker_extra_env' variable.
+#ceph_mgr_docker_memory_limit: 1g
+#ceph_mgr_docker_cpu_limit: 1
+
 #ceph_mgr_docker_extra_env: -e CLUSTER={{ cluster }} -e MGR_NAME={{ ansible_hostname }}
 #ceph_config_keys: [] # DON'T TOUCH ME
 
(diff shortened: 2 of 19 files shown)


leseb:
jenkins test luminous-ansible2.3-bluestore_cluster

leseb:
jenkins test luminous-ansible2.3-bluestore_docker_dedicated_journal

leseb:
jenkins test this please

leseb:
jenkins test this please

leseb:
jenkins test this please

leseb:
jenkins test luminous-ansible2.3-bluestore_cluster

leseb:
jenkins test luminous-ansible2.3-bluestore_dmcrypt_journal_collocation

leseb:
jenkins test luminous-ansible2.3-dmcrypt_journal

leseb:
jenkins test luminous-ansible2.3-dmcrypt_journal

leseb:
jenkins test luminous-ansible2.3-dmcrypt_journal_collocation

leseb:
jenkins test luminous-ansible2.3-update_dmcrypt

leseb:
jenkins test luminous-ansible2.3-switch_to_containers

leseb:
jenkins test luminous-ansible2.3-dmcrypt_journal

leseb:
jenkins test luminous-ansible2.3-dmcrypt_journal

leseb:
jenkins test luminous-ansible2.3-bluestore_cluster

bengland2:
concerns about default parameters for OSDs.

Memory:  I have seen OSDs use more than 1 GB RAM (in RSS within "top" display) when Tim and I tested OSP 10 on RHCS2.  This could happen in backfilling scenarios for example. What happens when an OSD tries to allocate more than 1 GB?  Does it get ENOMEM?  does it swap or core dump?  IMHO instead of this we should document OSD memory requirements and make it the responsibility of the system administrator to allocate enough memory for the OSDs to match their peak requirements.

all-flash configurations - seems like 1 CPU core will not be enough.  Either we'll have to partition the NVM device to allow multiple OSDs per NVM or we'll have to increase the CPU core count for an OSD that is in front of an NVM device.  Can we detect if the OSD device is non-rotational and adjust accordingly?  If not, we need to document this in RHCS and/or Openstack documentation, has that been done?

leseb:
I don't think there are any decent default values, that's why we stick with the minimum recommendation. We have variables so that the default can be changed. 
Changing defaults on the fly is not ideal as this is overriding user's choice.

bengland2:
if that's the way you want to go, at least allow OpenStack deployer to customize these values, is that possible?  Also 3 GB/OSD would be a good default, including kernel buffer cache.  

For HCI I suggested that OpenStack Nova has to avoid consuming memory needed for Ceph OSDs, and Ceph has to avoid consuming CPU needed for OpenStack VMs. [ John Fulton's ref arch](https://access.redhat.com/documentation/en-us/reference_architectures/2017/html-single/hyper-converged_red_hat_openstack_platform_10_and_red_hat_ceph_storage_2/#tuning) covers how to do the first.  We want to stop the OSD if it has a runaway memory leak, but otherwise we don't ever want to deny the OSD memory at runtime, which is a dangerous thing to do.  Instead we should a) constrain Nova VMs' total memory consumption, and b) constrain the number of OSDs per node, based on rules of thumb about how much memory the OSD would need.    a) is done when openstack is deployed, and b) is done when you purchase the hardware. @fultonj what's your opinion?

fultonj:
I missed this question from the past. FWIW, my opinion is that 3G/OSD is a better default.

fultonj:
Ben wrote:
> if that's the way you want to go, at least allow OpenStack deployer to customize these values, is that possible? 

Yes, I have added documentation showing a TripleO user (Pike and newer) how to do this at:

https://review.openstack.org/#/c/531214

PR #1864, opened 2017-09-07
switch-from-non-containerized-to-containerized: fix devices

leseb:
If devices is passed through an extra var this register won't work so
let's only register the var is devices is not defined. If devices is not
defined and lenght is > 0 then we trust its content.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1489099

### DIFF
--- a/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
+++ b/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
@@ -203,8 +203,18 @@
     - name: collect osd devices
       shell: |
         blkid | awk '/ceph data/ { sub ("1:", "", $1); print $1 }'
-      register: devices
+      register: collect_devices
       changed_when: false
+      when:
+        - devices is not defined or (devices is defined and devices == [])
+
+    - name: set devices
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
```diff
@@ -205,6 +205,9 @@
         blkid | awk '/ceph data/ { sub ("1:", "", $1); print $1 }'
       register: devices
       changed_when: false
+      when:
+        - devices is not defined
+        - devices | length > 0
```
guits: `devices is not defined` but his length would be `> 0` ?

leseb: I need to wake up I guess

leseb:
jenkins test luminous-ansible2.3-switch_to_containers

leseb:
After 2 hr 53 min of waiting for luminous-ansible2.3-switch_to_containers, the slave miserably failed with `Failed to connect to the host via ssh: `. Thanks OVH Cloud!

PR #1865, opened 2017-09-07
add new site-docker-tripleo.yml.sample for openstack

leseb:
OpenStack nodes don't need to run the ceph-common role since packages re
already installed.

Closes: #1876

### DIFF
--- a/site-docker-tripleo.yml.sample
+++ b/site-docker-tripleo.yml.sample
@@ -1,0 +1,98 @@
+---
+# Defines deployment design and assigns role to server groups
+
+- hosts:
+  - mons
+  - agents
+  - osds
+  - mdss
+  - rgws
+  - nfss
+  - restapis
+  - rbdmirrors
+  - clients
+  - iscsigws
+  - mgrs
+
+  tasks: []
+
+- hosts: mons
+  become: True
+  gather_facts: false
+  roles:
+    - ceph-defaults
+    - ceph-docker-common
+    - ceph-config
+    - ceph-mon
+  serial: 1 # MUST be '1' WHEN DEPLOYING MONITORS ON DOCKER CONTAINERS
+
+- hosts: osds
+  become: True
+  gather_facts: false
+  roles:
+    - ceph-defaults
+    - ceph-docker-common
+    - ceph-config
+    - ceph-osd
+
+- hosts: mdss
+  become: True
+  gather_facts: false
+  roles:
+    - ceph-defaults
+    - ceph-docker-common
+    - ceph-config
+    - ceph-mds
+
+- hosts: rgws
+  become: True
+  gather_facts: false
+  roles:
+    - ceph-defaults
+    - ceph-docker-common
+    - ceph-config
+    - ceph-rgw
+
+- hosts: nfss
+  become: True
+  gather_facts: false
+  roles:
+    - ceph-defaults
+    - ceph-docker-common
+    - ceph-config
+    - ceph-nfs
+
+- hosts: rbd_mirrors
+  become: True
+  gather_facts: false
+  roles:
+    - ceph-defaults
+    - ceph-docker-common
+    - ceph-config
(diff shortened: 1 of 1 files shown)


gfidente:
Would be nice to generate this with only the diffs from the regular site-docker playbook.

Was also wondering if we can rename it into -nopackages or something like that as it isn't really tripleo specific but rather covering a scenario where packages are not on the baremetal.

fultonj:
TripleO changes are pending but on hold [1] in case we switch to this playbook. I want to clarify that TripleO supports two scenarios: 

- deployed-servers: we would need ceph-ansible to install the packages
- image-based deploys: we would not need ceph-ansible to install the packages

In the first scenario, we would want users to not use the `--skip-tags package-install,with_pkg`, but the only reason we seem to need the package installed is so that the container host can run `ceph --version`. So, I would like to get your feedback on the following proposal: 

In ceph-common, if the scenario is for a containerized deployment, then determine the ceph version not by running the command but by starting a short-lived container to run the command to determine the version and then move on to the next task with that information. The original task which runs the command without the container, could just get a new "when not containerzed" added so that both scenarios are covered.

I found a container from ceph-ansible which just determines the version already running on my system so could we use something like this? 
```
[root@overcloud-cephstorage-2 ~]# docker ps -a
CONTAINER ID        IMAGE                                                   COMMAND                  CREATED             STATUS                    PORTS               NAMES
e623f6a61093        docker.io/ceph/daemon:tag-build-master-jewel-centos-7   "/usr/bin/ceph --vers"   11 hours ago        Exited (0) 11 hours ago                       backstabbing_heisenberg
[root@overcloud-cephstorage-2 ~]# UUID=e623f6a61093
[root@overcloud-cephstorage-2 ~]# docker logs $UUID
ceph version 10.2.9 (2ee413f77150c0f375ff6f10edd6c8f9c7d060d0)
[root@overcloud-cephstorage-2 ~]# 
```
If you like this idea, then please let me know and I can send in a PR. If you think I've not understood the full issue here then that's fine, just let me know. 

Thanks,
  John

[1] https://review.openstack.org/#/c/501743 and https://review.openstack.org/#/c/501748

site-docker-tripleo.yml.sample
```diff
@@ -0,0 +1,105 @@
...
+  gather_facts: false
+  roles:
+    - ceph-defaults
+    - ceph-docker-common
+    - ceph-config
+    - ceph-restapi
+
+- hosts: mgrs
+  become: True
+  gather_facts: false
```
fultonj: My testing of this with a client-only deployment failed with 'ansible_hostname' is undefined. When I set gather_facts to true it seemed to correctly configure the ceph client. 

2017-09-08 17:29:50,269 p=9448 u=mistral |  fatal: [192.168.24.22]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'ansible_hostname' is undefined\n\nThe error appears to have been in '/usr/share/ceph-ansible/roles/ceph-defaults/tasks/facts.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- set_fact:\n  ^ here\n"}

andrewschoen: @fultonj what version of ansible are you using? This playbook will require ``ansible>=2.3.1`` and this error kind of looks like one we see when ``delegate_facts: True`` is used with ansible versions older than ``2.3.1``.

fultonj: @andrewschoen TripleO uses Ansible 2.3.0 so I'm worried this will not work with it. I opened an issue on this here: https://github.com/ceph/ceph-ansible/issues/1876

leseb: Then I think we can solve #1876 with this PR. Let me update it.

PR #1890, opened 2017-09-13
tests: fix lvm_setup.yml for purge_cluster.yml

andrewschoen:
The ``dev-ansible2.3-purge_lvm_osds`` task has been failing because the ``purge-cluster.yml`` playbook does not destroy partitions used for lvm osd journals which causes the second run of ``lvm_setup.yml`` to fail during the purge scenario.

Proper support for purging partitions used for lvm osds will be added when ``ceph-volume lvm zap`` is available.

### DIFF
--- a/tests/functional/lvm_setup.yml
+++ b/tests/functional/lvm_setup.yml
@@ -42,11 +42,6 @@
         unit: '%'
         state: present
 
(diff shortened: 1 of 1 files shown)


andrewschoen:
The CI failures here are unrelated to this PR

PR #1898, opened 2017-09-14
defaults: restart docker daemon higher delay

leseb:
Use default delay since the mon (in particular) can take more time to
restart.
Solves error with:

STDERR:

Error response from daemon: No such container: ceph-mon-mon0


leseb:
@guits this should make the CI happy.

PR #1911, opened 2017-09-15
Introduce ceph_nfs_ceph_user

### LINKED ISSUE #1910: nfs-ganesha always connects as client.admin
In `roles/ceph-nfs/templates/ganesha.conf.j2` (in `stable-2.2`:  `roles/ceph-common/templates/ganesha.conf.j2`), the FSAL configuration looks like this:

```conf
        FSAL {
                Name = CEPH;
        }
```

Since this doesn't contain any `User_Id`, this will normally mean that nfs-ganesha connects to the Ceph cluster as `client.admin`, which is a rather needless privilege escalation. In analogy to using `ceph_nfs_rgw_user` for the RGW FSAL, this template should support `ceph_nfs_ceph_user` and use that for the `User_id` parameter. For backwards compatibility, it's probably necessary to have that variable default to `admin`, but at least it does give users the option to override the insecure behavior.

See nfs-ganesha/nfs-ganesha#104.

fghaas:
In analogy to ceph_nfs_rgw_user, we should be able to define a user
with which the nfs-ganesha Ceph FSAL connects to the cluster.

Introduce a ceph_nfs_ceph_user, setting its default to "admin" (which
preserves the prior behavior of always connecting as client.admin).

Fixes #1910.

### DIFF
--- a/group_vars/nfss.yml.sample
+++ b/group_vars/nfss.yml.sample
@@ -44,6 +44,7 @@ dummy:
 #ceph_nfs_ceph_pseudo_path: "/cephobject"
 #ceph_nfs_ceph_protocols: "3,4"
 #ceph_nfs_ceph_access_type: "RW"
+#ceph_nfs_ceph_user: "admin"
 
 ###################
 # FSAL RGW Config #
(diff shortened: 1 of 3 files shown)


leseb:

leseb:
@fghaas please run `generate_group_vars_sample`. Then we should be good to go.

fghaas:
@leseb Done, branch updated.

Issue #1907, opened 2017-09-15
master: nfs-ganesha installation broken for jewel, kraken?


fghaas:
According to http://docs.ceph.com/ceph-ansible/master/, the `master` branch should support `jewel`, `kraken`, and `luminous` deployments. But it appears to try to install `nfs-ganesha` from http://download.ceph.com/nfs-ganesha/, where `nfs-ganesha` builds are only available for `luminous`.

So this is any one of the following:

- documentation bug, because `master` is in fact *not* meant to be used for deploying anything earlier than `luminous`,
- role bug, because for `jewel` and `kraken` the same approach must be applied as in `stable-2.2` (which means, including the upstream `nfs-ganesha` PPAs),
- upstream packaging bug, because the `nfs-ganesha` builds really ought to exist for all 3 releases.

ktdreyer:
IMHO it is the latter (upstream packaging problem).

We really need nfs-ganesha builds for `jewel` and `luminous` (kraken is EOL and needs to be removed entirely from ceph-ansible).

One challenge is that nfs-ganesha 2.5 requires newer features in librgw than what `jewel` supports. Jewel has `LIBRGW_FILE_VERSION` 1.1.0, and nfs-ganesha 2.5 has `find_package(RGW 1.1.3)`. So we cannot build the 2.5 RGW FSAL with jewel; it has to be 2.4. And there are no 2.4 packages on download.ceph.com.

CephFS does not have this problem (as far as I know!)

Problems like this have made me think that we need to build and distribute ceph + nfs-ganesha via the CentOS Storage SIG and Ceph Launchpad groups so it is easier to identify and consume the matrix of supported versions.

ktdreyer:
That sounds good to me @fghaas . @alimaredia , as the nfs-ganesha maintainer, what do you think?

alimaredia:
@fghaas If you can remove the `nfs_obj_gw` variable entirely, and remove the part of the template in `ganesha.conf.j2`, that would be ideal. If removing `nfs_obj_gw` is too much of a pain, setting it to false with a comment saying to never turn it on works fine to.

alimaredia:
@fghaas now that I think about it, since you're working on a pull request, it might be best to just set `nfs_obj_gw` to false instead of making a lot of changes to the code base.

fghaas:
@alimaredia Turns out that the default for `nfs_obj_gw` already *is* `false`, so nothing to do there.

[...]

PR #1912, opened 2017-09-15
stable-2.2: Enable working nfs-ganesha deployment on Ubuntu

fghaas:
- Install `nfs-ganesha` packages on Ubuntu from the correct PPA (one with a build that actually contains a Ceph FSAL).
- Correct the `ganesha.conf` template to generate correct nfs-ganesha parameters.
- Enable nfs-ganesha to connect to Ceph using a CephX identity other than `client.admin` (but retain the default to connect as `client.admin`, to preserve compatibility).

### DIFF
--- a/roles/ceph-common/tasks/installs/debian_ceph_repository.yml
+++ b/roles/ceph-common/tasks/installs/debian_ceph_repository.yml
@@ -56,8 +56,8 @@
     repo: "{{ item }}"
     state: present
   with_items:
-    - ppa:gluster/libntirpc
-    - ppa:gluster/nfs-ganesha
+    - ppa:gluster/libntirpc-1.5
+    - ppa:gluster/nfs-ganesha-2.5
   changed_when: false
   when:
     - (nfs_obj_gw or nfs_file_gw)
--- a/roles/ceph-common/tasks/installs/install_on_debian.yml
+++ b/roles/ceph-common/tasks/installs/install_on_debian.yml
@@ -53,7 +53,7 @@
 
 - name: install NFS gateway
   apt:
-    pkg: nfs-ganesha-fsal
+    pkg: nfs-ganesha-ceph
     state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
     update_cache: yes
   when: nfs_group_name in group_names
--- a/roles/ceph-common/templates/ganesha.conf.j2
+++ b/roles/ceph-common/templates/ganesha.conf.j2
@@ -20,6 +20,7 @@ EXPORT
 
 	FSAL {
 		Name = CEPH;
+		User_Id = "{{ ceph_nfs_ceph_user }}";
 	}
 }
 {% endif %}
--- a/roles/ceph-common/templates/ganesha.conf.j2
+++ b/roles/ceph-common/templates/ganesha.conf.j2
@@ -4,7 +4,7 @@
 {% if nfs_file_gw %}
 EXPORT
 {
-	Export_ID={{ ceph_nfs_ceph_export_id }};
+	Export_id={{ ceph_nfs_ceph_export_id }};
 
 	Path = "/";
 
@@ -12,11 +12,11 @@ EXPORT
 
 	Access_Type = {{ ceph_nfs_ceph_access_type }};
 
-	NFS_Protocols = {{ ceph_nfs_ceph_protocols }};
+	Protocols = {{ ceph_nfs_ceph_protocols }};
 
-	Transport_Protocols = TCP;
+	Transports = TCP;
 
-	Sectype = sys,krb5,krb5i,krb5p;
+	SecType = sys,krb5,krb5i,krb5p;
 
 	FSAL {
 		Name = CEPH;
@@ -26,7 +26,7 @@ EXPORT
 {% if nfs_obj_gw %}
 EXPORT
 {
-	Export_ID={{ ceph_nfs_rgw_export_id }};
+	Export_id={{ ceph_nfs_rgw_export_id }};
 
 	Path = "/";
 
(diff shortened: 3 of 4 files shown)


ktdreyer:
We're pretty cautious about touching the `stable-2.2` branch at this point. Is this well-covered in CI (Jenkins)?

roles/ceph-common/tasks/installs/install_rh_storage_on_debian.yml
```diff
@@ -87,6 +87,6 @@
 
 - name: install red hat storage nfs gateway
   apt:
-    name: nfs-ganesha-fsal
+    name: nfs-ganesha-ceph
```
ktdreyer: Note: this will break on RH Ceph Storage 2 where the nfs-ganesha 2.4.5 sub-package is just named `nfs-ganesha-fsal` (matching the nfs-ganesha 2.4.5 upstream debian package)

fghaas:
@ktdreyer It's evidently *not* well covered in Jenkins, otherwise you'd have noticed much earlier that this doesn't actually install any Ceph FSAL on Ubuntu xenial. :)

roles/ceph-common/tasks/installs/install_rh_storage_on_debian.yml
fghaas: Which "upstream Debian" package are you referring to? I ask because https://packages.debian.org/search?keywords=nfs-ganesha-fsal&searchon=names&suite=stable&section=all turns up nothing.

ktdreyer: Yeah, the official Debian/Ubuntu ones are not well-maintained.

Here are the upstream (2.4.5) ones upon which RH Ceph Storage 2 was based: https://launchpad.net/~gluster/+archive/ubuntu/nfs-ganesha/+packages

fghaas: Yes, *these packages do not contain the Ceph FSAL.* That is exactly why I am sending this PR.

ktdreyer: The point I'm getting at is that `install_rh_storage_on_debian.yml` does not use any PPA; it's installing the downstream RHCS 2 product. For better or worse, the `stable-2.2` branch needs to support RH Ceph Storage 2.

fghaas: Oh, so you're saying in RHCS2, on Debian, you're not using any upstream binary packages but you're _rebuilding_ the ones from upstream, with the Ceph FSAL enabled? Fair enough, then the only thing that needs to happen is for me to roll back the change to `install_rh_storage_on_debian.yml` to where it continues to install `nfs-ganesha-fsal`, and the rest of the PR can remain as-is — correct?

ktdreyer: @fghaas you've got it now, please accept my apologies for not making this clear earlier.

Have all of these changes gone into master first?

fghaas: In master, I have no idea what the plan for nfs-ganesha is that doesn't break on Jewel or Kraken (which master ostensibly supports) — see #1907. So this PR just unbreaks something that is broken specifically in `stable-2.2`. In `master`, I don't know how to properly unbreak things.

fghaas:
@ktdreyer Branch fixed up, doesn't touch `install_rh_storage_on_debian.yml` anymore.

fghaas:
Must confess I have no idea why all the Jenkins runs break, but on a cursory glance it looks more like an infrastructure problem rather than something being wrong with the PR.

leseb:
@fghaas indeed, we do not test stable-2.2 AFAIR @andrewschoen right?
So perhaps we should just go ahead and merge it as it?

ktdreyer:
Changes need to be in master before going to any stable branch

fghaas:
@ktdreyer Care to elaborate on that last comment please? I don't quite follow.

- The `master` branch doesn't do PPAs for nfs-ganesha at all anymore. Instead it uses upstream package builds that exist only for luminous. As you commented in #1907, you consider that an upstream packaging problem. Fair enough, but I'd prefer for this not to be broken in a stable branch, while the upstream packaging issue gets sorted out.
- The configuration template already *is* fixed in `master`, where it also belongs to a different role. So I can't have this change go through master — depending on how you look at it, the file either no longer exists in `master`, or this patch already is a backport.
- For getting the `ceph_nfs_ceph_user` variable into `master`, there's #1911.

alimaredia:
@ktdreyer as @fghaas mentioned in detail, these changes couldn't go into master. I don't see how any of this nfs-ganesha related work could break the testing we already have for stable 2.2 right now because it wasn't being tested for this stable branch in the first place.

@fghaas Thank you so much for pointing out all the config variables that are off in the template. I'm going to go ahead and fix that in master as well.

fghaas:
Thanks @alimaredia — I forgot that issues don't auto-close when they get merged into a branch other than `master`. I'm closing those manually.

fghaas:
@alimaredia, I'm really sorry but there's a minor issue in the backport in that it does correctly add the new variable to `all.yml.sample`, but not to `nfss.yml.sample`. It's been correctly added to both in `master` in ada2f147f5c6537cf51171c995009b0b3288c027, but was missing in `nfss.yml.sample` in 72e32ae0bf1f060aaabd88ddb9c8dd000c0c5d97. 

I've now added `b6cb5f76fb6f40cd80476bac2e7d4ce6e58c4ce9` to my `1905` branch to fix this, would you mind re-pulling that one more time? Sorry about the hassle.

alimaredia:
@fghaas do mean you want me to revert the commit, then re-merge the pull request?

fghaas:
@alimaredia  No need to revert anything, just an additional patch to pull in. I'll just send an additional PR if that's OK.

Issue #1921, opened 2017-09-19
stable-2.2 - Ceph-rgw and ceph-mds restarts target inconsistently


andymcc:
The ceph-common/handlers uses a delegate_to with "run_once" that creates inconsistent behaviour.

This works some of the time, but it seems to fail occasionally by not  referencing the appropriate "ansible_hostname" or "mds_name" for the delegated to host, and instead uses the ansible_hostname from the initiating host.

RUNNING HANDLER [ceph.ceph-common : restart ceph rgws] *************************
changed: [rgw2 -> {{ ansible_addr_prefix }}.107] => (item=rgw1)
changed: [rgw2 -> {{ ansible_addr_prefix }}.108] => (item=rgw2)
changed: [rgw2 -> {{ ansible_addr_prefix }}.106] => (item=osdmonrgw)

E.g. this worked fine, however on a previous run with the same versions and inventory I saw this:
RUNNING HANDLER [ceph.ceph-common : restart ceph rgws] *************************
failed: [osdmonrgw -> {{ ansible_addr_prefix }}.107] (item=rgw1) => {"failed": true, "item": "rgw1", "msg": "Could not find the requested service ceph-radosgw@rgw.osdmonrgw: host"}
failed: [osdmonrgw -> {{ ansible_addr_prefix }}.108] (item=rgw2) => {"failed": true, "item": "rgw2", "msg": "Could not find the requested service ceph-radosgw@rgw.osdmonrgw: host"}
changed: [osdmonrgw -> {{ ansible_addr_prefix }}.106] => (item=osdmonrgw)

andymcc:
fixed in master

PR #1922, opened 2017-09-19
Restart service based on host specific vars

andymcc:
The current approach to restarting rgw/mds servers can result in
inconsistent behaviour. Sometimes this will attempt to restart the
initial hosts "ansible_hostname" service on the delegated_to host, which
will fail.

This PR changes to specifically use the delegate_to: host's hostvars.
This will prevent a situation where the incorrect hostvars are used.

This doesn't apply to master, since the approach to restart handlers has
moved to in role and no longer runs as a "run_once" with delegation.

Fixes: #1921

### DIFF
--- a/roles/ceph-common/handlers/main.yml
+++ b/roles/ceph-common/handlers/main.yml
@@ -52,7 +52,7 @@
 
 - name: restart ceph mdss
   service:
-    name: ceph-mds@{{ mds_name }}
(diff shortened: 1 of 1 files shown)


leseb:
Do you mind waiting for stable-3.0? We have that handled already but in a different way.

andymcc:
@leseb should be no problem.

andymcc:
closing this one out.

leseb:
@andymcc thanks!

Issue #1927, opened 2017-09-19
Automated Jenkins testing broken


fghaas:
This is a bit of a meta-issue, but I noticed it sending PRs over the last few days and it's rather annoyingly persistent: the Jenkins tests seem to be broken to the point of uselessness, which is probably a barrier to entry for new and occasional contributors.

leseb:
This comment suits so well Florian, always complaining and never thankful to the project.
To be honest, I don't like the way you talk, always cold and the impression everything is granted.

I value your technical skills but I can't so much about your social skills.
I think you should revisit the way you engage conversation; your writing lacks manner.
Otherwise, you won't gain much attention if you continue on that path.

I hope I'm mistaking, and that this is only a language issue.
Anyway, I hope things will change and that we can continue to collaborate in a more friendly manner.

Now, regarding the issue, yes we've noticed (we work on this project on a daily basis).
This is as frustrating for us as it is for you (to not say even more). We are currently addressing the issues.

fghaas:
Neither offense, nor complaint intended, and thanks for pointing out the wording. Let me clarify.

Here's the motivation for filing this: whenever I send a contribution (here or anywhere else), I always make a point of making sure that it's not breaking anything. That's what the integration tests are for, I assume, and it's good that they exist. In a bunch of projects it's also safe to assume that PRs don't ever get reviewed as long as they break the CI test suite. So, rather than attaching a comment to each of my PRs saying "I really *think* that the Jenkins run breakage *isn't* due to my patch, but I'm not 100% sure so please let me know if it is and I'll fix it", I thought I'd throw in this issue here. That way, interested parties (including myself) could watch the issue, and once it is marked fixed, we'd all understand that yes, breaking tests *is* a blocker. Right now that is not clear. If you prefer to handle this a different way, then please feel free to close this issue.

leseb:
I understand the concern, and I've been feeling the pain for weeks. We are currently in the process of fixing some of the major bugs in the CI, as well as introducing handy features too.

Reported issues:

- [ ] `uninitialized constant VagrantPlugins::ProviderLibvirt::Action::ShareFolders`

- [ ] `java.io.EOFException`

Errors are currently being investigated, this is mainly due to a resource issue. We quite often reach our quota limit on the cloud provider with boot jenkins slave...

So far this is what we came up with in terms of features:

[...]

PR #1935, opened 2017-09-21
stable-2.2: Eliminate RGW NFS

fghaas:
As evident from the discussion in #1907, it doesn't make much sense to keep the RGW NFS configuration around in the `stable-2.2` branch. Thus:

- Eliminate the RGW FSAL stanza from the `ganesha.conf` template
- Drop the `nfs_obj_gw` variable and all references
- Drop the (now-unused) remaining `ceph_nfs_rgw*` variables

This isn't intended to be merged as-is, just to give @alimaredia options to choose from as discussed in #1907.


alimaredia:
@fghaas I'm inclined to merge all the commits, I don't see any harm in any of them. However, I would rather let @leseb be the final call. 

Since The automated testing doesn't work her on jenkins I want to at least have his input before merging anything.

fghaas:
@alimaredia Sure, no problem at all.

alimaredia:
@fghaas knit, these two comments should be removed:

https://github.com/fghaas/ceph-ansible/blob/1907/group_vars/nfss.yml.sample#L26

https://github.com/fghaas/ceph-ansible/blob/1907/roles/ceph-nfs/defaults/main.yml#L18

fghaas:
@alimaredia Done, also rebased on current `stable-2.2`.

leseb:
@alimaredia feel free to merge if you feel it's good.

alimaredia:
lgtm

alfredodeza:
How is it possible that this got merged with 0 tests passing out of 31? 

* Does this even work on stable-2.2?
* what is the explanation on merging without CI passing?

guits:
@alfredodeza I think this obviously doesn't work on stable-2.2 because [this](https://github.com/ceph/ceph-ansible/pull/1935/files#diff-4fa370989543a95c8cc72971e1e4fdae) has been removed while [this](https://github.com/ceph/ceph-ansible/blob/gmeno-stable-2.2/roles/ceph-rgw/tasks/docker/main.yml#L42) hasn't.
As you mentioned, this should have been caught with CI testing.

fghaas:
@alfredodeza As it happens, I did create #1927 (before this one) in which I pointed out that the automated Jenkins testing was unreliable, and that that was a problem particularly for occasional contributors like me. Basically, all contributions that I made recently failed CI one way or another, because the CI was (is?) itself in a problematic state, and I was also told that `stable-2.2` isn't being CI-tested at all.

alfredodeza:
@fghaas the testing for ceph-ansible does not depend exclusively on Jenkins at all.

It was created to be able to run anywhere. It is entirely possible to run tests locally. This is very well documented here http://docs.ceph.com/ceph-ansible/master/testing/index.html and more specifically here http://docs.ceph.com/ceph-ansible/master/testing/running.html

I am not particularly placing you against the failures this has caused, but rather, I would like to see the leads to be able to verify things when contributors that might not know about details on how to properly test this, to thoroughly test/review/verify and to point the right way as to how to test things for the project.

Lastly, even if there was no CI at all, the questions are still valid, and should be commented/addressed before merging in such a state:

* Does this patch work? (any output with verifiable content could do)
* What is the explanation on merging a patch that fails everything

fghaas:
@alfredodeza Yup, sorry, I was entirely unaware of the alternate testing option and did my testing, as best as I knew, on a reference Ceph cluster. Would have been happy to post output if someone had asked to review, but didn't think to do that myself — apologies for that.

PR #1950, opened 2017-09-26
Docs err

yangyimincn:
ceph config templates localtion is now in roles/ceph-config/templates/ceph.conf.j2


yangyimincn:
Sorry about that. I will close this pr and pull it again

PR #1951, opened 2017-09-26
[skip ci] nfs: bindmount /var/lib/ganesha

leseb:
If we restart the container we need to keep /var/lib/ganesha intact for
data/session persistence.


leseb:
Is `/var/run/ganesha` needed as well?

ajarr:
Isn't it /var/lib/nfs/ganesha that you want to persist?

And I don't think you need to persist /var/run/ganesha.* 

@dang ?

PR #1939, opened 2017-09-21
tests: implement reboot after a deployment

leseb:
We don't test server reboot, a lot of things can happen after that. So
now, we deploy, reboot then we run testinfra.

Signed-off-by: Sébastien Han <seb@redhat.com>

PR #1967, opened 2017-09-29
Use systemd module instead of service.

guits:
Using systemd module allows us to do in one task what we did in three
tasks:

- enable unit file,
- issue a `daemon-reload`,
- start the service


leseb:
waits for https://github.com/ceph/ceph-ansible/pull/1939

PR #1976, opened 2017-10-03
Only perform actions on the rbd pool after it has been created

guits:
The rbd pool is the default pool that gets created during ceph cluster
initializaiton.  If we act on the rbd related operations too early, the
rbd pool does not exist yet.  Move the call to perform rbd operations
to a later stage after other pools have been created.

The rbd_pool.yml playbook has all the operations related to the rbd pool.

Replace the always_run (deprecated) directive with check_mode.

Most of the ceph related tasks only need to run once.  The run_once directive
executes the task on the first host.

The ceph sub-command to delete a pool is delete (not rm).

The changes submitted here were tested with this ceph version.
ceph version 0.94.9-9.el7cp (b83334e01379f267fb2f9ce729d74a0a8fa1e92c)

This upload includes these changes:
  - Use the fail module (instead of assert).
  - From luminous release, the rbd pool is no longer created by default.
    Delete the code to create the rbd pool for luminous release
  - Conform the .yml files to use the suggested syntax.

The commands are executed on the mcp nodes and I think shell ansible module
is the right one to use.  The command module is used to execute commands on
remote nodes.  I can make the change to use command module if that is
prefrerred.

### DIFF
--- a/roles/ceph-client/tasks/main.yml
+++ b/roles/ceph-client/tasks/main.yml
@@ -6,5 +6,4 @@
   include: create_users_keys.yml
   when:
     - user_config
-    - global_in_ceph_conf_overrides
-    - ceph_conf_overrides.global.osd_pool_default_pg_num is defined
+    - ceph_conf_overrides.get('global', {}).get('osd_pool_default_pg_num', False) != False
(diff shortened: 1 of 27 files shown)


guits:

PR #1953, opened 2017-09-27
[skip ci] Colocation

leseb:
wait for https://github.com/ceph/ceph-build/pull/873

PR #1982, opened 2017-10-04
set docker_exec_cmd fact early in each role

guits:
This is to ensure `docker_exec_cmd` fact is set with the correct value
in case of daemons collocation

### DIFF
--- a/roles/ceph-rgw/tasks/main.yml
+++ b/roles/ceph-rgw/tasks/main.yml
@@ -1,4 +1,10 @@
 ---
+- name: set_fact docker_exec_cmd
+  set_fact:
+    docker_exec_cmd: "docker exec ceph-rgw-{{ ansible_hostname }}"
+  when:
+    - containerized_deployment
+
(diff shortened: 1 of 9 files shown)


roles/ceph-rgw/tasks/main.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- name: set_fact docker_exec_cmd
+  set_fact:
+    docker_exec_cmd: "docker exec ceph-rg-{{ ansible_hostname }}"
```
leseb: s/rg/rgw/

leseb:
You need also:

```
- name: set_fact docker_exec_cmd nfs ganesha
  set_fact:
    docker_exec_cmd: "docker exec ceph-nfs-{{ ansible_hostname }}"
  when:
    - containerized_deployment
```

leseb:
cherry-picked in https://github.com/ceph/ceph-ansible/pull/1953

PR #1983, opened 2017-10-04
Allow to override systemd service instance id

jprovaznik:
It's useful to have constant service instance id when ceph-nfs
is managed by pacemaker.

### DIFF
--- a/roles/ceph-defaults/templates/restart_nfs_daemon.sh.j2
+++ b/roles/ceph-defaults/templates/restart_nfs_daemon.sh.j2
@@ -2,12 +2,12 @@
 
 RETRIES="{{ handler_health_nfs_check_retries }}"
 DELAY="{{ handler_health_nfs_check_delay }}"
-NFS_NAME="{{ ansible_hostname }}"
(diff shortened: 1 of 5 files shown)


jprovaznik:
@leseb done

roles/ceph-defaults/templates/restart_nfs_daemon.sh.j2
```diff
@@ -2,7 +2,7 @@
 
 RETRIES="{{ handler_health_nfs_check_retries }}"
 DELAY="{{ handler_health_nfs_check_delay }}"
-NFS_NAME="{{ ansible_hostname }}"
+NFS_NAME="ceph-nfs@{{ ceph_nfs_service_suffix | default(ansible_hostname) }}"
```
guits: isn't going to cause trouble with L10 below ?
will result in : `ceph-nfs@ceph-nfs@{{ ceph_nfs_service_suffix | default(ansible_hostname) }}`

jprovaznik: Good catch, fixed, thanks. It seems we currently don't check whether "systemctl restart" command passes or not which causes that ceph-ansible passes even with the bug above.

PR #1993, opened 2017-10-05
Fix bind mount for /var/lib/nfs/ganesha directory

jprovaznik:
It's useful to keep logs permanently, also this fixes bind mount
for /var/lib/nfs/ganesha directory.

### DIFF
--- a/roles/ceph-nfs/templates/ceph-nfs.service.j2
+++ b/roles/ceph-nfs/templates/ceph-nfs.service.j2
@@ -11,7 +11,7 @@ ExecStart=/usr/bin/docker run --rm --net=host \
   {% if not containerized_deployment_with_kv -%}
   -v /var/lib/ceph:/var/lib/ceph \
   -v /etc/ceph:/etc/ceph \
-  -v /var/lib/ganesha:/var/lib/ganesha \
+  -v /var/lib/nfs/ganesha:/var/lib/nfs/ganesha \
   -v /etc/ganesha:/etc/ganesha \
   {% else -%}
   -e KV_TYPE={{kv_type}} \


roles/ceph-nfs/templates/ceph-nfs.service.j2
```diff
@@ -11,7 +11,8 @@ ExecStart=/usr/bin/docker run --rm --net=host \
   {% if not containerized_deployment_with_kv -%}
   -v /var/lib/ceph:/var/lib/ceph \
   -v /etc/ceph:/etc/ceph \
-  -v /var/lib/ganesha:/var/lib/ganesha \
+  -v /var/lib/nfs/ganesha:/var/lib/nfs/ganesha \
+  -v /var/log/ganesha:/var/log/ganesha \
```
leseb: 👎  logs from the container are on stdout already, so they are on journald and appear on `docker logs` too.

jprovaznik:
OK, I've updated the patch to change only bind mount for /var/lib/nfs/ganesha

roles/ceph-nfs/templates/ceph-nfs.service.j2
```diff
@@ -11,7 +11,7 @@ ExecStart=/usr/bin/docker run --rm --net=host \
   {% if not containerized_deployment_with_kv -%}
   -v /var/lib/ceph:/var/lib/ceph \
   -v /etc/ceph:/etc/ceph \
-  -v /var/lib/ganesha:/var/lib/ganesha \
+  -v /var/lib/nfs/ganesha:/var/lib/nfs/ganesha \
```
leseb: so `/var/lib/ganesha` doesn't exist?

jprovaznik: Correct, at least on my setup there is only /var/lib/nfs/ganesha used by ganesha

PR #1995, opened 2017-10-05
jewel: remove rbd check

leseb:
The value of doing this is fairly low compare to the added value.
So we remove these tasks, if rbd pool on Jewel doesn't have the right PG
value you can always increase it.


leseb:

PR #1998, opened 2017-10-05
DNM: ci: test remove CEPH_STABLE_RELEASE

leseb:
The CEPH_STABLE_RELEASE, ceph_stable_release is hardcoded in tox, which
makes the playbook passing when sometimes it should fail.
So let's remove it since this variable is generated by the playbook
itself during the play.


andrewschoen:
Don't we need ``ceph_stable_release`` to know which upstream version to install?  Like we do here: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/installs/redhat_community_repository.yml#L14

leseb:
@andrewschoen right, this is commit is mostly a test to see how we can avoid some issue, I don't intend to merge this now, I just want to see how the playbook is going to react :)

guits:
jenkins test all

PR #2013, opened 2017-10-07
A couple of purge cluster fixes

b-ranto:
I have encourted a couple of bugs when testing ceph-ansible on my machine. These commits helped me fix them.

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -67,6 +67,7 @@
       name: ceph-mds@{{ ansible_hostname }}
       state: stopped
       enabled: no
+    failed_when: false
     when: ansible_service_mgr == 'systemd'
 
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/purge-cluster.yml
```diff
@@ -478,6 +483,7 @@
       name: ceph-mon@{{ ansible_hostname }}
       state: stopped
       enabled: no
+    failed_when: false
```
leseb: is it for idempotency purpose?

b-ranto: kinda, I had to add these lines mostly because it failed before (the zsh expansion failed) and it did not purge everything. Since it failed half-way, it failed to purge anything in the second round because it could not find the systemd services.

btw: if you look a couple lines down, you'll notice we were already doing this for upstart.

PR #2020, opened 2017-10-10
Ceph-nfs dynamic exports fixes

jprovaznik:
* DBus on host should include ganesha service file
* to allow ganesha container to respond on DBus it needs to run
  in --privileged mode (ganesha folks contacted to look at this)
* ceph_nfs_include_exports_dir variable replaced with more general
  ceph_nfs_dynamic_exports

### DIFF
--- a/roles/ceph-nfs/tasks/pre_requisite_container.yml
+++ b/roles/ceph-nfs/tasks/pre_requisite_container.yml
@@ -37,3 +37,19 @@
   when:
     - ansible_os_family == 'RedHat'
     - sestatus.stdout != 'Disabled'
+
+- name: create dbus service file
+  become: true
+  template:
+    src: "{{ role_path }}/templates/org.ganesha.nfsd.conf"
+    dest: /etc/dbus-1/system.d/org.ganesha.nfsd.conf
+    owner: "root"
+    group: "root"
+    mode: "0644"
+  when:
+    - ceph_nfs_dynamic_exports
+
+- name: reload dbus configuration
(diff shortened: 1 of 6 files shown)


roles/ceph-nfs/tasks/pre_requisite_container.yml
```diff
@@ -37,3 +37,20 @@
...
+    src: "{{ role_path }}/templates/org.ganesha.nfsd.conf"
+    dest: /etc/dbus-1/system.d/org.ganesha.nfsd.conf
+    owner: "root"
+    group: "root"
+    mode: "0644"
+  when:
+    - ceph_nfs_dynamic_exports
+
+- name: reload dbus configuration
+  shell:
```
leseb: you should use the `command` module instead.

jprovaznik: done

roles/ceph-nfs/templates/ceph-nfs.service.j2
```diff
@@ -13,6 +13,10 @@ ExecStart=/usr/bin/docker run --rm --net=host \
   -v /etc/ceph:/etc/ceph \
   -v /var/lib/nfs/ganesha:/var/lib/nfs/ganesha \
   -v /etc/ganesha:/etc/ganesha \
+  {% if ceph_nfs_dynamic_exports %}
+  --privileged \
```
leseb: I'm tempted to say we always need `--privileged` don't you think?

Issue #2017, opened 2017-10-09
ceph-ansible needs to allow configuring PG for metadata and data pools


batrick:
The metadata pool often uses far fewer PG than the data pool. The `osd_pool_default_pg_num` parameter is inappropriate for this.

leseb:
Ok so I think we should add one variable for the metadata pool only.

leseb:
And btw I'm not if this should be marked as a bug... This is more an improvement to me.

batrick:
TBH, I'm not thrilled by the presence of `osd_pool_default_pg_num `. There is no good default for placement groups -- it's the most challenging admin config option in the whole of Ceph. Deployment which includes setting up pools should require deliberate choice of PG for each pool.

Can we somehow require the user to specify the pools to create and the PGs for each pool? That would need to be done as a prereq for creating the file system (`ceph fs new`).

leseb:
Makes sense, let me work on something on that regard.
Thanks for the input.

zmc:
Any docs mentioning the addition of another mandatory var?

leseb:
@zmc nop :/, looks like we forgot that.

PR #2036, opened 2017-10-12
mds: precisely define cephfs pool

leseb:
We now have a variable called ceph_pools that is mandatory when
deploying a MDS.
It's a dictionnary that contains a pool name and a PG count. PG count is
mandatory and must be set, the playbook will fail otherwise.

Closes: https://github.com/ceph/ceph-ansible/issues/2017

### DIFF
--- a/group_vars/mons.yml.sample
+++ b/group_vars/mons.yml.sample
@@ -12,17 +12,11 @@ dummy:
 ###########
 # GENERAL #
 ###########
-
 #mon_group_name: mons
 
 # ACTIVATE BOTH FSID AND MONITOR_SECRET VARIABLES FOR NON-VAGRANT DEPLOYMENT
 #monitor_secret: "{{ monitor_keyring.stdout }}"
 #admin_secret: 'admin_secret'
-
-# CephFS
-#cephfs_data: cephfs_data
-#cephfs_metadata: cephfs_metadata
-#cephfs: cephfs
 
 # Secure your cluster
 # This will set the following flags on all the pools:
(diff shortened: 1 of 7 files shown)


roles/ceph-mon/defaults/main.yml
```diff
@@ -4,18 +4,12 @@
...
 mon_group_name: mons
 
 # ACTIVATE BOTH FSID AND MONITOR_SECRET VARIABLES FOR NON-VAGRANT DEPLOYMENT
 monitor_secret: "{{ monitor_keyring.stdout }}"
 admin_secret: 'admin_secret'
 
-# CephFS
-cephfs_data: cephfs_data
-cephfs_metadata: cephfs_metadata
-cephfs: cephfs
```
guits: I see here you removed cephfs variable although it's still used here https://github.com/ceph/ceph-ansible/pull/2036/files#diff-1ebfa6bd1a372462e995257b0e1f09e3R9

leseb: good catch, this should remain

leseb:
@batrick I haven't included multi fs support on purpose, I see default still disables this. Let me know if that makes sense for you.

group_vars/mons.yml.sample
```diff
@@ -43,9 +37,21 @@ dummy:
...
+##########
+# CEPHFS #
+##########
+#cephfs: cephfs # name of the ceph filesystem
+#cephfs_data: cephfs_data # name of the data pool for a given filesystem
+#cephfs_metadata: cephfs_metadata # name of the metadata pool for a given filesystem
+
+#ceph_pools:
+#  - { name: "{{ cephfs_data }}", pgs: "" }
+#  - { name: "{{ cephfs_metadata }}", pgs: "" }
```
batrick: The placement of `ceph_pools` under CEPHFS seemed odd. Do YAML parsers concatenate multiple definitions of a list? (Not really a big deal but I could see ceph_pools being moved later on to somewhere more generic...)

batrick:
multifs is still experimental so I'm okay with leaving that out for now.

group_vars/mons.yml.sample
leseb: It's just because I should have called this `cephfs_pools` :). Fixing this.

Issue #1574, opened 2017-05-31
Deprecation warning: alaways_run is deprecated. Use check_mode = no instead.


zealoussnow:
When I deploy a new ceph cluster, I got the following message:

```yml
TASK [ceph-mon : collect admin and bootstrap keys] *******************************************************************************************************************************
[DEPRECATION WARNING]: always_run is deprecated. Use check_mode = no instead..
This feature will be removed in version 2.4. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.
ok: [kvm-host1]
```

Shoud I remove always_run and add check_mode = no or set deprecation_warnings=False in ansible.cfg?

leseb:
That's not an issue for now. We are keeping `always_run` for backward compatibility. Once we move to 2.4 we will use `check_mode` instead. No rush here, but if you find this annoying feel free to set deprecation_warnings=False in ansible.cfg.

zealoussnow:
@leseb OK, Thank you

PR #2037, opened 2017-10-12
Avoid deprecated always_run

major:
The `always_run` key is deprecated and being removed in Ansible 2.4.
Using it causes a warning to be displayed:

    [DEPRECATION WARNING]: always_run is deprecated.

This patch changes all instances of `always_run` to use the `always`
tag, which causes the task to run each time the playbook runs.

### DIFF
--- a/roles/ceph-common-coreos/tasks/main.yml
+++ b/roles/ceph-common-coreos/tasks/main.yml
@@ -3,7 +3,8 @@
   raw: stat $HOME/.python
   register: need_python
   ignore_errors: true
-  always_run: true
+  tags:
+    - always
 
 - include: install_pypy.yml
   when: need_python | failed
@@ -12,7 +13,8 @@
   raw: stat $HOME/.pip
   register: need_pip
   ignore_errors: true
-  always_run: true
+  tags:
+    - always
 
 - include: install_pip.yml
   when: need_pip | failed and need_python | failed
--- a/roles/ceph-common/tasks/checks/check_firewall.yml
+++ b/roles/ceph-common/tasks/checks/check_firewall.yml
@@ -5,7 +5,8 @@
   failed_when: false
   register: nmapexist
   run_once: true
-  always_run: true
+  tags:
+    - always
 
 - name: inform that nmap is not present
   debug:
@@ -19,10 +20,11 @@
   changed_when: false
   failed_when: false
   register: monportstate
-  always_run: true
   when:
     - mon_group_name in group_names
     - nmapexist.rc == 0
+  tags:
+    - always
 
 - name: fail if monitor port is filtered
   fail:
@@ -37,10 +39,11 @@
   changed_when: false
   failed_when: false
   register: osdrangestate
-  always_run: true
   when:
     - osd_group_name in group_names
     - nmapexist.rc == 0
+  tags:
+    - always
 
 - name: fail if osd and mds range is filtered (osd hosts)
   fail:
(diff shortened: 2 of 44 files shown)


guits:
@major why not using `check_mode` as suggested in the warning message?

major:
@guits That's where it gets confusing. `check_mode: yes` means to always run the task in check mode, even if the playbook is called without `--check`.

`check_mode: no` means the task should always apply changes to the system, *even if* the playbook is called with `--check`. That could cause some problems if someone intended to have no changes made. :/

I've generally used the `always` tag to ensure the task always runs, but the `--check` argument will still be obeyed. Does that make sense?

guits:
@major thanks for this explanation!

major:
@leseb @guits You're welcome. Thanks for looking over these patches. ;)

towolf:
@major i'm afraid you misunderstand what `always_run` and `check_mode: no` are used for.

To acquire information, by running tasks, even in check mode.

Tasks that are a prerequisite for later tasks. For example here:

```- name: check if nmap is installed
  local_action: shell command -v nmap
  changed_when: false
  failed_when: false
  register: nmapexist
  run_once: true
  tags:
    - always
```
by using `always_run` or `check_mode: no` we store `nmapexist` for later use. This boolean usually used to make check mode work at all.

I'm afraid you broke check mode.

towolf:
`tags: always` is orthogonal to that, and only needed when using other tags to make sure that some prerequisite tasks are not skipped.

towolf:
@major, @leseb I've reviewed this PR and all instances of always_run were purely informational, making no changes on the system, in terms of "get", "print", "cat", "stat", "docker ps", and so on.

These tasks acquire info need for the execution of the following tasks. Now `--check` will not work anymore ...

`tags: always` is the wrong fix for this. I recommend a revert and repace with `check_mode: no`.

towolf:
@guits, please look.

guits:
Hi @towolf, I would tend to agree with what you said, this is why I initially asked why not using `check_mode: no` [here](https://github.com/ceph/ceph-ansible/pull/2037#issuecomment-336139070) but then, I guess I got confused by @major explanation. To be honest,
I'm still confused with `tags: always` vs `check_mode: no` because I don't have time to double check this and definitely make a fix right now (will have time on next week).

major:
I'll work on a fixed PR, @towolf and @guits.

towolf:
Just for context, there once was a PR to switch to `check_mode: no`, which was then rejected: 

> Once we move to 2.4 we will use check_mode instead.
https://github.com/ceph/ceph-ansible/issues/1574

And the official docs explain this also:

> Prior to version 2.2 only the equivalent of check_mode: no existed. The notation for that was always_run: yes.

_actually it should be "Prior to version 2.0"_

The equivalent for `check_mode: yes` did not exist before 2.0. This notation means, "run this task in check mode regardless of global `--check` flag".

towolf:
And, a year ago some instances of `check_mode: no` were replaced with `always_run: yes`.

https://github.com/ceph/ceph-ansible/commit/3023d0339ccbedf88d0b57264a91e039369e60dd

But it's up to the project to decide which Ansible version it wants to be compatible with.

major:
The roles currently have:

```
  min_ansible_version: 2.3
```

towolf:
Then your new PR is correct.

major:
@towolf Thanks. Sorry for the confusion. Feel free to comment on #2105. ;)

PR #2038, opened 2017-10-12
Suppress yum/dnf/rpm command warnings

major:
Ansible throws warnings when using yum/dnf/rpm with the command
module:

    [WARNING]: Consider using yum module rather than running yum

This patch adds the `warn: no` argument to suppress the warnings
in the Ansible output.


major:
jenkins test luminous-ansible2.3-xenial_cluster

PR #2041, opened 2017-10-12
ceph-osd: Remove jinja2 delimiters from when

major:
Modern versions of Ansible throw a warning when jinja2 delimiters
are used in `when:` keys:

    [WARNING]: when statements should not include jinja2 templating delimiters
    such as {{ }} or {% %}. Found: ceph_release_num.{{ ceph_release }} >=
    ceph_release_num.luminous

This patch sets a fact for `current_ceph_release_num` in the
`ceph-osd` role that can be re-used throughout the role.

### DIFF
--- a/roles/ceph-mgr/tasks/main.yml
+++ b/roles/ceph-mgr/tasks/main.yml
@@ -5,6 +5,12 @@
   when:
     - containerized_deployment
 
+- name: set a fact for the release number of the current release
+  set_fact:
+    current_ceph_release_num: "ceph_release_num.{{ ceph_release }}"
+  tags:
+    - always
+
 - name: include pre_requisite.yml
   include: pre_requisite.yml
   when: not containerized_deployment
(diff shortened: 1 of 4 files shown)


roles/ceph-mgr/tasks/main.yml
```diff
@@ -5,6 +5,12 @@
   when:
     - containerized_deployment
 
+- name: set a fact for the release number of the current release
```
leseb: We use this convention for set_fact:

```
- name: set_fact current_ceph_release_num - < explanation >
```

This is helpful when debugging is not turned on.

leseb: Also I think we want to have this fact in `ceph-common` so the rest of the role will inherit from it (when doing non-container deployments).

major: That makes sense. I wasn't sure if these roles were used independently without `ceph-common`.

I'll go back and add it to `ceph-common` first.

major:
Closing in favor of https://github.com/ceph/ceph-ansible/pull/2043.

PR #2043, opened 2017-10-12
Avoid jinja in `when` keys

major:
This series of patches does the following:

* Sets a `current_ceph_release_num` fact to hold the current ceph release number
* Adjusts tasks to use the `current_ceph_release_num` fact to avoid "jinja2 in when" warnings


major:
That's not a bad idea. Let me verify that.

major:
@guits That works! I'll adjust the patch.

major:
Revising the patch into something new.

PR #2050, opened 2017-10-13
ceph-defaults: fix handlers that are always triggered

guits:
Handlers are always triggered in ceph-ansible because ceph.conf file is
generated with a random order for the different keys/values pairs
in sections.

In python, a dict is not sorted. It means in our case each time we try
to generate the ceph.conf file it will be rendered with a random order
since the mechanism behind consist of rendering a file from a python dict
with keys/values. Therefore, as a quick workaround, forcing this dict to be
sorted before rendering the configuration file will ensure that it will be
rendered always the same way.

I don't know whether it should be fixed upstream, fixing this at ceph-ansible level means forking this plugin. However, it's not a huge modification and should be pretty easy to maintain or revert if ever we could find a better fix.
@leseb @andrewschoen any thoughts?

### DIFF
--- a/plugins/actions/_v1_config_template.py
+++ b/plugins/actions/_v1_config_template.py
@@ -26,7 +26,7 @@ from ansible import errors
 from ansible.runner.return_data import ReturnData
 from ansible import utils
 from ansible.utils import template
-
+from collections import OrderedDict
 
 CONFIG_TYPES = {
     'ini': 'return_config_overrides_ini',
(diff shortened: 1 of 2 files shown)


plugins/actions/_v1_config_template.py
```diff
@@ -27,6 +27,7 @@
 from ansible import utils
 from ansible.utils import template
 
+from collections import OrderedDict
```
leseb: why not putting it line 29?

plugins/actions/_v2_config_template.py
```diff
@@ -38,6 +38,7 @@
 from ansible import constants as C
 from ansible import errors
 
+from collections import OrderedDict
```
leseb: ditto

PR #2051, opened 2017-10-13
mds: fix fs pool creation

leseb:
1. add the variables to docker_collocation
2. trigger the check when a MDS is part of the inventory file, not when
we run on an MDS...


guits:
failures here are not related to this PR, merging anyway.

PR #2054, opened 2017-10-16
ceph-nfs - add bind address variable

jprovaznik:


roles/ceph-nfs/templates/ganesha.conf.j2
```diff
@@ -5,6 +5,14 @@
 %include /etc/ganesha/export.d/INDEX.conf
 {% endif %}
 
+NFS_Core_Param
+{
+{% if ceph_nfs_bind_addr %}
```
leseb: I think you want `if ceph_nfs_bind_addr is defined` instead, `if ceph_nfs_bind_addr` means `ceph_nfs_bind_addr` is true. This won't work since it's not.

jprovaznik:
fixed

PR #2055, opened 2017-10-16
upgrade: support for rbd mirror and nfs

leseb:
- Add upgrade support for rbd mirror and nfs daemons.
- Only works with systemd (remove sysvinit and upstart occurence)
- A bit of cleanup


leseb:

PR #2000, opened 2017-10-05
[skip ci] ci: new osd scenarios

leseb:
This commit add new osd scenarios, it aims to simplify the CI setup and
brings a better coverage on the OSD scenarios.
We decided to differentiate between filestore and bluestore, thinking
ahead when filestore won't be supported anymore.
So we now have two classes of tests:

* Filestore
* Bluestore

In each of those classes we have container and non-container.
Then for each we test the following:

* collocated
* collocated dmcrypt
* non-collocated
* non-collocated dmcrypt
* auto discovery collocated
* auto discovery collocated dmcrypt

This gives us a nice coverage and also reduces the footprint on the CI.
We are now up to 4 scenarios, each containing 6 OSD VMs.

Signed-off-by: Sébastien Han <seb@redhat.com>

PR #2060, opened 2017-10-17
Add set `devices` to empty list when `osd_auto_discovery` used

Arano-kai:
- `osd_auto_discovery` intend to setup OSD automatically, so one does not
need that `devices` contain any value.
- when `osd_auto_discovery` used, one may ommit `devices`

Signed-off-by: Arano-kai <captcha.is.evil@gmail.com>

### DIFF
--- a/roles/ceph-defaults/tasks/facts.yml
+++ b/roles/ceph-defaults/tasks/facts.yml
@@ -127,6 +127,13 @@
     - rbd_client_directory_mode is not defined
       or not rbd_client_directory_mode
 
+- name: set_fact devices to empty list
+  set_fact:
+    devices: []
+  when:
+    - osd_auto_discovery is defined
+    - osd_auto_discovery
+
 - name: resolve device link(s)
   command: readlink -f {{ item }}
   changed_when: false


roles/ceph-defaults/tasks/facts.yml
```diff
@@ -127,6 +127,13 @@
     - rbd_client_directory_mode is not defined
       or not rbd_client_directory_mode
 
+- name: set_fact devices to empty list
```
leseb: why not something like this? https://github.com/ceph/ceph-ansible/pull/2000/files#diff-2857b72f37476e1b5d16e7e81783ecbaR137

Arano-kai: Well, this is matter of taste... and work time =)
The `devices` may use elsewhere (not only in `ceph-defaults`), so, since we explicitly define var, no further changes needed.

roles/ceph-defaults/tasks/facts.yml
```diff
@@ -127,6 +127,13 @@
...
+  set_fact:
+    devices: []
+  when:
+    - osd_auto_discovery is defined
+    - osd_auto_discovery
```
berendt: I think ``not osd_auto_discovery|default(False)`` should be used here.

leseb:
@Arano-kai since handled in #2000, with https://github.com/ceph/ceph-ansible/pull/2000/files#diff-2857b72f37476e1b5d16e7e81783ecbaR137 would you consider closing this? #2000 will merge at the latest tomorrow. Thanks.

PR #2061, opened 2017-10-17
Fix `monitor_address_block` usage

Arano-kai:


roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -39,11 +39,9 @@ mon initial members = {% for host in groups[mon_group_name] %}
 mon host = {% if nb_mon > 0 %}
   {% for host in groups[mon_group_name] -%}
     {% if monitor_address_block | length > 0 %}
-      {% if ip_version == 'ipv4' -%}
-        {{ hostvars[host]['ansible_all_' + ip_version + '_addresses'] | ipaddr(monitor_address_block) | first }}
-      {%- elif ip_version == 'ipv6' -%}
-        [{{ hostvars[host]['ansible_all_' + ip_version + '_addresses'] | ipaddr(monitor_address_block) | first }}]
-      {%- endif %}
+      {%- set host_addrs = hostvars[host]['ansible_all_' + ip_version + '_addresses'] -%}
```
guits: @Arano-kai why did you remove the ipv4/ipv6 condition?
when ipv6, the ip address must be enclosed with '[]'

Arano-kai: ipv6 enclosure automatically done by `ipwrap` filter at 44

guits: @Arano-kai in that case, it could be nice to keep something consistent by fixing it everywhere

Arano-kai: That will be my next push request =)

leseb: Yes, please, let's try to not overload the CI if we can keep this in the same PR.
Thanks.

Arano-kai: Ok, but note that the changes will be in:

- `ceph-config/templates/ceph.conf.j2`
- `ceph-defaults/templates/restart_rgw_daemon.sh.j2`
- `ceph-mon/templates/ceph-mon.service.j2`
- `ceph-rgw/templates/ceph-radosgw.service.j2`

leseb: Sounds good to me.

roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -39,11 +39,9 @@ mon initial members = {% for host in groups[mon_group_name] %}
...
+      {%- set monitor_address_block = monitor_address_block | ipaddr('address') -%}
+      {{ monitor_address_block | difference(monitor_address_block | difference(host_addrs)) | first | ipwrap }}
```
guits: Are you sure this is actually working?
In any case, I think what drove you to write this, is the default value of `monitor_address_block` which is currently `[]` which is wrong. It should be `''`. Therefore, if `monitor_address_block` is a string you can keep something like that : 
```
{{ host_addrs | ipaddr(monitor_address_block) | first | ipwrap }}
```
Can you update your PR with what has been said?
Thanks!

Arano-kai: Could you give an example of the data that you expect in `monitor_address_block`?
This patch is tested on deploying CEPH from scratch on Debian (stretch)

guits: @Arano-kai 
for instance:
`monitor_address_block: '192.168.123.0/24'`

Arano-kai: Why not use `monitor_address` instead?
You can filter by `ipaddr('net')` and `ipaddr('host/address')` later and do corresponding branching.
In addition, the name `monitor_address_block` is misleading - no any reference to network...

guits: @Arano-kai `monitor_address_block` was introduced by bcb2e39.
Using `monitor_address` is not always possible because there are some cases where you can't predict which IP address will be assigned to your mons nodes.
Also, I just remind that we wanted to limit the use cases where `ipaddr()` filter would be used since there is a dependance with `python-netaddr` package.

I don't see what is misleading with `monitor_address_block`, it just means you have a block of ip addresses against which your mons will match an IP.

Given all we said, I'm not sure to understand what is your issue? Isn't enough using `monitor_address_block` as a string for your use case?

Arano-kai: Ok, let's clarify the situation.

> In any case, I think what drove you to write this, is the default value of monitor_address_block which is currently [] which is wrong. It should be ''. 

Currently it is not documented, defaults to `[]` (yes, by mistake) and evaluated by `length > 0` (which is common for lists/dicts, string common way is `var_name|trim == ''`)

> Also, I just remind that we wanted to limit the use cases where `ipaddr()` filter would be used since there is a dependance with `python-netaddr` package.
```
(remove 'l' to get in-line matches)
user@host:~/ansible/ceph-ansible$ grep -lr 'ipaddr' roles/
roles/ceph-config/templates/ceph.conf.j2
roles/ceph-defaults/templates/restart_rgw_daemon.sh.j2
roles/ceph-mon/tasks/docker/main.yml
roles/ceph-mon/templates/ceph-mon.service.j2
roles/ceph-rgw/templates/ceph-radosgw.service.j2
```
So, to move out from `python-netaddr`, we need replace` ipaddr` in the listed files by some crutches (and catch every bug in it) and use it along with ipv4 / 6 branching in every place, where ip addresses occures.
Or, we can use `ipaddr`, `ipwrap` and others to reduce code complexity (and bugs), lines etc., just placing notice in documentation and few asserts (in `ceph-defaults` and all affected roles).
Choose your pill, but be warned, that use of `python-netaddr` is mandatory for now, since `ceph-mon` affected (and it is core daemon).

> I don't see what is misleading with monitor_address_block, it just means you have a block of ip addresses against which your mons will match an IP.

Lets see:
- `'192.24.2.1/32', '::1/128', 'fe80::100/10'` is block of some ip addresses, but we can't use it in `ipaddr` - it's a list...
- `192.168.123.0/24` is *(sub)network address* <-- worldwide used terminology. You can get *range* of ip addresses from it. Also, you can replace word *range* by *block*, but it is not commonly used.

So, the `monitor_address_block` is misleading,  `monitor_address_net(work)` is much better name I think (same apply to `radosgw_address_block`).

> Given all we said, I'm not sure to understand what is your issue? Isn't enough using monitor_address_block as a string for your use case?

Sorry, English is not my native language, it is sometimes difficult to understand and be understood (Google translation also contributes (: ).
And yes, I think ability to use net address is great and must be implemented asap, but not in hard way, not increasing spaghetti in code and not expanding config by one more var when possible.

Thank you for understanding and sorry, if I broke something.

guits: @Arano-kai I don't mind changing the name of this variable and/or bring some changes in there. I guess there is some legacy and/or leftover behind this and I'm just trying to understand what is your need and how we can move forward with this without breaking anything existing.
Also, if you want, you can find me on IRC (oftc/#ceph) to chat further about that.

leseb:
Closing this due to inactivity, feel free to re-open, thanks!

PR #2062, opened 2017-10-17
[skip ci] Cleanup readme files in roles directories

berendt:
The contents of the README files are no longer up to date.
Documentation for all roles is located below the docs directory.


berendt:
To avoid any confusion, these old files should be removed.

berendt:
Not sure about the Ansible Galaxy integration. At least https://galaxy.ansible.com/ceph/ceph-common/ looks pretty old.

leseb:
Thanks

Issue #2087, opened 2017-10-20
test ceph-ansible with Ansible 2.4


ktdreyer:
RHEL 7 Extras has Ansible 2.4.0.0, and CentOS Extras will soon follow after that.

We should have Jenkins validate that 2.4.0.0 works. Looks like tox.ini only tests ansible 2.2 and 2.3 at the moment.

leseb:
Does that mean, stable-3.0 must work with Ansible 2.4.x?

ktdreyer:
Yes

PR #2090, opened 2017-10-23
[skip ci] Test ansible 2.4.1

leseb:
Closes: https://github.com/ceph/ceph-ansible/issues/2087

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,5 +1,4 @@
 # These are Python requirements needed to run the functional tests
-# 1.6.1 fails with 'testinfra is in an unsupported or invalid wheel'
-# see https://github.com/philpep/testinfra/issues/201
-testinfra==1.6.0
+# testinfra < 1.7.1 does not work Ansible 2.4, see https://github.com/philpep/testinfra/issues/249
+testinfra==1.7.1
 pytest-xdist
--- a/tox.ini
+++ b/tox.ini
@@ -149,6 +149,7 @@ setenv=
 deps=
   ansible2.2: ansible==2.2.3
   ansible2.3: ansible==2.3.1
+  ansible2.4: ansible==2.4.1
   -r{toxinidir}/tests/requirements.txt
 changedir=
   # tests a 1 mon, 1 osd, 1 mds and 1 rgw xenial cluster using non-collocated OSD scenario


leseb:
Blocked by https://github.com/ansible/ansible/issues/31483 fix is https://github.com/ansible/ansible/pull/31514 and apparently backported in stable-2.4 and will be in 2.4.1

GregMeno:
@leseb Would you please be a little more specific about "blocked", do you mean that we can't install, can't upgrade, something else?

leseb:
@GregMeno, we cannot install Ceph.

tests/requirements.txt
```diff
@@ -1,5 +1,3 @@
 # These are Python requirements needed to run the functional tests
-# 1.6.1 fails with 'testinfra is in an unsupported or invalid wheel'
-# see https://github.com/philpep/testinfra/issues/201
-testinfra==1.6.0
+testinfra==1.7.1
```
ktdreyer: Does testinfra 1.7 not work with the latest Ansible 2.4? The commit message makes no mention of this

leseb: Correct, it does not work with 1.6, https://github.com/philpep/testinfra/issues/249.
Will edit the commit. Thanks Ken.

tox.ini
```diff
@@ -148,7 +148,7 @@ setenv=
   lvm_osds: CEPH_STABLE_RELEASE = luminous
 deps=
   ansible2.2: ansible==2.2.3
-  ansible2.3: ansible==2.3.1
+  ansible2.3: ansible==2.4.0
```
ktdreyer: "ansible2.3" now points at "2.4.0". would you please explain why we don't add a new "ansible2.4" line here instead?

My guess is that it's so that Jenkins will run the "ansible2.3" scenarios quickly? What will the final commit to master look like?

ktdreyer:
As far as the the `rpm_key` module goes:

Looks like it's used two places, one to install the download.ceph.com GPG key, and one to install Red Hat's key when doing an RHCS ISO-based installation.

The step to install the GPG key from the ISO should be unnecessary. It's just installing `RPM-GPG-KEY-redhat-release` from the ISO, and that key should already be present on all RHEL systems. My guess is that this was an ancient hold-over from when we signed ISO contents with Inktank's key instead of Red Hat's main key. I think we could delete this `rpm_key` step entirely from `roles/ceph-common/tasks/installs/prerequisite_rhcs_iso_install.yml`

As far as the upstream goes (`roles/ceph-common/tasks/installs/redhat_community_repository.yml`):
I wonder if we could delete this `rpm_key` module statement too. We're already configuring Yum to trust it in the next step with `gpgkey:` in the `yum_repository` module. When Yum has a key configured like this in (for example) `/etc/yum.repos.d/foo.repo`, and you run "yum install <package>" with the "-y" option, Yum will automatically import that key if it is not present. There is no need to call `rpm --import` as a separate step.

tox.ini
```diff
@@ -148,7 +148,7 @@ setenv=
   lvm_osds: CEPH_STABLE_RELEASE = luminous
 deps=
   ansible2.2: ansible==2.2.3
-  ansible2.3: ansible==2.3.1
+  ansible2.3: ansible==2.4.1
```
andrewschoen: Could we please add an ``ansible2.4`` factor and leave ``ansible2.3`` as-is? Thanks.

andrewschoen: I've made this PR to enable ``ansible2.4`` in the CI: https://github.com/ceph/ceph-build/pull/897

leseb: yes yes, this was just a quick test to run 2.4.1, that PR is not complete.

leseb:
Adding `[skip ci] ` as the CI already demonstrated success on 2.4.

tox.ini
leseb: done

PR #2091, opened 2017-10-23
[skip ci] fix ansible_distribution is undefined bug

yanyixing:
fix ansible_distribution is undefined bug

### DIFF
--- a/site.yml.sample
+++ b/site.yml.sample
@@ -14,7 +14,7 @@
   - mgrs
   - iscsi-gws
 
-  gather_facts: false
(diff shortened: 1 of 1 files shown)


leseb:
That's not the right way to fix this error, which Ansible version are you using? We must use `2.3.1` or `2.3.2`.

yanyixing:
It's my ansible version problem, my ansible version is 2.3.0. After updating ansible version to 2.3.1, the problem is not found.
And please update the doc for branch stable-3.0

PR #2098, opened 2017-10-23
Add option to use jemalloc for Debian

andymcc:
Use "ceph_jemalloc_enabled" to enable jemalloc as the default memory
allocator for Ceph on Debian installs. This defaults to false, so no
change to existing deploys, but can be set to install and configure
jemalloc as the default memory allocator.

If the LD_PRELOAD line is changed in /etc/default/ceph it will cause a
restart of the ceph services so that jemalloc is in use.

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -460,6 +460,12 @@ dummy:
 #  - { name: vm.swappiness, value: 10 }
 #  - { name: vm.min_free_kbytes, value: "{{ vm_min_free_kbytes }}" }
 
+# For Debian & Red Hat/CentOS installs set jemalloc as the memory allocator for Ceph.
+# jemalloc is generally faster for small IO workloads and when
+# ceph-osd is backed by SSDs. However, memory usage is usally higher.
+# Don't enable this if using osd_objectstorage: bluestore.
+#ceph_jemalloc_enabled: false
+
 
 ##########
 # DOCKER #
--- a/group_vars/rhcs.yml.sample
+++ b/group_vars/rhcs.yml.sample
@@ -460,6 +460,12 @@ ceph_repository: rhcs
 #  - { name: vm.swappiness, value: 10 }
 #  - { name: vm.min_free_kbytes, value: "{{ vm_min_free_kbytes }}" }
 
+# For Debian & Red Hat/CentOS installs set jemalloc as the memory allocator for Ceph.
+# jemalloc is generally faster for small IO workloads and when
+# ceph-osd is backed by SSDs. However, memory usage is usally higher.
+# Don't enable this if using osd_objectstorage: bluestore.
+#ceph_jemalloc_enabled: false
+
 
 ##########
 # DOCKER #
--- a/roles/ceph-common/tasks/configure_memory_allocator.yml
+++ b/roles/ceph-common/tasks/configure_memory_allocator.yml
@@ -1,0 +1,50 @@
+---
+- name: install jemalloc for debian
+  apt:
+    name: "libjemalloc1"
+    update_cache: no
+    state: "{{ (upgrade_ceph_packages|bool) | ternary('latest','present') }}"
+  when:
+    - ansible_os_family == 'Debian'
+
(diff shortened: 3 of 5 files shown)


roles/ceph-common/tasks/installs/install_on_debian.yml
```diff
@@ -27,3 +27,9 @@
   when:
     - (ceph_origin == 'repository' or ceph_origin == 'distro')
     - ceph_repository == 'rhcs'
+
+- name: include install_debian_jemalloc.yml
+  include: install_debian_jemalloc.yml
+  when:
+    - (ceph_origin == 'repository' or ceph_origin == 'distro')
+    - ceph_jemalloc_enabled | default(False)
```
leseb: you can remove `| default(False)` and add `ceph_jemalloc_enabled` to roles/ceph-defaults/default/main.yml then run `generate_group_vars_sample.sh`.

Thanks.

andymcc:
Should be fixed now!
One note - the script makes some other changes to the rhcs.yml.sample, i reverted those changes, so the only changes here are to do with the ceph_jemalloc_enabled var.

leseb:
No worries, leave the modification as is, the `rhcs.yml.sample` should stay in sync :).
Thanks!

guits:

andymcc:
@guits Thanks for the review - I changed it up to use one task and ternary instead. We can avoid the with_items loop in that case (and the ignore_errors: True).

andymcc:
Ahh nice spot! will fix that up - i'm looking into adding RHEL/CentOS support for this.

andymcc:
I reworked it slightly - it now works for CentOS/RHEL too - and I included a conditional on bluestore.
Lastly, I'd like to add a follow-up (if this is accepted) to adjust tcmalloc thread cache size, which is why I've called the task file "configure_memory_allocator", I'll just add an additional task in there if that works.

The ordering works well too, since we don't need to stat /etc/default/ceph anymore since it will have run that already (tags: always).

ktdreyer:
Note Sage just dropped support for this from /etc/default/ceph in Ceph master/Mimic, because jemalloc crashes the OSD if Bluestore is enabled. https://github.com/ceph/ceph/pull/18486 https://github.com/ceph/ceph/pull/18487

That will probably be backported to Luminous too.

ktdreyer:
By the way, does jemalloc override tcmalloc when using the default builds from download.ceph.com? From talking with Kyle Bader a few years ago, the `--with-tcmalloc` build configure option overrides `LD_PRELOAD=....jemalloc.so`, so it might appear to work, but Ceph still uses tcmalloc.

andymcc:
@ktdreyer - good feedback, let me test, I thought I had run it with the download.ceph.com packages - and it was using the LD_PRELOAD, since jemalloc was being used instead of tcmalloc. I'll do another test deploy just to confirm that. Otherwise I think the bluestore issue is resolved in that the conditional avoids that.

Just as a point of clarity - the default is still to not deploy jemalloc, you would have to actively go and enable jemalloc.

ktdreyer:
It would be a good idea to check with the developers on the best way to ensure that jemalloc is in use and tcmalloc is truly disabled. When Kyle last looked at this (2015) the environment variable is not enough. It makes me wonder if jemalloc ever really worked with the main download.ceph.com builds.

andymcc:
@ktdreyer you can check using lsof. it'll show which memory allocator is being used. In hammer and before I don't think you could just enable it, although I haven't tested that. In jewel+ it definitely does work, I'll try confirm with the download.ceph.com packages but I think that is the case for those too.

andymcc:
You can also use ldd on the binary to see what libraries its using.
The ceph.download.com packages are compiled with tcmalloc only - with that in mind I'm going to close this although I think there is use, if you have to build your own packages for this to become useful it makes it much harder to see a point of putting it in ceph-ansible. I'll post the second half of what I had planned to add in the "tc_malloc" settings in a separate PR! Thanks for the feedback @ktdreyer

ktdreyer:
Thanks Andy for researching this!

PR #2103, opened 2017-10-25
Option to set TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES

andymcc:
Use "ceph_tcmalloc_max_total_thread_cache" to set the
TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES value inside /etc/default/ceph for
Debian installs, or /etc/sysconfig/ceph for Red Hat/CentOS installs.

By default this is not set, so the default package value will be used,
if specified this value will be changed to match the variable, and ceph
services will be restarted.

### DIFF
--- a/roles/ceph-common/tasks/configure_memory_allocator.yml
+++ b/roles/ceph-common/tasks/configure_memory_allocator.yml
@@ -1,0 +1,25 @@
+---
+- name: configure TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES for debian
+  lineinfile:
+    dest: "{{ etc_default_ceph.stat.isdir | ternary('/etc/default/ceph/ceph', '/etc/default/ceph') }}"
+    insertafter: EOF
+    create: yes
+    regexp: "^TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES="
+    line: "TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES={{ ceph_tcmalloc_max_total_thread_cache }}"
+  when:
+    - ansible_os_family == 'Debian'
+    - etc_default_ceph.stat.exists
+  notify:
+    - restart ceph osds
+
+- name: configure TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES for redhat
+  lineinfile:
+    dest: "/etc/sysconfig/ceph"
+    insertafter: EOF
+    create: yes
+    regexp: "^TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES="
+    line: "TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES={{ ceph_tcmalloc_max_total_thread_cache }}"
+  when:
+    - ansible_os_family == 'RedHat'
+  notify:
+    - restart ceph osds
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -90,3 +90,10 @@
 
 - name: include configure_cluster_name.yml
   include: configure_cluster_name.yml
+
+- name: include configure_memory_allocator.yml
+  include: configure_memory_allocator.yml
+  when:
+    - (ceph_tcmalloc_max_total_thread_cache | int) > 0
+    - osd_objectstore == 'filestore'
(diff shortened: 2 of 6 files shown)


roles/ceph-common/tasks/configure_memory_allocator.yml
```diff
@@ -0,0 +1,35 @@
...
+- name: configure TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES for debian
+  lineinfile:
+    dest: "{{ etc_default_ceph.stat.isdir | ternary('/etc/default/ceph/ceph', '/etc/default/ceph') }}"
+    insertafter: EOF
+    create: yes
+    regexp: "^TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES="
+    line: "TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES={{ ceph_tcmalloc_max_total_thread_cache }}"
+  when:
+    - etc_default_ceph.stat.exists
+    - ansible_os_family == 'Debian'
```
leseb: It's probably easier if you move the Debian condition above.

roles/ceph-defaults/defaults/main.yml
```diff
@@ -452,6 +452,9 @@ os_tuning_params:
   - { name: vm.swappiness, value: 10 }
   - { name: vm.min_free_kbytes, value: "{{ vm_min_free_kbytes }}" }
 
+# For Debian & Red Hat/CentOS installs set TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES
+#ceph_tcmalloc_max_total_thread_cache: 134217728
```
leseb: Please uncomment it otherwise the tasks above will fail.

roles/ceph-common/tasks/main.yml
```diff
@@ -90,3 +90,9 @@
 
 - name: include configure_cluster_name.yml
   include: configure_cluster_name.yml
+
+- name: include configure_memory_allocator.yml
+  include: configure_memory_allocator.yml
+  when:
+    - (ceph_origin == 'repository' or ceph_origin == 'distro')
+    - ceph_tcmalloc_max_total_thread_cache is defined
```
leseb: Also, we need a condition on filestore only `osd_objectstore: filestore`.

roles/ceph-common/tasks/configure_memory_allocator.yml
```diff
@@ -0,0 +1,35 @@
...
+  notify:
+    - restart ceph mons
+    - restart ceph osds
```
leseb: Let's restart the OSDs only since this is only affecting OSDs

roles/ceph-defaults/defaults/main.yml
andymcc: d'oh - of course, initially i had wanted to use the package defaults - which arent necessarily 128MB - will fix.

roles/ceph-common/tasks/main.yml
andymcc: Does bluestore not use tcmalloc? I thought it did - I havn't messed about with it to confirm - easy enough to fix this, just wanted to confirm.

leseb: Yup, please leave it only for filestore.

andymcc:
Fixed those up.

1 note, I used a default of 0 (which will basically mean it gets skipped) - I'm trying to avoid a situation where existing clusters (on upgrades) run this and have their osds restarted for no reason, or worse they have tuned that value to a different setting and it gets force changed, and osds restarted - happy to change that but I'd rather not cause impact to running/existing clusters without the deployer/operator making a distinct choice (e.g. setting the var to non-default value).

roles/ceph-osd/templates/ceph-osd-run.sh.j2
```diff
@@ -62,6 +62,9 @@ expose_partitions "$1"
   {% endif -%}
   -e CLUSTER={{ cluster }} \
   -e OSD_DEVICE=/dev/${1} \
+  {% if (ceph_tcmalloc_max_total_thread_cache | int) > 0 -%}
```
leseb: and `osd_objectstore == 'filestore'`

andymcc:
one of these days ;)

PR #2105, opened 2017-10-25
Really fix always run

major:
These patches revert the original `always_run` fix and use `check_mode: no` instead.


major:
This PR should fix the errant logic from #2037.

towolf:
LGTM.

Issue #2092, opened 2017-10-23
make keys['mode'] an optional, not mandatory key


fultonj:
The openstack_keys structure [1] requires a key a called mode and if it's not included the playbook has an error because it assumes the key always exists. It should be optional, not mandatory. This affects TripleO [2]. 

[1] https://github.com/ceph/ceph-ansible/commit/7d429410901601a0b29d2be079aaf2bcc7c2bb94#diff-8c864abe7cb1a4c982cda60067561a1a
[2] https://review.openstack.org/#/c/508975/

gfidente:
To minimize the change, would it be viable to to reorder the tasks so that chmod always runs before setfacl?

fultonj:
I'm testing a change in my fork [1] and will make it into a PR for this issue after I've verified it. 

[1] https://github.com/fultonj/ceph-ansible/commit/4ae0bc11560d347829fade6165c370a25af1d74d

fultonj:
I revised my change to remove "- openstack_keys.acls is defined" from the "setfacl for openstack key(s)" task and I am now able to handle all of the provided input without error. However, I have an issue with getting it to do the right thing with the mode param when tested in combinations: 

My testing process is: 

1. reset: `setfacl -b ceph.client.openstack.keyring ; chmod 000 ceph.client.openstack.keyring`
2. test
```
try-just-acl.sh				 good
try-just-mode.sh			 bad
try-mode-then-acl.sh			 bad
try-no-acl-no-mode.sh			 good
```
3. the good/bad above is a result of runing the following after: `getfacl ceph.client.openstack.keyring ; ls -l ceph.client.openstack.keyring`

So there's a problem with the mode work that I'm looking into. 

The scripts for the 4 scenarios are: 

```
(undercloud) [stack@undercloud ansible-directly]$ ls try-*
try-just-acl.sh  try-just-mode.sh  try-mode-then-acl.sh  try-no-acl-no-mode.sh
(undercloud) [stack@undercloud ansible-directly]$ for f in $(ls try-*); do (echo $f && cat $f) | curl -F 'sprunge=<-' http://sprunge.us ; done  
http://sprunge.us/iQCA
http://sprunge.us/CbDH
http://sprunge.us/eShJ
http://sprunge.us/LBXJ
(undercloud) [stack@undercloud ansible-directly]$ 
```
I think I can have this fixed soon (would have been sooner but got pinged many time today).

PR #2109, opened 2017-10-25
Issue/2092

fultonj:
Make acls and mode parameters of opentack_keys optional

Only chmod or setfacl the requested keyring(s) in the opentack_keys data structure when the mode or acls keys of that data structure exist.

User may specify four permission combinations for the keyring file(s): 1. only set ACL, 2. only set mode, 3. set neither mode nor ACL, 4. set mode and then ACL.

Fixes: #2092


fultonj:
This PR got messed up by how I re-based it. Sorry about that. I will send another.

PR #2110, opened 2017-10-25
Make acls and mode parameters of opentack_keys optional

fultonj:
Only chmod or setfacl the requested keyring(s) in the
opentack_keys data structure when the mode or acls keys
of that data structure exist.

User may specify four permission combinations for the
keyring file(s): 1. only set ACL, 2. only set mode,
3. set neither mode nor ACL, 4. set mode and then ACL.

Fixes: #2092

### DIFF
--- a/roles/ceph-client/tasks/create_users_keys.yml
+++ b/roles/ceph-client/tasks/create_users_keys.yml
@@ -70,6 +70,15 @@
     docker_exec_client_cmd:
   when: docker_exec_client_cmd == 'ceph'
 
+- name: chmod key(s)
+  file:
+    path: "/etc/ceph/{{ cluster }}.{{ item.name }}.keyring"
+    mode: "{{ item.mode|default(omit) }}" # if mode not in list, uses mode from ps umask
+  with_items: "{{ keys }}"
+  when:
+    - cephx
+    - keys | length > 0
+
 - name: setfacl for key(s)
   acl:
     path: "/etc/ceph/{{ cluster }}.{{ item.0.name }}.keyring"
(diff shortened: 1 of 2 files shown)


fultonj:
FYI @guits @leseb @gfidente

roles/ceph-client/tasks/create_users_keys.yml
```diff
@@ -70,6 +70,16 @@
...
 
+- name: chmod key(s)
+  file:
+    path: "/etc/ceph/{{ cluster }}.{{ item.name }}.keyring"
+    mode: "{{ item.mode|default(omit) }}" # if mode not in list, uses mode from ps umask
+  with_items: "{{ keys }}"
+  when:
+    - cephx
+    - keys | length > 0
+    - keys.mode is defined
```
gfidente: should this be item.mode ?

roles/ceph-mon/tasks/openstack_config.yml
```diff
@@ -50,6 +50,15 @@
...
 
+- name: chmod openstack key(s)
+  file:
+    path: "/etc/ceph/{{ cluster }}.{{ item.name }}.keyring"
+    mode: "{{ item.mode|default(omit) }}" # if mode not in list, uses mode from ps umask
+  with_items: "{{ openstack_keys }}"
+  when:
+    - openstack_config
+    - cephx
+
```
gfidente: should this account for item.mode as well?

fultonj: Thanks, for pointing this out as I had to clean out a "keys.mode is defined" in create_users_keys.yml which I did in this latest revision. 

In this case, I'm relying on default(omit) [1] to handle the case where the key is not in the data structure. 

[1] http://docs.ansible.com/ansible/latest/playbooks_filters.html#omitting-parameters

Issue #2139, opened 2017-11-05
For non containerized_deployment,  osd_crush_location cannot be set


mslovy:
we set those in the inventory like below:

[osds]
cephsata-0-rack2-1 ceph_crush_root=HDD ceph_crush_pod=HDD-Replication-Domain-0 ceph_crush_rack=HDD-RACK-2
cephsata-1-rack3-1 ceph_crush_root=HDD ceph_crush_pod=HDD-Replication-Domain-0 ceph_crush_rack=HDD-RACK-3
cephsata-2-rack4-1 ceph_crush_root=HDD ceph_crush_pod=HDD-Replication-Domain-0 ceph_crush_rack=HDD-RACK-4

and  enable the param:
crush_location: true
osd_crush_location: "\"root={{ ceph_crush_root }} rack={{ ceph_crush_rack }} host={{ ansible_hostname }}\""

however, osd_crush_location is not properly set for each osd. I find the code in master just config osd_crush_location for containerized_deployment, am I miss something or something wrong here, thanks?

guits:
Closed due to inactivity, feel free to re-open if needed. Thanks!

PR #2146, opened 2017-11-08
osd: fix crush location for non-containerized deployment

mslovy:
crush location only set for containerized deployment

Signed-off-by: yaoning <yaoning@unitedstack.com>


mslovy:
Related to this issus: https://github.com/ceph/ceph-ansible/issues/2139

PR #2151, opened 2017-11-09
Add openSUSE Leap 42.3 support

hwoarang:
This PR adds support for openSUSE Leap 42.3 support. The Ceph packages can either be installed from the official repositories or from the OBS one.

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -80,6 +80,12 @@ dummy:
 #  - hdparm
 #  - python-setuptools
 
+#suse_package_dependencies:
+#  - python-pycurl
+#  - python-xml
+#  - hdparm
+#  - python-setuptools
+
 # Whether or not to install the ceph-test package.
 #ceph_test: false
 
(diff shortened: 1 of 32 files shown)


guits:
@leseb @andrewschoen @alfredodeza 
any thoughts on this PR?

hwoarang:
Does anyone know what needs to be done to add a Jenkins job for openSUSE as well so we can test it in the same way as the rest of the distros?

alfredodeza:
we don't have the ability to test openSUSE as it is not even part of the Openstack provider we use to launch ephemeral builders. If we did, we would need to change a bunch of services to support new configuration

roles/ceph-docker-common/tasks/misc/ntp_suse.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: include ../checks/check_ntp_suse.yml
```
leseb: Where is  `check_ntp_suse.yml`?
Also, can you merge this with `ntp_redhat.yml`?

hwoarang:
@leseb I think it should be OK now
@alfredodeza oh that's a shame. However, we can still test openSUSE + Ceph in the OpenStack Ansible project so we can capture potential future failures.

PR #2153, opened 2017-11-09
osd: always run disk_list test

guits:
there is no need to have a condition on this task, this test should be
always run since the result will be interpreted later.


guits:
@leseb I don't think it's wrong, it's just no longer necessary since we ensure the test is always run.

it means `disk_list.get('rc')` is enough i think

PR #2156, opened 2017-11-09
config: add work-around for host-specific ceph_conf_overrides evaluation

eduardegorov:
This allows us to use host-specific variables in ceph_conf_overrides variable (e.g. for rados gateway configuration variables, see profiles/rgw-keystone-v3).

Signed-off-by: Eduard Egorov <eduard.egorov@icl-services.com>

### DIFF
--- a/roles/ceph-config/tasks/main.yml
+++ b/roles/ceph-config/tasks/main.yml
@@ -13,19 +13,28 @@
       - /etc/ceph/ceph.d/
 
   - name: template ceph_conf_overrides
-    local_action: copy content="{{ ceph_conf_overrides }}" dest="{{ fetch_directory }}/ceph_conf_overrides_temp"
-    become: false
-    run_once: true
+    copy:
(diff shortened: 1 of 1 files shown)


roles/ceph-config/tasks/main.yml
```diff
@@ -12,20 +12,31 @@
       - /etc/ceph/
       - /etc/ceph/ceph.d/
 
+### Work-around for host-specific ceph_config_overrides generation
```
leseb: please remove this comment

eduardegorov: @leseb done. Sorry, I've just re-pushed a branch... Should I add fixes as additional commits in a pull request in the future?

eduardegorov: Update commit's message as well

leseb: You can keep the same commit if that's fixing the same thing. Thanks.

PR #2163, opened 2017-11-10
rpm: require ansible 2.4.1.0

ktdreyer:
2.4.0.0 has some bugs, and we're going to ship with v2.4.1.0.


guits:
I think there is no need to wait for the CI for this

PR #2177, opened 2017-11-15
Allow to use rados for ganesha exports

jprovaznik:


roles/ceph-nfs/tasks/start_nfs.yml
```diff
@@ -1,4 +1,19 @@
 ---
+- name: check if rados index object exists
+  shell: "rados -p {{ cephfs_data }} ls|grep ganesha_rados_export_index"
```
ajarr: Are you hard coding the name of the object used as an index of RADOS export object URLS? It needs to be a configurable as in the manila driver, right?

group_vars/nfss.yml.sample
```diff
@@ -44,6 +44,11 @@ dummy:
 ######################
 #ceph_nfs_log_file: "/var/log/ganesha/ganesha.log"
 #ceph_nfs_dynamic_exports: false
+# if set to true then rados is used to store dynamic exports
```
ajarr: s/dynamic exports/ganesha exports?

Why are exports stored in RADOS being called dynamic?

jprovaznik: fixed

roles/ceph-nfs/tasks/start_nfs.yml
jprovaznik: Not necessarily (this value is currently hardcoded in templates) but you are right it would be better to make this configurable. I'll update it in this PR.

jprovaznik:
@leseb, this is a new feature (not even yet in released luminous version), I think it would be safer to hold on with making this default option.

leseb:
@jprovaznik fair enough

group_vars/nfss.yml.sample
```diff
@@ -44,6 +44,14 @@ dummy:
...
 #ceph_nfs_log_file: "/var/log/ganesha/ganesha.log"
 #ceph_nfs_dynamic_exports: false
+# If set to true then rados is used to store ganesha exports
+# and client sessions information, this is useful if you
+# run multiple nfs-ganesha servers in active/passive mode and
+# want to do failover
+#ceph_nfs_rados_backend: false
+# Name of the rados object used to store a list of the export rados
+# object URLS
+#ceph_nfs_rados_export_index: "ganesha_rados_export_index"
```
ajarr: s/ganesha_rados_export_index/ganesha-export-index/

'ganesha-export-index' is the default value I've set in the ganesha driver patch. Better to match?

ajarr:
You're hardcoding the ceph's conf file as '/etc/ceph/ceph.conf'. I'm not sure if that's OK in the ceph-ansible world. Otherwise looks OK to me. Thanks, Jan.

roles/ceph-nfs/tasks/start_nfs.yml
```diff
@@ -1,4 +1,19 @@
 ---
+- name: check if rados index object exists
+  shell: "rados -p {{ cephfs_data }} ls|grep {{ ceph_nfs_rados_export_index }}"
```
leseb: `--cluster {{ cluster }}`

roles/ceph-nfs/tasks/start_nfs.yml
```diff
@@ -1,4 +1,19 @@
...
+  changed_when: false
+  failed_when: false
+  register: rados_index_exists
+  check_mode: no
+  when:
+    - ceph_nfs_rados_backend
+
+- name: create an empty rados index object
+  shell: "echo | rados -p {{ cephfs_data }} put {{ ceph_nfs_rados_export_index }} -"
```
leseb: ditto

PR #2189, opened 2017-11-16
Make openstack_keys param support no acls list

fultonj:
A recent change [1] required that the openstack_keys
param always containe an acls list. However, it's
possible it might not contain that list. Thus, this
param sets a default for that list to be empty if it
is not in the structure as defined by the user.

[1] d65cbaa53952269ec9a2e76fca8203ce7ad22c2b


guits:
related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1506625

PR #2191, opened 2017-11-16
Make openstack_keys param support no acls list

guits:
A recent change [1] required that the openstack_keys
param always containe an acls list. However, it's
possible it might not contain that list. Thus, this
param sets a default for that list to be empty if it
is not in the structure as defined by the user.

[1] d65cbaa53952269ec9a2e76fca8203ce7ad22c2b

(cherry picked from commit d73f751b6372cbd60ddd6a24ae53232307b32fac)


guits:
failure here is due to yum repository failure, merging anyway

Issue #2192, opened 2017-11-16
firewalld needs to be disabled on Centos 7.4 and ceph ansible stable 3.0


Masber:
ceph cluster ends in ERROR state due to firewalld is blocking OSDs traffic. Several errors like 

> 2017-11-17 10:18:12.101651 7fb658ccb700 -1 osd.0 71 heartbeat_check: no reply from 10.0.32.58:6815 osd.7 ever on either front or back, first ping sent 2017-11-16 19:22:44.900585 (cutoff 2017-11-17 10:17:52.101594)

can be seen in the log files.

as a work around firewalld can be disabled

leseb:
This is not a ceph-ansible issue. At present, this is your own responsibility to configure firewalls. We are targetting 3.1 to configure firewalld if available on the system.

eduardegorov:
@leseb Hi, Sebastien! You might be interested on how we implemented it - please see https://github.com/ceph/ceph-ansible/pull/2196

guits:
Closed due to inactivity, feel free to re-open if needed. Thanks!

PR #2196, opened 2017-11-17
firewall: configure firewalld if it's already installed on the host (…

eduardegorov:
…#2192).

Signed-off-by: Eduard Egorov <eduard.egorov@icl-services.com>


eduardegorov:
@leseb , pushed as an additional commit(see above):
  - mds: As far as I understand, mds nodes requires the same range as osd ones. Added.
  - rbd-mirror: didn't find anything useful on this in the internet - are usual osd port ranges fine here too?
  - mgr: it looks like that input ports depend on what kind of plugins are installed, right? Skipped it.
  - iscsi -  opened 3260/tcp and 5000/tcp (api?) (took from (https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/2/html/block_device_guide/using_an_iscsi_gateway_technology_preview) )
  - nfs - enabled default 'nfs.xml' service definition (2049/tcp) and 111/tcp port additionally. Do we need UDP ones as well?
  - restapi: used 'restapi_port' variable from 'group_vars/all.yml'

PR #2207, opened 2017-11-22
common: install ceph-common on all the machines

leseb:
Since some daemons now install their own packages the task checking the
ceph version fails on Debian systems. So the 'ceph-common' package must
be installed on all the machines.

### DIFF
--- a/roles/ceph-common/tasks/installs/install_debian_rhcs_packages.yml
+++ b/roles/ceph-common/tasks/installs/install_debian_rhcs_packages.yml
@@ -1,4 +1,9 @@
 ---
+- name: install red hat storage ceph-common for debian
+  apt:
+    pkg: ceph-common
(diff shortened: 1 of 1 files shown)


guits:
merged without CI passing since it's not covered

ktdreyer:
Would you please clarify "some daemons now install their own packages"? I'm not able to parse that sentence

guits:
@ktdreyer it means that some roles in ceph-ansible handle themselves the package installation for the corresponding daemon, basically the task has moved from `ceph-common` role to `ceph-mds`

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mds/tasks/non_containerized.yml#L2-L17

ktdreyer:
Thanks @guits, that helps.

PR #2211, opened 2017-11-25
Set tighter permissions on keyrings when containerized

fultonj:
During a containerized deployment, set the permissions of ceph.client.admin.keyring and other keyrings to chmod 600 and chown it to ceph.

### DIFF
--- a/roles/ceph-defaults/tasks/facts.yml
+++ b/roles/ceph-defaults/tasks/facts.yml
@@ -151,3 +151,17 @@
     - inventory_hostname in groups.get(osd_group_name, [])
     - not osd_auto_discovery|default(False)
     - osd_scenario != 'lvm'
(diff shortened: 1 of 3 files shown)


fultonj:

fultonj:

fultonj:

fultonj:
jenkins test Docs

fultonj:

fultonj:

guits:
@fultonj user `ceph` doesn't exist on the host machine, we could use numerical value instead but it differs according the distribution you are running `64045` for debian based distributions vs. `167` for rhel based distributions. It means we would have to split the task to do that.

fultonj:
ok, thanks. I'll update this with that within the week.

PR #2214, opened 2017-11-27
handlers: restart daemons only if docker is running

guits:
In case where docker CLI is available but docker is not running, we
don't want to trigger the restart of the daemons.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1510555


dmesser:
I can confirm this patch works in my environment.

PR #2215, opened 2017-11-28
Add support for using loopback devices as OSDs

squidboylan:
This is particularly useful in CI environments where you dont have
the option of adding extra devices or volumes to the host. It is also
a simple change to support loopback devices.


guits:
@leseb this is also fixing a bug, should we backport it to stable-3.0?

leseb:
@guits yup!

squidboylan:
Backport PR for this is #2220

PR #2226, opened 2017-11-30
Skip mklabel gpt if already gpt

andymcc:


roles/ceph-osd/tasks/check_gpt.yml
```diff
@@ -17,3 +17,4 @@
   when:
     - not item.0.get("skipped")
     - item.0.get("rc", 0) != 0
+    - '"Partition Table: gpt" not in item[0].stdout'
```
leseb: I suspect this check could replace the one at line 19. It's more accurate and captures what the task wants to achieve.

guits:
was there a bug which led to this PR?

andymcc:
In some test scenarios it would fail to mklabel gpt, and would need a partprobe. I don't anticipate that would happen in any real world scenarios, but it seemed like a decent change anyway since there is no need to mklabel gpt if it already is gpt, and I'd be grateful too :)

guits:
@andymcc yup! sure this is a decent change, my concern is more about the way it's done. I'm usually not really enthusiastic about output parsing/screen-scraping even if I know sometimes there is no choice.

Perhaps this could help :

```
[root@ceph-osd0 ~]# blkid /dev/sda -t PTTYPE="gpt"
/dev/sda: PTTYPE="gpt"
[root@ceph-osd0 ~]# echo $?
0
[root@ceph-osd0 ~]# blkid /dev/sdb -t PTTYPE="gpt"
[root@ceph-osd0 ~]# echo $?
2
[root@ceph-osd0 ~]#
```

this could replace this [task](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/check_gpt.yml#L2)

it's just an idea, what do you think?

andymcc:
@guits - thanks for the response! Sorry I took a while getting back, I was  out last week with no access.

I'll give that a go - my initial test suggests it wont work for basically the same reason, it fails to initially set the partition table to gpt, so it returns a 2 (similar to using parted). I'll do a few more tests and see if i can come up with something else.

andymcc:
Ok I have a pretty good way to change this that I think is better for all.
The parted module does the trick pretty well.
I think we could probably skip the list before hand and the conditionals, but I'm guessing there was a pretty good reason for those, so I've left them in place for now.

PR #2230, opened 2017-12-04
[stable-3.0]Add information about current branch to index.rst

MrStupnikov:
It turns out that index.rst contains outdated information about
master branch and doesn't contain information about current branch
(stable-3.0). This patch fixes this issue.


guits:
@MrStupnikov actually we already have 7d239b690 which should have been backported to `stable-3.0`. We usually cherry-pick commits from master to stable branches to backport it (see https://github.com/ceph/ceph-ansible/pull/2231)

sorry for that @MrStupnikov 

however, thanks for having highlighted us on that missing backport

PR #2246, opened 2017-12-11
Add flags for OSD 'docker run --cpuset-{cpus,mems}'

fultonj:
Add the variables ceph_osd_docker_cpuset_cpus and
ceph_osd_docker_cpuset_mems, so that a user may specify
the CPUs and memory nodes of NUMA systems on which OSD
containers are run.

Provides an example for osds.yaml.sample to guide user
based on sample `lscpu` output since cpuset-mems refers
to the memory by NUMA node only while cpuset-cpus can
refer to individual vCPUs within a NUMA node.

### DIFF
--- a/roles/ceph-osd/defaults/main.yml
+++ b/roles/ceph-osd/defaults/main.yml
@@ -255,6 +255,14 @@ ceph_config_keys: [] # DON'T TOUCH ME
 ceph_osd_docker_memory_limit: 1g
 ceph_osd_docker_cpu_limit: 1
 
+# The next two variables are undefined, and thus, unused by default.
+# If `lscpu | grep NUMA` returned the following:
+#  NUMA node0 CPU(s):     0,2,4,6,8,10,12,14,16
+#  NUMA node1 CPU(s):     1,3,5,7,9,11,13,15,17
+# then, the following would run the OSD on the first NUMA node only.
+#ceph_osd_docker_cpuset_cpus: "0,2,4,6,8,10,12,14,16"
+#ceph_osd_docker_cpuset_mems: "0"
+
 # PREPARE DEVICE
 #
 # WARNING /!\ DMCRYPT scenario ONLY works with Docker version 1.12.5 and above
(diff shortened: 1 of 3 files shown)


roles/ceph-osd/defaults/main.yml
```diff
@@ -255,6 +255,14 @@ ceph_config_keys: [] # DON'T TOUCH ME
 ceph_osd_docker_memory_limit: 1g
 ceph_osd_docker_cpu_limit: 1
 
+# The next two variables are undefined, and thus, unused by default.
```
leseb: please run  `generate_group_vars_sample.sh`

roles/ceph-osd/templates/ceph-osd-run.sh.j2
```diff
@@ -54,6 +54,12 @@ expose_partitions "$1"
   {% else -%}
   --cpu-quota={{ ceph_osd_docker_cpu_limit * 100000 }} \
   {% endif -%}
+  {% if ceph_osd_docker_cpuset_cpus -%}
```
leseb: if `is defined` other the current check mean `ceph_osd_docker_cpuset_cpus` is True.

roles/ceph-osd/templates/ceph-osd-run.sh.j2
```diff
@@ -54,6 +54,12 @@ expose_partitions "$1"
...
+  --cpuset-cpus='{{ ceph_osd_docker_cpuset_cpus }}' \
+  {% endif -%}
+  {% if ceph_osd_docker_cpuset_mems -%}
```
leseb: ditto

fultonj:

roles/ceph-osd/defaults/main.yml
fultonj: OK, I did that and included the generated group_vars/osds.yml.sample in the updated PR

fultonj:

guits:
@fultonj I confirm, the failure here is unrelated to this patch

leseb:
@fultonj are you expecting this for 3.0 or 3.1?

fultonj:
@leseb if this could get backported to 3.0, then that would be great. thanks.

PR #2248, opened 2017-12-12
Rename fact docker_version to ceph_docker_version

berendt:
The name docker_version is very generic and is also used by other
roles. As a result, there may be name conflicts. To avoid this a
ceph_ prefix should be used for this fact. Since it is an internal
fact renaming is not a problem.


berendt:
Backport would be nice, we actually have a name conflict in one of our environments.

guits:

Issue #2250, opened 2017-12-12
Ceph user/group missing


berendt:
Ceph user & group missing when deploying with Ubuntu/Docker:

```
TASK [ceph-config : create ceph conf directory and assemble directory] *********
failed: [xxx] (item=/etc/ceph/) => {"failed": true, "gid": 64045, "group": "64045", "item": "/etc/ceph/", "mode": "0755", "msg": "chown failed: failed to look up user ceph", "owner": "64045", "path": "/etc/ceph/", "size": 4096, "state": "directory", "uid": 64045}
failed: [xxx] (item=/etc/ceph/ceph.d/) => {"failed": true, "gid": 0, "group": "root", "item": "/etc/ceph/ceph.d/", "mode": "0755", "msg": "chown failed: failed to look up user ceph", "owner": "root", "path": "/etc/ceph/ceph.d/", "size": 4096, "state": "directory", "uid": 0}
```

guits:
@berendt can you paste here the output of :
`grep -Ev '(^$|^#) group_vars/*.yml` ?

your nodes are running on Ubuntu?

berendt:
@guits Running on Ubuntu, yes. With ``containerized_deployment: true`` (Cannot paste the whole configuration.). There is no ``ceph`` user or group by default and no packages will be installed that will create a ``ceph`` user and group. Because of this the role has to ensure the existence of the user and group.

guits:
@berendt you shouldn't hit this error since you are running a containerized deployment. The failing tasks you mentioned here are played only for non containerized deployments.

guits:
@berendt can you link here the command you've used to launch the playbook?

berendt:
Learned a bit more about variable precedence in ansible. Can be closed as invalid.

howels:
Ran into this using the example site-docker.yml.sample file on CentOS Atomic. Install fails as user is not created before the local configuration directory is chown'd to that user. Looks like the commit is not yet included as making the same change to my local files fixes the issue; roles/ceph.ceph-common/tasks/installs/configure_redhat_local_installation.yml is not executed for Docker installs.

leseb:
@howels, as you mentioned ceph-common, does not run on containerized deployment. So it's unclear to me what's your error?

MrStupnikov:
Hi @leseb 

I have run into the same issue while installing containerized CEPH on CentOS machine. Please check error details below:

~~~
[...]

PR #2251, opened 2017-12-12
Make sure the ceph user and group is always present

berendt:
Some installation types do not have ceph user or Group. This change
ensures that user and group are always present and created if this
is not the case.

Closes: ceph/ceph-ansible#2250

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -98,6 +98,9 @@
     - ceph_current_fsid.rc == 0
     - mon_group_name in group_names
 
+- name: include create_ceph_user_and_group.yml
+  include: create_ceph_user_and_group.yml
+
 - name: include create_ceph_initial_dirs.yml
   include: create_ceph_initial_dirs.yml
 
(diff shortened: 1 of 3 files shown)


roles/ceph-common/tasks/main.yml
```diff
@@ -98,6 +98,9 @@
     - ceph_current_fsid.rc == 0
     - mon_group_name in group_names
 
+- name: include create_ceph_user_and_group.yml
+  include: create_ceph_user_and_group.yml
```
guits: either :
`include: ./installs/create_ceph_user_and_group.yml` 
or move `./installs/create_ceph_user_and_group.yml` to `roles/ceph-common/tasks`

berendt: Yes, of course. Copy & paste issue..

berendt: Move the file.

guits: From what I'm seeing in #2250 I'm not sure this PR makes sense

berendt:
Learned a bit more about variable precedence in ansible. Can be closed as invalid.

roles/ceph-common/tasks/main.yml
berendt: You're right.

PR #2253, opened 2017-12-12
docker: add missing condition for selinux tasks

guits:
on `client` and `mds` roles, it tries to set selinux even on non rhel
based distributions.`


guits:
the failure is unrelated to this patch

Issue #2254, opened 2017-12-13
ceph-default: fuser command is used without verifcation


sparktown:
In task file "check_socket_non_container.yml" the command fuser is used several times to check if a socket is used. But there is no verification, that the fuser command is available. If fuser is not available on a host the role removess all ceph-daemons on the host, because the when conditions applys to rc !=0 of the task with the fuser command (which is 127, command fuser not found).

Possible fixes:
a) Ensure package "psmisc" is installed (Ubuntu, don't know the packagename of centos)
b) Instead of daemon_name_socket.rc !=0 use daemon_name_socket.rc == 1 in the when condistions of the remove socket tasks

leseb:
@sparktown you're right using `daemon_name_socket.rc == 1` sounds the right approach. Will send a PR for this.

PR #2257, opened 2017-12-14
default: look for the right return code on socket stat in-use

leseb:
As reported in https://github.com/ceph/ceph-ansible/issues/2254, the
check with fuser is not ideal. If fuser is not available the return code
is 127. Here we want to make sure that we looking for the correct return
code, so 1.

Closes: https://github.com/ceph/ceph-ansible/issues/2254


guits:
failures here are unrelated to this patch

PR #2263, opened 2017-12-14
ceph.conf: fail if interface does not have any IP

leseb:
We want to fail if the interface provided for either mon/rgw does not
have any IP.


guits:
@leseb this feature could be addressed by the upcoming 'host var / pre check' validation module I think.
should we close this PR ?

leseb:
I'm not sure, it's hard to tell. The use case here is: a user provides a network interface for a monitor and that interface doesn't have any IP. I'm not sure if the upcoming module will solve that in the first iteration.

I'll leave this one open just so we don't forget about it.
What do you think?

guits:
if the upcoming module can access `hostvars` it should be able to handle this use case.
I'm ok to leave this open if we are not sure.

guits:
@andrewschoen any thoughts on this?

leseb:
Closing this in favor of https://github.com/ceph/ceph-ansible/pull/2915

PR #2279, opened 2017-12-19
nfs: fix package install for debian/suss systems

leseb:
This resolves the following error:
E: There were unauthenticated packages and -y was used without
--allow-unauthenticated


leseb:

leseb:
Xenial issue is different. But the current bug is fixed with this patch so merging.

Issue #1457, opened 2017-04-19
OSDs restarted outside play scope


Logan2211:
When running ceph-common with a `--limit` and a restart is notified on any of the ceph-ansible services, the new restart method has no concept of scoping its restarts to hosts targeted by the playbook run. Instead it will try to restart OSDs on every host in the group, not just the ones currently being operated on. This is really a huge regression because it completely removes the control of OSD restarts from the operator's control.

The conditionals in the file that control the host selection for restarts (ie. https://github.com/ceph/ceph-ansible/blob/e8e4245e434b5405d489714a31557252a9ab186d/roles/ceph-common/handlers/main.yml#L43) should be modified to reference play_hosts ex.

```
  when:
    - osd_group_name in group_names
    - item in play_hosts
```

Logan2211:
ps. love the host-serialized restarts though :)

leseb:
Thanks for reporting this @Logan2211 , we will try to address this in https://github.com/ceph/ceph-ansible/pull/1455

andrewschoen:
Fixed with #1455

PR #2287, opened 2017-12-20
Restart services if handler called

andymcc:
This patch fixes an issue where if hosts have different service lists,
it will prevent restarting changes on services that run later on.

For example, hostA in the mons and rgws group would initiate a config
change and restart of services on all mons and rgws hosts, even though
a separate hostB (which is only in the rgws group) has not had its
configuration changed yet. Additionally, when the second host has its
coniguration changed as part of the ceph-rgw role, it will not initiate
a restart since its inventory name != the first hosts.

To fix this we should run the restart once (which is already the case)
as long as the host is in the appropriate group, and filter to the other
hosts in that group.

### DIFF
--- a/roles/ceph-defaults/handlers/main.yml
+++ b/roles/ceph-defaults/handlers/main.yml
@@ -5,6 +5,14 @@
   when:
     - ansible_os_family == 'Debian'
 
+# We only want to restart on hosts that have called the handler.
+# This var is set when he handler is called, and unset after the
+# restart to ensure only the correct hosts are restarted.
+- name: set _mon_handler_called before restart
+  set_fact:
+     _mon_handler_called: True
+  listen: "restart ceph mons"
+
 - name: copy mon restart script
   template:
     src: restart_mon_daemon.sh.j2
@@ -15,7 +23,6 @@
   listen: "restart ceph mons"
   when:
     - mon_group_name in group_names
-    - inventory_hostname in play_hosts
 
 - name: restart ceph mon daemon(s) - non container
   command: /usr/bin/env bash /tmp/restart_mon_daemon.sh
@@ -24,11 +31,11 @@
     # We do not want to run these checks on initial deployment (`socket.rc == 0`)
     - mon_group_name in group_names
     - not containerized_deployment
-    - inventory_hostname == groups.get(mon_group_name) | last
-    - inventory_hostname in play_hosts
+    - hostvars[item]['_mon_handler_called'] | default(False)
     - mon_socket_stat.rc == 0
   with_items: "{{ groups[mon_group_name] }}"
   delegate_to: "{{ item }}"
+  run_once: True
 
 - name: restart ceph mon daemon(s) - container
   command: /usr/bin/env bash /tmp/restart_mon_daemon.sh
@@ -38,11 +45,21 @@
     - mon_group_name in group_names
     - containerized_deployment
     - ceph_mon_container_stat.get('rc') == 0
-    - inventory_hostname == groups.get(mon_group_name) | last
-    - inventory_hostname in play_hosts
+    - hostvars[item]['_mon_handler_called'] | default(False)
     - ceph_mon_container_stat.get('stdout_lines', [])|length != 0
   with_items: "{{ groups[mon_group_name] }}"
   delegate_to: "{{ item }}"
+  run_once: True
+
+- name: set _mon_handler_called after restart
+  set_fact:
+     _mon_handler_called: False
+  listen: "restart ceph mons"
+
+- name: set _osd_handler_called before restart
+  set_fact:
+     _osd_handler_called: True
+  listen: "restart ceph osds"
 
 # This does not just restart OSDs but everything else too. Unfortunately
 # at this time the ansible role does not have an OSD id list to use
@@ -57,7 +74,6 @@
   listen: "restart ceph osds"
   when:
     - osd_group_name in group_names
-    - inventory_hostname in play_hosts
 
 - name: restart ceph osds daemon(s) - non container
   command: /usr/bin/env bash /tmp/restart_osd_daemon.sh
@@ -65,16 +81,15 @@
   when:
     - osd_group_name in group_names
     - not containerized_deployment
-    - inventory_hostname == groups.get(osd_group_name) | last
     # We do not want to run these checks on initial deployment (`socket_osd_container.results[n].rc == 0`)
     # except when a crush location is specified. ceph-disk will start the osds before the osd crush location is specified
     - ((crush_location is defined and crush_location) or osd_socket_stat.rc == 0)
     - ceph_current_fsid.rc == 0
     - handler_health_osd_check
-    # See https://github.com/ceph/ceph-ansible/issues/1457 for the condition below
-    - inventory_hostname in play_hosts
+    - hostvars[item]['_osd_handler_called'] | default(False)
   with_items: "{{ groups[osd_group_name] }}"
   delegate_to: "{{ item }}"
+  run_once: True
 
 - name: restart ceph osds daemon(s) - container
   command: /usr/bin/env bash /tmp/restart_osd_daemon.sh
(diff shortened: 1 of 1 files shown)


roles/ceph-defaults/handlers/main.yml
```diff
@@ -24,7 +24,7 @@
     # We do not want to run these checks on initial deployment (`socket.rc == 0`)
     - mon_group_name in group_names
     - not containerized_deployment
-    - inventory_hostname == groups.get(mon_group_name) | last
+    - inventory_hostname in groups.get(mon_group_name)
```
guits: this line become redundant to L25

guits:
@andymcc can you rebase on master please?

roles/ceph-defaults/handlers/main.yml
guits: actually, we can't remove this line, otherwise, because of `serial: 1` we won't skip this task on first nodes.

guits:
@andymcc From what I've seen here, I'm not sure how far this PR is still relevant but perhaps I'm missing something, could you provide more context/information regarding this?

andymcc:
@guits sorry I was out for a few days!

I think the issue still stands: the issue is essentially that if a change happens on the node that isn't the last in the group (but doesn't happen on the node that IS the last in the group) the service will never get restarted. That would be an issue in two scenarios that i can think of off the top of my head:

1. A change (that would require a restart) has happened on only a subset of the hosts, either by using limit-hosts or a manual change - and the host that changed is not the last host in the group.
2. A host is in multiple groups, and is the last host in the group, meaning it will have its configuration changed as part of the mons group, then when it runs through the rgw or later roles, it will not initiate a restart on the other hosts that didn't have their configuration changed as part of the mons group, even though they call the handler - they will get skipped because they are not the last host.

You're right though, that line can probably just be removed since its there already, and serial = 1 will ensure this only gets run once which is correct, but it'll run on the first host that has had a change and calls the handler.

I believe the service_socker_stat.rc = 0 line ensures it doesn't run on the initial deploy - so i think that should still be ok. I may be missing something else though!

Does my explanation make sense? Happy to go into more detail.

andymcc:
I've updated the PR to just remove the line, I'm not sure we need to ensure its the last host in the group, given that we have serial=1

andymcc:
@guits @leseb - any further thoughts on this? I think it's still an issue.

leseb:
I think we are on the right track with this patch. I think we are just a bit hesitant since handlers have proved to be hard to maintain and easy to break things. Even though the CI has good coverage, we are 'afraid' we might be missing something.

However, as stated before we already use serial 1 upfront so this should just work.

andymcc:
@guits - any chance I can get you to take a second look? Appreciate it!

guits:
@andymcc sorry for the late response.

I think we better to remove the `run_once: true` instead.

According to the test I've made, simply using `run_once: true` has not the expected behavior.
[(see results)](https://gist.github.com/guits/6fc77c222dd54da3223a82c13a2c48be)

andymcc:
Hi @guits so the problem is that the handler gets called on each host if you run it in serial, and then restarts the other hosts?

We could change the group to be ansible_play_hosts in that case - serial would then not run against any hosts but the one in question?

The issue with |last for me, is still that if we have groups that are not unique - e.g. some hosts are in multiple groups, the restart handler does not work as expected. This is because there is a chance that no-change to the conf happened to the "last" host during the service run (because it has already had its conf updated in the previous role) - this in turn means that the restart handler doesnt execute and the other hosts that had their config, updated during the current role, are not restarted.

Perhaps by example I can explain better: If we have hosts: host1, host2, hostLast

host1 and host2 are just rgw hosts, and hostLast is a mon and an rgw:

hostLast will run through during the mon update, its configuration will change and it will run the notification handlers to restart services, and those of other hosts in that group. At this point the configuration for host1 and host2 has not changed.
The rgw role will then run, both host1 and host2 have their configuration changed - hostLast does not (it's already updated). The notification happens for host1 and host2 but because of "|last" and the fact that hostLast has not called the handler the restart never happens, leaving host1 and host2 in a state which does not match the configuration on disk.

We could maybe fix this by setting a var if the config changes (e.g. register: conf_changed) and then doing run_once, but only delegating if the host it is delegating to has "conf_changed = True" set?

IMO the current situation is a bug that needs to be fixed, otherwise we should document that groups need to be uniform and mixing groups is a bad idea.

andymcc:
Actually, we already have play_hosts set which is equivalent to ansible_play_batch:
which is available as a list of hostnames that are in scope for the current ‘batch’ of the play. The batch size is defined by serial, when not set it is equivalent to the whole play.

So we could change this to use "with_items: ansible_play_batch"
instead of groups[mon_names] for e.g.

guits:
@andymcc 
 agree, I think it could be the fix here, do you mind updating your PR ?

andymcc:
Thinking about this a bit more, and looking at the old issue: #1457 
I think we can simplify this by checking if the host it will delegate to is in the play_hosts instead of the current host. The handlers have changed such that the "inventory_hostname in play_hosts" is redundant - it has to be in play_hosts to run that task.

Put a new PR in, but it should now:

* Only run_once
* Delegate to each host in the group, as long as that host is in the play_hosts - which means in serial it will only restart the one host.
* Will run as long as one host calls the handler in the current list of play_hosts - which means in serial it will restart each host, but only on that hosts run.

guits:
@andymcc It looks good, indeed. I just need to double check something but I think we should be good to merge this.
I know this has been a long time, thanks for your patience!

andymcc:
Of course, no problem - its a pretty core change so I understand the need to take it slow.

guits:
@andymcc I took a look at your last patch but I still get some trouble with it :
see [here](http://pastefile.fr/3a3fcabb0b781c6e7501570e1c334f25) this log is the result of what you propose.
If you look well, you can see daemons are restarted several times.

#### inventory host file
```
[mons]
mon0 ansible_host=mon0 ansible_ssh_private_key_file='/Users/guits/.ssh/id_rsa' ansible_ssh_extra_args='-o StrictHostKeyChecking=no'
mon1 ansible_host=mon1 ansible_ssh_private_key_file='/Users/guits/.ssh/id_rsa' ansible_ssh_extra_args='-o StrictHostKeyChecking=no'
mon2 ansible_host=mon2 ansible_ssh_private_key_file='/Users/guits/.ssh/id_rsa' ansible_ssh_extra_args='-o StrictHostKeyChecking=no'

[osds]
mon0 ansible_host=mon0 ansible_ssh_private_key_file='/Users/guits/.ssh/id_rsa' ansible_ssh_extra_args='-o StrictHostKeyChecking=no'
osd0 ansible_host=osd0 ansible_ssh_private_key_file='/Users/guits/.ssh/id_rsa' ansible_ssh_extra_args='-o StrictHostKeyChecking=no
```

#### defaults/handlers/main.yml
```
- name: restart mons
  debug:
    msg: "restarting mons..."
  with_items: "{{ groups['mons'] }}"
  when: item in play_hosts
  delegate_to: "{{ item }}"
  run_once: true
  listen: restart mons

- name: restart osds
  debug:
    msg: "restarting osds..."
  with_items: "{{ groups['osds'] }}"
  when: item in play_hosts
  delegate_to: "{{ item }}"
  run_once: true
  listen: restart osds
```

With the following code, we can see the handlers are well triggered as supposed :
#### defaults/handlers/main.yml
```
- name: restart mons
  debug:
    msg: "restarting mons..."
  with_items: "{{ groups['mons'] }}"
  when: inventory_hostname in groups.get('mons') | last
  delegate_to: "{{ item }}"
  listen: restart mons

- name: restart osds
  debug:
    msg: "restarting osds..."
  with_items: "{{ groups['osds'] }}"
  when: inventory_hostname in groups.get('osds') | last
  delegate_to: "{{ item }}"
  listen: restart osds
```
which gives this [output](http://pastefile.fr/82d9e1d8a7780baae16b51e5ea80f07c). Here you can see the behavior is well the expected one.

andymcc:
Hey @guits - I'm trying to think of a good way to change it. I can't think of a way that is as simple as I'd hope, without tracking which hosts have changed (ansible should have a serial restart handler!) - e.g. I only delegate if the host has called the handler. Perhaps that is the only way to implement that properly, otherwise you will get a restart on ALL hosts in the same group, as long as one host in the group has called the handler - which ensures consistency but does wasteful restarts.

To show you the current issue though, if you were to change the inventory such that mon2 was an osd host and mon0 was not, and then change osd0 to be called aosd0 (forcing it to not be the |last host) - you'll see different results for the second run - which is essentially the problem.

andymcc:
I think we can streamline my current PR more by removing "item in play_hosts", since it will now only run if the host has the fact set, which means it already has to be in play_hosts. will fix that quickly

guits:
@andymcc 
That current patch is now ready to be merged.
Thanks for your patience and the time spent on this.

leseb:
Thanks @andymcc!

Issue #2228, opened 2017-12-01
Creates argument prevents bootstrap keys from being generated in luminous and greater releases


allen13:
When running the playbook with the luminous release the mons role fails to generate bootstrap keys. Since fetch doesn't have "fail_on_missing: yes" it continues on without failing. It finally fails when it gets the the osd role and can't find the fetch keys.

Removing the creates statement fixes the problem because the {{ cluster }}.client.admin.keyring gets generated in a previous step which means this command will never get run:

args:
    creates: /etc/ceph/{{ cluster }}.client.admin.keyring

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/ceph_keys.yml#L2

Apparently the creates arg wins out over always_run: yes and check_mode: no (I tested both)

guits:
@allen13 can you provide a full playbook run log with verbose?

vishalkanaujia:
@allen13 I also faced the same problem and resolved it by commenting `#creates: /etc/ceph/{{ cluster }}.client.admin.keyring`.  This file is first created in `./roles/ceph-mon/tasks/deploy_monitors.yml`. Since the deply_monitor.yml task runs before ceph_keys task, the latter _create_ operation fails.

allen13:
Everything works fine when I comment out the creates argument as well. Is there any reason for keeping it? ceph-create-keys is already idempotent. Running it again doesn't hurt anything.

PR #2289, opened 2017-12-20
mon: always run ceph-create-keys

leseb:
ceph-create-keys is idempotent so it's not an issue to run it each time
we play ansible. This also fix issues where the 'creates' arg skips the
task and no keys get generated on newer version, e.g during an upgrade.

Closes: https://github.com/ceph/ceph-ansible/issues/2228

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -1,8 +1,6 @@
 ---
 - name: collect admin and bootstrap keys
   command: ceph-create-keys --cluster {{ cluster }} -i {{ monitor_name }} -t 30
-  args:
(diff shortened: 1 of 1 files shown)


ktdreyer:
I'm curious what process you followed to merge this? The PR lists "ce323d5f57d7ee437f9f3c582a7d66c54edcc89a" as the commit here, but "0b55abe3d0fc6db6c93d963545781c05a31503bb" is what went into master.

ktdreyer:
Ok great. Is this the standard that all maintainers should follow? How do we communicate this for everyone who can merge changes to ceph-ansible?

leseb:
The rebase & merge is the github button when merging a PR so contributors don't have anything to do.

ktdreyer:
Oh I see, you've updated the settings at some point so all other options are "Not enabled for this repository". Got it!

Issue #2088, opened 2017-10-21
ceph_osd_docker_run_script_path /usr/share not writable / Atomic Host


megian:
The current default for the variable ceph_osd_docker_run_script_path is /usr/share. /usr is not writable on Atomic Host and cause the playbook to stop. After I manually changed the variable in osds.yml to ceph_osd_docker_run_script_path: /etc/ceph everything works. Not sure if this is the optimal path. Can the default be changed or at least for Atomic Host systems?

leseb:
Having a variable to define the path is the purpose of this variable. I'm not keen on changing the behavior based on the OS. However, if we can come up with a decent path that would make all distros happy I don't mind.
What are the writable paths on Atomic?

megian:
According the FHS http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html is /usr/share for "read-only architecture independent data files". So I would say, that it is anyway not recommended to write there independent from a Distribution based package management system, even it would work with most of the distributions.

Atomic Host has just mounted /etc and /var read-write.

leseb:
@megian so something like `/etc/ceph` maybe?

megian:
I would suggest so. Other startup scripts, like init scripts, are in /etc, too. Maybe, if more scripts would be required in the future a subdirectory like /etc/ceph/container or /etc/ceph/docker could make sense.

PR #2300, opened 2018-01-03
osd: set ceph_osd_docker_run_script_path to /etc/ceph

leseb:
The current default for the variable ceph_osd_docker_run_script_path is
/usr/share. /usr is not writable on Atomic Host and cause the playbook
to stop. So it's better to have a default location that works for all
the distro.

Closes: https://github.com/ceph/ceph-ansible/issues/2088


leseb:
Closing this due to inactivity, feel free to re-open, thanks!

PR #2301, opened 2018-01-05
tests: fixes to allow for stable-3.0 nightly testing of jewel

andrewschoen:
This fixes a few things related to our nightly tests with ``stable-3.0`` that deploy jewel clusters.

- when updating a jewel cluster ignore failures related to stopping ``nfs-ganesha`` 
- when switching a jewel cluster to containers ignore ``nfs`` and ``mgr``  service failures
- explicitly set ``UPDATE_CEPH_DOCKER_IMAGE_TAG`` for jewel tests so they upgrade to luminous and not latest.

These are really only an issue because we've setup our tests to work with both jewel and luminous.

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -535,11 +535,15 @@
   become: True
 
   pre_tasks:
+    # failed_when: false is here so that if we upgrade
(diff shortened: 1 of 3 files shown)


leseb:
CI failures unrelated, moving on.

PR #2306, opened 2018-01-09
mon: use crush rules for non-container too

leseb:
There is no reasons why we can't use crush rules when deploying
containers. So moving the inlcude in the main.yml so it can be called.


leseb:
CI failures unrelated.

PR #2318, opened 2018-01-12
ceph-volume: adds a crush_device_class option for lvm_volumes

andrewschoen:
A ``--crush-device-class`` feature is currently in development for ``ceph-volume``, but we need this in place first so that we can test the PR that introduces ``--crush-device-class``.

### DIFF
--- a/docs/source/osds/scenarios.rst
+++ b/docs/source/osds/scenarios.rst
@@ -205,6 +205,7 @@ The following keys are accepted for a ``filestore`` deployment:
 * ``data_vg`` (not required if ``data`` is a raw device or partition)
 * ``journal``
 * ``journal_vg`` (not required if ``journal`` is a partition and not a logical volume)
+* ``crush_device_class`` (optional, sets the crush device class for the OSD)
 
 The ``journal`` key represents the logical volume name or partition that will be used for your OSD journal.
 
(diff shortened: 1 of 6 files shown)


docs/source/osds/scenarios.rst
```diff
@@ -205,6 +205,7 @@ The following keys are accepted for a ``filestore`` deployment:
 * ``data_vg`` (not required if ``data`` is a raw device or partition)
 * ``journal``
 * ``journal_vg`` (not required if ``journal`` is a partition and not a logical volume)
+* ``crush_device_class`` (not required, sets the crush device class for the OSD)
```
alfredodeza: the above mention "not required" because there is a condition: *not required if <reason>*, but that is not the case here. I think it would be fine to say *optional* instead

leseb:
@andrewschoen this should be on hold until https://github.com/ceph/ceph/pull/19949 gets merged right?

andrewschoen:
@leseb we actually need this merged first so that we can test https://github.com/ceph/ceph/pull/19949 through the CI

This shouldn't break deploying ``ceph-volume`` OSDs with ``luminous`` though, if you don't include ``crush_device_class`` in ``lvm_volumes`` this new feature won't get used. If you like I could remove the commit here that updates the tests so that our lvm PR tests here will still pass for luminous.

leseb:
@andrewschoen thanks, I forgot we were testing with ceph-ansible on ceph/ceph.

PR #2322, opened 2018-01-16
ntp: followup cleanup

leseb:
Checking if ntp is present is not needed anymore, these tasks are not
used anymore. So let's remove them


roles/ceph-docker-common/tasks/misc/ntp_redhat.yml
```diff
@@ -1,9 +1,4 @@
 ---
-- name: include ../checks/check_ntp_redhat.yml
```
guits: it can't start the service anymore since the package installation is in this include

PR #2323, opened 2018-01-17
Resolv parent device

guits:


infrastructure-playbooks/purge-docker-cluster.yml
```diff
@@ -341,62 +341,76 @@
 
   - name: get ceph data partitions
     shell: |
-      blkid | awk -F: '/ceph data/ { print $1 }'
+      blkid -o device -t PARTLABEL="ceph data"
```
leseb: use `command`?

guits: true, just a legacy. thanks maggle

guits: done

PR #2326, opened 2018-01-18
Make CephFS pools creation optional.

mgornikov:


guits:
@mgornikov this commit doesn't exist in master, it should be merged in master first and then backported to stable-3.0

PR #2328, opened 2018-01-18
Skip devices that used as backed for other devices

mgornikov:
Devices that used as backed for logical layers (dmraid, bcache, LVM , etc) have "masters" in Ansible "devices" fact. This patch skip such devices on initial lookup.


guits:
@mgornikov this commit doesn't exist in master

PR #2336, opened 2018-01-19
Add rgw keystone implicit tenants = true to ceph.conf.j2

colonwq:
Set to true for private tenant for each new user (1).

1: http://docs.ceph.com/docs/master/radosgw/keystone/


roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -171,6 +171,7 @@ host = {{ hostvars[host]['ansible_hostname'] }}
 host = {{ hostvars[host]['ansible_hostname'] }}
 keyring = /var/lib/ceph/radosgw/{{ cluster }}-rgw.{{ hostvars[host]['ansible_hostname'] }}/keyring
 log file = /var/log/ceph/{{ cluster }}-rgw-{{ hostvars[host]['ansible_hostname'] }}.log
+rgw keystone implicit tenants = true
```
leseb: we don't accept PR on this file anymore, unless there is a case where `ceph_conf_overrides` can not be used. Please use `ceph_conf_overrides`.

PR #2345, opened 2018-01-23
ansible: set ssh retry option to 5

guits:
We noticed that sometime, ceph-ansible can fail with error :

`Failed to connect to the host via ssh:`

It can occurs after the task `restart firewalld` has been played.

Setting `retries` to 5 should prevent from unexcepted ssh failure.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -19,3 +19,6 @@ timeout = 30
 [ssh_connection]
 # see: https://github.com/ansible/ansible/issues/11536
 control_path = %(directory)s/%%h-%%r-%%p
+
+# Option to retry failed ssh executions if the failure is encountered in ssh itself
+retries = 5


ansible.cfg
```diff
@@ -11,6 +11,9 @@ log_path = /var/log/ansible.log
 # Disable them in the context of https://review.openstack.org/#/c/469644
 retry_files_enabled = False
 
+# Option to retry failed ssh executions if the failure is encountered in ssh itself
+retries = 5
```
alfredodeza: According to the docs at https://docs.ansible.com/ansible/latest/intro_configuration.html#openssh-specific-settings this would need to go under the `[ssh_connection]` header

guits:
CI isn't going to pass due to vagrant images error for centos

PR #2350, opened 2018-01-25
ceph-common: Don't check for ceph_stable_release for distro packages

guits:
When we consume the distribution packages, we don't have the choise on
which version to install, so we shouldn't require that variable to be
set. Distributions normally provide only one version of Ceph in the
official repositories so we get whatever they provide.

(cherry picked from commit dd6ee72547a4eca22c8c9b8691b910c2cfa821d3)


guits:
jenkins test luminous-ansible2.4-update_cluster

PR #2351, opened 2018-01-25
upgrade: skip luminous tasks for jewel minor update

guits:
These tasks are needed only when upgrading to luminous.
They are not needed in Jewel minor upgrade and by the way, they fail because
`ceph versions` command doesn't exist.

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -363,11 +363,15 @@
       command: "{{ docker_exec_cmd_update_osd|default('') }} ceph --cluster {{ cluster }} versions"
       register: ceph_versions
       delegate_to: "{{ groups[mon_group_name][0] }}"
+      when:
+        - not jewel_minor_update
 
     - name: set_fact ceph_versions_osd
       set_fact:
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/rolling_update.yml
```diff
@@ -363,20 +363,25 @@
...
+        - not jewel_minor_update
 
     # length == 1 means there is a single osds versions entry
     # thus all the osds are running the same version
     - name: complete osds upgrade
       command: "{{ docker_exec_cmd_update_osd|default('') }} ceph --cluster {{ cluster }} osd require-osd-release luminous"
       delegate_to: "{{ groups[mon_group_name][0] }}"
       when:
-        - (ceph_versions.stdout|from_json).osd | length == 1
+        - (ceph_versions.get('stdout', '{}')|from_json).get('osd', {}) | length == 1
```
andrewschoen: You need to add quotes to the ``{}`` in ``.get('osd', {})`` like you've done with the first one.

andrewschoen: nevermind, quotes on the first were needed because of ``from_json``, the next one does not need it.

PR #2356, opened 2018-01-29
Do not search osd ids if ceph-volume

leseb:
The 'get osd id' task goes through all the 10 times (and its respective
timeouts) to make sure that the number of OSDs in the osd directory
match the number of devices.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1537103


leseb:
@ErwanAliasr1 done.

PR #2367, opened 2018-01-31
Support multiple inventory files under individual inventory directory 

hswong3i:
Currently default caph-ansible example coming with following limitation:
* Didn't provide sample inventory file as reference, e.g. ``inventory/sample/hosts``
* {{ fetch_directory }} is relative to project top directory, so multiple deployments will result as shared ``fetch/`` folder generated
* ``group_vars`` is relative to project to directory, so multiple deployments will again sharing the same configuration
* We can't manage all deployment at once, inside centralized ansible controller

This patch try to improve above limitation by:
* Create sample inventory file: ``inventory/sample/hosts``
* {{ fetch_directory }} now relative to {{ inventory_dir }}: ``inventory/sample/fetch/``
* ``group_vars`` now relative to {{ inventory_dir }}: ``inventory/sample/group_vars/``
* Simplify sample configuration files from ``group_vars/*.yml.sample`` as ``inventory/sample/group_vars/*.yml``
* Update ``generate_group_vars_sample.sh`` according to above changes
* Update ``docs/source/index.rst`` according to above changes
* Add fetch of artifacts, e.g. `ceph.conf` and `ceph.client.admin.keyring`, into ansible host with a simple helper script, so if ansible host also deploy within group `clients` we will have Ceph CLI and able to connect into multiple deployed cluster at once.

## Example Result for Multiple Inventory Directories
With above changes user can simply copy entire ``inventory/sample/`` for individual deployment, start its own configuration and deploy without any conflict, e.g. 
```
ceph-ansible/
├── inventory
│   ├── dev
│   │   ├── fetch
│   │   │   └── ceph_cluster_uuid.conf
│   │   ├── group_vars
│   │   │   └── all.yml
│   │   └── hosts
│   ├── production
│   │   ├── fetch
│   │   │   └── ceph_cluster_uuid.conf
│   │   ├── group_vars
│   │   │   └── all.yml
│   │   └── hosts
│   └── staging
│       ├── fetch
│       │   └── ceph_cluster_uuid.conf
│       ├── group_vars
│       │   └── all.yml
│       └── hosts
├── site-docker.yml.sample
└── site.yml.sample
```

## Example Result for Multiple Artifacts
Loop though all inventory directories (e.g. c11/c21/d11/d21 here) and get its nodes with:
```
ls -1d inventory/*/artifacts/ | while read line
do 
	cd $line
	echo `pwd`
	./ceph.sh -s
	cd ../../../
done
```
Sample output:
```
/opt/ceph-ansible/inventory/c11/artifacts
  cluster:
    id:     3ad71047-c488-4728-9c6c-183ccac5f216
    health: HEALTH_OK
 
  services:
    mon: 3 daemons, quorum fs11,fs12,fs13
    mgr: fs11(active), standbys: fs12, fs13
    mds: cephfs-1/1/1 up  {0=fs11=up:active}, 2 up:standby
    osd: 3 osds: 3 up, 3 in
 
  data:
    pools:   2 pools, 136 pgs
    objects: 219k objects, 6321 MB
    usage:   43398 MB used, 257 GB / 299 GB avail
    pgs:     136 active+clean
```
```
/opt/ceph-ansible/inventory/c21/artifacts
  cluster:
    id:     1bc7193f-c477-4e22-9f26-84d6996c4cd4
    health: HEALTH_OK
 
  services:
    mon: 3 daemons, quorum fs21,fs22,fs23
    mgr: fs22(active), standbys: fs21, fs23
    mds: cephfs-1/1/1 up  {0=fs21=up:active}, 2 up:standby
    osd: 3 osds: 3 up, 3 in
 
  data:
    pools:   2 pools, 136 pgs
    objects: 21 objects, 34251 bytes
    usage:   3316 MB used, 296 GB / 299 GB avail
    pgs:     136 active+clean
```
```
/opt/ceph-ansible/inventory/d11/artifacts
  cluster:
    id:     3ad71047-c488-4728-9c6c-183ccac5f216
    health: HEALTH_OK
 
  services:
    mon: 1 daemons, quorum d11
    mgr: d11(active)
    mds: cephfs-1/1/1 up  {0=d11=up:active}
    osd: 3 osds: 3 up, 3 in
 
  data:
    pools:   2 pools, 136 pgs
    objects: 21 objects, 2246 bytes
    usage:   3170 MB used, 296 GB / 299 GB avail
    pgs:     136 active+clean
```
```
/opt/ceph-ansible/inventory/d21/artifacts
  cluster:
    id:     1bc7193f-c477-4e22-9f26-84d6996c4cd4
    health: HEALTH_OK
 
  services:
    mon: 1 daemons, quorum d21
    mgr: d21(active)
    mds: cephfs-1/1/1 up  {0=d21=up:active}
    osd: 3 osds: 3 up, 3 in
 
  data:
    pools:   2 pools, 136 pgs
    objects: 21 objects, 21740 bytes
    usage:   3188 MB used, 296 GB / 299 GB avail
    pgs:     136 active+clean
```

### Case 1: For existing deployment that move the original ``fetch`` **BEFORE** another run of playbook

No impact.

### Case 2: For existing deployment that **DIDN'T** move the origin ``fetch`` folder before another run of playbook

Since if the target ``inventory/mycluster/fetch`` folder not existing we will always regenerate a new one, then update all deployment target hosts accordingly, therefore the running of playbook will always works.

NOTE: The only concern is from the client use case point of view: because the ``fetch/__uuid__/etc/ceph/ceph.client.admin.keyring`` also get updated, so existing mount from client machine need to update accordingly, e.g. update its ``/etc/fstab``. Once this get done, everything should works as expected.

### Case 3: Already run playbook once as Case 2, then move ``fetch`` and re-run playbook

Again, no impact because all parameters rollback as original, so both server and client should works as expected ;-)


hswong3i:
I found that following tests always fail in latest master, whatever with/without the patch from here:
* luminous-ansible2.4-purge_bluestore_osds_container
* luminous-ansible2.4-xenial_cluster

Does this means these 2 tests should be get fixed by upstream?

guits:
@hswong3i i'm currently working on fixing these two tests, sorry for that.

hswong3i:
@guits thank you very much for the fix, just hope to understand if patch here getting something wrong as I am new with ceph-ansible ;-)

Also hope to contribute for the fix, please feel free to share about how to do so, so I could help running the test locally, too.

hswong3i:
@guits I had recently contribute a similar patch for Kubespray (https://github.com/kubernetes-incubator/kubespray/pull/2233) which already get merged. Since recent failed test in here are mainly not related to its changes, may we also have a review here and so get it merged?

leseb:
@hswong3i  I like this patch and definitely see the value here. I'm just concerned about the impact on existing deployments.

guits:
+1 with what @leseb said here, this is a decent feature, the only concern is the backward compatibility.

The minimum would be to add some documentation to warn users that have already deployed a cluster even if I'm not sure this would be enough.

group_vars/rhcs.yml.sample
```diff
@@ -29,7 +29,7 @@ dummy:
 #  mimic: 13
 
 # Directory to fetch cluster fsid, keys etc...
-fetch_directory: ~/ceph-ansible-keys
+fetch_directory: "{{ inventory_dir }}/fetch/"
```
leseb: this must remain `~/ceph-ansible-keys`, values are getting pickef from `rhcs_edits.txt`

hswong3i: @leseb also updated ``rhcs_edits.txt`` so changes from ``inventory/sample/group_vars/rhcs.yml`` is valid as generated result from ``generate_group_vars_sample.sh``

hswong3i:
@leseb .gitignore updated ;-)

hswong3i:
@guits basically documents already updated for new deployment.

### Case 1: For existing deployment that move the original ``fetch`` **BEFORE** another run of playbook

No impact.

### Case 2: For existing deployment that **DIDN'T** move the origin ``fetch`` folder before another run of playbook

Since if the target ``inventory/mycluster/fetch`` folder not existing we will always regenerate a new one, then update all deployment target hosts accordingly, therefore the running of playbook will always works.

NOTE: The only concern is from the client use case point of view: because the ``fetch/__uuid__/etc/ceph/ceph.client.admin.keyring`` also get updated, so existing mount from client machine need to update accordingly, e.g. update its ``/etc/fstab``. Once this get done, everything should works as expected.

### Case 3: Already run playbook once as Case 2, then move ``fetch`` and re-run playbook

Again, no impact because all parameters rollback as original, so both server and client should works as expected ;-)

hswong3i:
@guits I also remove some fake ``group_vars`` from previous commit, because tests should always failed unless we also update the test plan accordingly, so let's keep this patch simple stupid ;-)

guits:
jenkins test this please

group_vars/rhcs.yml.sample
hswong3i: @leseb shall this still keep as ``~/ceph-ansible-keys``?

``rhcs_edits.txt`` also updated accordingly.

andrewschoen:
Instead of modifying the ceph-ansible source have you tried creating your ``inventory`` directory outside of the ceph-ansible tree and then when running ``site.yml`` pointing to whichever inventory file you're trying to manage with the ``-i`` flag of ``ansible-playbook``? Ansible will then pick up all config that is defined relative to that inventory file. You can set ``fetch_directory`` to be stored in these directories as well so your clusters will be isolated. This is how we handle our CI tests for ceph-ansible. In the ``tests/functional`` directory we have many clusters defined, all isolated from each other with their own config and inventory.

For example, if you were deploying to your ``production`` cluster you might run a command like this from the root of ceph-ansible:

```
ansible-playbook site.yml -i /path/to/inventory_dirs/production/hosts
```

I don't think we need to modify ceph-ansible source at all to accomplish the kind of setup you want here.

hswong3i:
| You can set fetch_directory to be stored in these directories as well so your clusters will be isolated. This is how we handle our CI tests for ceph-ansible. 

Agree that we could alter this by changing those variables manually which also shown as works in CI, but this is simple just because we are developer or even contributor.

From normal non-contributor point of view (which I was before indeed study the code logic of ceph-ansible for many hours), the unexpected behavior which **by default** store all generated secrets under **shared** fetch/ directory when using different hosts file which placed in different locations looks insane for me... This behavior keep as is even if we place the hosts file **outside** ceph-Ansible source tree... This changed unless finally I could discover those variables and it’s **out of the box default values**...

That’s why this patch coming from: make use of those variables and provide a better user experience in a out of the box style for normal non-contributors use cases. For CI which may have another special needs in using a shared fetch/ directory when running with multiple tests, let’s keep it as contributor’s duty to understand how to do so by altering those default values ;-)

leseb:
@andrewschoen has a point here, however for newcomers/ non-well-versed in Ansible this is non-obvious to come up with such setup.
@hswong3i's approach changes the repo structure and yes this is not ideal, but it brings a better out of the box experience. Later, it could even improve multi-cluster deployments more easily.

There are apparently pros/cons to this approach, most of them were discussed already, and the cons are very limited (mostly docs) compare to the ease of use from a contributor/operator/developer perspective, which is IMHO a huge win.

infrastructure-playbooks/rolling_update.yml
```diff
@@ -79,6 +79,26 @@
 
     - set_fact: rolling_update=true
```
hswong3i: @leseb and @guits I just add some code here for migrate old fetch dir to new location if not yet exists, which clone similar logic from https://github.com/kubernetes-incubator/kubespray/pull/2345. Hopefully reduce some impact on existing deployments.

roles/ceph-client/tasks/create_users_keys.yml
```diff
@@ -92,3 +92,17 @@
   when:
     - cephx
     - keys | length > 0
+
```
hswong3i: @andrewschoen I add some code here for coping ceph.conf and ceph.client.admin.keyring to ansible host, so we could remotely manage all deployed cluster in a single ansible controller.

See reference: https://github.com/kubernetes-incubator/kubespray/pull/2347

roles/ceph-client/tasks/pre_requisite.yml
```diff
@@ -30,3 +30,33 @@
   when:
     - cephx
     - copy_admin_key
+
```
hswong3i: @andrewschoen I add some code here for coping ceph.conf and ceph.client.admin.keyring to ansible host, so we could remotely manage all deployed cluster in a single ansible controller.

See reference: kubernetes-incubator/kubespray#2347

hswong3i: @andrewschoen I add some code here for coping ceph.conf and ceph.client.admin.keyring to ansible host, so we could remotely manage all deployed cluster in a single ansible controller.

See reference: kubernetes-incubator/kubespray#2347

hswong3i:
Update description with example result of multiple artifacts.

alfredodeza:
I agree that this needed, having to understand ansible and how it works is cumbersome if all one wants to do is deploy a cluster. This is what ceph-deploy helped with, and what gdeploy does with Gluster.

Maybe we should rethink how to help ceph-ansible with a wrapper CLI project that can help, vs. just changing the whole project structure because a new comer might not be aware of how Ansible works?

This change is not going to make it easy for any other downstream product consuming ceph-ansible, and it will require hefty rewrites of all (most?) of the tests. Not sure why we are seeing only 8 tests running on this PR.

alfredodeza:
jenkins test all

hswong3i:
> Maybe we should rethink how to help ceph-ansible with a wrapper CLI project that can help, vs. just changing the whole project structure because a new comer might not be aware of how Ansible works?

@alfredodeza Are you sure this is just about **a new comer might not be aware of how Ansible works**? Please checkout https://github.com/ansible/ansible/issues/16956 for more information about how this also impact to Ansible experts, since Ansible 2.1.1 ;-)

See https://github.com/ansible/ansible/issues/16956#issuecomment-280626669
> Anyway everyone, there is no point in replying to this issue any more, because this will never be reverted back to the state it was in 2.0, because it was a bug. Just modify your playbook directory structure to make it work and be done with it.

IMHO, ceph-ansible current project structure may not as how Ansible expected :-S

andrewschoen:
@hswong3i if you look at the Ansible best practices page you'll see that both directory structures we've discussed here are valid: http://docs.ansible.com/ansible/latest/playbooks_best_practices.html#directory-layout

My original contention was that ceph-ansible should not be in the business of defining which layout or approach the end user should use. Doesn't this way force a user to define their inventory inside of the ceph-ansible project? Do they also always now need to define an ``inventory_dir`` variable to let ceph-ansible know where there inventory is located? This seems redundant as you are already telling ansible where your inventory is with the ``-i`` flag. IMHO, ceph-ansible should just provide the roles and then leave it up to the end user to best use those for their situation. This PR will certainly cause lots of change for our downstream usage.

I'm not really trying to block this PR, but I think it further complicates an already complicated project and introduces more complexity while limiting flexibility. Maybe I'm missing something though.

hswong3i:
First of all let's clarify about `inventory_dir`: it is one of the Ansible build-in [Magic Variables](http://docs.ansible.com/ansible/latest/playbooks_variables.html#magic-variables-and-how-to-access-information-about-other-hosts) which you could use it anytime, NO additional manual definition is required:

> Also available, inventory_dir is the pathname of the directory holding Ansible’s inventory host file, inventory_file is the pathname and the filename pointing to the Ansible’s inventory host file.

Yes this point to the right place, which I am now proposing with the [Alternative Directory Layout](http://docs.ansible.com/ansible/latest/playbooks_best_practices.html#alternative-directory-layout), due to this original needs:
> Alternatively you can put each inventory file with its group_vars/host_vars in a separate directory. This is particularly useful if your group_vars/host_vars don’t have that much in common in different environments.

Back to the original though of this issue:

> Currently default caph-ansible example coming with following limitation:
> * Didn't provide sample inventory file as reference, e.g. inventory/sample/hosts
> * {{ fetch_directory }} is relative to project top directory, so multiple deployments will result as shared fetch/ folder generated
> * group_vars is relative to project to directory, so multiple deployments will again sharing the same configuration
> * We can't manage all deployment at once, inside centralized ansible controller

The troublesome caused by **shared `{{ fetch_directory }}` just simply related to top dir**:
* In multiple inventories deployment use case, the generation of `{{ fetch_directory }}/ceph_cluster_uuid.conf` is now [hard coded](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/tasks/facts.yml#L75) as a shared location, so result as all different deployments **hard coded** with the same uuid
* With the shared uuid, all deployments will again shared with the `{{ fetch_directory }}/{{ fsid }}`, so all generated keyring will be **enforced to be the same**

Again, when combine with the troublesome caused by **shared `group_vars` just simply related to top dir**:
* In multiple inventories deployment use case, most likely both hosts structure and its `group_vars/all.yml` need slightly different, e.g. in my above description example:
    * Both c11 and c21 zone coming with 3 ceph servers (inventory related), each with 100GB SSD x1 (`group_vars` related)
    * Both d11 and d21 zone coming with 1 ceph server only, but with 100GB SSD x3
* With current ceph-ansible [no recommendation on how inventory file should be placed](https://github.com/ceph/ceph-ansible/blob/master/docs/source/index.rst#inventory) but **by default** given with a [`group_vars` relative to top dir](http://docs.ansible.com/ansible/latest/playbooks_best_practices.html#directory-layout), which means we are misleading users that we are **by default** only support shared `group_vars` with multiple inventory files style

> IMHO, ceph-ansible should just provide the roles and then leave it up to the end user to best use those for their situation

Assuming this is how we really expected (as what am I original though, too), when migrate into [Alternative Directory Layout](http://docs.ansible.com/ansible/latest/playbooks_best_practices.html#alternative-directory-layout) it should also works well and perfectly: but the truth is it won't works, as all important stuff inside `{{ fetch_directory }}` are now shared, including the keyring...

Finally, from another point of views, applying changes from this issue will support **both** [shared configuration directory layout](http://docs.ansible.com/ansible/latest/playbooks_best_practices.html#directory-layout) and [completely isolated inventory and configuration directory layout](http://docs.ansible.com/ansible/latest/playbooks_best_practices.html#alternative-directory-layout) perfectly, so why not back to the origin that *leave it up to the end user to best use those for their situation*?

guits:
Actually, I think this is a nice PR, but for a nonexistent blocking issue.
It looks like we are changing the whole repo layout while it appears its not really needed, therefore, I would say I tend to agree with @andrewschoen 

I guess, as @andrewschoen said, this is more a question of documentation lacking than something else.

hswong3i:
You have my changes and its commit message for what you are looking for ;-)

guits:
@hswong3i, I can't understand your reaction.

It's pretty fair that whoever feel involved in this project and those in charge of maintaining it give their opinion and/or their concerns about such a patch.

The more we discuss, the more we can move forward the right way.

I think you should have addressed @andrewschoen's comments here https://github.com/ceph/ceph-ansible/pull/2367#pullrequestreview-98188383 and wait to see if we can find a consensus instead of reacting that way.

leseb:
@hswong3i, please re-add your initial change, we haven't decided yet. So remove your last commit and avoid this kind of behavior in the future, that's not helping. We are listening to contributors and happy to see contributions, so we also expect a similar reaction in regards to maintainers. 

@andrewschoen I understand what you're saying, all the comments are valid. However, I'd like to raise a few points again, in favor of this PR and perhaps also clarify this repository's goal.

ceph-ansible (here, https://github.com/ceph/ceph-ansible) is the repository that you can clone and get started. Out of the box, we hope to provide everything for a smooth experience. Any newcomer shouldn't have to know anything about Ansible, nor Ceph and I guess that's what makes this repository so attractive. By merely following the doc, you should be good to go in less than ten configuration steps, grab your coffee and check the result; your Ceph cluster is up and running.
When it comes to fine-grained/tuned Ansible configuration, we have the Ansible Galaxy for that. So we are providing each role individually. It means that if you are integrating our roles in your main Ansible structure, you know what you're doing.

I believe, this PR (if original state is submitted again) will ease:

* newcomers (Ceph sandbox)
* consultant deploying Ceph for customers (most comfortable configuration for mirrored cluster/multi-site deployments)
* QE and performance engineers who usually implement more than one cluster

All of this boils down to usability, and that's what we have been targetting recently.
Please let me know if what I just said makes sense to you, @andrewschoen, @guits

andrewschoen:
@leseb I do understand, and in my review I was trying to highlight things that I thought would cause regressions or other issues. My intention was to try to find a middle ground, not shut down the PR. I might disagree with the opinion that a "newcomer shouldn't have to know anything about Ansible" to use this repo, but I certainly can't disagree with the fact that usability is important here for all users, regardless of their level of Ansible knowledge.

In general, these were my main points:

- we can't assume ``fetch_directory`` will have always been ``fetch`` for existing deployments. Most likely that will not be the case.
- I don't like the addition of the ceph.sh script, I think it's not needed and introducing ``artifacts_directory`` is just more config in a project that already has a lot of config options.
- we can't assume ceph-ansible will always be run from a ``client`` node or from a node that even has ceph installed

I also completely understand that multiple deployments using the same fetch directory will cause major issues, I dealt with this while working on ``ceph-installer``. I know that we do store most things in the`` fetch_directory`` within a folder named after the cluster's fsid. What I don't see in that fsid directory is ceph_cluster_uuid.conf and monitor_keyring.conf. Could we always just ensure that everything stored in ``fetch_directory`` is put in the fsid folder?

hswong3i:
@andrewschoen Let me try to answer these question:

> we can't assume fetch_directory will have always been fetch for existing deployments. Most likely that will not be the case.

In this case, which means the user ALREADY define `fetch_directory` within their current `all.yml`, so our new default location will not be effective, i.e. will not affect existing deployment.

P.S. I would like to say most end-user may not truly understand the importance of `{{ fetch_directory }}` because it is not well documented... Therefore they should simply keep it as default commented in `all.yml`... If they do already change this variable, which means they already spend nubmers-of-day-and-night to understand what is it and the value should be always valid to their existing deployment even we change the default value...

> I don't like the addition of the ceph.sh script, I think it's not needed and introducing artifacts_directory is just more config in a project that already has a lot of config options.

I try to propose the use of `artifacts_directory` because I am integrating the use of both cahp-ansible and [kubespray](https://github.com/kubernetes-incubator/kubespray/blob/master/docs/getting-started.md#accessing-kubernetes-api), so would like to sync the management style when contribute if possible.

If this looks too much for ceph-ansible, we could utilize the existing `fetch_directory` for similar use cases.

On the other hand, the helper `ceph.sh` is useful for simplify daily operation: I hate to keep typing something like `ceph -c ceph.conf -k ceph.client.admin.keyring -s`  for every operation... It really looks silly... `./ceph.sh -s` looks much elegant ;-)

> we can't assume ceph-ansible will always be run from a client node or from a node that even has ceph installed

That's why **if ceph-ansible is NOT running as a client node**, the ceph.conf + ceph.client.admin.keyring + ceph.sh will NOT installed into the `artifacts_directory` (ok, let's assume it should be the `fetch_directory` as mentioned above...)

This is because if we need to access the ceph API correctly, following 3 components MUST exists:
* ceph CLI (which installed by client group)
* ceph.conf
* ceph.client.admin.keyring

Back to the origin: ceph-ansible MUST run as client node in order to access ceph API, else no need to fetch those artifacts from remote to localhost.

> What I don't see in that fsid directory is ceph_cluster_uuid.conf and monitor_keyring.conf. Could we always just ensure that everything stored in fetch_directory is put in the fsid folder?

Yes this is a good question, but a [Chicken or the egg](https://en.wikipedia.org/wiki/Chicken_or_the_egg) issue: currently ceph-ansible first read the cluster uuid from `ceph_cluster_uuid.conf`and setup the `fsid` variable, then fetching the contents from `{{ fetch_directory}}/{{fsid}}/balabalbal...`. Therefore you CAN'T put the `ceph_cluster_uuid.conf` INSIDE `{{ fetch_directory}}/{{fsid}}`: you simply don't know the cluster uuid, yet.

From the other point of view, if `{{ fetch_directory }}` become relative to `{{ inventory_dir }}` which should always unique between multiple deployment, the redundancy use of `{{ fetch_directory}}/{{fsid}}` could now simplify as `{{ fetch_directory}}` only: simple and elegant ;-)

Again back to the origin: because `{{ fetch_directory }}` is now by default shared, so we try to keep stuff unique by using `{{ fetch_directory}}{{ fsid }}`; BTW if `{{ fetch_directory }}` already unique, the use of `{{ fetch_directory}}{{ fsid }}` become redundancy.

hswong3i:
ping @leseb and @andrewschoen, shall I assume enough information provided for your discussion?

leseb:
Closing this due to inactivity.

PR #2379, opened 2018-02-08
Add support for multipath disks

squidboylan:
Add support for using multipath disks as OSDs. This requires ceph-ansible knowing how the multipath disk partition names are formatted.


guits:
@squidboylan please, add a commit message.

PR #2397, opened 2018-02-16
client: use `ceph_uid` fact to set uid/gid on admin key

guits:
That task is failing on containerized deployment because `ceph:ceph`
doesn't exist.
The idea here is to use the `{{ ceph_uid }}` to set the ownerships for
the admin keyring when containerized_deployment.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1540578

### DIFF
--- a/roles/ceph-client/tasks/pre_requisite.yml
+++ b/roles/ceph-client/tasks/pre_requisite.yml
@@ -20,7 +20,7 @@
     - sestatus is defined
     - sestatus.stdout != 'Disabled'
 
-- name: copy ceph admin keyring
+- name: copy ceph admin keyring when non containerized deployment
   copy:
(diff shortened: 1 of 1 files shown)


roles/ceph-client/tasks/pre_requisite.yml
```diff
@@ -24,8 +24,8 @@
   copy:
     src: "{{ fetch_directory }}/{{ fsid }}/etc/ceph/{{ cluster }}.client.admin.keyring"
     dest: "/etc/ceph/"
-    owner: "ceph"
-    group: "ceph"
+    owner: "{{ ceph_uid }}"
```
leseb: This won't work since `ceph_uid` is only set for containerized deployment, this breaks non-containerized deployment gros, see: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/tasks/facts.yml#L173

guits:

PR #2403, opened 2018-02-19
Fix template reference for ganesha.conf

andymcc:
We can simply reference the template name since it exists within the
role that we are calling. We don't need to check the ANSIBLE_ROLE_PATH
or playbooks directory for the file.

### DIFF
--- a/roles/ceph-nfs/tasks/start_nfs.yml
+++ b/roles/ceph-nfs/tasks/start_nfs.yml
@@ -25,7 +25,7 @@
 - name: generate ganesha configuration file
   action: config_template
   args:
-    src: "{{ lookup('env', 'ANSIBLE_ROLES_PATH') | default (playbook_dir + '/roles', true) }}/ceph-nfs/templates/ganesha.conf.j2"
+    src: "ganesha.conf.j2"
     dest: /etc/ganesha/ganesha.conf
     owner: "root"
     group: "root"


guits:
@andymcc It looks like a leftover when `ganesha.conf.j2` used to be in `ceph-common`.
We had to do this because of the `playbook_dir` variable, if I'm not mistaking it might caused issues with OSP deployment for instance. I would say I approve this PR, just need to @leseb to  confirm whether what I said is correct.

andymcc:
@guits - sounds good! thanks!

leseb:
This is explained in this commit 76ac9b077b0c239c584a9e677a4cf809cdd93f0d, I believe the issue is still present.

andymcc:
Is there a way we can test if that issue is still present? I don't think it should be, since the primary location of the template is in the role if you call that role. It looks like in the link you made the other changes have changed to just be straight includes: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/tasks/main.yml#L45

The issue is that if your ANSIBLE_ROLES_PATH has more than one path (which is pretty common) then that doesn't work, and the playbook_dir may well not be the location of the roles to point to.

This means in order for this to work you have to set ANSIBLE_ROLES_PATH specifically for this task only, and/or ensure you run the playbook from inside the ceph-ansible dir only.

leseb:
@guits hum maybe, but the tripleo guys could try to reproduce the error with this patch.
@fultonj could you please run a deployment with this patch and let us know if you have any issues?
Thanks in advance for your help.

guits:

guits:

andymcc:
@fultonj did you get a chance to run this? The bug prevents you from using ceph-ansible (nfs role) as an imported role, would be good to get it fixed up! I can't see it causing any issues but happy for you to check first.

andymcc:
@leseb @guits - i know this is old now, any chance we can get this looked at again? It's basically a blocker to using the nfs ceph-ansible role as part of something bigger (importing the role) since ANSIBLE_ROLES_PATH would be different.

guits:
@andymcc in my opinion this is good to go but @leseb had concern that needed to be confirmed by @fultonj

guits:
although I approve this PR I still want to have @fultonj opinion

fultonj:
Sorry @andymcc I didn't see this until just now. Let me see about getting this tested.

tombarron:
I deployed OpenStack with manila and cephFS NFS back end with ceph-ansible patched with this PR.  The deployment was successful, /etc/ganesha/ganesha.conf was set up in the cephnfs (ganesha) container with the expected contents, and the ganesha service is running properly.

+1

PR #2404, opened 2018-02-19
[TEST] Test setting up correct systemd file for nfs-ganesha

andymcc:
Don't merge this.
Test to see if we copy over the nfs-ganesha-lock.service.debian8 file
properly, whether the Xenial CI job will work.

The upstream download.ceph.com nfs-ganesha package should be fixed for
xenial (which is in progress).

### DIFF
--- a/roles/ceph-nfs/tasks/start_nfs.yml
+++ b/roles/ceph-nfs/tasks/start_nfs.yml
@@ -68,6 +68,13 @@
   notify:
     - restart ceph nfss
 
+- name: copy over nfs-ganesha-lock.service file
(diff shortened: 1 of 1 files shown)


andymcc:
we know it works :) we just need to wait for the package to get fixed in download.ceph.com

guits:
@andymcc This is a good workaround, the real fix should reside in the package though

roles/ceph-nfs/tasks/start_nfs.yml
```diff
@@ -68,6 +68,13 @@
   notify:
     - restart ceph nfss
 
+- name: copy over nfs-ganesha-lock.service file
+  shell: "cp /lib/systemd/system/nfs-ganesha-lock.service.debian8 /lib/systemd/system/nfs-ganesha-lock.service"
```
leseb: `command`?

PR #2413, opened 2018-02-22
Make rule_name optional when defining items in openstack_pools

gfidente:
Previously it was necessary to provide a value (eventually an
empty string) for the "rule_name" key for each item in
openstack_pools. This change makes that optional and defaults to
empty string when not given.


ASBishop:
+1

leseb:
@gfidente should this be backported in stable-3.0?

leseb:

leseb:

gfidente:
hi @leseb, there is no pressure because 3.0 consumers are probably passing the empty string anyway but if we can backport it, it should be harmless so I'd do that

PR #2417, opened 2018-02-26
[skip ci] Clarify RGWs configuration when using ceph_conf_overrides.

jtudelag:
To avoid future misconfigurations, clarify that the only valid
scheme is [client.rgw.<instance-name>] instead of [client.radosgw.<instance-name>].


guits:

guits:

guits:

jtudelag:
@leseb @guits please review.

jtudelag:
Cool, I was not sure this variable templating could be done in a group_vars/ file: "client.rgw.{{ hostvars[inventory_hostname]['ansible_hostname'] }}":.

In that case, If you dont mind, I will replace client.rgw.<instance-name> with "{{ hostvars[groups.get(rgw_group_name)[0]]['ansible_hostname'] }}. It will save a a lot of time for many people. 

I would also add a note pointing to github.com/ceph/ceph-ansible/blob/master/profiles. I usually dont like pointing to master because eventually, the link will be broken in the future, but ok.

ktdreyer:
Agreed that it would be great to not point to a GitHub master URL. Why not just say "the profiles directory" ?

PR #2424, opened 2018-02-28
Create rgw pools containerized

jtudelag:
Adds RGWs pool creation to containerized installation.
    
ceph command has to be executed from one of the monitor containers.
Task has to be delegated.


roles/ceph-rgw/tasks/docker/main.yml
```diff
@@ -4,3 +4,6 @@
 
 - name: include start_docker_rgw.yml
   include: start_docker_rgw.yml
+
+- name: include rgw_pool_pgs_docker.yml
+  include: rgw_pool_pgs_docker.yml
```
guits: add:
```
when:
  - create_pools is defined
  - containerized_deployment
```

roles/ceph-rgw/tasks/docker/rgw_pool_pgs_docker.yml
```diff
@@ -0,0 +1,23 @@
...
+
+   # Execute the command on the first monitor, just once.
+   - name: create rgw pools
+     command: "{{ docker_exec_cmd }} ceph --connect-timeout 5 --cluster {{ cluster }} osd pool create {{ item.key }} {{ item.value.pg_num }}"
+     changed_when: false
+     with_dict: "{{ create_pools }}"
+     delegate_to: "{{ groups['mons'][0] }}"
+     run_once: true
+     become: true
+  when:
```
guits: replace this `when` with a `run_once: true`

roles/ceph-rgw/tasks/docker/rgw_pool_pgs_docker.yml
```diff
@@ -0,0 +1,23 @@
+---
+- name: create rgw pools if create_pools is defined
+  block:
+   # Get the short hostname of the first monitor to run the ceph command.
+   - name: find first monitor
```
guits: remove this task

roles/ceph-rgw/tasks/docker/rgw_pool_pgs_docker.yml
```diff
@@ -0,0 +1,23 @@
...
+     set_fact:
+       docker_exec_cmd: "docker exec ceph-mon-{{ ceph_first_monitor_hostname }}"
+
+   # Execute the command on the first monitor, just once.
+   - name: create rgw pools
+     command: "{{ docker_exec_cmd }} ceph --connect-timeout 5 --cluster {{ cluster }} osd pool create {{ item.key }} {{ item.value.pg_num }}"
+     changed_when: false
+     with_dict: "{{ create_pools }}"
+     delegate_to: "{{ groups['mons'][0] }}"
+     run_once: true
```
guits: remove this `run_once: true`

roles/ceph-rgw/tasks/docker/rgw_pool_pgs_docker.yml
```diff
@@ -0,0 +1,23 @@
...
+  block:
+   # Get the short hostname of the first monitor to run the ceph command.
+   - name: find first monitor
+     set_fact:
+       ceph_first_monitor_hostname: "{{ hostvars[item].ansible_hostname }}"
+     with_items: "{{ groups['mons'][0] }}"
+
+   - name: set_fact docker_exec_cmd
+     set_fact:
+       docker_exec_cmd: "docker exec ceph-mon-{{ ceph_first_monitor_hostname }}"
```
guits: replace `{{ ceph_first_monitor_hostname }}` with `{{ hostvars[groups.get(mon_group_name)[0]]['ansible_hostname'] }}`

roles/ceph-rgw/tasks/docker/rgw_pool_pgs_docker.yml
```diff
@@ -0,0 +1,23 @@
...
+
+   # Execute the command on the first monitor, just once.
+   - name: create rgw pools
+     command: "{{ docker_exec_cmd }} ceph --connect-timeout 5 --cluster {{ cluster }} osd pool create {{ item.key }} {{ item.value.pg_num }}"
+     changed_when: false
+     with_dict: "{{ create_pools }}"
+     delegate_to: "{{ groups['mons'][0] }}"
```
guits: do not hardcore `'mons'`, use the variable `mon_group_name`

guits:
@jtudelag could you please squash your commits ? there is no need to have 3 separated commits for this

roles/ceph-rgw/tasks/docker/main.yml
jtudelag: This task is inside tasks/docker/.   containerized_deployment has already been check here: https://github.com/jtudelag/ceph-ansible/blob/a551faca4036a0418d3c4cd7216281c856610203/roles/ceph-rgw/tasks/main.yml#L48

guits: true

roles/ceph-rgw/tasks/docker/rgw_pool_pgs_docker.yml
jtudelag: You are right. I guess sth like this would do the trick: {{ groups[mon_group_name][0] }}.

guits: yes

jtudelag: Btw @guits, main playbooks have hardcoded group names in the host section:
* https://github.com/ceph/ceph-ansible/blob/master/site.yml.sample#L4
* https://github.com/ceph/ceph-ansible/blob/master/site-docker.yml.sample#L4

Maybe I'm missing sth, but In that case, shouldn't they be changed for sth like this??

  hosts:
    - "{{ mon_group_name|default('mons') }}"
    - "{{ osd_group_name|default('osds') }}"
    - "{{ mds_group_name|default('mdss') }}"
....

guits: @jtudelag no, it won't work i think because if i'm not mistaking, at this point `mon_group_name` is still an unknown variable even if defined in `group_vars/`

roles/ceph-rgw/tasks/docker/main.yml
```diff
@@ -4,3 +4,8 @@
...
```
jtudelag: @guits rewritten, please review.

roles/ceph-rgw/tasks/rgw_pool_pgs.yml
```diff
@@ -2,7 +2,7 @@
 - name: create rgw pools if create_pools is defined
   command: ceph --connect-timeout 5 --cluster {{ cluster }} osd pool create {{ item.key }} {{ item.value.pg_num }}
```
guits: prefix with `{{ docker_exec_rgw_cmd }}`

roles/ceph-rgw/tasks/rgw_pool_pgs.yml
```diff
@@ -2,7 +2,7 @@
...
   changed_when: false
-  failed_when: false
+  run_once: true
   with_dict: "{{ create_pools }}"
```
guits: insert a `delegate_to: "{{ groups[mon_group_name][0] }}"`

roles/ceph-rgw/tasks/rgw_pool_pgs.yml
```diff
@@ -2,7 +2,7 @@
...
   when:
```
guits: remove this `when`

roles/ceph-rgw/tasks/main.yml
```diff
@@ -30,7 +30,7 @@
 - name: include rgw_pool_pgs.yml
   include: rgw_pool_pgs.yml
   when:
-    - create_pools is defined
+    - not containerized_deployment
```
guits: revert this change

roles/ceph-rgw/tasks/main.yml
```diff
@@ -30,7 +30,7 @@
 - name: include rgw_pool_pgs.yml
```
guits: move this task after the include of `docker/main.yml`

roles/ceph-rgw/tasks/docker/rgw_pool_pgs_docker.yml
```diff
@@ -0,0 +1,12 @@
+---
```
guits: keep the task `set_fact` rename the fact to `docker_exec_rgw_cmd` and move it in `roles/ceph-rgw/tasks/main.yml`, drop the rest (remaining task + file)

jtudelag: Isn't it better to keep the docker/containerized stuff all together inside tasks/docker instead of having the code spread all over the role?

Actually I was going to propose to move this fact inside tasks/docker/main.yml:
https://github.com/ceph/ceph-ansible/blob/v3.0.27/roles/ceph-rgw/tasks/main.yml#L2

guits: @jtudelag in fact, not really.
We aim to get rid of these kind of splits because this duplicates lot of code while it's not necessary. So let's write patches in that sense when it's possible (the `{{ docker_exec_cmd }}` facts were even created for that.).
By the way, this approach would fix the non-containerized case, which at first sight, seems to be broken.

roles/ceph-rgw/tasks/rgw_pool_pgs.yml
jtudelag: @guits Just curious. Whats the point of having "mon_group_name", and the rest of the group names variables,  if we are forced to hardcode the default group names (mons, rgws, etc) in the main playbook.

https://github.com/ceph/ceph-ansible/blob/master/site-docker.yml.sample#L4

jtudelag: Btw, per this PR https://github.com/ceph/ceph-ansible/pull/2428/files, which copies client.admin.keyring to RGWs, I think there is no need to delegate the task to monitors anymore as RGWs could execute it. Am I wrong?

guits: @jtudelag it let users the possibility to rename their groups as they want.
The 2 main playbooks in the repository are just provided as sample where you would have to rename the group in any case if you decide to use them with custom group names except if you keep the default values.

For instance, if you want to rename your monitor group you will have only few occurrences to edit whereas you will have more than 100 in the whole playbook as you can see below :

```
 guits@work:(master)~/GIT/ceph-ansible$ git grep 'mons' site*.sample | wc -l 
4
 guits@work:(master)~/GIT/ceph-ansible$ git grep 'mon_group_name' | wc -l
148
```

guits: @jtudelag there is a condition to copy the admin keyring so if you don't set `copy_admin_key: true` you won't have it on your rgw node.

jtudelag: @guits Thanks for the explanations! So I guess we can keep it this way. Running it delegated to the first mon guarantees the admin keyring is present and the command can be executed.

guits:

roles/ceph-rgw/tasks/rgw_pool_pgs.yml
```diff
@@ -2,7 +2,23 @@
 - name: create rgw pools if create_pools is defined
```
guits: You can remove this task

roles/ceph-rgw/tasks/rgw_pool_pgs.yml
```diff
@@ -2,7 +2,23 @@
...
+  when:
+    - containerized_deployment
+
+- name: create rgw pools if create_pools is defined
+  command: "{{ docker_exec_rgw_cmd }} ceph --connect-timeout 5 --cluster {{ cluster }} osd pool create {{ item.key }} {{ item.value.pg_num }}"
+  changed_when: false
+  run_once: true
+  with_dict: "{{ create_pools }}"
+  delegate_to: "{{ groups[mon_group_name][0] }}"
+  when:
```
guits: @jtudelag remove this when

roles/ceph-rgw/tasks/rgw_pool_pgs.yml
jtudelag: Sorry @guits Im no following your reasoning. Why removing this task? Its the one that creates pools when not containerized. I guess what you want is set a fact depnding on containerized/non conatinerized and then having just one task that uses the fact to execute the command. Am I wrong?

roles/ceph-rgw/tasks/rgw_pool_pgs.yml
jtudelag: If I do this, this task will be executed always. And when non-containerized, the fact docker_exec_rgw_cmd would be undefined.

guits: yes, this is the idea, when non-containerized the fact `docker_exec_rgw_cmd ` won't be defined and the command won't include the prefix `docker exec ceph-mon ...` so it can be executed in non-containerized deployment.
But somehow you are right, so I guess you have to add `| default('')` like this :
`command: "{{ docker_exec_rgw_cmd }} | default('') ceph --connect-timeout 5 --cluster {{ cluster }} osd pool create {{ item.key }} {{ item.value.pg_num }}"`
nice catch, I missed this.

That way you make that task common to both containerized and non-containerized deployment and avoid kind of duplicated code so you can remove the first task.

jtudelag: I can see. Isn't it a little hacky? Also, you would need to be carefull with trailing  spaces 

In that case, wouldn't it be better, at least less hacky, more readable and easy to maintain, to have a fact to set. For example `"ceph_command" : "ceph" `when non containerized, and `"ceph_comand": "docker exec..."`.  Then in the main tasks `command: "{{ ceph_command }} --cluster ...."`.

guits: @jtudelag At first sight, I don't see why having more task than necessary is more readable and easy to maintain, the more of code you have harder it is to maintain.

I think it's more logic to keep the common part of the command outside of the fact.
The only difference is the prefix `docker exec <name>` so this is well the variable.
By the way, it can be a bit annoying for debugging purpose to have the command split.

I'm not trying to block your suggestion, just sharing my own opinion.
So what I would propose is to keep consistency for this PR with what we already have in the playbook.
Then, if we should change the style of how we manage ceph command for containerized deployment vs. non-containerized, I think it should be addressed in a dedicated PR so it could be applied in the whole playbook, again, for consistency reasons, because I don't really see the point of having mixed styles to achieve a same goal at many places.

@jtudelag What do you think?

leseb:
@jtudelag up :).

roles/ceph-rgw/tasks/rgw_pool_pgs.yml
jtudelag: @guits @leseb sorry guys, long holidays completely isolated from anything with internet ;).

**"So what I would propose is to keep consistency for this PR with what we already have in the playbook."**

Agree with that, I will go back to my first approach. I will keep Docker related tasks inside docker/ folder.

**"Then, if we should change the style of how we manage ceph command for containerized deployment vs. non-containerized, I think it should be addressed in a dedicated PR so it could be applied in the whole playbook, again, for consistency reasons, because I don't really see the point of having mixed styles to achieve a same goal at many places."**

Agree, happy to help here and propose a few ideas. Every decision has Its trade-offs, so in the end, It is a matter of being consistent with what It's been agreed ;).

guits:
@jtudelag yes, that's correct, for container cluster scenario this is not covered, it should be added.
Do you mind adding a commit for this?

jtudelag:
@guits Yes, why not. I will create a separate PR for this.

guits:
@jtudelag I think you can add it in this current PR since this is related.

jtudelag:
Added tests for container scenarios. Let's see If I didn't screw it up.

jtudelag:
Actually I  think all docker tests are skipped for **docker_cluster_collocation** ;(.

roles/ceph-rgw/tasks/main.yml
```diff
@@ -48,3 +49,9 @@
     - containerized_deployment
   # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
   static: False
+
+- name: include rgw_pool_pgs.yml
```
leseb: why is `rgw_pool_pgs.yml` included twice?

roles/ceph-rgw/tasks/main.yml
```diff
@@ -48,3 +49,9 @@
...
+  include: rgw_pool_pgs.yml
+  when:
+    - create_pools is defined
```
leseb: `create_pools` is not really meaningful, please use something like `rgw_create_pools`. Thanks

roles/ceph-rgw/tasks/main.yml
jtudelag: Your right, my mistake.

roles/ceph-rgw/tasks/main.yml
jtudelag: Don't agree, this is the name of variable in the rgws.yml.sample , and not sure in how many places is called or even worst, if it is mentioned somewhere in the downstream documentation: 

https://github.com/ceph/ceph-ansible/blob/master/group_vars/rgws.yml.sample#L45

If you think this should be renamed, I think it should be done in a different PR.

guits: I guess it's a misunderstanding, this variable wasn't introduced in this patch so @jtudelag is just reusing it.
In any case, I agree with @leseb it should be renamed to avoid confusion.

I don't mind to either add a commit in this PR or in a dedicated one @jtudelag

guits:
@jtudelag I'll try to find some free time to take a look at collocated tests that seems to be failing :-/

roles/ceph-rgw/tasks/main.yml
leseb: Fair enough, but adding a commit to change the name won't hurt, there is no need for a separate PR.
You have a change to do anyway, so please include this as well.
Thanks.

guits: @jtudelag any news?

guits: @jtudelag we have a [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1550281) targeted for v3.0z4 linked to this patch, that would be awesome to get this patch merged very soon :-)

jtudelag:
@leseb @guits renamed the create_pools to rgw_create_pools.

roles/ceph-rgw/tasks/main.yml
jtudelag: @guits done ;)

guits:
@jtudelag thanks for this, one last thing, could you please fix your last commit by adding a sign-off and a commit message?

jtudelag:
@guits done!!

guits:
jenkins test luminous-ansible2.4-docker_collocation

guits:

guits:

guits:

guits:
no sure why it's complaining for the sign off, it looks good to me anyway

PR #2425, opened 2018-02-28
Choose disk

leseb:
Fresh resume of choose_disk with the content of the module + doc only.
@ErwanAliasr1


alfredodeza:
The objective of this module is great, we will need something like this to move
forward with LV support in ceph-ansible.

We would like for this to be reworked, from scratch, so that only the parts
that are relevant to logical volumes and ceph-volume are used.

The current implementation would need to have too much taken out or re-done in
order to work correctly. I will go ahead with a full review given the new
context that we need for ceph-volume so that it is clear what we are
envisioning moving forward.

ErwanAliasr1:
@alfredodeza if we only support LVs ,who is taking care of creating them ?

leseb:
I think we can use this module: https://docs.ansible.com/ansible/2.5/modules/lvg_module.html
Basically, `choose_disk` does the detection and passes around to the `lvg` module the devices discovered.

How does that sound?

alfredodeza:
@leseb @ErwanAliasr1 a module like this one is surely what we need as part of the larger objective of handling LVs. The argument here is that *although* this module has the right idea, we would need to re-work so much that it would be better to start over with ceph-volume and bluestore support.

So yes, something like this module is highly desirable, the "create lvs" would happen with the lv module from Ansible for sure (yet to be implemented here in ceph-ansible).

library/choose_disk.py
```diff
@@ -361,6 +363,17 @@ def disk_label(partition):
     if "ceph journal" in stdout:
         return "journal"
 
+    # Then, let's search for metadata coming from ceph-volume
+    output_cmd = subprocess.Popen("ceph-volume lvm list --format=json {}".format(partition),
```
alfredodeza: do not use `shell=True` here, it is safer to use itemized arguments on a list

library/choose_disk.py
```diff
@@ -361,6 +363,17 @@ def disk_label(partition):
...
+                                  shell=True, stdout=subprocess.PIPE)
+    raw_json, _ = output_cmd.communicate()
+    json = json.loads(raw_json)
```
alfredodeza: this is blindly trusting the output of the CLI. If the CLI breaks, it will do so with invalid JSON, this must be caught and dealt with, even if it means it must fail here.

library/choose_disk.py
```diff
@@ -361,6 +363,17 @@ def disk_label(partition):
...
+    for item in json:
+        current_item = json[item]
```
alfredodeza: If you need to look at values as well as the keys from the JSON, you should iterate over `items()` that will give you both, instead of asking the dictionary for `current_item`.

That would look like:

    for key, value in json.items():

library/choose_disk.py
```diff
@@ -361,6 +363,17 @@ def disk_label(partition):
...
+        if "type" in current_item[0]:
```
alfredodeza: this is only looking at the `0` index, which means it is assuming there will always be items in that list. Not entirely sure when/if we will see an empty JSON for a device, but it would be just safer to loop over the list and retrieve the values that you need

library/choose_disk.py
```diff
@@ -361,6 +363,17 @@ def disk_label(partition):
...
+            return current_item[0]["type"]
+        else:
+            return "undefined"
```
alfredodeza: what would be the use case for having 'undefined' here vs. `""` below?

library/choose_disk.py
```diff
@@ -412,10 +421,23 @@ def select_only_free_devices(physical_disks):
...
             if not ceph_disk:
                 logger.info('Ignoring %10s : device has existing partitions', physical_disk)
                 continue
 
+        # If we reach here and ceph_disk is not populated, it means it could be an LVM
+        # We didn't checked LVM before as some ceph disks could be under LVM
+        # disk_label is supposed to catch this case, so it only remain LVM made by the user
+        if not ceph_disk:
+            # Does the disk belongs to a LVM ?
+            return_code = os.system("pvdisplay -c /dev/{}".format(physical_disk))
```
alfredodeza: please avoid using `os.system`. This command will send all its output to stdout without having a way to silence it. This line only care about the return code, and subprocess can certainly handle that

library/choose_disk.py
ErwanAliasr1: Thx. Fixed.

library/choose_disk.py
ErwanAliasr1: Fixed it, thx.

library/choose_disk.py
ErwanAliasr1: I agree that if the output is empty, it have to be handled : fixed.
As I'm parsing the disk directly, it would be erratic have multiple entries. I've added a check for this case and parse the sole entry I should get.

ErwanAliasr1: And as its a list, I cannot access it without an integer index.

library/choose_disk.py
ErwanAliasr1: I should have been more explicit. Let's me clarify that

library/choose_disk.py
ErwanAliasr1: roger that. Fixed.
Thx for the insightful review.

library/choose_disk.py
```diff
@@ -606,7 +626,7 @@ def main():
     matched_devices = None
     lookup_disks = None
 
-    setup_logging()
+    setup_logging("/tmp/choose_disk.log")
```
alfredodeza: Please don't use a hard coded path for logging. This should be either a well known/established location or at the very least a configurable one. In systems where `/tmp/` doesn't exist, this would break as the `setup_logging` function doesn't deal with any errors for log locations

ErwanAliasr1: This should not have been commited .... Wrong staging here. Fixed by 508145e9128fe375c7a3cd8a57f1a3af4e6d7d2d

Thx !

library/choose_disk.py
```diff
@@ -602,7 +602,7 @@ def main():
     matched_devices = None
     lookup_disks = None
 
-    setup_logging("/tmp/choose_disk.log")
+    setup_logging()
```
alfredodeza: this commit should be squashed then?

library/choose_disk.py
```diff
@@ -353,46 +353,69 @@ def expand_disks(lookup_disks, ceph_type="", module=None):
...
+                        return "foreign : {}".format(value[0]["tags"]["ceph.cluster_fsid"])
+            if "type" in value[0]:
+                return value[0]["type"]
+            else:
+                # ceph-volume should give a type
+                # If not, we don't really what it is _but_ the disks is used
+                # So we should ignore it, let's report it undefined
+                return "undefined"
+
+    except subprocess.CalledProcessError:
```
alfredodeza: eating out any/all errors from a subprocess call is going to make it difficult to debug this, I would expect some logging with stderr and stdout handling at the very least

library/choose_disk.py
```diff
@@ -353,46 +353,69 @@ def expand_disks(lookup_disks, ceph_type="", module=None):
...
+        pass
+
+    # 2) let's try with ceph-disk
+    # If the disk have some metadata, the json is filled with information
+    for item in ceph_disk:
```
alfredodeza: the cyclomatic complexity for this block is too high, ideally you wouldn't want to nest these if/for/if/if/if/if in this way because it is hard to understand what is going on. Breaking up these conditionals will make it better.

library/choose_disk.py
```diff
@@ -353,46 +353,69 @@ def expand_disks(lookup_disks, ceph_type="", module=None):
...
+                if "path" in partition_iter:
+                    if partition_iter["path"] == partition:
+                        if "ceph_fsid" in partition_iter:
+                            if partition_iter["ceph_fsid"] != current_fsid:
+                                return "foreign : {}".format(partition_iter["ceph_fsid"])
+
+    # 3) let's search directly from the partitions labels
+    try:
+        stdout = subprocess.check_output(["blkid", "-s", "PARTLABEL", "-o", "value", "{}".format(partition)])
+    except subprocess.CalledProcessError:
```
alfredodeza: this is another instance of eating completely the stderr,stdout from the call, which is going to cause issues when trying to debug failures. `"failed"` is not sufficient here.

library/choose_disk.py
```diff
@@ -353,46 +353,69 @@ def expand_disks(lookup_disks, ceph_type="", module=None):
...
+                                      stdout=subprocess.PIPE)
+        raw_json, _ = output_cmd.communicate()
+        try:
+            json_dict = json.loads(raw_json)
+        except (ValueError, TypeError) as e:
+            fatal("Cannot parse ceph-volume properly : {}".format(e))
+
+        # If the disk have some metadata, the json is filled with information
+        for key, value in json_dict.items():
+            # As we parse a partition it should only have a single entry unless that's an error
```
alfredodeza: this is not correct, we don't force a single partition, why would that be an error?

library/choose_disk.py
```diff
@@ -353,46 +353,69 @@ def expand_disks(lookup_disks, ceph_type="", module=None):
...
+            if (len(value) > 1):
+                fatal("Error while parsing ceph-volume")
+            if "tags" in value[0]:
+                if "ceph.cluster_fsid" in value[0]['tags']:
+                    if value[0]["tags"]["ceph.cluster_fsid"] != current_fsid:
+                        return "foreign : {}".format(value[0]["tags"]["ceph.cluster_fsid"])
```
alfredodeza: what does "foreign" mean?

leseb: "foreign" is a common term used on RAID systems. It describes a disk that has an existing RAID configuration. Typically the controller will see that information when a new disk is inserted and will mark this drive has "foreign". It's caution mechanism helping the administrator that the disk is part of another RAID.

In our case, this means we have found a drive that has existing Ceph metadata but from a **different** cluster than the current one.

library/choose_disk.py
ErwanAliasr1: I do force the parsing of the partitio with "output_cmd = subprocess.Popen(["ceph-volume", "lvm", "list", "--format=json", partition"]

library/choose_disk.py
ErwanAliasr1: This code is running by successive exclusion. We might split the cases in various functions.

library/choose_disk.py
ErwanAliasr1: Failed is not a debugging message but a label to say "this is not free but we don't understand what was wrong with it".
An extra debugging info could be added.

library/choose_disk.py
ErwanAliasr1: Commit message of 3983331058042958d41aa375f8c50cdb24a5a1bc does give a complete description.

ErwanAliasr1:
@alfredodeza yes you can break the operator code. Is it the priority for now ? We agreed to make a fast track to get ansible/choose_disk/ceph-volume working all together. I don't have the feeling this part of the code is the top most priority. We will hardened it later when all the chain will work perfectly.

ErwanAliasr1:
@alfredodeza  I do also run mccabe and try to make code as efficient as possible.

andrewschoen:
I think as an alternative to the operator code we could add a ``max_size`` and a ``min_size`` parameter to the module and only require that one be set. If only ``max_size`` is set then it would be equal to or lesser than that value. If only ``min_size`` is set then it would be greater than or equal to that value. If both are set, it would be a range between the two values. I think this would accomplish the usecase we want here without the operator code. We could make it simple and always assume GB on these values as well.

@ErwanAliasr1 I do think the operator code is an important part of this, but the current solution is just not as robust as we need it to be and I believe will generate a lot of bug reports and user confusion. Please consider the solution I've mentioned above as a possible alternative.

ErwanAliasr1:
@andrewschoen I understand your position & arguments but as @alfredodeza your are only considering the size argument which is one of the most obvious.

But this operators can work on _any_ parameters than we have about disks : it could be a temperature, it could a number of bad sectors, it could be strictly anything the admin want to keep the control on.

"I'm looking for a disk below 35° which have no broken sectors, from Sandisk, which is an SSD and is size is between(100 & 300GB" -> that's the idea of this feature. If no function is mentioned then equal is used as a default meaning that unless an admin wants that advanced feature, it have no 'cost' for a regular user or test.

Yes, things can be improved, hardened, controlled etc.., but being able to control every single disk feature by adding functions in its value offer a great area of usage and advanced admin loves that. For every single admin I trained on that feature, it took some time for them to perfectly get but after they like it so much.

Another way to express that, this is a way to implement an hardware conformity matching at runtime and we need it to satisfy installing ceph on large clusters.

andrewschoen:
> "I'm looking for a disk below 35° which have no broken sectors, from Sandisk, which is an SSD and is size is between(100 & 300GB"

Why couldn't that be done with a parameters to the module as well? That seems like a very complicated selection criteria, but perhaps it'd be useful. I tend to think it might be too much to support on a first pass here, but adding ``max_temp``, ``min_temp``, ``sectors``, ``brand`` and ``type`` would probably satisfy your requirement.

The point still stands that the current implementation that uses regex is not robust enough. It will cause us trouble in the long run. I can not support this operator implementation unless it was rewritten as a proper parser with a lexer, tokenizer, etc. which is a large amount of work. I'm trying to find a middle ground here where we can get some, if not all, of the features you want without the custom operators. I believe adding more parameters to the module is that middle ground.

ErwanAliasr1:
@andrewschoen This is what you miss here. All the detected features are generic and depends on the hardware & ansible. My code is agnostic of that and so the operators. I don't have specific operators and yes it's useful for advanced end user. I'm always surprised when developers says that an end-user features is not useful. 

The can can be improved but proved to be working on customers without much troubles. This is a revamp from edeploy. Before reworking that code let's see if that generates as much issues as you think it does.

If It does, I'll handle the associate workload to support it, don't worry on that.

leseb:
@alfredodeza @andrewschoen @ErwanAliasr1  @guits you will find in https://github.com/ceph/ceph-ansible/pull/2425/commits/4504b6b29e1f8f1b6b42e0171f13265ccdd82f88 the new implementation of the operand that I hope should solve everyone's concerns.

Also, this is rough so you can get an idea on how things work, I still need to work on the exception handling properly by using our `fatal` function.

Thanks in advance for your reviews.

alfredodeza:
@leseb this is a step in the right direction. Looking forward to the rest

leseb:
Current unit test status:

```
====================================================================================================================================================== test session starts ======================================================================================================================================================
platform darwin -- Python 2.7.14, pytest-3.4.0, py-1.5.2, pluggy-0.6.0 -- /usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python
cachedir: .pytest_cache
rootdir: /Users/leseb/Documents/Ceph/ceph-ansible-1, inifile:
plugins: mock-1.7.1
collected 13 items

library/test_choose_disk.py::test_expand_disks_explicit_count_1 PASSED                                                                                                                                                                                                                                                    [  7%]
library/test_choose_disk.py::test_expand_disks_explicit_count_2 PASSED                                                                                                                                                                                                                                                    [ 15%]
library/test_choose_disk.py::test_fake_device PASSED                                                                                                                                                                                                                                                                      [ 23%]
library/test_choose_disk.py::test_expand_disks_legacy PASSED                                                                                                                                                                                                                                                              [ 30%]
library/test_choose_disk.py::test_units_gb PASSED                                                                                                                                                                                                                                                                         [ 38%]
library/test_choose_disk.py::test_find_match_matched PASSED                                                                                                                                                                                                                                                               [ 46%]
library/test_choose_disk.py::test_find_match_matched_gt PASSED                                                                                                                                                                                                                                                            [ 53%]
library/test_choose_disk.py::test_find_match_matched_lt PASSED                                                                                                                                                                                                                                                            [ 61%]
library/test_choose_disk.py::test_find_match_matched_between PASSED                                                                                                                                                                                                                                                       [ 69%]
library/test_choose_disk.py::test_find_match_matched_between_e PASSED                                                                                                                                                                                                                                                     [ 76%]
library/test_choose_disk.py::test_find_match_matched_gt_units PASSED                                                                                                                                                                                                                                                      [ 84%]
library/test_choose_disk.py::test_find_match_unmatched PASSED                                                                                                                                                                                                                                                             [ 92%]
library/test_choose_disk.py::test_select_only_free_devices FAILED                                                                                                                                                                                                                                                         [100%]
```

`test_select_only_free_devices` is more complex to test now as we introduced subprocesses. I'll look next week into extracting the subprocesses into individual functions they will be easier to test.

leseb:

ErwanAliasr1:
I wasn't pro for this change as it complexified the input for users.
As per the latest commits for having eq as default operator, the input is more relaxed for most of the cases.
I would have enjoyed keeping the aliases to get between() operator.
I don't see a major reason not to ack this patch set.

leseb:

leseb:

leseb:

leseb:
jenkins test pipeline

library/choose_disk.py
```diff
@@ -822,6 +838,9 @@ def main():
         logger.info("Legacy syntax")
         logger.info("devices : %s", devices)
         input_mode = "legacy"
+        for device in devices:
+            if len(getParentBlock(device)):
+                fatal("Device {} have a parent device which is not allowed in legacy mode".format(device), module)  # noqa E501
```
leseb: s/have/has/

library/choose_disk.py
```diff
@@ -832,6 +851,9 @@ def main():
         lookup_disks = expand_disks(fake_device(devices, "data"), "data", module)
         if dedicated_devices:
             logger.info("dedicated_devices : %s", dedicated_devices)
+            for dedicated_device in dedicated_devices:
+                if len(getParentBlock(dedicated_device)):
+                    fatal("Device {} have a parent device which is not allowed in legacy mode".format(dedicated_device), module)  # noqa E501
```
leseb: ditto

leseb:
jenkins test pipeline

ErwanAliasr1:
jenkins test pipeline

leseb:
jenkins test pipeline

leseb:
jenkins test pipeline

leseb:

leseb:
Current ceph-volume error:

```
TASK [ceph-osd : run choose disk detection] ************************************
task path: /home/jenkins-build/build/workspace/ceph-ansible-prs-luminous-ansible2.4-centos7_cluster/roles/ceph-osd/tasks/main.yml:39
Tuesday 17 July 2018  14:22:47 +0000 (0:00:00.144)       0:14:53.759 ********** 
fatal: [ceph-osd0]: FAILED! => {
    "changed": false, 
    "rc": 1
}

MSG:

MODULE FAILURE


MODULE_STDOUT:

  Failed to find device for physical volume "/dev/sda".
  Failed to find device for physical volume "/dev/sdb".
  Failed to find physical volume "/dev/sdc".
-->  MultiplePVsError: Got more than 1 result looking for physical volume: /dev/vda3
```


@alfredodeza This seems similar to what we saw yesterday in http://tracker.ceph.com/issues/24957, right?

alfredodeza:
If you are listing (not clear from the output) then yes, this looks similar. For debugging ceph-volume it is critical to have full logs with tracebacks. By default ceph-volume eats them up for the terminal, but writes them in full to /var/log/ceph/ceph-volume.log. They can also be enabled on the terminal with `CEPH_VOLUME_DEBUG=1`. It will otherwise be very hard to tell what/where this is coming from

leseb:

leseb:
jenkins test luminous-centos7_cluster

leseb:
jenkins test luminous-centos7_cluster

leseb:
todo: deprecate `osd_auto_discovery` as well.

PR #2426, opened 2018-02-28
Adds handy ceph aliases whe containerized installations.

jtudelag:
Same approach as openshift-ansible with etcdctl:
* https://github.com/openshift/openshift-ansible/blob/release-3.7/roles/etcd/tasks/auxiliary/drop_etcdctl.yml
* https://github.com/openshift/openshift-ansible/blob/release-3.7/roles/etcd/etcdctl.sh

Adds some ceph aliases to /etc/profile.d/ceph-aliases.sh so we can run ceph cli commands: ceph, radosgw-admin, rados and rbd directly from the host, without having to do "sudo docker exec ...".

### DIFF
--- a/roles/ceph-mon/tasks/docker/configure_ceph_command_aliases.yml
+++ b/roles/ceph-mon/tasks/docker/configure_ceph_command_aliases.yml
@@ -1,0 +1,8 @@
+---
+- name: configure ceph profile.d aliases
+  template:
+    src: ceph-aliases.sh.j2
+    dest: "/etc/profile.d/ceph-aliases.sh"
+    mode: 0755
+    owner: root
+    group: root
--- a/roles/ceph-mon/tasks/docker/main.yml
+++ b/roles/ceph-mon/tasks/docker/main.yml
@@ -6,6 +6,9 @@
 - name: include start_docker_monitor.yml
   include: start_docker_monitor.yml
 
+- name: include configure_ceph_command_aliases.yml
+  include: configure_ceph_command_aliases.yml
+
 - name: wait for monitor socket to exist
   command: "docker exec ceph-mon-{{ ansible_hostname }} sh -c 'stat /var/run/ceph/{{ cluster }}-mon.{{ ansible_hostname }}.asok || stat /var/run/ceph/{{ cluster }}-mon.{{ ansible_fqdn }}.asok'"
   register: monitor_socket
--- a/roles/ceph-mon/templates/ceph-aliases.sh.j2
+++ b/roles/ceph-mon/templates/ceph-aliases.sh.j2
@@ -1,0 +1,17 @@
+#!/bin/bash
+# Sets up handy aliases for ceph.
+ceph() {
+  sudo {{ docker_exec_cmd }} ceph --cluster {{ cluster }} ${@}
+}
+
+radosgw-admin() {
+  sudo {{ docker_exec_cmd }} radosgw-admin --cluster {{ cluster }} ${@}
+}
+
+rados() {
+  sudo {{ docker_exec_cmd }} rados --cluster {{ cluster }} ${@}
+}
+
+rbd() {
+  sudo {{ docker_exec_cmd }} rbd --cluster {{ cluster }} ${@}
+}
--- a/infrastructure-playbooks/purge-docker-cluster.yml
+++ b/infrastructure-playbooks/purge-docker-cluster.yml
@@ -469,6 +469,10 @@
     tags:
       remove_img
 
+  - name: remove ceph aliases
+    file:
+      path: /etc/profile.d/ceph-aliases.sh
+      state: absent
 
 - name: remove installed packages
 


roles/ceph-mon/templates/ceph-aliases.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
leseb: `ceph-aliases.sh.j2`?

guits:
Perhaps I'm mistaking but I'm not sure we can remove the different `{{ docker_exec_cmd }}` in the playbook because I think you can't call these functions from `command` or `shell` modules.

I assume the initial goal of this patch is not to remove the `{{ docker_exec_cmd }}` in the playbook but just to ease operations on nodes.

jtudelag:
You are right @guits, the idea is just to make everyone life easier. Im configuring this aliases in every cluster I work with, and my life is happier since then ;). Removing every instance of docker_exec_cmd was not my intention at all. Although It would mean a lot of changes.

leseb:
@guits you're right, forgot about that
@jtudelag your template is still wrong though...

guits:
@jtudelag there is a mismatch between the actual filename of the template and the call in the playbook

jtudelag:
Thanks @guits @leseb. I Know, I have to fix the template name, It's missing the .j2.

jtudelag:
@leseb @guits fixed.

roles/ceph-mon/templates/ceph-aliases.sh.j2
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# Sets up handy aliases for ceph.
+ceph() {
+  sudo {{ docker_exec_cmd }} ceph ${@}
```
leseb: you're missing the `--cluster {{ cluster }}` option

roles/ceph-mon/templates/ceph-aliases.sh.j2
```diff
@@ -0,0 +1,17 @@
...
+}
+
+radosgw-admin() {
+  sudo {{ docker_exec_cmd }} radosgw-admin ${@}
```
leseb: ditto

roles/ceph-mon/templates/ceph-aliases.sh.j2
```diff
@@ -0,0 +1,17 @@
...
+}
+
+rados() {
+  sudo {{ docker_exec_cmd }} rados ${@}
```
leseb: ditto

roles/ceph-mon/templates/ceph-aliases.sh.j2
```diff
@@ -0,0 +1,17 @@
...
+}
+
+rbd() {
+  sudo {{ docker_exec_cmd }} rbd ${@}
```
leseb: ditto

roles/ceph-mon/tasks/docker/drop_ceph_command_aliases.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: configure ceph profile.d aliases
+  template:
+    dest: "/etc/profile.d/ceph-aliases.sh"
```
leseb: please put `src` before `dest`, this feels more natural.

roles/ceph-mon/tasks/docker/main.yml
```diff
@@ -6,6 +6,9 @@
 - name: include start_docker_monitor.yml
   include: start_docker_monitor.yml
 
+- name: include drop_ceph_command_aliases.yml
```
leseb: Can we rename this simply `configure_ceph_command_aliases.yml`? The term 'drop' can be misleading.

jtudelag:
@leseb please review.

guits:

guits:

roles/ceph-mon/tasks/docker/configure_ceph_command_aliases.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: configure ceph profile.d aliases
+  template:
+    src: ceph-aliases.sh.j2
+    dest: "/etc/profile.d/ceph-aliases.sh"
+    mode: 0755
+    owner: root
+    group: root
+  become: true
```
leseb: Sorry, this slipped from my radar, I don't think this is needed since we are calling the play with `become: true` already. @guits what do you think?

jtudelag: You might be right, on the other hand I got this other template task as an example, which happens to have become: true

https://github.com/jtudelag/ceph-ansible/blob/3b0c6dc0172f660b59e79bc4c4c90b8a42b6c7f9/roles/ceph-mon/tasks/docker/start_docker_monitor.yml#L44

leseb: Hum this might need to be removed as well then, but not in this PR. Please remove it :).

guits: @leseb true, although it's not breaking things, it's not necessary to have this. I guess it's just a leftover that could be nice to clean indeed

leseb:

leseb:

leseb:
@jtudelag LGTM, let's wait for the CI now. Thanks!

PR #2428, opened 2018-03-01
Rgw fixes

leseb:


roles/ceph-rgw/handlers/main.yml
```diff
@@ -1,6 +1,7 @@
 ---
 - name: update period
-  command: radosgw-admin period update --commit
+  command: radosgw-admin --cluster {{ cluster }} period update --commit
```
guits: this command will be run on first mon node as many rgw nodes there is on the cluster, is that expected ?

roles/ceph-rgw/handlers/main.yml
```diff
@@ -1,6 +1,7 @@
...
+  delegate_to: "{{ groups[mon_group_name][0] }}"
```
guits: `run_once: true` ?

leseb: affirmatif

guits: ok maggle

Issue #2435, opened 2018-03-08
Mix ssd and sata disks on hosts (CRUSH Map)


tssvl:
Hello.
I have cluster with bunch of slow sata disks for backup.
Each server also has couple of ssd disks.
The idea is to build s3 compatible storage on slow disks and keep "rgw bucket index" on ssd pool.
So I need two crushmap roots SSD and SATA.

I am trying following:

```
ceph version: 12.2.4 luminous (stable)
ceph-ansible version: latest master
```

**/etc/host:**

```
192.168.0.10 my-ceph-s3-backup-1
192.168.0.10 my-ceph-s3-backup-1-ssd
```

**inventory.yml:**

```
osds:
  hosts:
    my-ceph-s3-backup-1:
      osd_crush_location: "{ 'root': 'HDD', 'rack': 'mon-rackkkk', 'pod': 'monpod', 'host': 'my-ceph-s3-backup-1' }"
      devices: ['/dev/sdd', '/dev/sde', '/dev/sdf', '/dev/sdg', '/dev/sdh', '/dev/sdi', '/dev/sdj', '/dev/sdk', '/dev/sdl']
      dedicated_devices: ['/dev/sda', '/dev/sda', '/dev/sda', '/dev/sda', '/dev/sda', '/dev/sda', '/dev/sda', '/dev/sda', '/dev/sda']

    my-ceph-s3-backup-1-ssd:
      osd_crush_location: "{ 'root': 'SSD', 'rack': 'mon-rackkkk', 'pod': 'monpod', 'host': 'my-ceph-s3-backup-1-ssd' }"
      devices: ['/dev/sdb']
      osd_scenario: collocated
```

**group_vars/all.yml:**

```
      osd_objectstore: filestore
      osd_scenario: non-collocated

      crush_rule_config: true

      crush_rule_hdd:
        name: HDD
        root: HDD
        type: host
        default: false

      crush_rule_ssd:
        name: SSD
        root: SSD
        type: host
        default: false

      crush_rules:
        - "{{ crush_rule_hdd }}"
        - "{{ crush_rule_ssd }}"

      create_crush_tree: true
```

**expected result:**

host my-ceph-s3-backup-1-ssd have SSD as root.

**actual result:**

```
 ceph osd tree
ID CLASS WEIGHT   TYPE NAME                                   STATUS REWEIGHT PRI-AFF
-7       49.97276 root HDD
-4       49.97276     pod monpod
[...]

PR #2432, opened 2018-03-05
mon: add support for expected-num-objects

leseb:
This commit adds the support for expected-num-objects when creating a pool.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1541520
Signed-off-by: Sébastien Han <seb@redhat.com>

PR #2440, opened 2018-03-12
[skip ci] Mons: add devices classes option to the crush rule creation task.

leseb:
use device_class: to define the device class the rule should use.
Default is to do not use it (empty string)
Added some comments on how to enable and create custom crush rules.

Co-authored-by: Greg Charot <gcharot@redhat.com>

### DIFF
--- a/roles/ceph-mon/defaults/main.yml
+++ b/roles/ceph-mon/defaults/main.yml
@@ -44,18 +44,24 @@ cephfs_pools:
 ###############
 # CRUSH RULES #
 ###############
+# Set crush_rule_config to true (default is false) if you want ceph-ansible to configure extra crush rules for you.
 crush_rule_config: false
 
+# Define your rules and enable them with the "crush_rules:" list
+# If crush_rule_config is set to true and crush_rules is not overwritten by the user then the following rules will be configured:
+
 crush_rule_hdd:
-  name: HDD
-  root: HDD
-  type: host
-  default: false
+  name: HDD  # Name of the crush rule
+  root: HDD  # Crush root the rule you take.
+  type: host  # Reaplication failure domain
+  device_class: ""  # Should this rule target a specific device class ?
+  default: false  # Should this rule be the default for new pools ? Only ONE default rule max.
 
 crush_rule_ssd:
   name: SSD
   root: SSD
   type: host
+  device_class: ""
   default: false
 
 crush_rules:
--- a/roles/ceph-mon/tasks/crush_rules.yml
+++ b/roles/ceph-mon/tasks/crush_rules.yml
@@ -12,7 +12,7 @@
     - hostvars[item]['osd_crush_location'] is defined
 
 - name: create configured crush rules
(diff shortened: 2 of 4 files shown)


leseb:
needs rebase on top of https://github.com/ceph/ceph-ansible/pull/2432
closes: https://github.com/ceph/ceph-ansible/issues/2435

scaronni:
Any chance to see this merged? We badly need it, and I have another PR which requires this.

roles/ceph-mon/defaults/main.yml
```diff
@@ -44,18 +44,24 @@ cephfs_pools:
 ###############
 # CRUSH RULES #
 ###############
-crush_rule_config: false
+# Set crush_rule_config to true (default is false) if you want ceph-ansible to configure extra crush rules for you.
+ crush_rule_config: false
```
scaronni: Please remove the extra space at the beginning of the line. Invalid syntax at playbook run.

roles/ceph-mon/tasks/crush_rules.yml
```diff
@@ -12,7 +12,7 @@
     - hostvars[item]['osd_crush_location'] is defined
 
 - name: create configured crush rules
-  command: "{{ docker_exec_cmd }} ceph --cluster {{ cluster }} osd crush rule create-simple {{ item.name }} {{ item.root }} {{ item.type }}"
+  command: "{{ docker_exec_cmd }} ceph --cluster {{ cluster }} osd crush rule create-simple {{ item.name }} {{ item.root }} {{ item.type }} {{ item.device_class | default('') }}"
```
scaronni: This should be `create-replicated`, not `create-simple`. The `create-simple` one does not support device classes.

roles/ceph-mon/defaults/main.yml
leseb: sorry linter issue on my side, fix

roles/ceph-mon/tasks/crush_rules.yml
leseb: right, I need to rework this

leseb:
@scaronni I guess we can merge this soon, no worries.

leseb:
Alright, the problem is that we cannot assign the `hdd` class since it does not exist yet. We don't have any OSDs so no device class. Investigting...

scaronni:
Any news on the topic? We have a fork with this applied to easily differentiate between drive types and would like to get back to tracking this repository.

Thanks.

leseb:
@scaronni how did you manage to get this working? As it is the problem lives in ceph itself so this can't work for the reason explained in https://github.com/ceph/ceph-ansible/pull/2440#issuecomment-381960959. I sent an email to get back the ability to create device classes on demand here: https://www.spinics.net/lists/ceph-devel/msg41152.html

alikhajeh1:
Any update on this?

leseb:
We might need to move some of the code after the OSD creation, none has been working on this. @scaronni can we get your solution for this? Thanks!

leseb:
Closing this in favour of https://github.com/ceph/ceph-ansible/pull/3186 since it will move the pool creation code to ceph-osd. Thanks.

Issue #2337, opened 2018-01-19
purge-cluster wi bluestore: deleting LVM volume doesn't clean it up


bengland2:
I was using purge-cluster.yml with a LVM ceph-volume-based config involving multiple OSDs/SSD, and switching between filestore and bluestore for performance testing purposes.    When I tried to switch back from filestore to bluestore, it complained that the OSDs already existed even though I had deleted and recreated the LVM volumes.  Apparently that doesn't actually clean out the signatures on the storage.  So I saw the note about switching to ceph-volume lvm zap when it becomes available, which it is now in Luminous 12.2.2, and did this patch and it worked after I did purge-cluster and site.yml.  Can't do a PR right now, sorry.

```
diff --git a/infrastructure-playbooks/purge-cluster.yml b/infrastructure-playbooks/purge-cluster.yml
index c04c408..ebb4874 100644
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -320,23 +320,14 @@
     failed_when: false
     register: ceph_lockbox_partition_to_erase_path
 
-  # this should go away once 'ceph-volume lvm zap' is available
-  - name: remove osd logical volumes
-    command: "lvremove -f {{ item.data_vg }}/{{ item.data }}"
-    with_items: "{{ lvm_volumes }}"
-    when:
-      - osd_scenario == "lvm"
-
-  # this should go away once 'ceph-volume lvm zap' is available
-  - name: remove osd lvm journals
-    command: "lvremove -f {{ item.journal_vg }}/{{ item.journal }}"
+  - name: zap lvm data
+    command: "ceph-volume lvm zap {{ item.data_vg }}/{{ item.data}}"
     with_items: "{{ lvm_volumes }}"
     # journals might be logical volumes, but they could also be
     # devices so fail silently if this doesn't work
     failed_when: false
     when:
       - osd_scenario == "lvm"
-      - item.journal_vg is defined
 
   - name: get ceph block partitions
     shell: |
```

leseb:
This implies we know about `lvm_volumes`. You should try to detect it instead.
Thanks for letting us know anyway :).

bengland2:
**ceph_volume lvm zap** takes care of this as long as  you run it before deleting the LVM volumes.

PR #2442, opened 2018-03-12
adds a new ceph_volume_zap module

andrewschoen:
This module is used in ``purge-cluster.yml`` to zap devices, partitions and logical volumes that were used by ``ceph-volume`` previously so that they can be reused.

fixes: #2337

### DIFF
--- a/library/ceph_volume_zap.py
+++ b/library/ceph_volume_zap.py
@@ -1,0 +1,140 @@
+#!/usr/bin/python
+import datetime
+
+ANSIBLE_METADATA = {
+    'metadata_version': '1.0',
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = '''
+---
+module: ceph_volume_zap
+
+short_description: Zap devices used by ceph OSDs with ceph-volume
+
+description:
+    - Using the ceph-volume utility available in Ceph this module
+      can be used to zap devices used by ceph OSDs.
+    - These devices, partitions or logicial volumes can then be reused.
+    - All partitions and logical volumes will be kept intact.
+
+options:
+    device:
+        description:
+            - The logical volume name, device or partition to zap.
+        required: true
+    device_vg:
+        description:
+            - If device is a lv, this must be the name of the volume group it belongs to.
+        required: false
+    destroy:
+        description:
+            - If set to True any logical volumes existing on a given device or partition will be destroyed.
+        required: false
+        default: true
+
+
+author:
+    - Andrew Schoen (@andrewschoen)
+'''
+
+EXAMPLES = '''
+- name: zap a raw device
+  ceph_volume_zap:
+    device: /dev/sdc
+    destroy: True
+
+- name: zap a partition
+  ceph_volume_zap:
+    device: /dev/sdc1
+    destroy: True
+
+- name: zap a logical volume
+  ceph_volume_zap:
+    device: lv1
+    device_vg: vg_name
+'''
+
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def get_device(device, device_vg):
+    if device_vg:
+        device = "{0}/{1}".format(device_vg, device)
+    return device
+
+
+def run_module():
+    module_args = dict(
+        device=dict(type='str', required=True),
+        device_vg=dict(type='str', required=False),
+        destroy=dict(type='bool', required=False, default=True),
+    )
+
+    module = AnsibleModule(
+        argument_spec=module_args,
+        supports_check_mode=True
+    )
+
+    device = module.params['device']
+    device_vg = module.params.get('device_vg', None)
+    destroy = module.params['destroy']
+
+    cmd = [
+        'ceph-volume',
+        'lvm',
+        'zap',
+    ]
+
+    device_to_zap = get_device(device, device_vg)
+    cmd.append(device_to_zap)
+
+    if destroy:
+        cmd.append("--destroy")
+
+    result = dict(
+        changed=False,
+        cmd=cmd,
+        stdout='',
+        stderr='',
+        rc='',
+        start='',
+        end='',
+        delta='',
+    )
+
+    if module.check_mode:
+        return result
+
+    startd = datetime.datetime.now()
+
+    rc, out, err = module.run_command(cmd, encoding=None)
+
+    endd = datetime.datetime.now()
+    delta = endd - startd
+
+    result = dict(
+        cmd=cmd,
+        stdout=out.rstrip(b"\r\n"),
+        stderr=err.rstrip(b"\r\n"),
+        rc=rc,
+        start=str(startd),
+        end=str(endd),
+        delta=str(delta),
+        changed=True,
+    )
+
+    if rc != 0:
+        module.fail_json(msg='non-zero return code', **result)
+
+    module.exit_json(**result)
+
+
+def main():
+    run_module()
+
+
+if __name__ == '__main__':
+    main()
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -328,23 +328,52 @@
     failed_when: false
     register: ceph_lockbox_partition_to_erase_path
 
-  # this should go away once 'ceph-volume lvm zap' is available
-  - name: remove osd logical volumes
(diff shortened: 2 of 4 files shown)


andrewschoen:
jenkins test luminous-ansible2.4-lvm_osds

andrewschoen:
jenkins test luminous-ansible2.4-purge_lvm_osds

andrewschoen:
jenkins test luminous-ansible2.4-purge_lvm_osds

library/ceph_volume_zap.py
```diff
@@ -0,0 +1,140 @@
...
+---
+module: ceph_volume_zap
+
+short_description: Zap devices used by ceph OSDs with ceph-volume
+
+description:
+    - Using the ceph-volume utility available in Ceph this module
+      can be used to zap devices used by ceph OSDs.
+    - These devices, partitions or logicial volumes can then be reused.
+    - All partitions and logical volumes will be kept intact.
```
alfredodeza: This is not true all the time. Logical volumes may be destroyed if `destroy: True` is used

library/ceph_volume_zap.py
```diff
@@ -0,0 +1,140 @@
...
+options:
+    device:
+        description:
+            - The logical volume name, device or partition to zap.
+        required: true
+    device_vg:
+        description:
+            - If device is a lv, this must be the name of the volume group it belongs to.
+        required: false
+    destroy:
```
leseb: Perhaps we could call this `destroy_lv` or `destroy_data`? The distinction between zap and destroy is not clear so using a zap module with a destroy option is not a bit weird. So we need to clarify that.

andrewschoen:
> Why do we need a new module for this? Why not extend the current ceph_volume module? A new module feels a bit strange to me.

I originally started by extending the current ``ceph_volume`` module but realized it was adding a ton of complication to the module with little value. The only thing these two modules are sharing is the common ansible module boiler plate code. Trying to handle the differing inputs for ``lvm zap`` and ``lvm create`` was becoming difficult. For example, zap doesn't care about ``objectstore`` at all (or a majority of the inputs into the ``ceph_volume`` module). IMHO, keeping them simple and separate makes them both easier to maintain in the long run.

library/ceph_volume_zap.py
andrewschoen: I understand this flag can be confusing, I've only named it ``destroy`` so it is the same as the ``--destroy`` flag of ``ceph-volume lvm zap``.  If ``lvm zap`` is run against an lv then ``--destroy`` is a noop. The ``-destroy`` flag is used when zapping a partition or raw device that ceph-volume has previously used and created lvs on. This was introduced because users were providing a raw device for ``--data`` and needed an easy way to remove the lvs that ceph-volume creates in that situation.

I'm not really against changing the name of the flag, but maybe we can clarify better with the documentation? It might be confusing if it's not named the same as the ``ceph-volume lvm zap`` flag too.

andrewschoen:
@leseb I'm not sure the ``service`` module is a fair comparison as it's trying to solve one problem, management of services. I was trying to follow the approach of rabbitmq where there are specific ansible modules per task, creating vhosts, creating users, creating exchanges, etc. These modules I suspect are all using ``rabbitmqctl`` under the hood, but require different input and perform different tasks. http://docs.ansible.com/ansible/latest/list_of_messaging_modules.html

With these modules I'm dealing with two features, 1) creating OSDs and 2) zapping devices. Really the only similarity there is that ``ceph-volume`` is the base command. I could give it a try with them combined but I'm curious what benefit are we gaining by doing so?

leseb:
@andrewschoen then perhaps I should have mentioned some of the existing [storage modules](https://docs.ansible.com/ansible/latest/list_of_storage_modules.html#storage-modules). If you look at their description you will see they do everything using a single module (Create, Delete or Modify NFS).

Looking at RabbitMQ, they indeed have different modules but for sub-entities and each entity has create/delete/modify. So again, this is similar to what all modules are doing.

The benefit I see is that we only have one module to maintain and not a different module for each subcommand.

alfredodeza:
@leseb the "everything in one module" idea doesn't sound like it is a requirement for modules. You mention the storage modules, but does this mean we need to find ourselves (regardless on how this translates to code) matching the "one module per command" ? The link also shows a bunch of storage modules that are separate even for sub-commands, for example:

* netapp_e_volume
* netapp_e_volume_copy

The rabbitmq that Andrew linked is another example, there are others like the Git submodule and git_config, which are separate.

 To your point of maintainability, I would encourage you to take another look at the logic in the ceph_volume module and on this module, and how would those would work if they were in a single module. The reason to avoid this would be added complexity that is unnecessary (flags/options just don't map). Adding that complexity just to have everything in one module *will* cause a maintainability burden since a tremendous amount of logic will have to be added to deal with the duality.

andrewschoen:
> Looking at RabbitMQ, they indeed have different modules but for sub-entities and each entity has create/delete/modify. So again, this is similar to what all modules are doing.

We're not creating and deleting an OSD here though. If this zap module had anything to do with OSDs or with ceph I might agree, but zapping a device with ``ceph-volume lvm zap`` runs no ceph commands and could be used on a device that was never used by ceph. 

If we wanted to add a ``state`` argument to the ``ceph_volume`` module that could ensure that an OSD is present or absent then perhaps we could use ``ceph-volume lvm zap`` as part of that process, but it would also mean running other various ceph cli commands to remove the OSD from the cluster. That addition wouldn't get us anymore functionality than what we already have with these two modules and ``purge-cluster.yml`` though.

> The benefit I see is that we only have one module to maintain and not a different module for each subcommand.

I contend that two simple modules is easier to maintain then one larger module that does everything. Adding ``state`` to ``ceph_volume`` might be a decent approach but I'm still unsure if the value added there outweighs the additional complexity it'll add to the ``ceph_volume`` module to support it.

guits:
My first thought seeing this thread is that if the design of that module is really a concern perhaps it should have been discussed earlier instead of in a github PR.

@alfredodeza I'm not sure the example you mentioned is the best, one module manages volumes where the second manages snapshot images, both modules aim at creating AND/OR deleting.

I can hear that having two separated modules could be easier from a maintainability point of view but I tend to agree with @leseb, I think it should be a single module, IMHO it sounds more logic for usability.

alfredodeza:
I think we are all in agreement that a single module would look best and have a cleaner user-facing experience.

But from the technical side of things, it gets complicated by the fact that ansible modules are a single file, there is no leverage for sharing like in normal Python, and will make things cumbersome to maintain in the long run.

Having said that, if @leseb and @guits both understand the caveats of going down that path and are absolutely against two modules, then I don't know what else we can do here other than scrap the current implementation and do it all in one module.

andrewschoen:
> My first thought seeing this thread is that if the design of that module is really a concern perhaps it should have been discussed earlier instead of in a github PR.

I honestly didn't believe it would cause so much concern, but I certainly don't mind discussing these things in a PR. It's the whole point of code reviews, IMHO.

>@alfredodeza I'm not sure the example you mentioned is the best, one module manages volumes where the second manages snapshot images, both modules aim at creating AND/OR deleting.

``ceph-volume`` has no support for deleting or removing OSDs. Do you envision the removal of the OSD to happen within this module as well before zap is completed? How do you see the interface working here? Would the ``ceph_volume`` module take all the information about the OSD, ``ceph osd destroy`` the OSD and then zap all partitions, raw devices and lvs that might have been used in it's creation?

> I can hear that having two separated modules could be easier from a maintainability point of view but I tend to agree with @leseb, I think it should be a single module, IMHO it sounds more logic for usability.

@leseb @guits would you please elaborate on how you envision this would work in a combined module? Do we want to go with a ``state`` argument that can either be ``present`` or ``absent``? I tend to think my original addition of ``subcommand`` to the ``ceph_volume`` module was a mistake and trying to switch functionality by changing a string in ``subcommand`` wouldn't be terribly user friendly so I'd like to remove that and avoid changing ``subcommand`` to something like ``zap`` to get this to work.

guits:
@andrewschoen should we close this PR in favor of #2446 ?

andrewschoen:
@guits I wanted to leave it open until you, @leseb and @alfredodeza had a chance to review #2446 

If you think that #2446 is definitely the way to go then feel free to close.

andrewschoen:
closing in favor of #2446

PR #2444, opened 2018-03-13
Cleanup plugins directories and references

andymcc:
Having callback_plugins, and action plugins in random locations causes
a lot of disparity.

We should centralize this into one place in the plugins directory and
fix up the ansible.cfg to reflect this.

Additionally, since the ansible.cfg already reflects action_plugins, we
don't need a link to action_plugins in the base of the repository.


andymcc:
It would be nice to do that to library too and keep all plugins in the library directory - but didn't include that here because its been around ages and would require a change from anybody using this as a role where they reference the ceph-ansible library directory using env_vars or ansible.cfg

andymcc:
(It would also be nice to rename actions --> action, but same reason)

PR #2445, opened 2018-03-13
[WIP] Cleanup fsid generation

andymcc:
The current approach makes fsid generation complicated by utilising a
file and fetch directory. If you delegate a set_fact to localhost it
will set the fact for all other hosts.

In this way we can set the fsid based on the running cluster (if it's
running), or generate it as before - but only perform this once and all
hosts can consume that setting.

Additionally - to allow this to run as part of the rolling updates, we
can run the task to check if ceph is running against either mons[0] or
mons[1] if we are currently using mons[0].

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -28,7 +28,7 @@ dummy:
 #  luminous: 12
 #  mimic: 13
 
-# Directory to fetch cluster fsid, keys etc...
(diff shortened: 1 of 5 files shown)


leseb:
@andymcc any plan to resurrect this at some point? :)

andymcc:
I'll close this one out. Sadly haven't had a chance to look at it again. I'll reopen a new one if that happens!

PR #2446, opened 2018-03-14
ceph_volume: adds zap support

andrewschoen:
Adds zap support to the existing ``ceph_volume`` module, an alternative approach to #2442

Fixes: #2337

### DIFF
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -362,6 +362,7 @@ def run_module():
         wal_vg=dict(type='str', required=False),
         crush_device_class=dict(type='str', required=False),
         dmcrypt=dict(type='bool', required=False, default=False),
+        zap=dict(type='bool', required=False, default=False),
     )
 
     module = AnsibleModule(
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -351,7 +351,7 @@ def run_module():
     module_args = dict(
         cluster=dict(type='str', required=False, default='ceph'),
         objectstore=dict(type='str', required=False, choices=['bluestore', 'filestore']),
-        state=dict(type='str', required=True, choices=['present', 'absent'], default='present'),
+        state=dict(type='str', required=False, choices=['present', 'absent'], default='present'),
         data=dict(type='str', required=True),
         data_vg=dict(type='str', required=False),
         journal=dict(type='str', required=False),
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -329,8 +329,8 @@ def remove_osd(module):
 
             cmd_result = dict(
                 cmd=cmd,
-                stdout=out.rstrip(b"\r\n"),
-                stderr=err.rstrip(b"\r\n"),
+                stdout_lines=out.split("\n"),
+                stderr_lines=err.split("\n"),
                 rc=rc,
                 start=str(startd),
                 end=str(endd),
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -311,12 +311,12 @@ def remove_osd(module):
         result = dict(
             changed=False,
             stdout='There was no OSD to destroy and zap was set to False.',
-            rc=1,
+            rc=0,
         )
     else:
         result = dict(
             changed=True,
-            rc=1,
+            rc=0,
         )
         command_results = []
         for cmd in commands:
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -28,7 +28,8 @@ options:
     objectstore:
         description:
             - The objectstore of the OSD, either filestore or bluestore
-        required: true
+            - Required if state is 'present'
+        required: false
         choices: ['bluestore', 'filestore']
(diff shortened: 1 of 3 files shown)


andrewschoen:
jenkins test luminous-ansible2.4-lvm_osds

library/ceph_volume.py
```diff
@@ -250,6 +240,146 @@ def run_module():
...
+
+    commands = []
+
+    data = get_data(data, data_vg)
+
+    # check to see if osd already exists
+    # FIXME: this does not work when data is a raw device
+    # support for 'lvm list' and raw devices was added with https://github.com/ceph/ceph/pull/20620 but
+    # has not made it to a luminous release as of 12.2.4
+    rc, out, err = module.run_command(["ceph-volume", "lvm", "list", "--format=json", data], encoding=None)
```
alfredodeza: this is mixing two things that might not work well in all situations:

* `lvm list` can list OSDs that are present in the node, but are absent from the cluster
* `osd destroy` will tell the cluster that the OSD ID has to be removed (keeping the ID around for re-use) but will fail if the ID is no longer active in the cluster

In this case it would need to do something similar to what ceph-volume does, which is check the cluster:

    bootstrap_keyring = '/var/lib/ceph/bootstrap-osd/%s.keyring' % conf.cluster
    stdout, stderr, returncode = process.call(
        [
            'ceph',
            '--cluster', conf.cluster,
            '--name', 'client.bootstrap-osd',
            '--keyring', bootstrap_keyring,
            'osd',
            'tree',
            '-f', 'json',
        ],
        show_command=True
    )
    if returncode != 0:
        raise RuntimeError('Unable check if OSD id exists: %s' % osd_id)

    output = json.loads(''.join(stdout).strip())
    osds = output['nodes']
    return any([str(osd['id']) == str(osd_id) for osd in osds])

Since `state=absent` is trying make sure the OSD ID is removed from both the node and the cluster, either error checking the return of `destroy_cmd` or checking with the cluster first would be required.

andrewschoen:
jenkins test luminous-ansible2.4-purge_lvm_osds

andrewschoen:

andrewschoen:

andrewschoen:
jenkins test luminous-ansible2.4-purge_lvm_osds

andrewschoen:
jenkins test luminous-ansible2.4-lvm_osds

andrewschoen:
jenkins test luminous-ansible2.4-bluestore_lvm_osds

library/ceph_volume.py
```diff
@@ -250,6 +235,114 @@ def run_module():
...
+
+    module.exit_json(**result)
+
+
+def run_module():
+    module_args = dict(
+        cluster=dict(type='str', required=False, default='ceph'),
+        objectstore=dict(type='str', required=False, choices=['bluestore', 'filestore']),
+        action=dict(type='str', required=False, choices=['create', 'zap'], default='create'),
+        data=dict(type='str', required=True),
```
leseb: You need to add `default=` so you don't need to run  `module.params.get('data_vg', None)` and apply a default when trying the get the variable.

library/ceph_volume.py
```diff
@@ -250,6 +235,114 @@ def run_module():
...
+
+    result["commands"] = command_results
+
+    module.exit_json(**result)
+
+
+def run_module():
+    module_args = dict(
+        cluster=dict(type='str', required=False, default='ceph'),
+        objectstore=dict(type='str', required=False, choices=['bluestore', 'filestore']),
```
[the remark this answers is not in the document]
andrewschoen: default in ``ceph-volume`` is ``bluestore``, I suppose I could add that here as well. It's a required param when ``action`` is ``create``, so I'm not so worried about it though. This used to be ``required=True`` before this refactor and you can not specify a default on required params.

library/ceph_volume.py
andrewschoen: There would be no way to set a default for ``data_vg`` that made any sense. I'm unsure if Ansible will accept a default of ``None`` for a ``type='str'``, but I guess I could try.

Something like ``module.params.get('data_vg', None)`` is a very common pattern in python, are you not liking the extra bit of verbosity there?

leseb: Yes I know it's common, I have the impression that `module.params.get('data_vg', None)` achieves the same. That's what I observed but I might be wrong.

If that happens not being the case, then I'm okay with your code.

library/ceph_volume.py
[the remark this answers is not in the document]
andrewschoen: I've made ``objectstore`` now default to ``bluestore``.

library/ceph_volume.py
andrewschoen: Yeah, we don't gain anything by making that change. The ``, None)`` bit there is actually not needed but I like to be verbose. A ``.get()`` will return ``None`` if it doesn't find the key you're looking for.

andrewschoen:

PR #2454, opened 2018-03-20
Fix config_template to consistently order sections

andymcc:
In ec042219e64a321fa67fce0384af76eeb238c645 we added OrderedDict and
sorted to be able to preserve order for config_template k,v pairs inside
a section.

This patch adds a similar ordering for the sections themselves, which
could still change order and intiiate handler restarts.

OrderedDict isn't needed because we use .items() to return a list that
can then be sorted().

(cherry picked from commit fe4ba9d1353abb49775d5541060a55919978f45f)


leseb:
backport of https://github.com/ceph/ceph-ansible/pull/2450

PR #2463, opened 2018-03-22
Fix install of nfs-ganesha-ceph for Debian/SuSE

andymcc:
The Debian and SuSE installs for nfs-ganesha on the non-rhcs repository
requires you to allow_unauthenticated for Debian, and disable_gpg_check
for SuSE. The nfs-ganesha-rgw package already does this, but the
nfs-ganesha-ceph package will fail to install because of this same
issue.

This PR moves the installations to happen when the appropriate flags are
set to True (nfs_obj_gw & nfs_file_gw), but does it per distro (one for
SuSE and one for Debian) so that the appropriate flag can be passed to
ignore the GPG check.

### DIFF
--- a/roles/ceph-nfs/tasks/pre_requisite_non_container.yml
+++ b/roles/ceph-nfs/tasks/pre_requisite_non_container.yml
@@ -99,41 +99,33 @@
     - ceph_repository != 'rhcs'
     - ansible_os_family == 'Debian'
 
-# debian installation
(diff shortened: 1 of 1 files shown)


guits:

andymcc:

guits:
luminous-ansible2.4-purge_bluestore_osds_container will fail anyway because of the race condition

Issue #2457, opened 2018-03-21
vars validation


alfredodeza:
The current amount of variables in ceph-ansible, in addition to their relationships with one another makes it very complicated to try enforcing constraints with YAML as it is currently done in `check_mandatory_vars.yml`

The mandatory vars file does a good job at identifying a few, but it doesn't ever validate the tremendous amount of other options there are.

For ceph-installer [0] we used a validation engine called Notario [1] that allows to describe schemas for Python dictionaries. Since Ansible exposes these vars as a Python dictionary as well, Notario would be a good fit.

There are a couple of ways that Ansible grants access to the vars. For callback plugins it is a bit obscure:

    self.play._variable_manager._hostvars

But for action plugins it is straightforward if using `**kw` in the run method.

Using an Action plugin will allow ceph-ansible to define a lax or strict mode where variables can generate warnings, or halt execution of the whole play.

To fully take advantage of this approach, the action plugin would need to be created with Notario integration, and the schemas should be put in place. Maybe a good initial approach will be to have the action plugin with a few schemas to demonstrate the execution

[0] http://notario.cafepais.com/
[1] https://github.com/ceph/ceph-installer/blob/master/ceph_installer/schemas.py

andrewschoen:
I like this idea. Should we start by picking an install option and/or osd scenario and writing schemas for those? We could pick a testing scenario and write the schemas needed for that test to validate, maybe https://github.com/ceph/ceph-ansible/tree/master/tests/functional/centos/7/lvm-osds or https://github.com/ceph/ceph-ansible/tree/master/tests/functional/centos/7/cluster

guits:
since ceph-validate has been merged, we can close this issue.

PR #2468, opened 2018-03-23
[skip ci] DNM Host vars validation

alfredodeza:
This (preview) pr is to demonstrate how validation could work in a way that addresses issue #2457 

It provides an "action plugin" that when called in a task, will validate against the host vars available.

It only has 3 items in the schema, since that is enough to demonstrate the use case. Here is some output on a few potential use cases.

Permissive mode, with `vvvv` flag and `ANSIBLE_STDOUT_CALLBACK=debug`, shows lots info, with the key that failed
and for what host (in this case, `local`)

    TASK [validate vars mode=permissive] *********************************************************************************
    task path: /Users/alfredo/tmp/validate/local.yml:8
    Notario Failure: -> ceph_conf_key_directory -> True did not pass validation against callable: string (not of type string)
     [WARNING]: [local] Validation failed for variable: ceph_conf_key_directory
    
    ok: [local] => {
        "changed": false,
        "failed": false
    }
    
    MSG:
    
    Invalid variable assignment in host: local
        ceph_conf_key_directory = True
                                  ^^^^
    Reason: not of type string
    
    META: ran handlers
    META: ran handlers

Strict mode makes the module fail:

    TASK [validate vars mode=strict] *************************************************************************************
    task path: /Users/alfredo/tmp/validate/local.yml:8
    Notario Failure: -> ceph_conf_key_directory -> True did not pass validation against callable: string (not of type string)
     [WARNING]: [local] Validation failed for variable: ceph_conf_key_directory
    
    fatal: [local]: FAILED! => {
        "changed": false,
        "failed": true
    }
    
    MSG:
    
    Invalid variable assignment in host: local
        ceph_conf_key_directory = True
                                  ^^^^
    Reason: not of type string
    
    

Without the `debug` mode for stdout, the formatting is kept by using `stderr_lines`:

    TASK [validate vars mode=strict] *************************************************************************************
    task path: /Users/alfredo/tmp/validate/local.yml:8
    Notario Failure: -> ceph_conf_key_directory -> True did not pass validation against callable: string (not of type string)
     [WARNING]: [local] Validation failed for variable: ceph_conf_key_directory
    
    fatal: [local]: FAILED! => {
        "changed": false,
        "failed": true,
        "msg": "Invalid variable assignment in host: local\n    ceph_conf_key_directory = True\n                              ^^^^\nReason: not of type string",
        "stderr_lines": [
            "Invalid variable assignment in host: local",
            "    ceph_conf_key_directory = True",
            "                              ^^^^",
            "Reason: not of type string"
        ]
    }

### DIFF
--- a/ceph-ansible.spec.in
+++ b/ceph-ansible.spec.in
@@ -21,6 +21,7 @@ BuildRequires: python2-devel
 
 Requires: ansible >= 2.4.2.0
 Requires: python-netaddr
+Requires: python-notario
 
 %description
 Ansible playbooks for Ceph
--- a/library/validate
+++ b/library/validate
@@ -1,0 +1,4 @@
+# This is a placeholder file so that the ``validate`` action plugin can work.
+# An action plugin is prefered because that gets executed on the master. If at
+# some point the validation engine needs to run on remote nodes, then this
+# placeholder can implement the ``module`` part of the plugin.
--- a/plugins/actions/validate.py
+++ b/plugins/actions/validate.py
@@ -1,12 +1,9 @@
-import sys
-import os
 
 from ansible.plugins.action import ActionBase
 
 import notario
 from notario.exceptions import Invalid
-from notario.decorators import optional
-from notario.validators import types, recursive
+from notario.validators import types, chainable
 
 try:
     from __main__ import display
@@ -18,8 +15,10 @@ except ImportError:
 class ActionModule(ActionBase):
 
     def run(self, tmp=None, task_vars=None):
-        # we must use hostvars, since task_vars will have un-processed variables
-        hostvars = task_vars['hostvars']
+        # we must use vars, since task_vars will have un-processed variables
+        host_vars = task_vars['vars']
+        display.warning(host_vars)
+        host = host_vars['ansible_hostname']
         mode = self._task.args.get('mode', 'permissive')
 
         self._supports_check_mode = False # XXX ?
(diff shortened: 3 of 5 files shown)


alfredodeza:
A valid error from the tests is that there is no module `notario`. I am unsure how to add it as a dependency for the project as a whole for Python. As an Ansible project, there aren't any Python packaging files, like setup.py, or the pip-friendly: requirements.txt

Error from CI is:

    ERROR! Unexpected Exception, this is probably a bug: No module named notario

andrewschoen:
I really like this approach. I suspect in practice we're going to need to validate multiple schemas per playbook run. For example, ``lvm_volumes`` is required for ``osd_scenario: lvm`` but not for any other osd scenarios. How do you see this type of validation working?

alfredodeza:
The validation engine will most commonly address a 1:1 key to value assignment, even if the `value` here means a lot of other things (like an array of multiple values).

The use case you present would require custom validators that would need some state to be preserved and later checked on. For example, as the engine traverses the data structure it could have a validator that will "store" the `osd_scenario` so that when `lvm_volumes` is validated, the custom validator can see if it needs to blow up because `collocated` was used instead of `lvm`.

notario has a thread-safe storage that can be used for this purpose, it might even be desirable to implement a feature in notario where the whole incoming data can be stored so that checks can poke around it directly *aside* from having the special store for validators.

Each of these approaches have caveats: 

* capturing the whole data means that validators would need to know the exact path (that should not change) to the value they need to poke at. This is not going to be possible always (although it would work in this case)
* storing flags/values when reaching custom validators in the store means that these need to be called before the other validators that are required to read those values. If the order is done in an unexpected way, then the store is not going to be useful.

The `store` capabilities in notario are (unfortunately) undocumented, but it is very simple to play with: https://github.com/alfredodeza/notario/blob/master/notario/store.py

guits:
@alfredodeza I don't see any notario package in RHEL repositories, that's why we were wondering with sebastien how it will be shipped in ceph-ansible from downstream point of view.
Are we simply missing the right package name ?
If I'm not mistaking, I'm not sure we are going to use the fedora repository you mentioned ?

alfredodeza:
@guits @leseb what is the concern here? That there are no packages for notario? ceph-installer *has been* part of RHEL, there are packages!

In the case that there weren't any, I am confident we could package it.

Is this the only concern for moving forward with this?

leseb:
@alfredodeza yes that's the concern until we are not 100% sure we have the package we can't merge this (even though it's a wip). So in // let's check with @ktdreyer how we can get a package for notario downstream so it becomes a requirement of ceph-ansible's rpm.

andrewschoen:
I'm working on this PR and will reopen when it's ready to review again.

PR #2471, opened 2018-03-25
Refer to expected-num-ojects as expected_num_objects, not size

fultonj:
Follow up patch to PR 2432 [1] which replaces "size" (sorry if
the original bug used that term, which can be confusing) with
expected_num_objects as is used in the Ceph documentation [2].

[1] https://github.com/ceph/ceph-ansible/pull/2432/files
[2] http://docs.ceph.com/docs/jewel/rados/operations/pools

### DIFF
--- a/group_vars/clients.yml.sample
+++ b/group_vars/clients.yml.sample
@@ -21,7 +21,7 @@ dummy:
 #  rule_name: "replicated_rule"
 #  type: "replicated"
 #  erasure_profile: ""
-#  size: ""
+#  expected_num_objects: ""
 #test2:
 #  name: "test2"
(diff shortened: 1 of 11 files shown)


ktdreyer:
John what is the impact to OSP 12 and 13 if this gets backported to stable-3.0?

leseb:
@fultonj ^

ktdreyer:
I just want to be sure OSP 12 is not relying on "size" and it's ok that there is no backwards compatibility

leseb:
Unfortunately, this is not backportable as this relies on https://github.com/ceph/ceph-ansible/pull/2432 which is targetted for 3.1.

Issue #2474, opened 2018-03-28
ImportError: cannot import name to_bytes


Futvin:
I ran into this error while running ansible 2.5.0 on task *TASK [ceph-config : create ceph conf directory]*

I "fixed" this by updating ./ceph-ansible/plugins/actions/_v2_config_template.py

In the imports I commented out the offending line, and replaced it with two lines that work for ansible 2.5

#from ansible.utils.unicode import to_bytes, to_unicode
from ansible.module_utils._text import to_text as to_unicode
from ansible.module_utils._text import to_bytes

Asher256:
Same here.
```
Traceback (most recent call last):
  File "/var/lib/ceph-ansible/.pyenv_ansible/bin/ansible-playbook", line 118, in <module>
    exit_code = cli.run()
  File "/var/lib/ceph-ansible/.pyenv_ansible/local/lib/python2.7/site-packages/ansible/cli/playbook.py", line 122, in run
    results = pbex.run()
  File "/var/lib/ceph-ansible/.pyenv_ansible/local/lib/python2.7/site-packages/ansible/executor/playbook_executor.py", line 159, in run
    result = self._tqm.run(play=play)
  File "/var/lib/ceph-ansible/.pyenv_ansible/local/lib/python2.7/site-packages/ansible/executor/task_queue_manager.py", line 290, in run
    play_return = strategy.run(iterator, play_context)
  File "/var/lib/ceph-ansible/.pyenv_ansible/local/lib/python2.7/site-packages/ansible/plugins/strategy/linear.py", line 217, in run
    action = action_loader.get(task.action, class_only=True)
  File "/var/lib/ceph-ansible/.pyenv_ansible/local/lib/python2.7/site-packages/ansible/plugins/loader.py", line 370, in get
    self._module_cache[path] = self._load_module_source(name, path)
  File "/var/lib/ceph-ansible/.pyenv_ansible/local/lib/python2.7/site-packages/ansible/plugins/loader.py", line 346, in _load_module_source
    module = imp.load_source(full_name, path, module_file)
  File "/var/lib/jenkins/workspace/deploy-ceph/plugins/actions/config_template.py", line 29, in <module>
    from _v2_config_template import *
  File "/var/lib/jenkins/workspace/deploy-ceph/plugins/actions/_v2_config_template.py", line 37, in <module>
    from ansible.utils.unicode import to_bytes, to_unicode
ImportError: cannot import name to_bytes
```

[...]

PR #2475, opened 2018-03-28
Replace ansible.utils.unicode to ansible.module_utils._text

Recluse:
This feature was removed in Ansible version 2.4+ (and don't working)
ansible.utils.unicode. is deprecated, use ansible.module_utils._text instead

Fixes: #2474

### DIFF
--- a/plugins/actions/_v2_config_template.py
+++ b/plugins/actions/_v2_config_template.py
@@ -34,7 +34,8 @@ import yaml
 
 
 from ansible.plugins.action import ActionBase
-from ansible.utils.unicode import to_bytes, to_unicode
+#from ansible.utils.unicode import to_unicode # ansible.utils.unicode. is deprecated, use ansible.module_utils._text instead
+from ansible.module_utils._text import to_bytes, to_text as to_unicode
 from ansible import constants as C
 from ansible import errors
 


guits:
I noticed this while I was taking a look at the upstream version and was wondering whether we should resync the whole file since it seems there is some other differences, [see here](https://github.com/openstack/openstack-ansible-plugins/blob/master/action/config_template.py)

@leseb ?

guits:
@Recluse can you add "Fixes: #2474" in your commit message ?

leseb:
@guits I think we should always resync with what's in gerrit. Also, the best thing to do IMO is ti submit this in OpenStack ansible. Then do the resync.
@guits how does that sound?

@Recluse do you mind submitting this on https://review.openstack.org/#/q/project:openstack/openstack-ansible-plugins

fyi @andymcc

plugins/actions/_v2_config_template.py
```diff
@@ -34,7 +34,8 @@
 
 
 from ansible.plugins.action import ActionBase
-from ansible.utils.unicode import to_bytes, to_unicode
+#from ansible.utils.unicode import to_unicode # ansible.utils.unicode. is deprecated, use ansible.module_utils._text instead
```
leseb: I don't think this is necessary

guits:
@leseb these modifications are already present in openstack-ansible-plugins project. That's why I've suggested to simply resync with what there is upstream.
But in the end, yes, the process should be what you described.

plugins/actions/_v2_config_template.py
guits: should disappear since we are going to resync with upstream

leseb:
@guits oh I see! let's just resync then, @Recluse do you want to care of that? Thanks!

plugins/actions/_v2_config_template.py
Recluse: Oh, that ok! I was needed to deploy the cluster urgently, and I showed the solution to the problem, thank you.

leseb:
resynced here: https://github.com/ceph/ceph-ansible/pull/2476

PR #2478, opened 2018-03-28
purge-docker: added conditionals needed to successfully re-run purge-docker-cluster.yml.

utp887:
Added 'ignore_errors: true' to multiple tasks which run docker commands; even in cases where docker is no longer installed. Because of this, certain tasks in the purge-docker-cluster.yml will cause the playbook to fail if re-run and stop the purge. This leaves behind a dirty environment, and a playbook which can no longer be run.
Fix Regex line 275: Sometimes 'list-units' will output 4 spaces between loaded+active. The update will account for both scenarios.
purge fetch_directory: in other roles fetch_directory is hard linked ex.: "{{ fetch_directory }}"/"{{ somedir }}". That being said, fetch_directory will never have a trailing slash in the all.yml so this task was never being run(causing failures when trying to re-deploy).

Signed-off-by: Randy J. Martinez <ramartin@redhat.com>

### DIFF
--- a/infrastructure-playbooks/purge-docker-cluster.yml
+++ b/infrastructure-playbooks/purge-docker-cluster.yml
@@ -48,7 +48,6 @@
       ceph_docker_registry: "docker.io"
     when: ceph_docker_registry is not defined
 
-
 - name: purge ceph mds cluster
 
   hosts:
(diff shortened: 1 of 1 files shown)


utp887:
@leseb thoughts?

leseb:
@utp887 sorry the wait :), thanks

PR #2481, opened 2018-03-29
ceph-defaults: fix ceph_uid fact on container deployments

utp887:
Red Hat is now using tags[3,latest] for image rhceph/rhceph-3-rhel7. Because of this, the ceph_uid conditional passes for Debian when 'ceph_docker_image_tag: latest' on RH family deployments. Updating the conditional to ansible_os_family solves the problem.

Signed-off-by: Randy J. Martinez <ramartin@redhat.com>

### DIFF
--- a/roles/ceph-defaults/tasks/facts.yml
+++ b/roles/ceph-defaults/tasks/facts.yml
@@ -191,7 +191,14 @@
     ceph_uid: 167
   when:
     - containerized_deployment
-    - ceph_docker_image_tag | search("latest") or ceph_docker_image_tag | search("centos") or ceph_docker_image | search("rhceph")
+    - ceph_docker_image_tag | search("latest") or ceph_docker_image_tag | search("centos") or ceph_docker_image_tag | search("fedora")
+
+- name: set_fact ceph_uid for Red Hat
+  set_fact:
+    ceph_uid: 167
+  when:
+    - containerized_deployment
+    - ceph_docker_image | search("rhceph")
 
 - name: check if selinux is enabled
   command: getenforce
@@ -200,5 +207,3 @@
   check_mode: no
   when:
     - ansible_os_family == 'RedHat'
-
-


guits:
@leseb I think we have talked about that few weeks ago we agreed that we shouldn't do that kind of introspection if I remember well?

guits:
Actually I think this is misleading, this will check the OS on the host while we want to know the OS inside the container.

@leseb ?

leseb:
We have to do the introspection on the tag because we don't have the image yet. That's not the right approach, as @guits mentioned testing the OS you're running on won't work.

@utp887, I think you need to add a new task that has `ceph_docker_image | search("rhceph")` in its conditionnal so you will get `167` for `ceph_uid`.

utp887:
@leseb @guits I see your points... I went ahead and checked what centos/fedora/ubuntu are tagging, and the original check wouldn't have matched, so lets work with the image names instead. See update.

roles/ceph-defaults/tasks/facts.yml
```diff
@@ -184,11 +184,11 @@
     ceph_uid: 64045
   when:
     - containerized_deployment
-    - ceph_docker_image_tag | search("latest") or ceph_docker_image_tag | search("ubuntu")
+    - ceph_docker_image | search("ubuntu")
```
leseb: `ceph_docker_image` is either `ceph/daemon` or `rhceph` downstream so this won't work. My suggestion was to add an extra task that will do:

```
 - name: set_fact ceph_uid for Red Hat based system
   set_fact:
     ceph_uid: 167
   when:
     - containerized_deployment
     - ceph_docker_image | search("rhceph")
```

utp887: Hey @leseb, after talking with some other folks about this; how would you feel about having `ceph_uid: 167` configured in the `all.yml` under Docker section? Then we could just keep only the debian task with the following conditionals: 
```
when:
  - containerized_deployment
  - ceph_uid is not defined
```
So basically, if you're running a RH family deployment ceph_uid will need to be set. Otherwise ceph_uid would get configured w/64045 if not defined. 

I think this is the best approach since we really have no way of controlling tags in other registries outside of docker.io/redhat.... which is what I think you were getting at.

Also, the task on 189 doesn't overwrite ceph_uid because ansible doesn't like the conditional on line 194 for some reason.... The or conditional I submitted uses the same var throughout and works fine.

leseb: @utp887 we now use centos by default in our container image upstream so I'm sure if we should stick with 64045.
We don't really want to set the uid in group_vars because this will mean an extra config option. We can easily detect that.
If line 189 is not behaving properly then we need to fix that instead of changing the whole behavior.

If you want to override `ceph_uid` you can do so by running ansible with `-e ceph_uid=167` if your registry is not docker or redhat.

Do you mind running the test again with the line I suggested?
Thanks.

utp887:
@leseb This is working as expected for me now w/your suggested updates.

leseb:

utp887:
@leseb it looks like the test failed because ceph/daemon & tag: latest was used.... Recommendations?

utp887:

leseb:

PR #2488, opened 2018-04-02
nfs: ensure nfs-server server is stopped

alimaredia:
NFS-ganesha cannot start is the nfs-server service
is running. This commit stops nfs-server in case it
is running on a node before the nfs-ganesha service
starts up

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1508506

### DIFF
--- a/roles/ceph-nfs/tasks/pre_requisite_non_container.yml
+++ b/roles/ceph-nfs/tasks/pre_requisite_non_container.yml
@@ -223,3 +223,10 @@
     owner: "root"
     group: "root"
     mode: "0755"
+
+- name: stop nfs server service
+  systemd:
+    name: "{%- if ansible_os_family == 'RedHat' 'nfs-server' %}{%- endif %}{%- if ansible_os_family == 'Suse' 'nfsserver' %}{%- endif %}{%- if ansible_os_family == 'Debian' 'nfs-kernel-server' %}{%- endif %}"
+    state: stopped
+    enabled: no
+  ignore_errors: true


roles/ceph-nfs/tasks/pre_requisite_non_container.yml
```diff
@@ -223,3 +223,10 @@
     owner: "root"
     group: "root"
     mode: "0755"
+
+- name: stop nfs server service
+  systemd:
+    name: nfs-server
```
guits: the unit name on ubuntu is `nfs-kernel-server`

guits: @alimaredia sorry, I think i misled you. `nfs-server` seems to be ok on Debian based os (actually, it looks like both are ok).
The issue here is that if you try to stop a service which doesn't exist with module `systemd` its going to fail.

alimaredia:
jenkins retest this please

roles/ceph-nfs/tasks/pre_requisite_non_container.yml
```diff
@@ -223,3 +223,42 @@
     owner: "root"
     group: "root"
     mode: "0755"
+
+- name: check if nfs-utils is installed on Red Hat
```
leseb: do no use uppercase

roles/ceph-nfs/tasks/pre_requisite_non_container.yml
```diff
@@ -223,3 +223,42 @@
...
+  shell: rpm -q nfs-utils | grep 'package nfs-utils is not installed'
+  register: rpm_nfs_utils_check
+  when:
+    - ansible_os_family == 'RedHat'
+
+- name: check if nfs-kernel-server is installed on Debian
```
leseb: ditto

roles/ceph-nfs/tasks/pre_requisite_non_container.yml
```diff
@@ -223,3 +223,42 @@
...
+  shell: dpkg-query -s nfs-kernel-server | grep 'install ok installed'
+  register: dpkg_nfs_kernel_server_check
+  when:
+    - ansible_os_family == 'Debian'
+
+- name: stop nfs server service on Red Hat
```
leseb: ditto

roles/ceph-nfs/tasks/pre_requisite_non_container.yml
```diff
@@ -223,3 +223,42 @@
...
+  systemd:
+    name: nfs-server
+    state: stopped
+    enabled: no
+  when:
+    - ansible_os_family == 'RedHat'
+    - rpm_nfs_utils_check.rc == 1
+
+- name: stop nfs server service on Suse
```
leseb: ditto

roles/ceph-nfs/tasks/pre_requisite_non_container.yml
```diff
@@ -223,3 +223,42 @@
...
+  systemd:
+    name: nfsserver
+    state: stopped
+    enabled: no
+  when:
+    - ansible_os_family == 'Suse'
+      
+- name: stop nfs nfs server service on Debian
```
leseb: ditto

guits:
@alimaredia In that specific case, I'm not sure it's worth to check for the status of the package.
I think you could do everything in one task using `ignore_errors: true`
eg:
```
- name: stop nfs server service
  systemd:
    name: "{%- if ansible_os_family == 'RedHat' 'nfs-server' %}{%- endif %}{%- if ansible_os_family == 'Suse' 'nfsserver' %}{%- endif %}{%- if ansible_os_family == 'Debian' 'nfs-kernel-server' %}{%- endif %}"
    state: stopped
    enabled: no
  ignore_errors: true
```

alimaredia:
@guits I think this is ready to go, can you have another look?

alimaredia:
@leseb should be fine now

PR #2487, opened 2018-04-02
create keys and pools for client nodes only on first node

guits:
This PR aims to move the client pools and keys creation for from ceph-client to ceph-mon so we do not have to pull container image for client nodes.

PR #2493, opened 2018-04-04
backport of 2487

leseb:
backport of #2487


gfidente:
From TripleO perspective, I am not sure if we want to backport this change into stable-3.0

Assuming we want to backport into stable-3.0 only those changes which are known to fix specific issues, this seems more a generic performance improvement.

leseb:
@gfidente thanks for the clarification.

PR #2499, opened 2018-04-05
Remove deprecated allow_multimds

fullerdj:
allow_multimds will be officially deprecated in Mimic, specify it
only for all versions of Ceph where it was declared stable. Going
forward, specify only max_mds.

Version checks should make this safe to merge.

### DIFF
--- a/roles/ceph-mon/tasks/create_mds_filesystems.yml
+++ b/roles/ceph-mon/tasks/create_mds_filesystems.yml
@@ -22,12 +22,11 @@
   changed_when: false
   when:
     - ceph_release_num[ceph_release] >= ceph_release_num.jewel
-    - mds_allow_multimds
+    - ceph_release_num[ceph_release] < ceph_release_num.mimic
 
 - name: set max_mds
   command: "{{ docker_exec_cmd }} ceph --cluster {{ cluster }} fs set {{ cephfs }} max_mds {{ mds_max_mds }}"
   changed_when: false
   when:
     - ceph_release_num[ceph_release] >= ceph_release_num.jewel
-    - mds_allow_multimds
     - mds_max_mds > 1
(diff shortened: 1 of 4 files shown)


fullerdj:
cf. https://github.com/ceph/ceph/pull/16608

roles/ceph-mon/tasks/create_mds_filesystems.yml
```diff
@@ -21,13 +21,12 @@
   command: "{{ docker_exec_cmd }} ceph --cluster {{ cluster }} fs set {{ cephfs }} allow_multimds true --yes-i-really-mean-it"
   changed_when: false
   when:
-    - ceph_release_num[ceph_release] >= ceph_release_num.jewel
-    - mds_allow_multimds
+    - ceph_release_num[ceph_release] => ceph_release_num.jewel
```
leseb: I think you meant `>=` instead of `=>`

roles/ceph-mon/tasks/create_mds_filesystems.yml
```diff
@@ -22,12 +22,11 @@
   changed_when: false
   when:
     - ceph_release_num[ceph_release] >= ceph_release_num.jewel
-    - mds_allow_multimds
+    - ceph_release_num[ceph_release] < ceph_release_num.mimic
```
guits: from what I understand you want to skip this task only for releases prior to jewel, right?
is that something that will disappear after mimic ?

fullerdj: Yes, allow_multimds will be deprecated in mimic and will disappear afterward (as multiple MDS will always be allowed). It was not stable prior to jewel, so this commit keeps ceph-ansible's prior behavior of not enabling it for versions prior to jewel.

guits:
failure for `luminous-ansible2.4-purge_bluestore_osds_container` is unrelated to this PR.
failure for `luminous-ansible2.4-docker_cluster_collocation` will be addressed by #2517

PR #2497, opened 2018-04-05
tox: use container latest tag for upgrades

leseb:
Currently tag-build-master-luminous-ubuntu-16.04 is not used anymore.
Also now, 'latest' points to CentOS so we need to make that switch here
too.

We know have latest tags for each stable release so let's use them and
point tox at them to deploy the right version.

Signed-off-by: Sébastien Han <seb@redhat.com>

PR #2501, opened 2018-04-06
Use the CentOS repo for Red Hat dev packages

zmc:
No use even trying to use something that doesn't exist.


leseb:
CI failures are unrelated and are fixed in https://github.com/ceph/ceph-ansible/pull/2497

PR #2503, opened 2018-04-09
ceph-nfs: allow disabling ganesha caching

ajarr:
Add a variable, ceph_nfs_disable_caching, that if set to true
disables ganesha's directory and attribute caching as much as
possible.

Also, disable caching done by ganesha, when 'nfs_file_gw'
variable is true, i.e., when Ganesha is used as CephFS's gateway.
This is the recommended Ganesha setting as libcephfs already caches
information. And doing so helps avoid cache incoherency issues
especially with clustered ganesha over CephFS.

Fixes: https://tracker.ceph.com/issues/23393

Signed-off-by: Ramana Raja <rraja@redhat.com>

### DIFF
--- a/roles/ceph-nfs/defaults/main.yml
+++ b/roles/ceph-nfs/defaults/main.yml
@@ -27,6 +27,10 @@ ceph_nfs_enable_service: true
 #######################
 # These are currently in ceph-common defaults because nfs_obj_gw shared with ceph-rgw
 # Enable NFS File access
+# If set to true, then ganesha is set up to export the root of the
+# Ceph filesystem, and ganesha's attribute and directory caching is disabled
+# as much as possible since libcephfs clients also caches the same
+# information.
 nfs_file_gw: false
 # Enable NFS Object access
 nfs_obj_gw: true
@@ -48,6 +52,14 @@ ceph_nfs_rados_export_index: "ganesha-export-index"
 # addresses. (Note: ganesha ignores this parameter in current version due to
 # this bug: https://github.com/nfs-ganesha/nfs-ganesha/issues/217)
 # ceph_nfs_bind_addr: 0.0.0.0
+
+# If set to true, then ganesha's attribute and directory caching is disabled
+# as much as possible. Currently, ganesha caches by default.
+# When using ganesha as CephFS's gateway, it is recommended to turn off
+# ganesha's caching as the libcephfs clients also cache the same information.
+# Note: Irrespective of this option's setting, ganesha's caching is disabled
+# when setting 'nfs_file_gw' option as true.
+ceph_nfs_disable_caching: false
 
 ####################
 # FSAL Ceph Config #
--- a/roles/ceph-nfs/templates/ganesha.conf.j2
+++ b/roles/ceph-nfs/templates/ganesha.conf.j2
@@ -12,6 +12,18 @@ NFS_Core_Param
 {% endif %}
 }
 
+{% if ceph_nfs_disable_caching or nfs_file_gw %}
+CACHEINODE {
+	Dir_Max = 1;
+	Dir_Chunk = 0;
+
+	Cache_FDs = false;
+
+	NParts = 1;
+	Cache_Size = 1;
+}
+{% endif %}
+
 {% if ceph_nfs_rados_backend %}
 RADOS_URLS {
    ceph_conf = '/etc/ceph/{{ cluster }}.conf';
@@ -45,6 +57,8 @@ EXPORT
 	Transports = TCP;
 
 	SecType = sys,krb5,krb5i,krb5p;
+
+	Attr_Expiration_Time = 0;
 
 	FSAL {
 		Name = CEPH;


ajarr:
For reference,
see,

 https://github.com/nfs-ganesha/nfs-ganesha/commit/720b1466a7c982604c24c439e22a4c4d461eed4c#diff-d750277ebcac38a79b101cc0d2ed3f00R57

https://github.com/nfs-ganesha/nfs-ganesha/commit/720b1466a7c982604c24c439e22a4c4d461eed4c#diff-d750277ebcac38a79b101cc0d2ed3f00R102
and

https://github.com/nfs-ganesha/nfs-ganesha/blob/next/src/doc/man/ganesha-export-config.rst#export_defaults-

group_vars/nfss.yml.sample
```diff
@@ -56,6 +56,11 @@ dummy:
 # addresses. (Note: ganesha ignores this parameter in current version due to
 # this bug: https://github.com/nfs-ganesha/nfs-ganesha/issues/217)
 # ceph_nfs_bind_addr: 0.0.0.0
+# Minimize or turn off ganesha's caching such as dentry and fd caching.
+# Currently, the caching is turned on by default.
+# When using ganesha as CephFS's gateway, it's recommended to turn off
+# ganesha's caching as the libcephfs clients also cache the same information.
+# ceph_nfs_caching_off: false
```
batrick: The sample is generated from `roles/ceph-nfs/defaults/main.yml`. You don't need to change this.

roles/ceph-nfs/defaults/main.yml
```diff
@@ -49,6 +49,12 @@ ceph_nfs_rados_export_index: "ganesha-export-index"
 # this bug: https://github.com/nfs-ganesha/nfs-ganesha/issues/217)
 # ceph_nfs_bind_addr: 0.0.0.0
 
+# Minimize or turn off ganesha's caching such as dentry and fd caching.
+# Currently, the caching is turned on by default.
+# When using ganesha as CephFS's gateway, it's recommended to turn off
+# ganesha's caching as the libcephfs clients also cache the same information.
+ceph_nfs_caching_off: false
+
```
batrick: From the code it appears that `ceph_nfs_caching_off` is ignored when `ceph_nfs_file` is true. (I think this is the right thing.) The comment doesn't make that clear though.

roles/ceph-nfs/templates/ganesha.conf.j2
```diff
@@ -46,6 +58,8 @@ EXPORT
 
 	SecType = sys,krb5,krb5i,krb5p;
 
+	Attr_Expiration_Time = 0;
```
batrick: Should this also be addded to rados gw export but protected by `ceph_nfs_caching_off`?

roles/ceph-nfs/defaults/main.yml
batrick: I think we can improve the option name: `ceph_nfs_cache`.

group_vars/nfss.yml.sample
ajarr: will do

leseb: AS @batrick suggested you need to edit `roles/ceph-nfs/defaults/main.yml` only and then run `generate_group_vars_sample.sh`.

roles/ceph-nfs/defaults/main.yml
ajarr: > From the code it appears that ceph_nfs_caching_off is ignored when ceph_nfs_file is true. (I think this is the right thing.) The comment doesn't make that clear though.

OK. Added more info in the comments.

ajarr: > I think we can improve the option name: ceph_nfs_cache.

OK, changed to 'ceph_nfs_caching'. So if 'ceph_nfs_caching = true' , then doesn't it imply that ceph nfs caching is turned on by ceph-ansible by setting config options, which ceph-ansible doesn't. What if ganesha project decides to turn off caching by default, then 'ceph_nfs_caching = true' should set options to turn on caching?

I named is as 'ceph_nfs_caching_off', to be more explicit about what the option does, ceph-ansible only tries to disable caching and does not try to turn on caching. I think I should've named it 'ceph_nfs_disable_caching' instead.
@batrick what do you think?

roles/ceph-nfs/templates/ganesha.conf.j2
ajarr: I didn't want to protect 'Attr_Expiration_Time = 0' setting. Ideally, it should've been a EXPORT_DEFAULT option, to be effective for all exports when 'cephfs_nfs_caching_off = True'. But found that with at least NFS-Ganesha v2.5.5, 
'Attr_Expiration_Time' is only recognized as a per export option and not a EXPORT_DEFAULTS option.

I'm not sure about the caching model with Ganesha over RGW.  Also, I don't want to change the default export block settings (since 'nfs_file_gw' is true by default) unless we're certain.

guits:
jenkins test luminous-ansible2.4-bluestore_osds_container

guits:
jenkins test luminous-ansible2.4-bluestore_osds_container

ajarr:
@leseb can we merge this?

leseb:
@ajarr yes!

PR #2511, opened 2018-04-10
To support copy_admin_key 

VasishtaShastry:
With this patch, ceph-ansible will copy admin keyring to osd, mds, nfs nodes if copy_admin_key is set to true.
This also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1544720
(fixes  OSD - Both containerized and non containerized initialization, mds and nfs - only containerized scenario)


guits:
jenkins test luminous-ansible2.4-bluestore_osds_container

PR #2513, opened 2018-04-10
Changes which allow us install Ceph in version 10.2.5

kimbarbel:


kimbarbel:
Please ignore that that was by mistake.

PR #2517, opened 2018-04-12
tests: update tests for mds to cover multimds case

guits:
in case of multimds we must check for the number of mds up instead of
just checking if the hostname of the node is in the fsmap.

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #2521, opened 2018-04-12
Backport of 2517

leseb:
Backport of #2517


leseb:
This relies on #2499, which hasn't been backported so this one is not needed.

PR #2537, opened 2018-04-20
[skip ci] shrink-osd: ability to shrink NVMe drives

leseb:
Now if the service name contains nvme we know we need to remove the last
2 character instead of 1.

If nvme then osd_to_kill_disks is nvme0n1, we need nvme0
If ssd or hdd then osd_to_kill_disks is sda1, we need sda

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1561456


leseb:
jenkins test luminous-ansible2.4-shrink_osd_container

PR #2543, opened 2018-04-26
Move apt cache update to individual task per role

andymcc:
The apt-cache update can fail due to transient issues related to the
action being a network operation. To reduce the impact of these
transient failures this patch adds a retry to the update_cache task.

However, the apt_repository tasks which would perform an apt_update
won't retry the apt_update on a failure in the same way, as such this PR
moves the apt_update into an individual task, once per role.

Finally, the apt_repository tasks no longer have a changed_when: false,
and the apt_cache update is only performed once per role, if the
repositories change. Otherwise the cache is updated on the "apt" install
tasks if the cache_timeout has been reached.

### DIFF
--- a/roles/ceph-nfs/tasks/pre_requisite_non_container.yml
+++ b/roles/ceph-nfs/tasks/pre_requisite_non_container.yml
@@ -41,12 +41,27 @@
   apt_repository:
     repo: "deb {{ nfs_ganesha_stable_deb_repo }} {{ ceph_stable_distro_source | default(ansible_lsb.codename) }} main"
     state: present
-  changed_when: false
+    update_cache: no
+  register: add_ganesha_apt_repo
   when:
     - ansible_os_family == 'Debian'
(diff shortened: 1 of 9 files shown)


roles/ceph-nfs/tasks/pre_requisite_non_container.yml
```diff
@@ -41,12 +41,27 @@
...
+  when:
+    - ansible_os_family == 'Debian'
+    - nfs_ganesha_stable
+    - ceph_origin == 'repository'
+    - ceph_repository == 'community'
+
+- name: update apt cache
+  apt:
+    update_cache: yes
+  register: update_ganehsa_apt_cache
```
odyssey4me: spelling error here, resulting in the registered var and the used var not matching

andymcc: 👍 fixed! thanks

Issue #2504, opened 2018-04-09
RFE: install or upgrade all packages in one task


ktdreyer:
`roles/ceph-common/tasks/installs/install_redhat_packages.yml` runs several distinct Yum operations.

Instead of executing `yum install ceph-common ceph-base ceph-mon`, for example, it executes each of these serially: `yum install ceph-common`, `yum install ceph-base`, `yum install ceph-mon`.

This method is sensitive to packaging changes in some cases (eg. https://github.com/ceph/ceph-ansible/pull/2500).

A better method would be to build up a list of packages so that Ansible executes one single `yum` command.

This will allow the package manager to gracefully handle the sort of dependency solving issues described in PR 2500.

This will also provide a speed benefit because we're exec'ing Yum once instead of doing it several times.

Regarding implementation, note that [Ansible's yum module docs](http://docs.ansible.com/ansible/latest/modules/yum_module.html) say "When used with a loop: each package will be processed individually, it is much more efficient to pass the list directly to the name option."

ktdreyer:
Looks like this went into master (will be v3.2) in https://github.com/ceph/ceph-ansible/pull/2582

PR #2534, opened 2018-04-19
switch: fix ceph_uid fact

leseb:
Latest is now centos not ubuntu anymore so the condition was wrong.

Signed-off-by: Sébastien Han <seb@redhat.com>

PR #2544, opened 2018-04-26
Install Apt packages as a list

andymcc:
We can reduce the number of tasks required, and install all packages in
one go, rather than looping through them or installing them as
individual tasks. This will be more efficient since a list can be passed
to the apt install process.

This PR adds a list creation based on the conditionals used, and then
installs the packages in one go.

### DIFF
--- a/roles/ceph-common/tasks/installs/install_debian_packages.yml
+++ b/roles/ceph-common/tasks/installs/install_debian_packages.yml
@@ -1,29 +1,18 @@
 ---
-- name: install ceph for debian
+- name: set fact for packages that are always installed
+  set_fact:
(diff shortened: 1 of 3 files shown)


leseb:
Closes: https://github.com/ceph/ceph-ansible/issues/2504

leseb:
This needs a rebase on https://github.com/ceph/ceph-ansible/pull/2534 once merged.

andymcc:
should be np, i'll update it to include the others.

leseb:
@andymcc thanks mate!

andymcc:
#2552 is the included one, I also improved it slightly - you won't have to loop through to create the list (which in hindsight is mostly pointless). It'll create the list var dynamically outside of a task.

PR #2549, opened 2018-04-30
[skip ci] [WIP] rolling_upgrade: follow MDS upgrade procedure

batrick:
From: http://docs.ceph.com/docs/master/cephfs/upgrading/#upgrading-the-mds-cluster

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -445,6 +445,9 @@
         - not jewel_minor_update
 
 
+# See also:
+#   - [master] http://docs.ceph.com/docs/master/cephfs/upgrading/
+#   - [pre-mimic] http://docs.ceph.com/docs/luminous/cephfs/upgrading/
 - name: upgrade ceph mdss cluster
 
   vars:
@@ -457,6 +460,28 @@
   become: True
 
   pre_tasks:
+    - name: get fsmap
+      command: "{{ docker_exec_cmd_update_osd|default('') }} ceph --cluster {{ cluster }} fs dump --format=json"
+      delegate_to: "{{ groups[mon_group_name][0] }}"
+      register: fetched_fsmap
+
+    - name: set_fact fsmap
+      set_fact:
+        pre_upgrade_fsmap: "{{ fetched_fsmap.stdout | from_json }}"
+
+    # JSON looks like:
+    # {..., "filesystems": [{"fs_name": "foo", ...}, {"fs_name": "bar", ...}]}
+    - name: reduce ranks to 1
+      command: "{{ docker_exec_cmd_update_osd|default('') }} ceph --cluster {{ cluster }} fs set {{ item.fs_name }} max_mds 1"
+      delegate_to: "{{ groups[mon_group_name][0] }}"
+      with_items: "{{ pre_upgrade_fsmap.filesystems | select('mds_map.fs_name') }}"
+
+    - name: cephfs preupgrade sequence
+      command: "{{ docker_exec_cmd_update_osd|default('') }} python /tmp/preupgrade_cephfs.py {{ cluster }}"
+      delegate_to: "{{ groups[mon_group_name][0] }}"
+
+    # XXX From "get fsmap" to "cephfs preup...", these should be executed ONCE on the mons for EACH CephFS file system, THEN all mdss are restarted. THEN "restore max_mds" is done once.
+
     - name: stop ceph mds
       systemd:
         name: ceph-mds@{{ ansible_hostname }}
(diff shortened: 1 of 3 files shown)


batrick:
I don't expect this code to work as-is. I need help rapidly learning how to test this. Also, the procedure is different for pre-Mimic clusters so the code needs to add a check for that and do a slightly differnet procedure:

http://docs.ceph.com/docs/luminous/cephfs/upgrading/

infrastructure-playbooks/rolling_update.yml
```diff
@@ -401,46 +401,35 @@
...
       when: containerized_deployment
 
-    - name: wait for stable fsmap
-      command: ceph --cluster {{ cluster }} fs dump --format=json
-      delay: 5
-      delegate_to: "{{ groups[mon_group_name][0] }}"
-      register: get_fsmap
-      until: {{ get_fsmap.stdout | from_json }}['mds_map']['in'] == [0]
-      when: not containerized_deployment
+    - name: deactivate all non-zero ranks
```
batrick: @leseb, I need some ansible expertise for this PR. This loop needs to try to repeatedly deactivate the highest rank and then wait for the FSMap to stabilize (i.e. the highest non-zero rank fully deactivates). This requires running a sequence of tasks (mds_pre_upgrade.yml) until a condition is met: `highest_rank == 0`. Unfortunately, it doesn't seem possible to run a sequence of tasks in a loop: there appears to be a bug where `include_tasks` works as expected but the `when:` clause is ignored so the "loop" always runs once.

How should I go about fixing this?

guits: @batrick according to this issue https://github.com/ansible/ansible/issues/17098 it seems this is still not supported...
At the moment, I don't see an easier workaround than what is proposed by @nerdvegas in the issue

infrastructure-playbooks/fs_status.yml
```diff
@@ -0,0 +1,15 @@
+- name: get fsmap
+  command: ceph --cluster {{ cluster }} fs dump --format=json
+  delegate_to: "{{ groups[mon_group_name][0] }}"
+  register: get_fsmap
+  when: not containerized_deployment
+
+- name: get fsmap (containerized)
```
guits: Isn't it possible to merge this task with the one above using `{{ docker_exec_cmd }}` logic like we do at many places in the playbook?

infrastructure-playbooks/rolling_update.yml
batrick: Do we have a suggessted approach for rurnning a bash script on the mons? e.g. where do we store the script? Some existing code to look at?

@guits @leseb

infrastructure-playbooks/fs_status.yml
batrick: Maybe? I have no idea. I'm just following code I copy-pasted from somewhere else.

infrastructure-playbooks/rolling_update.yml
leseb: You can write a jinja template similar to this https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/templates/restart_mon_daemon.sh.j2 and then run it like this: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/handlers/main.yml#L28

Does that help?

leseb:
@batrick what's the status of this? How can we help? Thanks

batrick:
We need to generate the bash script which can do the loop deactivating MDS ranks. I haven't had time to work on it yet.

batrick:
@leseb @guits I have the CephFS parts of this done but I'm not sure how to get the ansible bits right. Can I get some help? I left some comments in.

There's now a python script (py2/3 compat) that does the preupgrade sequence.

infrastructure-playbooks/rolling_update.yml
```diff
@@ -457,6 +460,29 @@
   become: True
 
   pre_tasks:
+    - name: get fsmap
```
leseb: move this after the osd upgrade is done or actually create its own section

roles/ceph-defaults/handlers/main.yml
```diff
@@ -24,6 +24,17 @@
   when:
     - mon_group_name in group_names
 
+- name: copy cephfs preupgrade script
```
leseb: move this into rolling_update.yml

leseb:
@batrick can you allow write access to that PR so I get push into your branch? Thanks! https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

batrick:
@leseb, the PR shows you have permission (that check box is already clicked).

leseb:
@batrick thanks I'll give it a try then.

batrick:
@leseb, if that doesn't work, feel free to just create a new PR.

leseb:
@batrick nop, this hasn't moved yet because I didn't have time yet. Probably next week.

batrick:
ACK, no problem

PR #2555, opened 2018-05-02
Config validation action plugin

andrewschoen:
This is a continuation of https://github.com/ceph/ceph-ansible/pull/2468

The action plugin will validate that the provided config is correct very early in the playbook run so that no time is wasted deploying a cluster that has incorrect config.

Here's a validation failure you'd see if ``osd_objectstore`` was set to an incorrect value.

```
TASK [ceph-validate : validate provided configuration] ******************************************************************************************************************************
task path: /Users/andrewschoen/dev/ceph-ansible/roles/ceph-validate/tasks/main.yml:3
Wednesday 02 May 2018  13:48:16 -0500 (0:00:06.984)       0:00:18.803 *********
 [ERROR]: [mon0] Validation failed for variable: osd_objectstore

 [ERROR]: [mon0] Given value for osd_objectstore: foo

 [ERROR]: [mon0] Reason: osd_objectstore must be either 'bluestore' or 'filestore'

fatal: [mon0]: FAILED! => {
    "changed": false
}

MSG:

[mon0] Validation failed for variable: osd_objectstore
[mon0] Reason: osd_objectstore must be either 'bluestore' or 'filestore'
[mon0] Given value for osd_objectstore: foo

 [ERROR]: [osd0] Validation failed for variable: osd_objectstore

 [ERROR]: [osd0] Given value for osd_objectstore: foo

 [ERROR]: [osd0] Reason: osd_objectstore must be either 'bluestore' or 'filestore'

fatal: [osd0]: FAILED! => {
    "changed": false
}

MSG:

[osd0] Validation failed for variable: osd_objectstore
[osd0] Reason: osd_objectstore must be either 'bluestore' or 'filestore'
[osd0] Given value for osd_objectstore: foo
```

Here's a failure with a filestore cluster that using the ``lvm`` osd scenario, but has not provided a journal for one of the osds defined in ``lvm_volumes``.

```
TASK [ceph-validate : validate provided configuration] ******************************************************************************************************************************
task path: /Users/andrewschoen/dev/ceph-ansible/roles/ceph-validate/tasks/main.yml:3
Wednesday 02 May 2018  13:50:01 -0500 (0:00:06.524)       0:00:18.444 *********
ok: [mon0] => {
    "changed": false
}
 [ERROR]: [osd0] Validation failed for variable: lvm_volumes

 [ERROR]: [osd0] Reason: -> lvm_volumes -> list[0] -> item[2] did not match 'journal' (required item in schema is missing: journal)

fatal: [osd0]: FAILED! => {
    "changed": false
}

MSG:

[osd0] Validation failed for variable: lvm_volumes
[osd0] Reason: -> lvm_volumes -> list[0] -> item[2] did not match 'journal' (required item in schema is missing: journal)
```


andrewschoen:
@vasukulkarni you might be interested in this PR. Also note that this will require the installation of ``python-notario`` when running ``site.yml.sample`` or ``site-docker.yml.sample``. This will affect your teuthology testing both upstream and downstream.

alfredodeza:
we could maybe consider a flag/option to issue a warning instead of bailing out if notario is not available/installed

andrewschoen:
@alfredodeza good idea, I could add that warning if notario is not installed. I'd want it to fail the play completely and tell them to install notario, do you agree? I'm not sure we should give people and 'out' here or an easy way to avoid config validation maybe even without their knowledge.

andrewschoen:
Looks like I missed some things in regards to validating a containerized deployment. I'll fix that.

vasukulkarni:
@andrewschoen thanks for the heads up, we will have to take care of install when doing git clone, but I guess rpm installation should be fine and whatever is required by ceph-ansible would be installed?

plugins/actions/validate.py
```diff
@@ -0,0 +1,233 @@
...
+    assert value in ["filestore", "bluestore"], "objectstore must be set to 'filestore' or 'bluestore'"
+
+
+def validate_lvm_volumes(value):
+    if notario_store['osd_objectstore'] == "filestore":
+        assert isinstance(value, basestring), "lvm_volumes must contain a 'journal' key when the objectstore is 'filestore'"
+
+
+def validate_ceph_stable_release(value):
+    assert value in ['jewel', 'kraken', 'luminous', 'mimic'], "ceph_stable_release must be set to 'jewel', 'kraken', 'lumious' or 'mimic'"
```
leseb: I think ['jewel', 'kraken', 'luminous', 'mimic'] should be a list somewhere so we can easily reference it.
It's easier than always adding a new release name in here.

roles/ceph-validate/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
+---
+
```
leseb: empty is not needed

site.yml.sample
```diff
@@ -68,6 +68,27 @@
         - ansible_distribution == 'Fedora'
         - ansible_distribution_major_version|int >= 23
 
+- hosts:
```
leseb: can we make line 25 https://github.com/ceph/ceph-ansible/pull/2555/files#diff-5e11e06bb360175675e3ae901c5d08dbR25 a pre task and then run the validation right after instead of re-declaring a `hosts` block again?

site-docker.yml.sample
```diff
@@ -70,6 +70,27 @@
...
+  - mgrs
+  - iscsi-gws
+
+  gather_facts: false
+  any_errors_fatal: true
+  become: true
+
+  roles:
+    - ceph-defaults
+    - ceph-validate
```
leseb: The validation should be the first thing we do, here we potentially try to fetch container images first and then run the validation.

plugins/actions/validate.py
andrewschoen: This is the list, but I could maybe add a constant or something at the top of this file to make it easier to find?

roles/ceph-validate/tasks/main.yml
andrewschoen: will fix

site-docker.yml.sample
andrewschoen: That container fetching happens in ``ceph-defaults``? I do need the ``ceph-defaults`` role before ``ceph-validate`` of the action plugin doesn't have access to the variables I'm trying to validate. I ran into issues when I removed ``ceph-defaults``.

guits: @leseb it's sort of what we have been already facing before coming up with this validate module.

plugins/actions/validate.py
leseb: I know it's a list, what I meant is that the list could be a constant that's right :), thanks

site-docker.yml.sample
```diff
@@ -54,6 +55,7 @@
         - containerized_deployment | bool
         - not (is_atomic | bool)
         - (not (inventory_hostname in groups.get('clients', [])) or (inventory_hostname == groups.get('clients', [''])|first))
+    - role: ceph-validate
```
leseb: I think the `when` should be removed so we are sure to run this all the time

andrewschoen: Ok, I've removed it.

andrewschoen:
I've got one more bug here to fix related to default variables not being expanded, but I know the solution.

The failures I'm seeing are on ``containerized_deployment`` because it still has a value of ``"{{ True if mon_containerized_deployment or osd_containerized_deployment or mds_containerized_deployment or rgw_containerized_deployment else False }}"``.

andrewschoen:

andrewschoen:

andrewschoen:

andrewschoen:

andrewschoen:

andrewschoen:

guits:
```
fatal: [osd4]: FAILED! => {
    "changed": false
}

MSG:

[osd4] Validation failed for variable: item[0]
[osd4] Reason: -> item[0] key did not match 'devices' (required item in schema is missing: devices)
```

when `osd_auto_discovery` is set to `True` the variable `devices` becomes optional

guits:
@andrewschoen I didn't sift through all commits but I think some of them could be squashed
For instance, this one could be easily avoided: https://github.com/ceph/ceph-ansible/pull/2555/commits/43b87b6991d62f237815d15278aa7ca745527b20 by amending https://github.com/ceph/ceph-ansible/commit/197b4d84edd90d279633b5cf1e7d0acc471a31da I guess

This one is weird: https://github.com/ceph/ceph-ansible/pull/2555/commits/4a5a9a59f1b768a9ac7acaf1c36b77bf8d2dcbd1

guits:
@andrewschoen I tend to be +1 on removing `osd_auto_discovery`

leseb:
The thing is this is something we test and support downstream so this has to stay a bit longer. I think this will naturally go away once `choose_disk` is out. However, we'd have to maintain backward compatibility. So please, let's keep it for now.

andrewschoen:
@guits I'm not sure what you're expecting here. All commits now how a meaningful title as far as I can see. Are you expecting extended messages in all of them or are there still some that are confusing you as to their intent?

guits:
@andrewschoen I was just noticing from a traceability point of view it can be a bit difficult to understand the workflow of all these commits for someone who hasn't worked on this.

guits:

roles/ceph-defaults/defaults/main.yml
```diff
@@ -503,8 +503,7 @@ mon_containerized_deployment: False # backward compatibility with stable-2.2, wi
 osd_containerized_deployment: False # backward compatibility with stable-2.2, will disappear in stable 3.1
 mds_containerized_deployment: False # backward compatibility with stable-2.2, will disappear in stable 3.1
 rgw_containerized_deployment: False # backward compatibility with stable-2.2, will disappear in stable 3.1
-containerized_deployment: "{{ True if mon_containerized_deployment or osd_containerized_deployment or mds_containerized_deployment or rgw_containerized_deployment else False }}" # backward compatibility with stable-2.2, will disappear in stable 3.1
-
+containerized_deployment: False
```
guits: Perhaps it should be part of a patch removing all the backward compatibility with stable-2.2 ?

roles/ceph-defaults/defaults/main.yml
```diff
@@ -492,8 +503,7 @@ mon_containerized_deployment: False # backward compatibility with stable-2.2, wi
 osd_containerized_deployment: False # backward compatibility with stable-2.2, will disappear in stable 3.1
```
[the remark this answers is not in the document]
andrewschoen: I did try to remove these and it caused a bunch of breakage in the roles. I think fixing that should be done in a separate PR.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+ansible>=2.4.2
```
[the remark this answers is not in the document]
andrewschoen: We have requirement.txt files for the CI tests, but now we need one for the project so that ``notario`` can be easily installed along with our supported ansible version.

roles/ceph-defaults/defaults/main.yml
andrewschoen: Without at least this much done in this PR the config won't validate, because ``containerized_deployment`` should be a boolean and when the validate module gets the config it's not expanded, so it reads as a string and fails validation.

I propose we remove the rest of the backwards compatibility stuff in a separate PR.

roles/ceph-defaults/defaults/main.yml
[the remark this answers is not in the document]
leseb: Oh I see, sounds good.

roles/ceph-defaults/defaults/main.yml
guits: Ok, I guess because of what you said above
> I did try to remove these and it caused a bunch of breakage in the roles. I think fixing that should be done in a separate PR.

leseb:
jenkins test this please

roles/ceph-common/tasks/main.yml
```diff
@@ -2,9 +2,6 @@
 - name: include checks/check_system.yml
   include: checks/check_system.yml
```
leseb: This should be moved in `roles/ceph-validate/tasks/` as in include.

ktdreyer:
That's right @leseb . We should indicate this in the RPM packaging.

andrewschoen:
@leseb @ktdreyer I've got this change made to ``ceph-ansible.spec.ini`` https://github.com/ceph/ceph-ansible/pull/2555/files#diff-b6d431966a0d33b2f60ad62b8cc3dfe5R24

Is there something else I need to add to that file?

PR #2557, opened 2018-05-03
[TEST2] Install packages as a list

andymcc:
This is a test approach to see timing when generating the list as a task
and then installing once as part of a yum install, vs generating the
list in vars/main.yml - this will help us determine the time of the
actual install excluding the list generation and see if that is where
the time is being lost.

To make the package installation more efficient we should install
packages as a list rather than as individual tasks or using a
"with_items" loop. The package managers can handle a list passed to them
to install in one go.

In order to reduce the need for a task to sort the packages we can
create a var that will filter the packages and produce the list of
packages to install.


andymcc:
Closing in favour of #2582

PR #2558, opened 2018-05-03
common: copy iso files if rolling_update

leseb:
If we are in a middle of an update we want to get the new package
version being installed so the task that copies the repo files should
not be skipped.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1572032


guits:
@leseb should we wait for the CI since It's not covered?

leseb:
Rahhh I forgot, no need to wait, feel free to merge :).

PR #2576, opened 2018-05-10
Makefile: bail out on unknown Git tag formats

ktdreyer:
Prior to this change, if we created entirely new Git tags patterns like "3.2.0alpha" or "3.2.0foobar", the Makefile would incorrectly translate the Git tag name into a Name-Version-Release that would prevent upgrades to "newer" versions.

This happened for example in https://bugs.centos.org/view.php?id=14593, "Incorrect naming scheme for a build of ceph-ansible prevents subsequent updates to be installed"

If we encounter a new Git tag format that we cannot parse, pessimistically bail out early instead of trying to build an RPM.

The purpose of this safeguard is to prevent Jenkins from building RPMs that cannot be easily upgraded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,13 @@ ifneq (,$(findstring rc,$(VERSION)))
     RELEASE := 0.$(RC).$(RELEASE)
     VERSION := $(subst $(RC),,$(VERSION))
 endif
+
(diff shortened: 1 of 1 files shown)


leseb:
Please use `[skip ci]` next time.

PR #2582, opened 2018-05-11
Install packages as a list

andymcc:
To make the package installation more efficient we should install
packages as a list rather than as individual tasks or using a
"with_items" loop. The package managers can handle a list passed to them
to install in one go.

We can use a specified list with the "omit" filter to indicate when
packages should be installed or not.


andymcc:
I agreed - I'll change the commit message and close the others.

guits:

PR #2588, opened 2018-05-15
rolling_update: fix dest path for mgr keys fetching

guits:
the role `ceph-mgr` that is played later in the playbook fails because
the destination path for the fetched keys is wrong.
This patch fix the destination path used in the task `fetch ceph mgr
key(s)` so there is no mismatch.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1574995

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #2590, opened 2018-05-15
bkp of 2588: rolling_update: fix dest path for mgr keys fetching

leseb:
the role `ceph-mgr` that is played later in the playbook fails because
the destination path for the fetched keys is wrong.
This patch fix the destination path used in the task `fetch ceph mgr
key(s)` so there is no mismatch.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1574995

(cherry picked from commit 1b4c3f292d8779158ea445a8c9a11c8ed26abe11)

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -192,7 +192,48 @@
   become: True
 
   pre_tasks:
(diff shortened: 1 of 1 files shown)


leseb:
backport of #2588

PR #2595, opened 2018-05-16
[skip ci] switch: disable ceph-disk units

leseb:
During the transition from jewel non-container to container old ceph
units are disabled. ceph-disk can still remain in some cases and will
appear as 'loaded failed', this is not a problem although operators
might not like to see these units failing. That's why we remove them if
we find them.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1577846


leseb:
jenkins test luminous-ansible2.4-switch_to_containers

leseb:
jenkins test luminous-ansible2.4-switch_to_containers

PR #2610, opened 2018-05-17
[skip ci] Do nothing when mgr module is in good state

mvollman:
Check whether a mgr module is supposed to be disabled before disabling
it and whether it is already enabled before enabling it.


leseb:
Thanks @mvollman for the follow-up.

mvollman:
Added sign off

Issue #2618, opened 2018-05-21
ceph_release is no longer set for dev or rhcs 


alfredodeza:
The validation PR moved the mandatory vars before ceph-common gets executed, which has an include for `release-rhcs.yml` that in turn overrides the `ceph_stable_release` *which in turn* gets assigned to `ceph_release`.  

That whole process allowed one to have a dev installation from master that could get inferred as a mimic `ceph_stable_release`.  According to the comment on the task, this will break rhcs as well:

    # override ceph_stable_release for ceph_dev and rhcs installations since ceph_stable_release is not mandatory
    - name: include release-rhcs.yml
      include: release-rhcs.yml
      when:
        - ceph_repository in ['rhcs', 'dev']
      tags:
        - always
    
The failure looks like:

    The conditional check 'ceph_release_num[ceph_release] < ceph_release_num.luminous' failed. The error was: error while evaluating conditional (ceph_release_num[ceph_release] < ceph_release_num.luminous): 'dict object' has no attribute u'dummy'
    
    The error appears to have been in '/Users/alfredo/python/ceph/src/ceph-volume/ceph_volume/tests/functional/lvm/.tox/xenial-bluestore-dmcrypt/tmp/ceph-ansible/roles/ceph-validate/tasks/main.yml': line 47, column 3, but may
    be elsewhere in the file depending on the exact syntax problem.
    
    The offending line appears to be:
    
    
    - name: check if osd_scenario lvm is supported by the selected ceph version
      ^ here

PR #2619, opened 2018-05-21
ceph-validate: do not check ceph version on dev or rhcs installs

andrewschoen:
A dev or rhcs install does not require ceph_stable_release to be set and
instead generates that by looking at the installed ceph-version.
However, at this point in the playbook ceph may not have been installed
yet and ceph-common has not be run.

Fixes: https://github.com/ceph/ceph-ansible/issues/2618


andrewschoen:
jenkins test dev-ansible2.4-lvm_osds

PR #2620, opened 2018-05-21
ceph-radosgw: disable NSS PKI db when SSL is disabled

ltoscano-rh:
The NSS PKI database is needed only if radosgw_keystone_ssl
is explicitly set to true, otherwise the SSL integration is
not enabled.

It is worth noting that the PKI support was removed from Keystone
starting from the Ocata release, so some code paths should be
changed anyway.

Signed-off-by: Luigi Toscano <ltoscano@redhat.com>


ltoscano-rh:
If the idea is correct and the patch approved, I'd like to backport it to stable-3.1.

ltoscano-rh:
I updated the review, is it expected that it does not trigger again the CI? I see that you changed the title.

leseb:

leseb:
@ltoscano-rh thanks, yes, I updated the title so we run the CI when this is ready for review.

leseb:

leseb:
https://bugzilla.redhat.com/show_bug.cgi?id=1581573

Issue #2631, opened 2018-05-23
top-level requirements.txt does not pin dependencies


alfredodeza:
The top-level `requirements.txt` does not pin Ansible version, using instead:

    ansible>=2.4.2

At the time of this writing, this will install `2.5.3` which is causing breakage on vanilla deployments for both Xenial and CentOS7. Other projects that install ceph-ansible for testing that want to make use of the dependencies listed for ceph-ansible will break (teuthology, ceph-medic, ceph-volume, etc...) 

I think that the top level requirements.txt should go hand in hand with whatever the branch is willing to support. For example, if the stable-2.5 branch works with ansible 2.2, then it should pin to that version (or a range if that is desired). 

For the *master* branch, it should be probably the latest Ansible version that will be used in the next stable release. In this case, I think that would be Ansible 2.5

Having these dependencies pinned will ease the burden of other projects to just blindly depend on what ceph-ansible can support, regardless of the branch.

It however imposes that burden on the ceph-ansible team, as this would be another thing that needs to get updated continuously to make sure that dependencies are correct as time goes by.

ktdreyer:
For the RH Ceph Storage product, it would be very helpful if the pinned values matched the latest packages that are available in RHEL's repositories. For example `rhel-7-server-ansible-2.4-rpms` has 
ansible-2.4.4.0-1.el7ae.noarch.rpm.

Maybe the ceph-ansible CI system could use ansible RPMs instead of PyPI packages.

Issue #2470, opened 2018-03-24
Add Ansible 2.5 testing


alfredodeza:
2.5 was released March 23rd, we should add this to the tox.ini asap, and start looking for issues

https://pypi.python.org/pypi/ansible/2.5.0

alfredodeza:
I don't see any test runs for 2.5 in the CI, and although pr #2476 adds the 2.5 factor, there are no jenkins jobs to use that.

We just hit this problem in ceph-volume tests because we started using the top level `requirements.txt` that does not pin Ansible, and we were getting Ansible 2.5.3 which fails on something that doesn't fail on 2.4

A nightly job or even some PR jobs that use 2.5 would be helpful to identify these early in ceph-ansible master

guits:
@leseb shouldn't we test all PRs on master against ansible 2.5 and keep ansible 2.4 for nightlies job on stable-3.x ?

leseb:
I think we should test 2.5 on master, then keep 2.4 for stable-3.x. However, it really depends on the version downstream is shipping.

@ktdreyer please let us which version of Ansible we should test with respect to the ceph-ansible stable version.

stable-3.0: ansible version?
stable-3.1: ansible version?
master: ansible version?

Thanks.

ktdreyer:
stable-3.0: ansible 2.4.latest
stable-3.1: ansible 2.4.latest and ansible 2.5.latest
master: ansible 2.5.latest

ErwanAliasr1:
@ktdreyer why does stable-3.1 would have to support two versions of ansible ?

leseb:
@ktdreyer How are we supposed to test on 2 Ansible versions for stable-3.1? We can only run tests on a single Ansible version at the moment.

alfredodeza:
@leseb what do you mean? the ansible version is a tox "factor" and it shows all the multiple ansible versions that are ready to be tested:

    [stable-3.1]$ tox -l | grep rhcs
    rhcs-ansible2.2-xenial_cluster
    rhcs-ansible2.2-centos7_cluster
    rhcs-ansible2.2-docker_cluster
    rhcs-ansible2.2-update_cluster
    rhcs-ansible2.2-cluster
    rhcs-ansible2.2-update_docker_cluster
    rhcs-ansible2.2-switch_to_containers
    rhcs-ansible2.2-purge_filestore_osds_container
    rhcs-ansible2.2-purge_filestore_osds_non_container
[...]

PR #2632, opened 2018-05-23
pin version of ansible to 2.4.2 in requirements.txt

andrewschoen:
This is the latest version that we support. If we don't pin this we
get a 2.5.x version installed that causes the playbook to fail in
various ways.

Fixes: https://github.com/ceph/ceph-ansible/issues/2631

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-ansible>=2.4.2
+ansible~=2.4
 notario>=0.0.13
 netaddr>=0.7.19


andrewschoen:
Should we set this to ``ansible==2.5.0`` instead? I see that version is in the testing matrix.

guits:
@andrewschoen I think we should test with 2.5 for PRs going to master and 2.4 for PRs going to stable-3.x branches @alfredodeza has re-opened a PR (#2470) where I have already raised this question

alfredodeza:
I think this is a two-step process: we need to pin, so use what works today with ceph-ansible (2.4) and lets add the job(s) needed to ensure that 2.5 works with master. 

Once ceph-ansible is known to work with 2.5, lets update this. Otherwise, we are pinning to something that doesn't work for anyone.

andrewschoen:
@guits I agree, we should certainly test each branch against their supported version. I agree with @alfredodeza though in that we should get this in now to fix the immediate breakage and then follow up with the ``ansible2.5`` testing and CI support.

requirements.txt
```diff
@@ -1,2 +1,2 @@
-ansible>=2.4.2
+ansible==2.4.2
```
leseb: Looks like 2.4.5 is there https://github.com/ansible/ansible/blob/stable-2.4/CHANGELOG.md.
So shouldn't we try something like > 2.4 and < 2.5?

leseb: or just `ansible==2.4` to get the last one maybe

ktdreyer: :+1: to somehow pinning to "latest 2.4"

guits: the correct syntax seems to be : `ansible~=2.4`

`ansible==2.4` will give you ansible version `2.4.0` even if `2.4.5` is available

andrewschoen:
@leseb @guits I've got this changed to ``ansible~=2.4`` now

guits:

leseb:

leseb:
@andrewschoen please confirm whether this needs a backport or not.
Thanks!

andrewschoen:
@leseb it looks like the validate module isn't included in ``stable-3.1`` at all. I would assume that's something we wanted backported, yeah?

guits:
We have already released stable-3.1 and backporting validate would mean backporting a feature, is that something expected? Would an RFE be enough to consider backporting this module?

leseb:
@andrewschoen the rule is to not backport feature, but only bug fixes, so validate won't be part of 3.1.

PR #2644, opened 2018-05-24
[skip ci] iscsi purge process needed sudo privileges

pcuzner:
Without the become:yes, the purge tasks fail with either access issues to the rados config object
or the log file in /var/log.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1549004


pcuzner:
@leseb not sure what to do here. I resubmitted, and added a closes and signed-off-by but it's still failing dco check. What am I doing wrong? :/

leseb:
Use git commit --amend -s
That should do it.

PR #2652, opened 2018-05-28
tests: avoid yum failures

guits:
In the CI we can see at many times failures like following:

`Failure talking to yum: Cannot find a valid baseurl for repo:
base/7/x86_64`

It seems the fastest mirror detection is sometimes counterproductive and
leads yum to fail.

This has been added in the `setup.yml`.
This playbook was used until now only just before playing `testinfra`
and could be used before running ceph-ansible so we can add some
provisionning tasks.

### DIFF
--- a/tests/functional/setup.yml
+++ b/tests/functional/setup.yml
@@ -22,3 +22,11 @@
         state: present
       when:
         - not is_atomic
+
+    - name: disable fastest mirror detection
+      ini_file:
+        path: /etc/yum/pluginconf.d/fastestmirror.conf
+        section: main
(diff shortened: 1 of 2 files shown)


guits:

guits:

ktdreyer:
Did we report these at https://bugs.centos.org/ ?

guits:
Before reporting this at bugs.centos.org I would like to see if it has been actually fixed/improved in ceph-ansible.
In the case we can still see this failure, it's better anyway to have a 'static' environment using same constants in a CI context.

Issue #2666, opened 2018-05-30
yum repositories failures in CI


guits:
For a while, we are seeing too many yum repositories failure in the CI.

#2665 and #2652 have been created in the hope of fixing these failures but this is probably not enough.

This issue aims to track what has been done or is ongoing in order to fix these failures

djgalloway:
http://apt-mirror.front.sepia.ceph.com/epel7/ can be used for epel.  CentOS is still syncing.

ktdreyer:
Have we reported this to the Fedora / EPEL projects?

ErwanAliasr1:
@djgalloway will the resync will be automatic ?
Thanks for the quick response

ErwanAliasr1:
@ktdreyer we are not so sure about the real reason for now but it sounds like some mirrors might have quality issue. I'm not so sure that Fedora/Epel is handling the quality of every mirror. And that is not uncommon to have a local repo for better performance.

guits:
@djgalloway thanks!

djgalloway:
> will the resync will be automatic ?

After this initial sync is finished, I'll have the repos updated daily.

guits:
@djgalloway just curious, how will they be updated?

djgalloway:
Using `reposync` and `createrepo` for epel and a simple rsync for CentOS.

ErwanAliasr1:
@djgalloway it's important that rpms doesn't vanish during the rsync (which might take some time). How do you handle repo's integrity during the update ?

djgalloway:
It's a problem you're going to run into with any repo mirror AFAIK.  The best I can do is add `--delete-after` to the rsync so packages only get deleted all at once at the end of the rsync.

ErwanAliasr1:
That would be much better ! Thx.

djgalloway:
For EPEL, the repo is synced to a local dir then rsynced (locally) to the publicly-accessible directory so that should minimize problems.

The CentOS sync just finished as well: http://apt-mirror.front.sepia.ceph.com/centos/7.5.1804/os/x86_64/

guits:
we didn't see these failures for a while now, looks like we've improved things 👍 

special thanks to @djgalloway @ErwanAliasr1 for your more than quick reactivity, really appreciate!

PR #2685, opened 2018-06-01
ceph-defaults: Enable local epel repository

ErwanAliasr1:
During the tests, the remote epel repository is generating a lots of
errors leading to broken jobs (issue #2666)

This patch is about using a local repository instead of a random one.
To achieve that, we make a preliminary install of epel-release, remove
the metalink and enforce a baseurl to our local http mirror.

That should speed up the build process but also avoid the random errors
we face.

This patch is part of a patch series that tries to remove all possible yum failures.

### DIFF
--- a/tests/functional/setup.yml
+++ b/tests/functional/setup.yml
@@ -31,6 +31,34 @@
         value: 0
       when: ansible_distribution == 'CentOS'
 
+    - name: install epel
+      package:
+        name: epel-release
+        state: present
+      when:
+        - ansible_distribution == 'CentOS'
+        - not is_atomic
+
+    - name: enable local epel repository
+      ini_file:
+        path: /etc/yum.repos.d/epel.repo
+        section: epel
+        option: baseurl
+        value: http://apt-mirror.front.sepia.ceph.com/epel7/
+      when:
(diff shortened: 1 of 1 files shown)


tests/functional/setup.yml
```diff
@@ -31,6 +31,30 @@
...
+        - not is_atomic
+        - ansible_distribution == 'CentOS'
+
+    - name: enable local epel repository
+      ini_file:
+        path: /etc/yum.repos.d/epel.repo
+        section: epel
+        option: baseurl
+        value: http://apt-mirror.front.sepia.ceph.com/epel7/
+      when: ansible_distribution == 'CentOS'
```
guits: it's not needed to change this when atomic i think

tests/functional/setup.yml
```diff
@@ -31,6 +31,30 @@
...
+
+    - name: disable remote epel repository
+      ini_file:
+        path: /etc/yum.repos.d/epel.repo
+        section: epel
+        option: metalink
+        state: absent
+      when: ansible_distribution == 'CentOS'
```
guits: ditto

ErwanAliasr1:
@guits fixed

Issue #2697, opened 2018-06-05
Generate ceph.conf locally from ansible host


haphan:
We currently fetching keyrings from monitor using `ceph-fetch-keys`, it'd be useful to fetch `ceph.conf` and `ceph.d/*` configuration to ansible server node.

The use case at the moment is that `clients` group is unnecessary and can be easily replaced by a container which mounted a working `ceph.conf`.
@andrewschoen  @leseb

leseb:
The `ceph-fetch-keys` script gets Ceph's keys from the monitors and writes them on the filesystem.
Why should we fetch it when we can generate it?

Are you trying to simplify the client role?

haphan:
@leseb You are right. I think it is better  to generate `ceph.conf` locally. 

My goal is to enable ansible host to manage ceph cluster. Right now we already have the keyring, the only thing missing is `ceph.conf`

guits:
Fixed by #2700

PR #2700, opened 2018-06-05
Generate a copy of ceph.conf locally

haphan:
Refers to #2697 

This change creates a copy of `ceph.conf` in ansible server.

Room for improvement:  a toggle flag to control this behavior.

@leseb What do you think?

Signed-off-by: Ha Phan <thanhha.work@gmail.com>

### DIFF
--- a/roles/ceph-config/tasks/main.yml
+++ b/roles/ceph-config/tasks/main.yml
@@ -26,6 +26,30 @@
       - restart ceph rgws
       - restart ceph mgrs
       - restart ceph rbdmirrors
+
+  - name: "ensure fetch directory exists"
+    run_once: true
+    become: false
+    local_action:
+      module: file
+      path: "{{ fetch_directory }}/{{ fsid }}/etc/ceph"
+      state: directory
+      mode: "0755"
+    when:
+      - ceph_conf_local
+
+  - name: "generate {{ cluster }}.conf configuration file locally"
+    local_action: config_template
+    become: false
+    run_once: true
+    args:
+      src: "ceph.conf.j2"
+      dest: "{{ fetch_directory }}/{{ fsid }}/etc/ceph/{{ cluster }}.conf"
+      config_overrides: "{{ ceph_conf_overrides }}"
+      config_type: ini
+    when:
+      - inventory_hostname in groups[mon_group_name]
+      - ceph_conf_local
   when:
     - not containerized_deployment|bool
 
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -78,6 +78,8 @@ dummy:
 #ceph_rbdmirror_firewall_zone: public
 #ceph_iscsi_firewall_zone: public
 
+# Generate local ceph.conf in fetch directory
+#ceph_conf_local: false
 
 ############
 # PACKAGES #
--- a/group_vars/rhcs.yml.sample
+++ b/group_vars/rhcs.yml.sample
@@ -78,6 +78,8 @@ fetch_directory: ~/ceph-ansible-keys
 #ceph_rbdmirror_firewall_zone: public
 #ceph_iscsi_firewall_zone: public
 
+# Generate local ceph.conf in fetch directory
+#ceph_conf_local: false
 
 ############
 # PACKAGES #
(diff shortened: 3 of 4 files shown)


haphan:
done :)

roles/ceph-config/tasks/main.yml
```diff
@@ -34,6 +34,9 @@
       dest: "{{ fetch_directory }}/{{ fsid }}/etc/ceph/{{ cluster }}.conf"
       config_overrides: "{{ ceph_conf_overrides }}"
       config_type: ini
+    when:
+      - inventory_hostname in groups.get(mon_group_name, [])
```
guits: `inventory_hostname == groups[mon_group_name][0]`

guits:
if the `when` is limiting to the last mon only, the `run_once: true` is not needed then

@leseb @haphan

haphan:
done

guits:
just to clarify:

when limiting to a particular mon, you shouldn't use `run_once: true`

eg:

```
when:
  - inventory_hostname == groups[mon_group_name][-1:]
run_once: true
```

if the first node played is the first mon in `mon_group_name` this will make the task skipped and only run once, it means the task won't be ever played

the better approach if you want to use `run_once: true` and you don't need to care on which mon the task will be executed is this approach:

```
when:
  - inventory_hostname in groups[mon_group_name]
run_once: true
```

guits:
@haphan sorry for these multiple request changes, but would you please update your patch to fit the 2nd approach I've mentioned in my previous comment?
Thanks for your patience!

haphan:
@guits I've changed back to 2nd approach. 

@andymcc In my case, Ceph cluster is deployed from a container e.g. `haphan/ceph-installer`. This container role is similar to ansible server; containing this `ceph-ansible` playbook along with python dependencies, etc..

Now I want to extend this container so that it can manage the Ceph cluster like create pools, managing osds, etc.. And for that, besides keyrings which we already have locally, the container also needs access to `ceph.conf`

I am still experimenting this approach and until now it works for me. Do share your thoughts.

haphan:
Any chance to backport this feature to 3.0 branch?

guits:
@haphan this looks more like a feature/enhancement than a critical bug fixing, therefore, I tend to say we can't backport this to stable.

guits:

andymcc:
Does setting your container up as a client not work? Or you would rather not use the client role for some reason?
I'm guessing once the ceph.conf is in your fetch directory you will either have to move it or do some other action in order for it to be useful?

haphan:
@andymcc client role works perfectly fine except it assumes you dedicate a server for this purpose, and this does not scale well for multiple clusters.

What I really need is an on-demand client; only when I need to do manage ceph.

Once ceph.conf is in fetch directory, I can safely push them to git along with keyrings.

andymcc:
Ok I see, so the plan is to push it to git to use at some other point. Thanks for explaining the change to me.
The change seems fine, but I'd like to see a toggle variable that defaults to false.
"ceph_conf_local" or similar, that way it won't generate the conf locally for normal use cases but can be set to configure it when the variable is set to true.

guits:
@andymcc what's the concern by generating the file locally in any case?
we are already bound to the fetch directory, I would say it doesn't matter if you have the ceph.conf inside in addition?

guits:
@andymcc since it doesn't seem that getting rid of that fetch directory is something ongoing I wasn't really sure of the benefits of what you suggested, otherwise I agree with what you said

haphan:
Hmm looks like `{{ fsid }}/etc/ceph` is not created. 

Shall we add a task to ensure it is created?

leseb:
@haphan yes please, along with a variable `ceph_conf_local` just like @andymcc suggested.
Thanks.

haphan:
@leseb @guits Please review

haphan:
@guits What is the expected outcome of running `./generate_group_vars_sample.sh`?

```bash
[root@dev work]# ./generate_group_vars_sample.sh
[root@dev work]# echo $?
0
```

guits:
@haphan no output expected, you should see a diff using `git diff`

haphan:
Well, no new changes found. I was actually expecting new variable in `group_vars/all.yaml.sample` but it is not the case.

haphan:
@guits my bad. Ignore my previous comment. Pushing an update now.

haphan:
@guits Can you take a look and merge this PR?

PR #2693, opened 2018-06-04
ceph-common: move firewall checks after package installation

andrewschoen:
We need to do this because on dev or rhcs installs ceph_stable_release
is not mandatory and the firewall check tasks have a task that is
conditional based off the installed version of ceph. If we perform those
checks after package install then they will not fail on dev or rhcs
installs.

Signed-off-by: Andrew Schoen <aschoen@redhat.com>

PR #2716, opened 2018-06-05
tests: add a dummy value for 'dev' release

guits:
Functional tests are broken when testing against 'dev' release (ceph).
Adding a dummy value here will fix them.

Typical error:

```
>       if request.node.get_marker("from_luminous") and ceph_release_num[ceph_stable_release] < ceph_release_num['luminous']:
E       KeyError: 'dev'
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -25,7 +25,8 @@ def node(host, request):
       'jewel': 10,
       'kraken': 11,
       'luminous': 12,
-      'mimic': 13
+      'mimic': 13,
+      'dev': 99
     }
     if not request.node.get_marker(node_type) and not request.node.get_marker('all'):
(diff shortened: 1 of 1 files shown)


guits:
jenkins test dev-ansible2.4-centos7_cluster

andrewschoen:
I was working on this in #2693 already. I don't think we want to have the tests think that anything 'dev' is a bogus release. I think it's causing trouble in the playbook run as well according to the CI tests here. I also made a patch to ceph-build to set the release to nautilus when dev tests are run. https://github.com/ceph/ceph-build/pull/1040

I think in the context of PR testing we can always assume dev tests are going to be run against the latest ceph master. This would need to be updated each time there is a version bump in ceph master, which really isn't ideal, but I can't think of a quick and robust approach here. We gotta get our tests running against ceph master again.

PR #2723, opened 2018-06-06
client: use dummy created container when there is no mon in inventory

guits:
the `docker_exec_cmd` fact set in client role when there is no monitor
in inventory is wrong, `ceph-client-{{ hostname }}` is never created so
it will fail anyway.


leseb:
Failure is unrelated.

PR #2733, opened 2018-06-07
tests: increase ssh timeout and retries in ansible.cfg

andrewschoen:
We see quite a few failures in the CI related to testing nodes losing
ssh connection. This modification allows ansible to retry more times and
wait longer before timing out. This seems to really affect testing
scenarios that use a large amount of testing nodes. The centos7_cluster
scenario specifically has 12 nodes and suffered from these failures
often.


andrewschoen:
jenkins test dev-ansible2.4-centos7_cluster

andrewschoen:
The failure in ``dev-ansible2.4-centos7_cluster`` failed in the same was as https://github.com/ceph/ceph-ansible/pull/2693#issuecomment-395390088

No lose in ssh connection though, which is good news.

Issue #2709, opened 2018-06-05
re-run of the roles in the rolling_update.yml for each node type


TamizharasiS:
file-name: ceph-ansible/infrastructure-playbooks/rolling_update.yml

My concern here is, why do we have to run the roles again for each cluster component during the rolling_update. Shouldn't the rolling update be just a service stop, upgrade ceph packages and then start the service?

In the playbook, rolling_update.yml we have the roles to be run for each node type.
Example:

for ceph-osd

 roles:
    - ceph-defaults
    - { role: ceph-common, when: not containerized_deployment }
    - { role: ceph-docker-common, when: containerized_deployment }
    - ceph-config
    - ceph-osd

for ceph-mon

roles:
    - ceph-defaults
    - { role: ceph-common, when: not containerized_deployment }
    - { role: ceph-docker-common, when: containerized_deployment }
    - ceph-config
    - ceph-mon
	
Similarly for mgrs, rgws, and all other types of hosts.

These roles must have already been run on the specific nodes during the cluster setup.
And a rolling update is performed only on an established and configured cluster, 
isn't re-running the roles redundant and time consuming?
regenerating the config on every node, is it really required even for a minor upgrade?
Probably, it is required incase of a major upgrade??

Also, I have tested the rolling_update.yml on a test cluster to upgrade from
12.2.4 to 12.2.5 (luminous minor version upgrade) with roles removed. ie., a service stop, update packages and a service stop. And it worked perfectly and it was much faster too relatively.

 Please help me understand if am missing something 
Otherwise, we should enhance rolling_update.yml to not do redundant jobs.

leseb:
Although this makes sense, running the entire role for new daemons makes sense, e.g: the ceph-mgr and maybe more in the future. Of course, this could be added as an exception and rework the other pieces.
Running the following might just be enough:

```
roles:
- ceph-defaults
- { role: ceph-common, when: not containerized_deployment }
- { role: ceph-docker-common, when: containerized_deployment }
- ceph-config
```

We have to investigate that, but we have no bandwidth to provide some code at the moment, so PR welcome.
Thanks.

TamizharasiS:
Correct, however adding new daemons happens in a major release upgrade.

adding new things incase of a major release can be split into 2 steps,
1. add new daemons i.e., run only that specific tasks/roles
2. then run rolling_update(without roles) on all nodes.
Or
Probably, we should have a flag to decide if roles has to be run or not for each node type.

[...]

PR #2734, opened 2018-06-07
[skip ci] rolling_update: removes the re-run of mon,osd,rgw roles

TamizharasiS:
Fixes: #2709

Since a rolling update is performed on an already configured cluster,
re-run of these roles is actually not required.

Signed-off-by: TamizharasiS <t.shanmugam@flipkart.com>


infrastructure-playbooks/rolling_update.yml
```diff
@@ -115,13 +115,6 @@
       when:
         - not containerized_deployment
 
-  roles:
-    - ceph-defaults
```
andrewschoen: I think ``ceph-defaults``, ``ceph-common``, ``ceph-docker`` and ``ceph-config`` all need to be here so that the packages actually get upgraded. That package upgrade happens in ``ceph-common`` and ``ceph-docker``. ``ceph-config`` I'm unsure about because that's going to write a new config file, which might be needed depending on if there are config changes needed for the upgrade. Most likely it's harmless though if it's not needed for the upgrade.

andrewschoen:
jenkins test luminous-ansible2.4-update_cluster

infrastructure-playbooks/rolling_update.yml
andrewschoen: Everything should be skipped in ceph-common and ceph-docker that doesn't need to change, it's harmless. Those roles will upgrade and restart services as well while handling the logic for different package names, ceph version differences and os differences. I do think we can lose the ceph-$daemon roles though.

TamizharasiS: My concern is more of ceph-osd role, creating devices and partitions stuff. 
I will think about some simpler option to just upgrade the packages in a common way for all node types as required in a rolling update and do the testing and update the PR.
Thanks

andrewschoen: I understand that concern, but I don't think we need the ceph-osd role here. Removing those daemon specific roles might be the quickest path and should take care of your concern.

leseb:
jenkins test luminous-ansible2.4-update_cluster

infrastructure-playbooks/rolling_update.yml
TamizharasiS: Thanks for the insights. Am new to ceph/ceph-ansible and thanks for the insights.
I have updated the PR to remove the ceph-osd alone for now.
I have faced issues with raising PR too, then I realized I had to fork and can push only to that forked branch.
I have updated a revert commit, probably rebasing could have been better.
Anyway please have a look at the code change.

Also, is there any way that I can use to run the test cases myself to verify the code before pushing it to the repo?

leseb:
Please squash your commits and let us know when this is ready for review.

TamizharasiS:
I am facing issues while squashing the commits.

leseb:
@TamizharasiS next time you open a PR add `[skip ci]` in the title.

PR #2740, opened 2018-06-08
test: only on containerized iscsi

leseb:
We don't have the same service running on non-container for now, this
will change soon but for let's only run the test on container.


guits:

PR #2751, opened 2018-06-11
common: ability to enable/disable fw configuration

leseb:
Prior to this patch if you were running on a Red Hat system,
ceph-ansible would try to configure firewalld for you without the
operators's consent.
Now you can enable or disable the fw configuration by setting
configure_firewall to either true or false.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1589146


leseb:
backport here: https://github.com/ceph/ceph-ansible/pull/2756

Issue #2768, opened 2018-06-14
ceph-mgr does not restart


czunker:
In our setup, we experience the problem, that at the end of the site playbook, ceph-mgr does not restart.
We use a modified version of site.yml.sample. Because of crush map modifications and openstack config, we run the mon section multiple times. Because ceph-mon and ceph-mgr are colocated on the same servers, this leads to multiple restarts of ceph-mgr.

According to the systemd unit file, only 3 restarts in 30 minutes are allowed:
```
StartLimitInterval=30min
StartLimitBurst=3
```
And according to [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1016548), systemd does not differentiate between restart because of a failure or a `systemctl restart` execution.
On our special case we hit these limitations. ceph-ansible is just to fast.

Our workaround is this task at different positions in the playbook:
```
- name: Reset restart counter of ceph-mgr
  become: yes
  shell: "systemctl reset-failed ceph-mgr@{{ ansible_hostname }}"
  changed_when: false
```

Please don't get me wrong. I don't blame ceph-ansible for this behavior. When you think, this ticket should be closed, it is at least documented.

When you think this is worth a PR, I would be glad to contribute. Just point me to the correct position where it might fit.

leseb:
Sounds valid if we can avoid failure, where did you put this in the playbook? Do you mind pushing a PR so we can continue the discussion over it? Thanks!

czunker:
I will create a PR for site.yml.sample

PR #2787, opened 2018-06-19
reset failed count of ceph-mgr

czunker:
Depending on your setup, ceph-mgr might get restarted multiple times.
When this is done to fast, systemd will prevent further restarts because of
configured limits in the ceph-mgr systemd unit file.

Resetting the failure count will prevent this problem. The reset is done before
the restart so in case of a real problem during the restart it still fails.

Fixes: #2768


guits:
hi @czunker , could you please add a sign-off and squash your commits please? thanks!

czunker:
@guits Sorry, for that. Fixed it.

PR #2792, opened 2018-06-20
tests: fix broken test when collocated daemons scenarios

guits:
At the moment, a lot of tests are skipped when daemons are collocated.
Our tests consider a node belong to only 1 group while it's possible for
certain scenario it can belong to multiple groups.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -89,8 +89,6 @@ def node(host, request):
         num_devices = len(ansible_vars.get("devices", []))
     if not num_devices:
         num_devices = len(ansible_vars.get("lvm_volumes", []))
-    num_osd_hosts = len(ansible_vars["groups"]["osds"])
-    total_osds = num_devices * num_osd_hosts
     cluster_name = ansible_vars.get("cluster", "ceph")
     conf_path = "/etc/ceph/{}.conf".format(cluster_name)
     if "osds" in group_names:
@@ -116,8 +114,6 @@ def node(host, request):
         osd_ids=osd_ids,
         num_mons=num_mons,
         num_devices=num_devices,
-        num_osd_hosts=num_osd_hosts,
-        total_osds=total_osds,
         cluster_name=cluster_name,
         conf_path=conf_path,
         cluster_address=cluster_address,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -61,7 +61,7 @@ def node(host, request):
     if "nfss" in group_names and ceph_stable_release == "jewel":
         pytest.skip("nfs nodes can not be tested with ceph release jewel")
 
-    if "iscsigws" in group_names and ceph_stable_release == "jewel":
+    if group_names == ["iscsigws"] and ceph_stable_release == "jewel":
         pytest.skip("iscsigws nodes can not be tested with ceph release jewel")  # noqa E501
 
     if request.node.get_closest_marker("from_luminous") and ceph_release_num[ceph_stable_release] < ceph_release_num['luminous']:  # noqa E501
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,7 +18,7 @@ def node(host, request):
     # because testinfra does not collect and provide ansible config passed in
     # from using --extra-vars
     ceph_stable_release = os.environ.get("CEPH_STABLE_RELEASE", "luminous")
-    node_type = ansible_vars["group_names"][0]
+    group_names = ansible_vars["group_names"]
     docker = ansible_vars.get("docker")
     osd_auto_discovery = ansible_vars.get("osd_auto_discovery")
     lvm_scenario = ansible_vars.get("osd_scenario") == 'lvm'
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -18,7 +18,7 @@ def node(host, request):
     # because testinfra does not collect and provide ansible config passed in
     # from using --extra-vars
     ceph_stable_release = os.environ.get("CEPH_STABLE_RELEASE", "luminous")
-    node_type = ansible_vars["group_names"][0]
+    node_type = ansible_vars["group_names"]
```
andrewschoen: Just a small nit, but maybe this should be ``node_types`` now? Changing it to ``group_names`` might make even more sense actually.

guits: @andrewschoen true, that makes more sense. I've updated it

tests/conftest.py
```diff
@@ -29,41 +29,43 @@ def node(host, request):
...
         pytest.skip(
             "Not a valid test for containerized deployments or atomic hosts")
 
-    if request.node.get_marker("docker") and not docker:
+    if request.node.get_closest_marker("docker") and not docker:
         pytest.skip(
             "Not a valid test for non-containerized deployments or atomic hosts")  # noqa E501
 
-    if node_type == "mgrs" and ceph_stable_release == "jewel":
+    if group_names == "mgrs" and ceph_stable_release == "jewel":
```
andrewschoen: This needs to be ``if "mgrs" in group_names``. Same for ``nfss`` and ``iscsigws`` below.

guits: @andrewschoen updated, good catch, thanks

guits:
thanks @alfredodeza for helping here

tests/conftest.py
```diff
@@ -29,41 +29,53 @@ def node(host, request):
...
 
-    if request.node.get_marker("no_lvm_scenario") and lvm_scenario:
+    # capture the initial/default state
+    valid_markers = []
+    for marker in request.node.iter_markers():
+        if marker.name not in group_names and marker.name != 'all':
+            valid_markers.append(False)
+        else:
+            valid_markers.append(True)
+        # Check if any of the markers. This is tricky to do with all the negations, let me think a bit here
```
andrewschoen: This comment looks like it could be removed or reworded. I think we want to say something like:

"Check if any markers on the test method exist in the nodes group_names. If they do not, this test is not valid for the node being tested."

tests/conftest.py
```diff
@@ -29,41 +29,53 @@ def node(host, request):
         'mimic': 13,
         'dev': 99
     }
-    if not request.node.get_marker(node_type) and not request.node.get_marker('all'):  # noqa E501
-        pytest.skip("Not a valid test for node type: %s" % node_type)
 
-    if request.node.get_marker("no_lvm_scenario") and lvm_scenario:
+    # capture the initial/default state
+    valid_markers = []
```
andrewschoen: I know why you made this a list, but if we treated it like a boolean I think the logic would be easier to follow.

Why don't we rename this to something like ``run_test_on_node`` or ``test_is_applicable`` and default it to ``False``. Then below we just assign a ``True`` value to it if we find a marker that exists in group_names or if the marker is "all". This way we could avoid ``any()`` and just do something like ``if not run_test_on_node:`` on line 43 below.

tests/conftest.py
```diff
@@ -29,41 +29,51 @@ def node(host, request):
...
     }
-    if not request.node.get_marker(node_type) and not request.node.get_marker('all'):  # noqa E501
-        pytest.skip("Not a valid test for node type: %s" % node_type)
 
-    if request.node.get_marker("no_lvm_scenario") and lvm_scenario:
+    # capture the initial/default state
+    test_is_applicable = False
+    for marker in request.node.iter_markers():
+        if marker.name not in group_names and marker.name != 'all':
+            test_is_applicable = False
```
alfredodeza: you can't have this line here, because you could be reverting a `test_is_applicable` being `True`. Since this is a loop you might get True, True, False

tests/conftest.py
```diff
@@ -29,41 +29,49 @@ def node(host, request):
...
         'dev': 99
     }
-    if not request.node.get_marker(node_type) and not request.node.get_marker('all'):  # noqa E501
-        pytest.skip("Not a valid test for node type: %s" % node_type)
 
-    if request.node.get_marker("no_lvm_scenario") and lvm_scenario:
+    # capture the initial/default state
+    test_is_applicable = False
+    for marker in request.node.iter_markers():
+        if marker.name not in group_names and marker.name != 'all':
```
andrewschoen: I think the logic needs to be reversed here now right? Something like:

``if marker.name in group_names or marker.name == 'all':``

tests/conftest.py
```diff
@@ -29,41 +29,49 @@ def node(host, request):
...
         'dev': 99
     }
-    if not request.node.get_marker(node_type) and not request.node.get_marker('all'):  # noqa E501
-        pytest.skip("Not a valid test for node type: %s" % node_type)
 
-    if request.node.get_marker("no_lvm_scenario") and lvm_scenario:
+    # capture the initial/default state
+    test_is_applicable = False
+    for marker in request.node.iter_markers():
+        if marker.name in group_names and marker.name != 'all':
```
andrewschoen: Almost, you need: ``if marker.name in group_names or marker.name == 'all'``

tests/conftest.py
```diff
@@ -29,41 +29,49 @@ def node(host, request):
...
     }
-    if not request.node.get_marker(node_type) and not request.node.get_marker('all'):  # noqa E501
-        pytest.skip("Not a valid test for node type: %s" % node_type)
 
-    if request.node.get_marker("no_lvm_scenario") and lvm_scenario:
+    # capture the initial/default state
+    test_is_applicable = False
+    for marker in request.node.iter_markers():
+        if marker.name in group_names or marker.name == 'all':
+            test_is_applicable = True
```
guits: I think we must break this loop as soon as we enter in this condition, otherwise, we can reset it to `False`.
I guess this is kind of what @alfredodeza said [here](https://github.com/ceph/ceph-ansible/pull/2792#discussion_r197156568)

alfredodeza: Breaking out of the loop isn't really necessary, as this assignment is not going to change it back to `False`. If you really want to break out of it though, that is fine. It wouldn't be any different

guits:
jenkins test pipeline

alfredodeza:
mergify?

PR #2799, opened 2018-06-21
Fix RHEL based Ansible installation

mykaul:
It is now on its own channel, not extras.


guits:
@mykaul could you please add a sign-off in your commit?
thanks!

mykaul:
Will add SoB line, though I was hoping that there would be a commit hook for it...

PR #2801, opened 2018-06-25
Add keyring_permissions var  to set permissions for keyring files

amarao:
Add keyring_permissions variable to control permissions for keyring
files in /etc/ceph. Default value is the same as it was (0600),
but this variable allows user to override it (f.e. set it to 0640).

Rationale: We are configuring a monitoring for Ceph. It runs under a separate user (telegraf), but this user is included into the ceph group. To have access to the admin keyring it need to have `/etc/ceph/client.admin.keyring` to be at least `rw-r-----` (0640). I added a variable into defaults with the same value as current value in the task, but allow to override it through variable in the inventory/group_vars/host_vars.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,10 @@
+rules:
+  default:
+    protection:
+      required_status_checks:
+        strict: true
+        contexts:
+          - DCO
+          - "Testing: ceph-ansible PR Pipeline"
+      required_pull_request_reviews:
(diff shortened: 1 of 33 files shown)


guits:
jenkins test pipeline

amarao:
I apologie for the mess. I've tried to rebase and screwed. I'll open a new pull request.

PR #2809, opened 2018-06-26
Add ceph_keyring_permissions var to set permissions for keyring files

amarao:
Add `keyring_permissions` variable to control permissions for keyring
files in `/etc/ceph`. Default value is the same as it was (0600),
but this variable allows user to override it (f.e. set it to 0640).

Rationale: We are configuring a monitoring for Ceph. It runs under
a separate user (telegraf), but this user is included into the ceph
group. To have access to the admin keyring it need to have
/etc/ceph/client.admin.keyring to be at least rw-r----- (0640).
I added a variable into ceph-defaults with the same value as current
value in all related tasks.  This will allow to override it through variable
in the inventory/group_vars/host_vars.

### DIFF
--- a/group_vars/rhcs.yml.sample
+++ b/group_vars/rhcs.yml.sample
@@ -296,6 +296,9 @@ ceph_repository: rhcs
 
 #ceph_conf_key_directory: /etc/ceph
 
+# Permissions for keyring files in /etc/ceph
+#ceph_keyring_permissions: '0600'
+
 #cephx: true
 
 ## Client options
--- a/roles/ceph-defaults/defaults/main.yml
+++ b/roles/ceph-defaults/defaults/main.yml
@@ -288,6 +288,9 @@ generate_fsid: true
 
 ceph_conf_key_directory: /etc/ceph
 
+# Permissions for keyring files in /etc/ceph
+ceph_keyring_permissions: '0600'
+
 cephx: true
 
 ## Client options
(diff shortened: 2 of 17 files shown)


roles/ceph-defaults/defaults/main.yml
```diff
@@ -288,6 +288,9 @@ generate_fsid: true
 
 ceph_conf_key_directory: /etc/ceph
 
+# Permissions for keyring files in /etc/ceph
+keyring_permissions: '0640'
```
leseb: Do you mind prefixing with `ceph_`? Thanks.

group_vars/rhcs.yml.sample
```diff
@@ -296,6 +296,9 @@ ceph_repository: rhcs
 
 #ceph_conf_key_directory: /etc/ceph
 
+# Permissions for keyring files in /etc/ceph
+#keyring_permissions: '0640'
```
andymcc: Default should be 0600 to ensure no change from current setup.

andymcc:
Thanks for changing it up - looks better this way.
2 minor changes and then lgtm!

group_vars/rhcs.yml.sample
amarao: Oh! Sorry, slipped from my tests. I'll set it to `0600`.

roles/ceph-defaults/defaults/main.yml
amarao: done.

leseb:
@andymcc lgty?

amarao:
Thank you for advice! I've learned a lot.

PR #2815, opened 2018-06-27
mon: honour mon_docker_net_host option

leseb:
--net=host was hardcoded in the startup line so even though
mon_docker_net_host was set to False the net option would always be
activated.
mon_docker_net_host is set to True by default so this commit does not
change the behaviour.


leseb:
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1595741

PR #2829, opened 2018-06-29
App pool

leseb:


roles/ceph-mds/tasks/create_mds_filesystems.yml
```diff
@@ -20,6 +20,17 @@
   when:
     - check_existing_cephfs.rc != 0
 
+- name: assign application to cephfs pools
+  command: "{{ hostvars[groups[mon_group_name][0]]['docker_exec_cmd'] | default('') }} ceph --cluster {{ cluster }} osd pool application enable {{ item }} {{ cephfs }}"
```
andymcc: The only thing I can see that could be changed is setting the application pool to "cephfs_application" - if for some reason the  application pool was to be different to the filesystem name, you wouldn't be able to do that. Although I can't really think of a usecase for that so happy for this to merge.

PR #2830, opened 2018-06-29
ceph-client: do not kill the dummy container

leseb:
The container runs for 300 sec, then dies and removes itself thanks to
the '--rm' option, so there is no point of removing it. Also this is
causing failure under some circonstances.

Closing: https://bugzilla.redhat.com/show_bug.cgi?id=1568157


leseb:
jenkins test pipeline

leseb:
jenkins test pipeline

leseb:
jenkins test pipeline

PR #2839, opened 2018-07-03
Remove zone from zonegroup before deleting zone in multisite

smanjara:
Signed-off-by: Shilpa Jagannath <smanjara@redhat.com>

### DIFF
--- a/roles/ceph-rgw/tasks/multisite/destroy.yml
+++ b/roles/ceph-rgw/tasks/multisite/destroy.yml
@@ -4,6 +4,11 @@
   run_once: true
   failed_when: false
 
+- name: remove zone from zonegroup
+  command: radosgw-admin zonegroup remove --rgw-zonegroup={{ rgw_zonegroup }} --rgw-zone={{ rgw_zone }}
+  run_once: true
+  failed_when: false
+
 - name: delete the zone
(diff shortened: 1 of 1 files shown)


roles/ceph-rgw/tasks/multisite/destroy.yml
```diff
@@ -4,6 +4,11 @@
   run_once: true
   failed_when: false
 
+- name: remove zone from zonegroup
+  command: radosgw-admin zonegroup remove --rgw-zonegroup={{ rgw_zonegroup }} --rgw-zone={{ rgw_zone }}
+  run_once: true
+  failed_when: false
```
leseb: why?

smanjara: When you delete a zone without removing from zonegroup, the perios update --commit would fail since that command needs to load the zone and zonegroup to be able to update the master. Period update would fail with an error like this:

radosgw-admin period update --commit
-1 Cannot find zone id= (name=), switching to local zonegroup configuration
-1 Cannot find zone id= (name=)

leseb: Ok, please add this explanation into your commit message. Thanks

smanjara: Reworded. Thanks

PR #2852, opened 2018-07-05
fix ansible_devices for old cciss driver-based devices

eb4x:
The ansible_devices lists out cciss driver-based devices as cciss!c0d0,
which tends to go wrong later in the playbooks when trying to do
something to /dev/cciss!c0d0. Just replace the '!' with a '/' and we're
good to go.

Signed-off-by: Erik Berg <openstack@slipsprogrammor.no>

### DIFF
--- a/roles/ceph-config/templates/ceph.conf.j2
+++ b/roles/ceph-config/templates/ceph.conf.j2
@@ -75,6 +75,7 @@ mon host = {% if nb_mon > 0 %}
 {% if containerized_deployment %}
 fsid = {{ fsid }}
 log file = /dev/null
+mon cluster log file = /dev/null
 mon host = {% if nb_mon > 0 %}
 {% for host in groups[mon_group_name] -%}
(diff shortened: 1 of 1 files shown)


eb4x:
Apologies for the messy PRs, not really well-versed in git.

leseb:
@eb4x no problem.

eb4x:
So... this probably needs more thought to it. Because in this place we actually trim out devices from ansible_devices that contain partitions to avoid installing on say your boot-disk. ansible_devices "cciss!c0d0" shows no partitions listed, yet it is partitioned. And luckily there's a check later that warns/bails on installing on a disk with partitions.

I can look at fixing that in ansible itself perhaps? Or do some weird additional checks in the code to see if a cciss device is actually partitioned and trim it out from devices. And/or in the mean-time just add a 

when:
  - item.value is match('cciss') 

or something similar and just completely disregard these pesky old devices.

Issue #2867, opened 2018-07-10
creating keys for client is unnecessarily tied to copy_admin_key


no2a:
In roles/ceph-client/tasks/create_users_keys.yml,  import_key parameter for ceph_key is determined by copy_admin_key. So, if copy_admin_key is false, the generated keys will not actually be registered to the ceph cluster, and vice versa. But that is odd since there is no relation between copying an admin key to hosts in the cluster and to registering client keys to the cluster. import_key should be always true, or a new variable should be introduced for import_key.
```
- name: create cephx key(s)
  ceph_key:
    state: present
    name: "{{ item.name }}"
    caps: "{{ item.caps }}"
    secret: "{{ item.key | default('') }}"
    containerized: "{{ docker_exec_cmd | default('') }}"
    cluster: "{{ cluster }}"
    dest: "{{ ceph_conf_key_directory }}"
    import_key: "{{ copy_admin_key }}"    # <== THIS
    mode: "{{ item.mode|default(omit) }}"
  with_items: "{{ keys }}"
  delegate_to: "{{ delegated_node }}"
  when:
    - cephx
    - keys | length > 0
    - inventory_hostname == groups.get(client_group_name) | first
```

guits:
@no2a since you need the admin key on the node from which you are supposed to import your created keys, linking `import_key` to `copy_admin_key` is a way to ensure you will be able to do it.
(`delegated_node` is not always a mon node.)

no2a:
In my case, delegated_node is a mon node and I don't want to expose an admin key to client nodes (I believe it is a not-so-uncommon situation). But ceph-ansible cannot import client keys without unnecessarily storing the admin key on client hosts.

Although I can set copy_admin_key to false and import the generated keys by hand of cource, it would be nice if I could do it by just running ceph-ansible.

no2a:
So, is there something wrong with setting import_keys to true always? If a deployer sets copy_admin_key to false and delegated_node has no admin key, then the task fails. It is a natural behavior. The deployer can recognize that he should have copied the admin key in his deployment.

guits:
@no2a I get what you are reporting here.

I don't see why it should be wrong with always setting `import_keys` to true since it's already its default value.

In any case I think it's worth to have a dedicated variable for this, I'll send a patch.

Thanks

guits:
@no2a Actually I think there is already a variable for this, it looks more like something I've missed when I've written this part of the code:
[...]

PR #2874, opened 2018-07-11
client: do not rely on copy_admin_key to import keys

guits:
Relying on `copy_admin_key` to import created keys on client nodes makes
us obliged to copy admin key on those nodes which is not something we might
want.
We should use the fact `condition_copy_admin_key` which will be set to
`True` when the delegated node is a mon which means we can import keys
without taking care of admin keyring.

Fixes: #2867


guits:
jenkins test pipeline

guits:
jenkins test pipeline

Issue #2843, opened 2018-07-04
generage ceph configuration file failed on ubuntu


like-inspur:
I install ceph cluster successfully on centos with version 3.0.36, but failed on ubuntu, error is this:
TASK [ceph-config : generate ceph configuration file: ceph.conf] ********************************************************************************
Wednesday 04 July 2018  10:32:07 +0800 (0:00:00.494)       0:00:53.794 ******** 
fatal: [openstack-202]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (copy) module: original_basename Supported parameters include: _original_basename, attributes, backup, checksum, content, delimiter, dest, directory_mode, follow, force, group, local_follow, mode, owner, regexp, remote_src, selevel, serole, setype, seuser, src, unsafe_writes, validate"}

I check installation configuration is no problem, so what is the reason? Thank you! @hunter @gbraad @jd @dang @duritong

vishalkanaujia:
What is the ansible playbook version?
`$ ansible-playbook --version`

like-inspur:
2.6.0, is it too high?

vishalkanaujia:
yes, I guess so. The ceph-ansible requirements say `ansible~=2.4,<2.5`.

Could you please try using `ansible-playbook 2.4`?

jd:
Why the hell do you mention me @like-inspur? 🤨

leseb:
@andymcc has this been reported already? Thanks.

andymcc:
I believe this is fixed in master already: https://github.com/ceph/ceph-ansible/commit/eb836e7c3101aa89bb64f1515605cf717fd3d9d2 - although I only tested against ansible 2.5.0 and not ansible 2.6.0.

On a side note I wouldn't recommend using 2.5.0+ on stable 3.0 branch.

andymcc:
yeah its broken on 2.6
I'll try fix this up.

leseb:
@andymcc thanks, looking forward to it.

kennethredler:
@andymcc will this be ported to `stable-3.1` branch, by chance?

guits:
@kennethredler I don't think stable-3.1 is intended to run against ansible 2.6.

kennethredler:
😿

PR #2878, opened 2018-07-12
Sync config_template with upstream for Ansible 2.6

andymcc:
The original_basename option in the copy module changed to be
_original_basename in Ansible 2.6+, this PR resyncs the config_template
module to allow this to work with both Ansible 2.6+ and before.

Additionally, this PR removes the _v1_config_template.py file, since
ceph-ansible no longer supports versions of Ansible before version 2,
and so we shouldn't continue to carry that code.

Closes: #2843


andymcc:
One point of note - this will fix the config_template module to work with Ansible 2.6, but there may be other issues with Ansible 2.6 that are unrelated to config_template and aren't dealt with by this PR.

M0les:
So I notice there's a clone of `plugins/actions/_v[12]_config_template.py` inside `roles/ceph-common` and more notably, this did not get your Ansbile 2.6 fixes.

I'm still pretty low-down on the Ansible+Ceph-Ansible+Ceph learning curve, but I'm wondering if the fix should be:

1. Copy the patches into the clones (easy, but possibly kicking some technical debt down the road)
1. Work-out how to remove the `roles/ceph-common` copies of these files.

guits:
@M0les ceph-common used to be the role where the configuration were generated before we create the ceph-config role, my assumption is that from ansible galaxy point of view, it was needed to provide the plugin in ceph-common. So it's probably a legacy and I see 2 options:

1/ it should be totally removed because it's legacy,
2/ or it should probably be moved to ceph-config role + synced with this patch

@leseb could you clarify?

andymcc:
We should be able to remove those - the module path points to plugins/actions for action_plugins, so unless you are running with an ansible.cfg that points to the one in ceph-common it shouldn't use that plugin - additionally the ceph-common role doesn't have any config_template tasks (which lines up with what you said above @guits)
In my opinion we should just remove them because they aren't being used and are just creating confusion. (I've put a PR up to do that #2932)

guits:
@andymcc I'd like to get @leseb confirmation regarding ansible galaxy use case before merging #2932

andymcc:
@guits yeah definitely.

PR #2897, opened 2018-07-18
validate: only run osd test on osd node

leseb:
Do not run device validation on every hosts, only on OSD nodes.

Signed-off-by: Sébastien Han <seb@redhat.com>

PR #2902, opened 2018-07-18
ceph-validate: the lvm osd scenario does not need to check devices

andrewschoen:
This fixes a bug introduced by cf01e596b that was causing ceph-volume CI test to fail.


andrewschoen:
@guits That works, I'll close this. I didn't realize the fix was also in #2897

Issue #2901, opened 2018-07-18
test with ansible 2.6


ktdreyer:
Currently we test upstream in Jenkins with Ansible 2.4 and 2.5, as discussed in https://github.com/ceph/ceph-ansible/issues/2470

Please add testing for Ansible 2.6 as well as 2.4 and 2.5.

Ansible 2.4 is currently end-of-life, but it's the only one I know is well-tested with the RH Ceph Storage product.

Ansible 2.5 will go end-of-life after Ansible 2.7 is released (https://access.redhat.com/support/policy/updates/ansible-engine)

It would be great to use Ansible 2.6 in RH Ceph Storage 3.2.

guits:
as per #2908, it looks like master passes testing against ansible 2.6

ktdreyer:
That's great news, thanks @guits !

guits:
> Currently we test upstream in Jenkins with Ansible 2.4 and 2.5, as discussed in #2470
> 
> Please add testing for Ansible 2.6 as well as 2.4 and 2.5.
> 
> Ansible 2.4 is currently end-of-life, but it's the only one I know is well-tested with the RH Ceph Storage product.
> 
> Ansible 2.5 will go end-of-life after Ansible 2.7 is released (https://access.redhat.com/support/policy/updates/ansible-engine)
> 
> It would be great to use Ansible 2.6 in RH Ceph Storage 3.2.

@ktdreyer we've been asked to test ceph-ansible stable-3.2 against ansible 2.6, from what I'm seeing today, correct me if I'm wrong but it seems RHCS 3.2 ships ansible 2.4 downstream?

ktdreyer:
RH Ceph Storage 3.2 should be compatible with Ansible 2.6 because that is what we document for our customers to use ([rhbz#1613941](https://bugzilla.redhat.com/show_bug.cgi?id=1613941))

guits:
@ktdreyer from what I can see, RHCS 3.2 ships with ansible 2.4 though

ktdreyer:
Right, unfortunately there have been several changes in the way that the Ansible team preferred to have layered products consume their builds:

1.  Originally Ansible was [shipped](https://access.redhat.com/errata/RHEA-2017:2366) in the RHEL 7 Extras repository.
2. Later, the team [pulled the build](https://access.redhat.com/articles/3359651) out of RHEL 7 Extras.
[...]

PR #2908, opened 2018-07-19
tests: test master against ansible 2.6

guits:
Ansible 2.4 is currently end-of-life.
Ansible 2.5 will go end-of-life after Ansible 2.7 is released.

Fixes: #2901

### DIFF
--- a/roles/ceph-validate/tasks/check_system.yml
+++ b/roles/ceph-validate/tasks/check_system.yml
@@ -67,10 +67,10 @@
 
 - name: fail on unsupported ansible version
   fail:
-    msg: "Ansible version must be between 2.4.x and 2.5.x!"
+    msg: "Ansible version must be between 2.5.x and 2.6.x!"
   when:
     - ansible_version.major|int == 2
-    - (ansible_version.minor|int < 4 or ansible_version.minor|int > 5)
+    - (ansible_version.minor|int < 5 or ansible_version.minor|int > 6)
 
 - name: fail if systemd is not present
   fail:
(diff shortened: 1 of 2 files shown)


guits:
jenkins test pipeline

ktdreyer:
I've emailed dev@lists.rdoproject.org to find out when RDO will move to Ansible 2.6.

alfredodeza:
The commit here started breaking everything for ceph-volume tests: af82e7523

I understand that the ceph-ansible team wants to test against Ansible 2.6, but this is also making it impossible to use 2.4.x

Moreover, this non-backwards compatible change went unannounced anywhere I could look (I am subscribed to the ceph-ansible mailing list). There is still no changelog in the ceph-ansible project which also makes it tremendously complicated to see what things are no longer supported or planned to be deprecated.

There was no deprecation warning here, which means that in order to find out why ceph-volume tests are breaking (and all the other projects that depend on ceph-ansible) we had to find the exact task that failed, check out the project, do a git blame on the line(s) that seem to be at fault, and then go into Github to see if there is a PR associated with it.

Please consider a changelog file, adding deprecation warnings if possible, and announcing non-backwards compatible changes like this one to as many places as possible (ceph-ansible mailing list only would be fine)

alfredodeza:
@guits where can we find information about 2.4 being EOL ? The Ansible docs say that 2.4 is still supported (versions 2.2 onward are still supported). The latest EOL version is 2.1

https://docs.ansible.com/ansible/2.4/release_and_maintenance.html

This change has only been enforced for testing, which is not what the ceph-ansible project says it supports in its requirements.txt:

https://github.com/ceph/ceph-ansible/blob/master/requirements.txt

    ansible~=2.4,<2.5

ktdreyer:
Here's the lifecycle table that shows Ansible 2.4 as End-of-Life: https://access.redhat.com/support/policy/updates/ansible-engine

PR #2928, opened 2018-07-24
Run creation of empty rados index object to first monitor

gfidente:
When distributing ceph-nfs role, creation of rados index object
fails as it assumes availability of client.admin locally.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1607970


guits:
failure here is unrelated, as usual it's the race condition ...

PR #2932, opened 2018-07-25
Remove unrequired config_template from ceph-common

andymcc:
config_template tasks are not called in ceph-common, and the
config_template module itself lives in the plugins/actions directory of
the base ceph-ansible repository itself - this is referenced in the
ansible.cfg that is setup with ceph-ansible.

The config_template in the ceph-common role was no longer being used and
was not referenced in the action_plugins path so we should remove it to
reduce confusion.


guits:
Not covered by the CI.
I'm not sure it's worth to trigger the CI for this

guits:
looks good to me.
however I'd like @leseb to confirm we can merge this from ansible galaxy point of view (see https://github.com/ceph/ceph-ansible/pull/2878#issuecomment-407649946)

leseb:
If we drop these from ceph-common, there is no way we can provide them through the ansible-galaxy. @andymcc could it be the right time to consider packaging config_template in pip so we can put in our requirements.txt file accordingly?

andymcc:
I like that idea - @cloudnull @evrardjp - thoughts? That also gets around the constant need to sync it and makes it more consumable.

guits:
@leseb does that mean we should move this into ceph-config role until it gets packaged in pip ?

leseb:
@andymcc yup, eliminating the constant resync is a win, also, we will have more control in the requirements.txt where we would pin particular versions depending on which ansible version we are running on.

@guits yes in the meantime I think we should.

andymcc:
@leseb - agreed on the benefits, that said does anybody really use the galaxy roles? They are pretty outdated and aren't being synced up from what i can see - it seems like if you are using that approach you are not getting any updates. I still like the idea of moving to a pip install though!

leseb:
@andymcc the galaxy roles are relatively up-to-date since they have `v3.0.33`. I have no idea who's using them, to be honest, but I guess it's cool to keep this around.

andymcc:
I'll change the PR - we should probably still sync the files up in that case, otherwise you'll get different results depending on whether you use galaxy or not

leseb:
@andymcc Thanks!

leseb:
jenkins test pipeline

andymcc:
@guits - honestly I'm not too sure how the galaxy bits work - I don't see how removing the files causes it to fail so I'm not 100% sure. I had thought you needed the directory to be "action_plugins": https://docs.ansible.com/ansible/2.5/plugins/action.html#enabling-action-plugins - which would suggest a dir called plugins/actions wouldn't work without adjusting ansible.cfg - but if removing it caused a failure then it obviously does use it.

Happy to move it, but again - not 100% sure what the outcome will be.

guits:
@andymcc just like you, I'm not really familiar with galaxy stuff, I was just thinking that if we must keep this from a galaxy point of view, it would be better to have it in ceph-config instead of ceph-common.

leseb:
@andymcc if we remove the module from one of the roles then there is no way we provide the module when fetching the role from the Galaxy. We cannot call `config_template` if there is no module. That's why I suggested to have it in pip so we could pull it from anywhere.

You might be right about the "wrong" directory though, but that's a different issue :)

andymcc:
Ahh I see, so it's just a place to carry the module, you would still have to do some manual bits to use the module? Let me prod @cloudnull and crew again to see their thoughts.

leseb:
@andymcc thanks :)

evrardjp:
The idea of moving config_template outside the openstack-ansible-plugins was to ease the consumption of config_template as a role, that only carry the relevant files for the plugin with no tasks, and no extra plugins.

A role "x" requiring the config_template action plugin can simply list the ansible-config_template role in its meta/main.yml in dependencies, which would load the ansible-config_template role's content (having no tasks/defaults, no task will get executed) and therefore enabling the plugins shipped in config_template to be consumed by subsequent tasks in the same playbook calling the role "x".

For users, it means as long as they have the ansible-config_template role in their roles folder (that depends on their configuration), it will be able to be picked up by roles requiring it. (You can name the role whatever you want, as long as you are consistent with the meta/main.yml dependencies).

If you have a mechanism to fetch roles to your roles folder, like ansible-galaxy, you can reuse that mechanism to still fetch config_template in the right location.

leseb:
@evrardjp that makes perfect sense, thank you. It is handy upstream but for downstream product this is much more complex since there is nothing to package... I guess we should keep on syncing until we can get a pip package. Thanks!

PR #2944, opened 2018-07-26
ceph ansible 3.1 igw: fix rbd-target-gw startup

mikechristie:
The problem is rbd-target-gw needs the rbd pool to be created, keyring to be copied over, and the iscsi-gateway.cfg to be setup before starting the rbd-target-gw service.

In the master branch this is fixed by this commit:
    
        commit 91bf53ee932a6748c464bea762f8fb6f07f11347
        Author: Sébastien Han <seb@redhat.com>
        Date:   Fri Mar 23 11:24:56 2018 +0800
    
            ceph-iscsi: support for containerize deployment
    
where the needed setup tasks are done in common.yml which is done before prerequisites.yml.
    
To avoid porting all those changes to 3.1 this patch just moves the rbd-target-gw startup to configure_iscsi.yml after everything has been setup.

This fixes red hat bz:
 
https://bugzilla.redhat.com/show_bug.cgi?id=1601325


ktdreyer:
Mike, can you edit this pull request in GitHub to be targeted to "stable-3.1" instead of "master"?

ktdreyer:
Oh, I was able to do it, here.

PR #2749, opened 2018-06-11
rolling_update: remove the re-run of just the ceph-osd role

TamizharasiS:
Fixes: #2709

Since a rolling update is performed on an already configured cluster,
re-run of this roles is actually not required.

Signed-off-by: TamizharasiS <t.shanmugam@flipkart.com>

PR #2947, opened 2018-07-27
add ceph-handler role

leseb:
The role contains all the handlers for Ceph services. We decided to
leave ceph-defaults role with variables and a few facts only. This is
useful when organizing the site.yml files and also adding the known
variables to infrastructure-playbooks.


leseb:
As a followup to once this one merges, we need to resurrect this https://github.com/ceph/ceph-ansible/pull/2749 so we don't play the daemon roles during the upgrade.

guits:
jenkins test pipeline

PR #2954, opened 2018-07-30
[skip ci] config: enforce socket name

leseb:
This was introduced by
https://github.com/ceph/ceph/commit/59ee2e8d3b14511e8d07ef8325ac8ca96e051784
and made our socket checks impossible to run. The PID could be found,
but the cctid cannot.
This happens during upgrade to mimic and on cluster running on mimic.

So let's force the admin socket the way it was so we can properly check
for existing instances also the line $cluster-$name.$pid.$cctid.asok
is only needed when running multiple instances of the same daemon,
thing ceph-ansible cannot do at the time of writing

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1610220

### DIFF
--- a/roles/ceph-config/templates/ceph.conf.j2
+++ b/roles/ceph-config/templates/ceph.conf.j2
@@ -17,6 +17,12 @@ fsid = {{ fsid }}
 {% if common_single_host_mode is defined and common_single_host_mode %}
 osd crush chooseleaf type = 0
 {% endif %}
+{% if ceph_version not in ['jewel', 'kraken', 'luminous'] and containerized_deployment %}
+# let's force the admin socket the way it was so we can properly check for existing instances
+# also the line $cluster-$name.$pid.$cctid.asok is only needed when running multiple instances
+# of the same daemon, thing ceph-ansible cannot do at the time of writing
+admin socket = "$run_dir/$cluster-$name.asok"
+{% endif %}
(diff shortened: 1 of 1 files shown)


roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -17,6 +17,12 @@ fsid = {{ fsid }}
 {% if common_single_host_mode is defined and common_single_host_mode %}
 osd crush chooseleaf type = 0
 {% endif %}
+{% if ceph_version not in ['jewel', 'kraken', 'luminous'] and containerized_deployment %}
```
leseb: note for myself, perhaps it's better to do this for each daemon that is affected, eg: rgw and rbd-mirror

guits:
jenkins test luminous-update_docker_cluster

roles/ceph-config/templates/ceph.conf.j2
leseb: Actually no because this will be overridden by https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/templates/ceph.conf.j2#L128-L132
This is good to go!

leseb:
screen shot 2018-07-31 at 10 55 38

We passed CI, but I just amended my commit. So merging without new run of the CI.

PR #2959, opened 2018-07-31
defaults: backward compatibility with fqdn deployments

guits:
This commit ensures we are backward compatible with fqdn deployments.
Since ceph-container enforces deployment to be done with shortname, we
must keep backward compatibility with clusters already deployed with
fqdn configuration


guits:
jenkins test pipeline

PR #2960, opened 2018-07-31
tox: test upgrades to luminous only on stable-3.0

leseb:
stable-3.0 must not attempt to run an upgrade from Luminous to Mimic, so
if running on stable-3.0 we should only do:

* Jewel to Jewel minor
* Jewel to Luminous
* Luminous to Luminous minor


leseb:
jenkins test luminous-update_docker_cluster

leseb:
Running the pipeline is not necessary here.

PR #2974, opened 2018-08-07
docs: overall syntax improvements

berendt:
* add some missing dots and ``
* add/remove line breaks
* consistent use of shell prompt in consoles outpus
* fix block indents Bearbeiten
* use code blocks

Signed-off-by: Christian Berendt <berendt@b1-systems.de>


leseb:
@berendt thank you, really valuable contribution.

PR #2995, opened 2018-08-09
osd: generate device list for osd_auto_discovery on rolling_update

leseb:
rolling_update relies on the list of devices when performing the restart
of the OSDs. The task that is builind the devices list out of the
ansible_devices dict only runs when there are no partitions on the
drives. However during an upgrade the OSD are already configured, they
have been prepared and have partitions so this task won't run and thus
the devices list will be empty, skipping the restart during
rolling_update. We now run the same task under different requirements
when rolling_update is true and build a list when:

* osd_auto_discovery is true
* rolling_update is true
* ansible_devices exists
* no dm/lv are part of the discovery
* the device is not removable
* the device has more than 1 sector

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1613626

### DIFF
--- a/roles/ceph-osd/tasks/build_devices.yml
+++ b/roles/ceph-osd/tasks/build_devices.yml
@@ -12,6 +12,18 @@
     - item.value.holders|count == 0
     - "'dm-' not in item.key"
 
+- name: set_fact devices generate device list when osd_auto_discovery - rolling_update
+  set_fact:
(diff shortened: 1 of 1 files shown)


leseb:
Failing LVM scenarios will be fixed by https://github.com/ceph/ceph-ansible/pull/2997#pullrequestreview-144904653

PR #3005, opened 2018-08-12
igw: valid client CHAP settings.

mikechristie:
The linux kernel target layer, LIO, does not support the iscsi target to
mix ACLs that have chap enabled and disabled under the same tpg. This
patch adds a check and fails if this type of setup is detected.

This fixes Red Hat BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1615088

Review question:

Please review my loop. The implementation is not optimal. I am looping over the clients twice. It was not clear how I can do this with vars and/or maybe a register in one loop.


roles/ceph-iscsi-gw/tasks/common.yml
```diff
@@ -5,6 +5,35 @@
...
+  with_items: "{{ client_connections }}"
+  when:
+    - item.status == "present"
+    - item.chap == ""
+
+- name: Check if igw client chap settings are valid
+  fail:
+    msg: "All ACLs must use the same auth type. Cannot mix CHAP and non CHAP."
+  when:
+    - igw_client_chap_enabled|default(false)|bool == true
```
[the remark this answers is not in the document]
mikechristie: Thanks for the help! That is so much nicer than my patch. There is just one problem that I still can't figure out how to fix using your code as a base.

The code above is too strict. We only want to detect if there item.chap items not set (item.chap == "") mixed with entries that are set (item.chap == "any_valid_string"). If any_valid_string is different for different clients it is ok.

So:

client_connections:
  - { client: 'iqn.2005-03.com.ceph:ini1', image_list: 'rbd.a1', chap: 'asdfaiiououwiea/asdfaduoiuaosifduosi', status: 'present' }
  - { client: 'iqn.2005-03.com.ceph:ini3', image_list: 'rbd.a1', chap: 'asweeeeww33333/2323232k4l4lkjlklj', status: 'present' }

is ok.

But this:

client_connections:
  - { client: 'iqn.2005-03.com.ceph:ini1', image_list: 'rbd.a1', chap: '', status: 'present' }
  - { client: 'iqn.2005-03.com.ceph:ini3', image_list: 'rbd.a1', chap: 'asweeeeww33333/2323232k4l4lkjlklj', status: 'present' }

Is this possible with one loop in ansible?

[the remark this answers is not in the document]
leseb: Sorry I missed your update, so basically `chap` can have any value, they don't need to be identical but `chap` can be ignored if `status` is `absent` and `chap != ''`. Correct?

Something like this?

```
    - name:
      fail:
        msg: "bad chap settings"
      with_items: "{{ client_connections }}"
      when:
        - item.status == "present"
        - item.chap != ''
```

leseb:
@guits yes it should.

leseb:
@mikechristie any update on this? Thanks

mikechristie:
I tried the last chunk from @guits  and it failed to run with the error:

> ERROR! Syntax Error while loading YAML.
>   did not find expected '-' indicator
> 
> The error appears to have been in '/usr/share/ceph-ansible/roles/ceph-iscsi-gw/tasks/check_mandatory_vars.yml': line 15, column 10, but may
> be elsewhere in the file depending on the exact syntax problem.
> 
> The offending line appears to be:
> 
>     - item.chap == ''
>     - '' in client_connections | map(attribute='chap') | list
>          ^ here
> 

The '' seemed to throw it off. I tried to fix that up, but could not figure out another way that gives the correct end result.

mikechristie:
Do you guys just want me to give up on trying to do it in one loop? I can just fix up the patch in the PR to handle the review comments and repush that. If so, I had a question about this comment:

> @leseb shouldn't it be moved in role ceph-validate or /plugins/actions/validate.py ?

leesb wrote yes, but it was not clear if it should be ceph-validate or in the plugins? Is it my choice?

guits:
@mikechristie I think it's just a syntax issue

I'll investigate this but you might be missing double quotes like this :

```
- "'' in client_connections | map(attribute='chap') | list"
```

guits:
@mikechristie I confirm adding double quote as mentioned fixes your error.

mikechristie:
@guits Thanks. That fixed the problem.

I had to slightly modify it so when we make sure to only scan present clients. I did a re-push with the code moved to ceps-validate.

roles/ceph-validate/tasks/main.yml
```diff
@@ -82,3 +82,9 @@
     - radosgw_interface != "dummy"
     - radosgw_address == "0.0.0.0"
     - radosgw_address_block == "subnet"
+
+- name: include check_iscsi.yml
+  include: check_iscsi.yml
+  when:
+    - iscsi_gw_group_name is defined
```
leseb: This is always defined so we don't need it.

roles/ceph-validate/tasks/check_iscsi.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: fail if bad chap settings
+  fail:
+    msg: "Mixing clients with CHAP enabled and disabled is not supported."
+  with_items: "{{ client_connections }}"
+  when:
+    - item.status == "present"
+    - item.chap != ''
+    - " '' in client_connections | selectattr('status', 'match', '^present$') | map(attribute='chap') | list"
```
guits: I think you don't need to write`'^present$'`.
`'present'` is enough since you are using `'match'` rather than `'search'`

mikechristie:
@leseb @guits 

Both your review comments should be fixed now. I have done a re-push of this PR with those changes.

Thanks for the help! Hopefully, I never have to bug you about ansible stuff this much :)

guits:
@mikechristie no worries, happy to help if needed :+1:

roles/ceph-validate/tasks/check_iscsi.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: fail if unsupported chap configuration
+  fail:
+    msg: "Mixing clients with CHAP enabled and disabled is not supported."
+  with_items: "{{ client_connections }}"
```
leseb: `client_connections` is undefined because the var is part of ceph-iscsi, so I guess we either move the vars in ceph-defaults or temporarly do this in this code: `"{{ client_connections | default ({}) }}"`

And we need to add to the `when:`

`- item.status is defined`

mikechristie:
@leseb 
I was not sure which you preferred. I added the with_items/when additions.

If you prefer I can move stuff to ceph-defaults. For that, do you normally only add vars that are checked in ceph-validate or should all defaults from roles/ceph-iscsi-gw/defaults/main.yml be moved there?

guits:
@mikechristie We usually move vars in `ceph-defaults` when they are referenced multiple times in other roles.
In that case, I'd say it's ok to use `| default({})` since this is the only reference outside of `ceph-iscsi`.

guits:
`centos7_cluster` failing because of an unrelated yum repo failure...

PR #3014, opened 2018-08-14
Use /var/lib/ceph/osd folder to filter osd mount point

jeffrey4l:
In some case, use may mount a partition to /var/lib/ceph, and umount
it will be failure and no need to do so too.

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -222,9 +222,7 @@
       timeout: 500
 
   - name: remove data
-    file:
-     path: /var/lib/ceph
-     state: absent
+    command: rm -rf /var/lib/ceph/*
 
   tasks:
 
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/purge-cluster.yml
```diff
@@ -222,9 +222,7 @@
       timeout: 500
 
   - name: remove data
-    file:
-     path: /var/lib/ceph
-     state: absent
+    command: rm -rf /var/lib/ceph/*
```
leseb: what's wrong with the `file` module?

jeffrey4l: if /var/lib/ceph is mounted from a device, it can not be removed. whereas, ansible file module try to remove this folder rather than remove it, then error happends.

leseb: Makes sense.

leseb:
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1615872

PR #3015, opened 2018-08-14
roles: ceph-defaults: Handle missing 'ceph' binary failures

hwoarang:
Since commit f422efb1d6b56ce56a7d39a21736a471e4ed357 ("config: ensure
rgw section has the correct name") we observe the following failures in
new Ceph deployment with OpenStack-Ansible

fatal: [aio1_ceph-rgw_container-fc588f0a]: FAILED! => {"changed": false,
"cmd": "ceph --cluster ceph -s -f json", "msg": "[Errno 2] No such file
or directory"

The problem is because of the following two problems:

- The commit added the following conditional: 'ansible_hostname !=
ansible_fqdn' which for a properly configured system it evaluates to
'True' since ansible_hostname is not supposed to return an FQDN.

- The task executes 'ceph' but at this point no package installation has
happened. Packages are normally installed in the 'ceph-common' role
which runs after the 'ceph-defaults' one.

As such, we can handle failures in the first task so the subsequent
ones will not fail.

### DIFF
--- a/roles/ceph-defaults/tasks/facts.yml
+++ b/roles/ceph-defaults/tasks/facts.yml
@@ -221,6 +221,14 @@
 - block:
     - name: get current cluster status (if already running)
       command: "{{ docker_exec_cmd }} ceph --cluster {{ cluster }} -s -f json"
+      changed_when: false
+      failed_when: false
+      check_mode: no
+      run_once: true
+      delegate_to: "{{ groups[mon_group_name][0] }}"
+      when:
+        - not rolling_update
+        - groups.get(mon_group_name, []) | length > 0
       register: ceph_current_status
 
     - name: set_fact ceph_current_status (convert to json)
--- a/roles/ceph-defaults/tasks/facts.yml
+++ b/roles/ceph-defaults/tasks/facts.yml
@@ -238,6 +238,7 @@
     - name: set_fact rgw_hostname
       set_fact:
         rgw_hostname: "{% for key in ceph_current_status['servicemap']['services']['rgw']['daemons'].keys() %}{% if key == ansible_fqdn %}{{ key }}{% endif %}{% endfor %}"
+      when: ceph_current_status['servicemap']['services']['rgw'] is defined
   when:
     - ceph_current_fsid.get('rc', 1) == 0
     - inventory_hostname in groups.get(rgw_group_name, [])


roles/ceph-defaults/tasks/facts.yml
```diff
@@ -221,15 +221,20 @@
 - block:
     - name: get current cluster status (if already running)
       command: "{{ docker_exec_cmd }} ceph --cluster {{ cluster }} -s -f json"
+      changed_when: false
+      failed_when: false
+      check_mode: no
```
leseb: Actually, this command should be executed on a monitor node:

```
  run_once: true
  when:
    - not rolling_update
    - groups.get(mon_group_name, []) | length > 0
```

roles/ceph-defaults/tasks/facts.yml
```diff
@@ -221,15 +221,20 @@
...
       set_fact:
         ceph_current_status: "{{ ceph_current_status.stdout | from_json }}"
+      when: ceph_current_status.rc == 0
 
     - name: set_fact rgw_hostname
       set_fact:
         rgw_hostname: "{% for key in ceph_current_status['servicemap']['services']['rgw']['daemons'].keys() %}{% if key == ansible_fqdn %}{{ key }}{% endif %}{% endfor %}"
+      when: ceph_current_status.rc == 0
   when:
     - ceph_current_fsid.get('rc', 1) == 0
```
leseb: This task should prevent us from entering the block if no cluster is running, so it's still unclear why you ran into this.

hwoarang: This is the output of the task that is supposed to set `ceph_current_fsid`. But the task is delegated to `"{{ groups[mon_group_name][0] }}"` and that's why it's returning `rc=0`.

```
ok: [aio1_ceph-rgw_container-9474601c -> 172.29.238.208] => {
    "changed": false, 
    "cmd": [
        "timeout", 
        "5", 
        "ceph", 
        "--cluster", 
        "ceph", 
        "fsid"
    ], 
    "delta": "0:00:00.524653", 
    "end": "2018-08-15 08:23:15.892404", 
    "failed_when_result": false, 
    "invocation": {
        "module_args": {
            "_raw_params": "timeout 5  ceph --cluster ceph fsid", 
            "_uses_shell": false, 
            "chdir": null, 
            "creates": null, 
            "executable": null, 
            "removes": null, 
            "stdin": null, 
            "warn": true
        }
    }, 
    "rc": 0, 
    "start": "2018-08-15 08:23:15.367751", 
    "stderr": "", 
    "stderr_lines": [], 
    "stdout": "391701ff-eaf3-4b93-91f6-4b2f57937f74", 
    "stdout_lines": [
        "391701ff-eaf3-4b93-91f6-4b2f57937f74"
    ]
}
```

However, the problematic task in this PR is not delegated as you said and it is running on the wrong node.

roles/ceph-defaults/tasks/facts.yml
hwoarang: done

leseb:
@hwoarang Thanks!

Issue #2998, opened 2018-08-09
site.yml: The use of 'static' has been deprecated.


leseb:
Hello,

Running a syntax check on 'site.yml' using ansible 2.6.2 produces a 
deprecation warning:

```
ceph-ansible]$ ansible-playbook site.yml --syntax-check
[DEPRECATION WARNING]: The use of 'static' has been deprecated. Use 
'import_tasks' for static inclusion,
  or 'include_tasks' for dynamic inclusion. This feature will be removed 
in a future release. Deprecation
  warnings can be disabled by setting deprecation_warnings=False in 
ansible.cfg.

playbook: site.yml
```

I did not found 'static' in the playbook itself. It may be set in one of 
the referenced roles and tasks.

The playbook respectively the files in question should be changed 
accordingly before the feature will be removed.

This messages seems not to be worthy to create a bug report or feature 
request on GitHub. Unfortunately I'm not able to solve this by myself. 
So I would appreciate if the reason for this warning could be discussed 
and fixed by the members of this list.

Best regards,
Joerg K.

leseb:
@rishabh-d-dave whenever you have time :). Thanks.

rishabh-d-dave:
Sure.

rishabh-d-dave:
Fixed at - https://github.com/ceph/ceph-ansible/pull/3016.

rishabh-d-dave:
Closing this issue since it got fixed with PR #3016

PR #3016, opened 2018-08-14
don't use "static" field while including roles

rishabh-d-dave:
See - https://github.com/ceph/ceph-ansible/issues/2998

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -1,55 +1,39 @@
 ---
-- name: include installs/install_on_redhat.yml
-  include: installs/install_on_redhat.yml
-  when:
-    - ansible_os_family == 'RedHat'
+- name: include_tasks installs/install_on_redhat.yml
+  include_tasks: installs/install_on_redhat.yml
+  when: ansible_os_family == 'RedHat'
   tags:
     - package-install
-  # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
-  static: False
 
-- name: include installs/install_on_suse.yml
-  include: installs/install_on_suse.yml
-  when:
-    - ansible_os_family == 'Suse'
+- name: include_tasks installs/install_on_suse.yml
+  include_tasks: installs/install_on_suse.yml
+  when: ansible_os_family == 'Suse'
   tags:
     - package-install
-  # Hard code this so we will skip the entire file instead of individual tasks (Default isn't Consistent)
-  static: False
 
-- name: include installs/install_on_debian.yml
-  include: installs/install_on_debian.yml
-  when:
-    - ansible_os_family == 'Debian'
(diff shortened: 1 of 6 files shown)


roles/ceph-common/tasks/main.yml
```diff
@@ -1,55 +1,43 @@
 ---
-- name: include installs/install_on_redhat.yml
-  include: installs/install_on_redhat.yml
+- include_role:
```
leseb: Shouldn't this be `include_tasks` or `import_tasks` instead? You can read https://docs.ansible.com/ansible/2.4/playbooks_reuse_includes.html

rishabh-d-dave: > Shouldn't this be include_tasks or import_tasks instead?

This task has [`static` option set to `False`](https://github.com/ceph/ceph-ansible/pull/3016/files/19d13314c96f9a92e0096c154165c7ba77febe1d#diff-fbf6d60f8ff9940f0621a8a5c45c392eL9) so `include_task` is correct. Quoting [docs](https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse.html#dynamic-vs-static) -

```
If you use any import* Task (import_playbook, import_tasks, etc.), it will be static. If you use any include* Task (include_tasks, include_role, etc.), it will be dynamic.
```

leseb: Ok let's change it to `include_tasks` then.

rishabh-d-dave:
retest this please

rishabh-d-dave:
@guits @leseb
Can we merge this PR? It looks good to me,

roles/ceph-common/tasks/main.yml
```diff
@@ -1,55 +1,33 @@
 ---
-- name: include installs/install_on_redhat.yml
-  include: installs/install_on_redhat.yml
-  when:
-    - ansible_os_family == 'RedHat'
+- include_tasks: installs/install_on_redhat.yml
```
guits: @rishabh-d-dave can you check and keep this syntax if possible ?
```
- name: include_tasks installs/install_on_redhat.yml
  include_tasks: installs/install_on_redhat.yml
```

guits: @rishabh-d-dave I was concerned because of the output in the playbook but I've just made some tests, not sure it's worth to keep this syntax.

rishabh-d-dave: For most tasks, the description in the `name` field looked redundant. So, IMO it's okay to let them go for now.

guits: @leseb that was my concern

rishabh-d-dave: @leseb 
Yes. I'll modify the patch accordingly.

PR #3022, opened 2018-08-15
mgr: improve/fix disabled modules check

dardok:
Follow up on 36942af6983d60666f3f8a1a06b352a440a6c0da

"disabled_modules" is always a list, it's the items in the list that
can be dicts in mimic.  Many ways to fix this, here's one.


dardok:
Whoops - done.

PR #3041, opened 2018-08-20
iscsi group name preserve backward compatibility

leseb:
Recently we renamed the group_name for iscsi iscsigws where previously
it was named iscsi-gws. Existing deployments with a host file section
with iscsi-gws must continue to work.

This commit adds the old group name as a backoward compatility, no error
from Ansible should be expected, if the hostgroup is not found nothing
is played.

Close: https://bugzilla.redhat.com/show_bug.cgi?id=1619167

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -697,6 +697,7 @@
 
   hosts:
     - "{{ iscsi_gw_group_name|default('iscsigws') }}"
+    - iscsi-gws # for backward compatibility only!
 
   serial: 1
   become: True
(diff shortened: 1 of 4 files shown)


infrastructure-playbooks/rolling_update.yml
```diff
@@ -697,6 +697,7 @@
 
   hosts:
     - "{{ iscsi_gw_group_name|default('iscsigws') }}"
```
GregMeno: Why is this different than all the other instances ?

leseb: What do you mean?

leseb: Oh I see, well that's a different way to write it.

leseb:
No need to wait for the osd scenario since updates/purges and default scenarios have validated the change.

PR #3077, opened 2018-08-29
roles: ceph-rgw: Enable the ceph-radosgw target

hwoarang:
If the ceph-radosgw target is not enabled, then enabling the
ceph-radosgw@ service has no effect since nothing will pull
it on the next reboot. As such, we need to ensure that the
target is enabled.


hwoarang:
@leseb hello. Do you have any thoughts on that? This fixes in the openSUSE Ceph deployments with OpenStack-Ansible.

leseb:
@hwoarang sorry for the late reply, sounds good to me.

leseb:
I guess I was just wondering if we needed a backport or not, but we do.

PR #3081, opened 2018-08-30
Use a tempfile to store restart scripts

mcv21:
Make a tempfile directory and copy the restart scripts there (and then
execute them from there), rather than using insecure known filenames
in /tmp/

This is a partial fix for ceph/ceph-ansible#2937

Signed-off-by: Matthew Vernon <mv3@sanger.ac.uk>

Issue #2937, opened 2018-07-25
Unsafe /tmp/ file use in ceph-ansible


mcv21:
**Bug Report**

Hi,

I've been migrating our local setup to v3.0.39, and discovered that ceph-ansible is now using predictable files/filenames in `/tmp` instead of e.g. the Ansible `tempfile` module in several places This is not best security practice (cf https://cwe.mitre.org/data/definitions/377.html and more generally https://cwe.mitre.org/data/definitions/376.html )

eg 1: roles/ceph-config/tasks/main.yml
```YAML
  - name: template ceph_conf_overrides
    copy:
      content: "{{ ceph_conf_overrides }}"
      dest: "/tmp/ceph_conf_overrides_temp_{{ ansible_hostname }}"
```

eg 2: roles/ceph-docker-common/tasks/fetch_image.yml
```YAML
- name: export local ceph dev image
  local_action:
    module: command
      docker save -o "/tmp/{{ ceph_docker_username }}-{{ ceph_docker_imagename }}-{{ ceph_docker_image_tag }}.tar" "{{ ceph_docker_username }}/{{ ceph_docker_imagename }}:{{ ceph_docker_image_tag }}"
  when:
    - (ceph_docker_dev_image is defined and ceph_docker_dev_image)
  run_once: true
                                                                               -
- name: copy ceph dev image file
  copy:
    src: "/tmp/{{ ceph_docker_username }}-{{ ceph_docker_imagename }}-{{ ceph_docker_image_tag}}.tar"
    dest: "/tmp/{{ ceph_docker_username }}-{{ ceph_docker_imagename }}-{{ ceph_docker_image_tag }}.tar"
  when:
    - (ceph_docker_dev_image is defined and ceph_docker_dev_image)
```

eg 3: roles/ceph-defaults/handlers/main.yml
```YAML
- name: copy mon restart script
  template:
    src: restart_mon_daemon.sh.j2
    dest: /tmp/restart_mon_daemon.sh
    owner: root
    group: root
    mode: 0750
```
[and many examples in that file]

`git grep /tmp` is probably a good way to find most/all of these. The right thing to do, I think, is instead use `tempfile` and register the `path` it returns.

**Environment**:
* Ubuntu 16.04
* kernel  4.4.0-93-generic
* Ansible 2.4.4.0
* ceph-ansible version v3.0.39
* ceph version 10.2.9 (Jewel)

leseb:
[...]

PR #3082, opened 2018-08-30
Use a tempfile to store restart scripts

mcv21:
Make a tempfile directory and copy the restart scripts there (and then
execute them from there), rather than using insecure known filenames
in /tmp/

This is a partial fix for ceph/ceph-ansible#2937

This is a backport of the equivalent fix to master (ceph/ceph-ansible#3081) for stable-3.0

### DIFF
--- a/roles/ceph-defaults/handlers/main.yml
+++ b/roles/ceph-defaults/handlers/main.yml
@@ -5,6 +5,21 @@
   when:
     - ansible_os_family == 'Debian'
 
+- name: make tempdir for scripts
+  tempfile:
+    state: directory
+    prefix: ceph_ansible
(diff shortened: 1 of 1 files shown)


leseb:
We never push directly into stable branches if the initial patch is not in master yet. Typically, you push a patch in master, add the label backport-stable-3.0, once the PR is merged on master a backport will be created automatically.

mcv21:
OK, but the master patch needed a bit of tweaking to apply to stable-3.0...

leseb:
I see, I guess I read too quickly, re-opening, sorry about that.

mcv21:
retest this please

PR #3083, opened 2018-08-30
Use a tempdir for docker images

mcv21:
Make a tempfile directory (both on localhost and on targets), and use
that to store docker images, rather than using insecure known
filenames in /tmp/

This is the other part of the fix for ceph/ceph-ansible#2937

I've not tested this, as I don't currently have a dockerised
ceph-ansible set up in use here.

### DIFF
--- a/roles/ceph-docker-common/tasks/fetch_image.yml
+++ b/roles/ceph-docker-common/tasks/fetch_image.yml
@@ -274,27 +274,55 @@
     - ceph_nfs_container_inspect_before_pull.get('rc') == 0
     - ceph_nfs_image_repodigest_before_pulling != image_repodigest_after_pulling
 
+- name: make local tempdir for ceph dev images
+  local_action:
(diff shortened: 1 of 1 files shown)


roles/ceph-docker-common/tasks/fetch_image.yml
```diff
@@ -274,27 +274,56 @@
...
+    module: tempdir
+    state: directory
+    prefix: ceph_ansible
+  register: localtmpdir
+  when:
+    - (ceph_docker_dev_image is defined and ceph_docker_dev_image)
+  run_once: true
+
+- name: make target tempdir for ceph dev images
+  tempdir:
```
leseb: typo?

mcv21: Yes, sorry; pushed a fix (squashed into the previous commit)

PR #3084, opened 2018-08-30
Dont run client dummy container on ppc64le hosts

andymcc:
The dummy client container currently wont work on ppc64le hosts.
This PR creates a filtered client group that contains only hosts
that are not ppc64le - which can then be the group to run the
dummy container against.

This is for the specific case of a containerized_deployment where
there is a mixture of non-ppc64le hosts and ppc64le hosts.
Currently ppc64le is not supported with Ceph server components.

### DIFF
--- a/roles/ceph-client/tasks/create_users_keys.yml
+++ b/roles/ceph-client/tasks/create_users_keys.yml
@@ -12,6 +12,19 @@
   when:
     - keys_tmp is defined
 
+# dummy container setup is only supported on x86_64
+# when running with containerized_deployment: true this task
+# creates a group that contains only x86_64 hosts.
+# when running with containerized_deployment: false this task
+# will add all client hosts to the group (and not filter).
+- name: create filtered clients group
+  add_host:
+    name: "{{ item }}"
+    groups: _filtered_clients
+  with_items: "{{ groups[client_group_name] }}"
+  when:
+    - (hostvars[item]['ansible_architecture'] == 'x86_64') or (not containerized_deployment)
+
 - name: run a dummy container (sleep 300) from where we can create pool(s)/key(s)
   command: >
     docker run \
@@ -25,7 +38,7 @@
   changed_when: false
   when:
     - containerized_deployment
-    - inventory_hostname == groups.get(client_group_name) | first
+    - inventory_hostname == groups.get('_filtered_clients') | first
 
 - name: set_fact delegated_node
   set_fact:
@@ -57,7 +70,7 @@
   when:
     - cephx
     - keys | length > 0
-    - inventory_hostname == groups.get(client_group_name) | first
+    - inventory_hostname == groups.get('_filtered_clients') | first
 
 - name: slurp client cephx key(s)
   slurp:
@@ -69,7 +82,7 @@
   when:
     - cephx
     - keys | length > 0
-    - inventory_hostname == groups.get(client_group_name) | first
+    - inventory_hostname == groups.get('_filtered_clients') | first
 
 - name: list existing pool(s)
   command: >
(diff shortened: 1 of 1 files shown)


tbreeds:
RHBZ#[1624216](https://bugzilla.redhat.com/show_bug.cgi?id=1624216)

tbreeds:
Works for me!

roles/ceph-client/tasks/create_users_keys.yml
```diff
@@ -12,6 +12,16 @@
...
 
+# dummy container run is currently not supported for ppc64le
+# this task creates a group that contains only non-ppc64le clients
+- name: create filtered clients group
+  add_host:
+    name: "{{ item }}"
+    groups: _filtered_clients
+  with_items: "{{ groups[client_group_name] }}"
+  when:
+    - hostvars[item]['ansible_architecture'] != 'ppc64le'
```
tbreeds: I wonder if it's worth expressing this as:

`- hostvars[item]['ansible_architecture'] == 'x86_64'`

that way if someone tries to do the client setup on aarch64 they don't run into the same issue.

leseb: +1

andymcc:
Works for me, I can add in to specify only x86_64 instead.
I do wonder though, is that container required at all? E.g. for clients, why would we need a dummy container (that will be removed) to run client commands, when the clients should have been setup when ceph-common is installed?

leseb:
@andymcc typically in a fully containerized env, clients do not have any package manager. So to get the binary they need a container.

andymcc:
Ok fixed that up - I made one additional change, which is that if you are running iwith containerized_deployment: false - it won't try filter the list and will instead just add all hosts in to the filtered group.

That'll mean a use-case where you have ppc64le clients only - but are not running in containerized_deployment mode, it wont filter those hosts out and will run normally.

andymcc:
@leseb - ahh ok - and there isn't a container that is just the client binaries?
I'm a bit confused by the fact its a short lived container though - or rather, what does a client look like in a containerized_deployment that we would need to setup a temporary container to do client actions.

roles/ceph-client/tasks/create_users_keys.yml
```diff
@@ -12,6 +12,19 @@
...
+# creates a group that contains only x86_64 hosts.
+# when running with containerized_deployment: false this task
+# will add all client hosts to the group (and not filter).
+- name: create filtered clients group
+  add_host:
+    name: "{{ item }}"
+    groups: _filtered_clients
+  with_items: "{{ groups[client_group_name] }}"
+  when:
+    - (hostvars[item]['ansible_architecture'] == 'x86_64') or (not containerized_deployment)
```
leseb: Shouldn't this be?

```
when:
  - containerized_deployment
  - hostvars[item]['ansible_architecture'] == 'x86_64'
```

andymcc: @leseb - i think that would create a situation where on containerized_deployment: false that the filtered_clients group contains no hosts, we want the opposite which is when containerized_deployment is false or its an x86_64 host, then add it to the group.

leseb:
@andymcc no dedicated container for the binaries only, just the same one. This was a quick and dirty solution which I hope will disappear for 3.2. Actually, 2 PRs have code to remove that dependency already, I just need to resurrect them :).

andymcc:
@leseb - ahh nice! That makes sense. Thanks for the update, I figured it wasn't a permanent fix.

roles/ceph-client/tasks/create_users_keys.yml
leseb: Good point, can't we then do something like `groups.get('_filtered_clients', 'client_group_name')`? So if `_filtered_clients` does not exist we use `client_group_name`?

andymcc: We could - but it makes line 130 difficult because we reference host [0] in the group, rather than using groups.get.

Issue #3064, opened 2018-08-27
mon fails due to systemd unit file error


alitvak69:
**Bug Report**

What happened:
systemd failed to start 1st ceph mon
What you expected to happen:
systemd successfully started 1st ceph mon
How to reproduce it (minimal and precise):

Run deployment and observe it fail at this task
2018-08-26 09:15:33,260 p=604 u=vagrant |  TASK [ceph-mon : systemd start mon container] ***************************************************************************************
2018-08-26 09:15:33,260 p=604 u=vagrant |  task path: /home/vagrant/ansible/alex/ceph-ansible/roles/ceph-mon/tasks/docker/start_docker_monitor.yml:54
2018-08-26 09:15:33,261 p=604 u=vagrant |  Sunday 26 August 2018  09:15:33 +0000 (0:00:01.559)       0:04:34.476 ********* 
2018-08-26 09:15:33,304 p=604 u=vagrant |  Using module file /usr/lib/python2.7/dist-packages/ansible/modules/system/systemd.py
2018-08-26 09:15:33,627 p=604 u=vagrant |  fatal: [storage1n1-chi]: FAILED! => {
    "changed": false, 
    "invocation": {
        "module_args": {
            "daemon_reload": true, 
            "enabled": true, 
            "force": null, 
            "masked": null, 
            "name": "ceph-mon@storage1n1-chi", 
            "no_block": false, 
            "state": "started", 
            "user": false
        }
    }, 
    "msg": "Error loading unit file 'ceph-mon@storage1n1-chi': System.Error.ENOEXEC \"Exec format error\""

system logs show:

Aug 26 09:33:43 storage1n1-chi systemd[1]: **/etc/systemd/system/ceph-mon@.service:8: Executable path is not absolute: $(command -v mkdir) -p /etc/ceph /var/lib/ceph/mon**
Aug 26 09:33:43 storage1n1-chi systemd[1]: **ceph-mon@storage1n1-chi.service: Unit configuration has fatal error, unit will not be started**

I fixed by changing a line in template ceph-mon.service.j2

 [Service]
EnvironmentFile=-/etc/environment 
ExecStartPre=-/usr/bin/docker rm ceph-mon-%i 
--  ExecStartPre=$(command -v mkdir) -p /etc/ceph /var/lib/ceph/mon 
++ ExecStartPre=/bin/bash -c '$$(command -v mkdir) -p /etc/ceph /var/lib/ceph/mon' 
[...]

PR #3091, opened 2018-09-03
ceph-mon: remove unnessary command

leseb:
It appears it's not necessary to find the path of mkdir since it's the same most of the time (as far as I can tell).

Closes: https://github.com/ceph/ceph-ansible/issues/3064


leseb:
Hum I guess we had a good reason in the end, we don't test on Ubuntu anymore so we missed this.

guits:
Closed in favor of #3101

Issue #3063, opened 2018-08-25
rgw_hostname variable becomes empty


maxkochubey:
**Bug Report**

What happened:

After second (and several further) runs of playbook for deploying of my test Ceph cluster, I noticed that my radosgw section in `ceph.conf` became to look such way:

```
root@ceph-node-1:~# cat /etc/ceph/ceph.conf 
[client.rgw.]
host = 
keyring = /var/lib/ceph/radosgw/ceph-rgw./keyring
log file = /var/log/ceph/ceph-rgw-ceph-node-1.log
rgw frontends = civetweb port=10.156.0.2:80 num_threads=32
```

After first run, everything was fine.

Well, as far as I can see, new variable `rgw_hostname` is responsible for this - It becomes empty after two tasks in https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/tasks/facts.yml#L228-L243.

Output of ansible-playbook with `-vvv`:
```
2018-08-26 00:00:25,413 p=17241 u=user |  TASK [ceph-defaults : set_fact rgw_hostname - fqdn] ******************************************************************************************************************************************************************************************
2018-08-26 00:00:25,414 p=17241 u=user |  task path: /home/user/ceph-ansible/roles/ceph-defaults/tasks/facts.yml:225
2018-08-26 00:00:25,414 p=17241 u=user |  Sunday 26 August 2018  00:00:25 +0700 (0:00:00.021)       0:00:12.385 ********* 
2018-08-26 00:00:25,500 p=17241 u=user |  ok: [ceph-node-1] => {
    "ansible_facts": {
        "rgw_hostname": ""
    }, 
    "changed": false, 
    "failed": false
}
2018-08-26 00:00:25,506 p=17241 u=user |  TASK [ceph-defaults : set_fact rgw_hostname - no fqdn] ***************************************************************************************************************************************************************************************
2018-08-26 00:00:25,506 p=17241 u=user |  task path: /home/user/ceph-ansible/roles/ceph-defaults/tasks/facts.yml:235
2018-08-26 00:00:25,506 p=17241 u=user |  Sunday 26 August 2018  00:00:25 +0700 (0:00:00.092)       0:00:12.477 ********* 
2018-08-26 00:00:25,537 p=17241 u=user |  skipping: [ceph-node-1] => {
    "changed": false, 
    "skip_reason": "Conditional result was False", 
    "skipped": true
}
```
My `group_vars`:
```
[...]

PR #3095, opened 2018-09-05
Issue #3063: rgw_hostname

maxkochubey:
#3063
I don't know if it can affect other stages, but it works with such an additional condition.


maxkochubey:
Sorry for the mess, pushed from tag :(

PR #3096, opened 2018-09-05
defaults: add a default value to rgw_hostname

guits:
let's add ansible_hostname as a default value for rgw_hostname if no
hostname in servicemap matches ansible_fqdn.

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #3097, opened 2018-09-05
One more condition to check if rgw_hostname is empty

maxkochubey:
#3063
I don't know if it can affect other stages, but it works with such an additional condition.


maxkochubey:
#3096 is better implementation.

PR #3101, opened 2018-09-06
mon: fix `ExecStartPre` option in systemd unit file

guits:
This command line is not supported.
According to official documentation:

```
Note that shell command lines are not directly supported.
If shell command lines are to be used,
they need to be passed explicitly to a shell implementation of some kind.
```

We must run this using /bin/sh instead.


leseb:
No backport?

PR #3108, opened 2018-09-07
tests: set ceph_docker_image_tag to CEPH_DOCKER_IMAGE_TAG env variable

guits:
Using `UPDATE_CEPH_DOCKER_IMAGE_TAG` here will make an upgrade of the
ceph release when running switch_to_containers scenario which is not
correct.

Eg:
If ceph luminous was first deployed, then we should switch to ceph
luminous containers, not to mimic.


guits:
jenkins test luminous-switch_to_containers

PR #3109, opened 2018-09-09
tests: do not upgrade ceph release for switch_to_containers scenario

guits:
Using `UPDATE_*` environment variables here will make an upgrade of the ceph release when running switch_to_containers scenario which is not correct.

Eg:
If ceph luminous was first deployed, then we should switch to ceph luminous containers, not to mimic.


guits:
jenkins test luminous-switch_to_containers

PR #3113, opened 2018-09-12
set osd max memory based on host memory

neha-ojha:
This PR addresses https://bugzilla.redhat.com/show_bug.cgi?id=1595003 by adding an `osd_memory_target` option to ceph-ansible.

Signed-off-by: Neha Ojha <nojha@redhat.com>


roles/ceph-config/tasks/main.yml
```diff
@@ -9,6 +9,24 @@
       group: "ceph"
       mode: "0755"
 
+  - name: count number of osds for non-lvm scenario
+    set_fact:
+      num_osds: "{{ devices | length }}"
+    when:
+      - devices|length > 0 and (osd_scenario == 'collocated' or osd_scenario == 'non-collocated')
```
andrewschoen: ``(osd_scenario == 'collocated' or osd_scenario == 'non-collocated')`` should be on a new line

```
when:
  - devices|length > 0
  - (osd_scenario == 'collocated' or osd_scenario == 'non-collocated')
```

roles/ceph-config/tasks/main.yml
```diff
@@ -9,6 +9,24 @@
...
+
+  - name: count number of osds for lvm scenario
+    set_fact:
+      num_osds: "{{ lvm_volumes | length }}"
+    when:
+      - lvm_volumes|length > 0 and osd_scenario == 'lvm'
```
andrewschoen: Same here, two lines are needed in the ``when`` condition.

roles/ceph-config/tasks/main.yml
```diff
@@ -9,6 +9,24 @@
...
+
+  - name: count number of osds for lvm-batch scenario
+    set_fact:
+      num_osds: "{{ (ceph-volume lvm batch --report devices | from_json()).osds | length }}"
+    when:
+      - devices|length > 0 and osd_scenario == 'lvm-batch'
```
andrewschoen: Same here, two lines are needed. Also, there is no ``lvm-batch`` scenario, this should still check for the ``lvm`` scenario.

roles/ceph-config/tasks/main.yml
```diff
@@ -9,6 +9,24 @@
...
+
+  - name: count number of osds for lvm scenario
+    set_fact:
+      num_osds: "{{ lvm_volumes | length }}"
+    when:
+      - lvm_volumes|length > 0 and osd_scenario == 'lvm'
+
+  - name: count number of osds for lvm-batch scenario
+    set_fact:
+      num_osds: "{{ (ceph-volume lvm batch --report devices | from_json()).osds | length }}"
```
andrewschoen: the ``ceph-volume lvm batch --report`` will need to be run in a separate task using the command module, with the output from the command being registered into a variable. Then that variable can be used here to set the ``num_osds`` fact.

roles/ceph-config/tasks/main.yml
neha-ojha: I used `and` based on https://github.com/neha-ojha/ceph-ansible/commit/9e203db2334ee04449451e240473860e4749d454#r30481853. I'll put them in two lines.

neha-ojha: I used and based on neha-ojha@9e203db#r30481853. I'll put them in two lines.

roles/ceph-config/tasks/main.yml
```diff
@@ -13,19 +13,22 @@
...
+      - lvm_volumes|length > 0
+      - osd_scenario == 'lvm'
 
   - name: count number of osds for lvm-batch scenario
     set_fact:
       num_osds: "{{ (ceph-volume lvm batch --report devices | from_json()).osds | length }}"
     when:
-      - devices|length > 0 and osd_scenario == 'lvm-batch'
+      - devices|length > 0
+      - osd_scenario == 'lvm-batch'
```
andrewschoen: This needs to be ``- osd_scenario == 'lvm'``

neha-ojha: yeah, updating it in my next commit with the lvm-batch changes.

roles/ceph-config/tasks/main.yml
```diff
@@ -9,6 +9,27 @@
...
+  - name: count number of osds for lvm scenario
+    set_fact:
+      num_osds: "{{ lvm_volumes | length }}"
+    when:
+      - lvm_volumes|length > 0
+      - osd_scenario == 'lvm'
+
+  - name: count number of osds for lvm-batch scenario
+    set_fact:
+      num_osds: "{{ (ceph-volume lvm batch --report devices | from_json()).osds | length }}"
```
guits: adding this here would mean it'd be run on all host, I think it must be run on an OSD.

By the way, you have to do it in two steps :
```
- name: get number of osds for lvm-batch scenario
  command: "{{ docker_exec_cmd }} ceph-volume lvm batch --report devices"
  register: lvm_batch_devices
  when:
    - devices|length > 0 and osd_scenario == 'lvm-batch'
```

```
- name: set_fact num_osds
  set_fact:
    num_osds: "{{ (lvm_batch_devices | from_json).osds | length }}"
  when:
    - devices|length > 0 and osd_scenario == 'lvm-batch'
```

roles/ceph-config/tasks/main.yml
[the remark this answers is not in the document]
andrewschoen: @guits would we just use ``delegate_to`` and let the first OSD host take care of it. We'd probably only need to run this once I'd imagine, not on every OSD host.

roles/ceph-config/tasks/main.yml
andrewschoen: When https://github.com/ceph/ceph-ansible/pull/3111 merges this command will also need the ``--osds-per-device osds_per_device`` flag added.

andrewschoen: You're also missing ``--format=json`` here.

neha-ojha:
@guits @andrewschoen I've updated the PR, could you guys take another look please.

roles/ceph-config/tasks/main.yml
```diff
@@ -9,6 +9,36 @@
...
+      num_osds: "{{ lvm_volumes | length }}"
+    when:
+      - lvm_volumes|length > 0
+      - osd_scenario == 'lvm'
+
+  - name: get number of osds for lvm-batch scenario
+    command: "ceph-volume lvm batch --report --format=json {{ devices| join(' ') }}"
+    register: lvm_batch_devices
+    run_once: true
+    delegate_to: "{{ groups[osd_group_name][0]"
```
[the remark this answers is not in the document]
neha-ojha: fixing it. I see that https://github.com/ceph/ceph-ansible/pull/3111 got merged, I'll add `--osds-per-device osds_per_device` as well.

roles/ceph-config/tasks/main.yml
```diff
@@ -9,6 +9,36 @@
...
+
+  - name: count number of osds for lvm scenario
+    set_fact:
+      num_osds: "{{ lvm_volumes | length }}"
+    when:
+      - lvm_volumes|length > 0
+      - osd_scenario == 'lvm'
+
+  - name: get number of osds for lvm-batch scenario
+    command: "ceph-volume lvm batch --report --format=json --osds-per-device osds_per_device {{ devices| join(' ') }}"
```
guits: @andrewschoen is this supposed to be run only in non containerized deployment?

neha-ojha:
@guits squashed.

guits:
jenkins test pipeline

roles/ceph-defaults/defaults/main.yml
```diff
@@ -352,6 +352,10 @@ cephfs_pools:
 
 ## OSD options
 #
+is_hci: false
```
guits: @neha-ojha you have to run `./generate_group_vars_sample.sh` to update sample files in `group_vars/`

roles/ceph-config/tasks/main.yml
andrewschoen: I don't know, but I suspect it should work on containerized deployments as well. ``docker_exec_cmd`` is expecting a mon at this point, right?

guits: this is exactly what i'm raising this.
It means, we would need to set a new fact like `docker_exec_cmd_osd`

roles/ceph-defaults/defaults/main.yml
neha-ojha: added it.

roles/ceph-config/tasks/main.yml
guits: @neha-ojha you have to add this task before L26

```
- name: set_fact docker_exec_cmd_osd
  set_fact:
    docker_exec_cmd_osd: "docker exec ceph-osd-{{ hostvars[groups[osd_group_name][0]]['ansible_hostname'] }}"
  when:
    - containerized_deployment
    - groups.get(osd_group_name, []) | length > 0
```

then 

`command: "{{ docker_exec_cmd_osd | default('') }} ceph-volume lvm batch --report --format=json --osds-per-device osds_per_device {{ devices| join(' ') }}"`

neha-ojha: @guits will this `command:` be added as a part of a new task for `containerized_deployment`or will it replace https://github.com/ceph/ceph-ansible/pull/3113/files#diff-408c9554f34e961d163f68b645ee6da3R27?

neha-ojha: @guits updated it, let me know if it looks alright. Managed to get https://github.com/ceph/ceph-ansible/pull/3113/commits/5d9f269798c995aeb3baec22744ef0a5da27c7d4#diff-2857b72f37476e1b5d16e7e81783ecbaL241 as a by-product of that change.

roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -151,6 +151,17 @@ filestore xattr use omap = true
 {# else, default is false #}
 {% endif %}
 {% endif %}
+{% if osd_objectstore == 'bluestore' %}
+[osd]
+osd memory target = {{ osd_memory_target }}
+{% if is_hci is defined %} # is_hci should be defined in group_vars/all.yml
+{% if num_osds > 0 and ansible_memtotal_mb * hci_safety_factor / num_osds  > osd_memory_target %} # hci_safety_factor is the safety factor for HCI deployments
+osd memory target = {{ ansible_memtotal_mb * hci_safety_factor / num_osds }}
+{% endif %}
```
guits: @neha-ojha something is wrong here, I think you want a `{% elif %}`?

guits: don't pay attention to my last comment, I think you are just missing a `{% endif %}`

neha-ojha: yeah, I noticed it when I wrote a pythonic version of these ifs :) fixing it

jdurgin:
@markhpc do you think the default of using 0.5 of the node memory is too conservative? In your earlier graphs it looked like RSS could be ~10% over the target

roles/ceph-config/templates/ceph.conf.j2
guits: @neha-ojha perhaps I'm missing something but as it is

If osd_objectstore = 'bluestore' and is_hci is defined, it means you are going to add `osd memory target` twice in the config, L156 + either L159 or L163

neha-ojha: I need the default value of `osd memory target` to be set to 4000000000. If those conditions are met, we will override this value else not.

neha-ojha:
ok, this is passing tests now.

guits:
jenkins test pipeline

alfredodeza:
This PR is now breaking ceph-volume tests. Ideally, we would like to see some tests pass with devel (ceph-volume, ceph master, etc...):

    TASK [ceph-config : count number of osds for non-lvm scenario] *****************
    task path: /tmp/tox.fjM8P3TRe0/xenial-filestore-dmcrypt/tmp/ceph-ansible/roles/ceph-config/tasks/main.yml:13
    skipping: [mon0] => {
        "changed": false, 
    }
    fatal: [osd0]: FAILED! => {}
    
    MSG:
    
    The conditional check 'devices | length > 0' failed. The error was: error while evaluating conditional (devices | length > 0): 'devices' is undefined
    
    The error appears to have been in '/tmp/tox.fjM8P3TRe0/xenial-filestore-dmcrypt/tmp/ceph-ansible/roles/ceph-config/tasks/main.yml': line 13, column 7, but may
    
      - block:
        - name: count number of osds for non-lvm scenario
          ^ here

PR #3115, opened 2018-09-12
igw: stop daemons on purge all calls

mikechristie:
When purging the entire igw config (lio and rbd) stop the api and gw
daemons. Note that the api and gw daemons are bound so you only need
to stop the gw one explicitly.

Fixes Red Hat BZ
https://bugzilla.redhat.com/show_bug.cgi?id=1621255

### DIFF
--- a/infrastructure-playbooks/purge-iscsi-gateways.yml
+++ b/infrastructure-playbooks/purge-iscsi-gateways.yml
@@ -36,4 +36,9 @@
       when: igw_purge_type == 'all'
 
     - name: restart rbd-target-gw daemons
+      service: name=rbd-target-gw state=stopped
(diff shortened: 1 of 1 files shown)


leseb:
@mikechristie what was the issue with that PR?

mikechristie:
The patch in this PR only stopped the daemon. It did not disable it so if you rebooted the system then it started up again.

When I fixed up the patch I messed up my git trees and had to just open a new PR with the updated patch because I couldn't do a forced repush. Here is the new PR:

https://github.com/ceph/ceph-ansible/pull/3116

leseb:
@mikechristie thanks for the clarification.

Issue #3135, opened 2018-09-18
`osds_per_device` is used literally, breaks all `batch` ceph-volume tests


alfredodeza:
**Bug Report**

What happened: Undefined variables are being used, still coming from PR #3113 from commit 27027a17d 

In this case, the following error is raised:

    ceph-volume lvm batch: error: argument --osds-per-device: invalid int value: 'osds_per_device'

Because the command being sent to ceph-volume is not using a integer:

          command: "ceph-volume lvm batch --report --format=json --osds-per-device osds_per_device {{ devices | join(' ') }}"

In that case `osds_per_device` is being used literally.

What you expected to happen: the correct number of osds per device is used

How to reproduce it (minimal and precise): Attempt to run any `batch` test case for ceph-volume using the current master version of ceph-ansible

alfredodeza:
TASK [ceph-config : get number of osds for lvm-batch scenario] *****************
    task path: /tmp/tox.hTs4ghLxdR/centos7-bluestore-mixed_type/tmp/ceph-ansible/roles/ceph-config/tasks/main.yml:27
    skipping: [mon0] => {
        "changed": false, 
        "skip_reason": "Conditional result was False"
    }
    fatal: [osd0]: FAILED! => {
        "changed": true, 
        "cmd": [
            "ceph-volume", 
            "lvm", 
            "batch", 
            "--report", 
            "--format=json", 
            "--osds-per-device", 
            "osds_per_device", 
            "/dev/sdb", 
            "/dev/sdc", 
            "/dev/nvme0n1"
        ], 
        "delta": "0:00:00.296023", 
        "end": "2018-09-18 18:26:49.871305", 
        "rc": 2, 
        "start": "2018-09-18 18:26:49.575282"
    }
    
    STDERR:
    
    usage: ceph-volume lvm batch [-h] [--bluestore] [--filestore] [--report]
                                 [--yes] [--format {json,pretty}] [--dmcrypt]
                                 [--crush-device-class CRUSH_DEVICE_CLASS]
                                 [--no-systemd]
                                 [--osds-per-device OSDS_PER_DEVICE]
                                 [DEVICES [DEVICES ...]]
    ceph-volume lvm batch: error: argument --osds-per-device: invalid int value: 'osds_per_device'

alfredodeza:
Testing out with these changes:

    -      command: "ceph-volume lvm batch --report --format=json --osds-per-device osds_per_device {{ devices | join(' ') }}"
[...]

PR #3136, opened 2018-09-18
ceph-config: do not use osds_per_device literally

andrewschoen:
Fixes: https://github.com/ceph/ceph-ansible/issues/3135

### DIFF
--- a/roles/ceph-config/tasks/main.yml
+++ b/roles/ceph-config/tasks/main.yml
@@ -10,12 +10,12 @@
       mode: "0755"
 
   - block:
-    - name: count number of osds for non-lvm scenario
+    - name: count number of osds for ceph-disk scenarios
       set_fact:
         num_osds: "{{ devices | length | int }}"
       when:
         - devices | default([]) | length > 0
-        - (osd_scenario == 'collocated' or osd_scenario == 'non-collocated')
+        - osd_scenario in ['collocated', 'non-collocated']
 
     - name: count number of osds for lvm scenario
       set_fact:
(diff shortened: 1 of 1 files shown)


alfredodeza:
So this is not going to work at all. The ceph.conf file is needed to define journal sizing, and it is optional in bluestore for block.db sizing. The command is still wrong because --filestore needs to be passed in and it isn't.

At this point I would recommend to do a "best guess" of number of OSDs should be roughly the length of `devices`

alfredodeza:
@neha-ojha @jdurgin would it be OK to default here to the number of devices to infer the number of OSDs? The caveat being that 10 HDDs and 1 SSD would produce 9 OSDs, not 11.

andrewschoen:
> @neha-ojha @jdurgin would it be OK to default here to the number of devices to infer the number of OSDs? The caveat being that 10 HDDs and 1 SSD would produce 9 OSDs, not 11.

The only solution I can think of here that might be more accurate is to write a temporary ceph.conf and then point the ``ceph-volume`` command at it so we could retrieve the number of OSDs and then use that information to render the correct configuration file. This might be too much of a hassle though.

andrewschoen:
For now I think the best solution is to count ``devices`` and multiply by ``osds_per_devices`` for a best guess estimate. We've cut http://tracker.ceph.com/issues/36088 to add functionality into ``ceph-volume`` that will allow us to get the number of osds without the need for a ceph.conf. Once that's complete we can change things here.

This branch is now passing ceph-volume tests again: https://jenkins.ceph.com/job/ceph-volume-scenario/951

neha-ojha:
@andrewschoen @alfredodeza sounds good to me.

Issue #2008, opened 2017-10-06
Luminous : "restart ceph osds daemon" error :  "PGs were not reported as active+clean"


mistur:
the condition in ./roles/ceph-defaults/templates/restart_osd_daemon.sh.j2 expects all PGs are active+clean, it fails if there is scrub or deep-scrub running during the check

```
RUNNING HANDLER [ceph-defaults : restart ceph osds daemon(s) - non container] *******************************************************************************************************************************************************************************************************************************************
Using module file /usr/lib/python2.7/dist-packages/ansible/modules/commands/command.py
<iccluster015.iccluster.epfl.ch> ESTABLISH SSH CONNECTION FOR USER: None
<iccluster015.iccluster.epfl.ch> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=30 -o ControlPath=/root/.ansible/cp/%h-%r-%p iccluster015.iccluster.epfl.ch '/bin/sh -c '"'"'echo ~ && sleep 0'"'"''
<iccluster015.iccluster.epfl.ch> (0, '/root\n', '')
<iccluster015.iccluster.epfl.ch> ESTABLISH SSH CONNECTION FOR USER: None
<iccluster015.iccluster.epfl.ch> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=30 -o ControlPath=/root/.ansible/cp/%h-%r-%p iccluster015.iccluster.epfl.ch '/bin/sh -c '"'"'( umask 77 && mkdir -p "` echo /root/.ansible/tmp/ansible-tmp-1507287643.94-139261965346823 `" && echo ansible-tmp-1507287643.94-139261965346823="` echo /root/.ansible/tmp/ansible-tmp-1507287643.94-139261965346823 `" ) && sleep 0'"'"''
<iccluster015.iccluster.epfl.ch> (0, 'ansible-tmp-1507287643.94-139261965346823=/root/.ansible/tmp/ansible-tmp-1507287643.94-139261965346823\n', '')
[...]

PR #2965, opened 2018-08-02
osd: Fixed bug in regular expression in restart_osd_daemon.sh

artur-fijalkowski:
OSDs can have numbers >99 which means regular expression in `restart_osd_daemon.sh` should catch OSD number of any non-zero length.

This solves https://github.com/ceph/ceph-ansible/issues/2964
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1612853

PR #3137, opened 2018-09-19
restart_osd_daemon.sh.j2 - consider scrubbing pgs as OK

mcv21:
After restarting each OSD, restart_osd_daemon.sh checks that the
cluster is in a good state before moving on to the next one. One of
the checks it does is that the number of pgs in the state
"active+clean" is equal to the total number of pgs in the cluster.

On large clusters (e.g. we have 173,696 pgs), it is likely that at
least one pg will be scrubbing and/or deep-scrubbing at any one
time. These pgs are in state "active+clean+scrubbing" or
"active+clean+scrubbing+deep", so the script was erroneously not
including them in the "good" count.

Fix this by summing the total of pgs in these 3 good states, and
comparing that to the num_pgs in pgmap; do this as an integer
comparison.

(could this be backported to at least stable-3.0 please?)

Closes: #2008

### DIFF
--- a/roles/ceph-defaults/templates/restart_osd_daemon.sh.j2
+++ b/roles/ceph-defaults/templates/restart_osd_daemon.sh.j2
@@ -10,7 +10,7 @@ check_pgs() {
     return 0
   fi
   while [ $RETRIES -ne 0 ]; do
-    test "[""$($docker_exec ceph $CEPH_CLI -s -f json | python -c 'import sys, json; print(json.load(sys.stdin)["pgmap"]["num_pgs"])')""]" = "$($docker_exec ceph $CEPH_CLI -s -f json | python -c 'import sys, json; print [ i["count"] for i in json.load(sys.stdin)["pgmap"]["pgs_by_state"] if i["state_name"] == "active+clean"]')"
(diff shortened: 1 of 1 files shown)


mcv21:
As a quibble, I think `{1,}` is unconventional regex style, and `+` would be clearer - the `{1,}` were introduced in #2965 (previously they were `{1,2}`). Would you be open to my changing `{1,}` to `+`? If so, in this PR, or as a subsequent one?

guits:
@mcv21 I think you can add a commit in this PR for that

kcwong-verseon:
> As a quibble, I think `{1,}` is unconventional regex style, and `+` would be clearer - the `{1,}` were introduced in #2965 (previously they were `{1,2}`). Would you be open to my changing `{1,}` to `+`? If so, in this PR, or as a subsequent one?

I'm used to `+` from my Perl days as well, but believe it or not, it's not universally supported (I've encountered failure with some version of `sed`, and so `{1,}` actually has broader support. That said, we are already using the `grep -E` and `+` is in the alternate regex.

mcv21:
Having looked at this further, there are other good states (e.g. active+clean+snaptrim); so instead modify the check to be "is there 'active+clean' in the state of this pg", since that catches all the good states.

Issue #3148, opened 2018-09-21
Add Ansible 2.7 testing


alfredodeza:
Ansible 2.7 is about to be finalized, according to schedule that should be October 4th [0]

Release candidates are already available from PyPI. This project should start testing against this version to catch issues as soon as possible.

[0] https://docs.ansible.com/ansible/devel/roadmap/ROADMAP_2_7.html#expected

alfredodeza:
This was closed almost a month after 2.7 was officially released. The idea here is to be able to test *in advance* before the actual release, so that fixes can be implemented (or at least issues are known) beforehand. 

As soon as 2.8 RC's are announced, the work to add the tests should be enabled.

PR #3151, opened 2018-09-21
tests: test master against ansible 2.7

guits:
Let's test ceph-ansible master against ansible 2.7 to catch early any
potential issue with this ansible version.

Closes: #3148

### DIFF
--- a/ceph-ansible.spec.in
+++ b/ceph-ansible.spec.in
@@ -16,10 +16,10 @@ Obsoletes:      ceph-iscsi-ansible <= 1.5
 
 BuildArch:      noarch
 
-BuildRequires: ansible >= 2.4.2.0
+BuildRequires: ansible >= 2.7
 BuildRequires: python2-devel
 
-Requires: ansible >= 2.4.2.0
+Requires: ansible >= 2.7
 Requires: python-netaddr
 Requires: python-notario >=  0.0.13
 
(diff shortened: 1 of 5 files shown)


guits:
we must wait for the first 2.7 release to be available in pip

guits:
@ktdreyer what is the ansible version available in rpm repository for 2.7 ?

ktdreyer:
The places where we ship ceph-ansible 3.0 (RHCS 2 and 3.0) have Ansible 2.4.

The places where we ship ceph-ansible 3.1 (RHCS 3.1) [will have](https://bugzilla.redhat.com/show_bug.cgi?id=1576102) Ansible 2.4.

The places where we ship ceph-ansible 3.2 (RHCS 3.2) [will have](https://bugzilla.redhat.com/show_bug.cgi?id=1613941) Ansible 2.6 ... although there's discussion ongoing with @gfidente regarding TripleO compatibility with Ansible 2.6. It seems like it would be wise to keep Ansible 2.5 compatibility in ceph-ansible if it is not hurting anything.

gfidente:
@ktdreyer +1

tripleo uses luminous in queens and rocky; we could try using ansible 2.6 (or 2.7) in rocky but it's much more difficult to bump it in queens

I would vote for keeping 2.5 compatibility as well

gfidente:
pretty sure at least [1] needs to be fixed to work past 2.5

1. https://github.com/ceph/ceph-ansible/blob/v3.1.6/plugins/actions/_v2_config_template.py#L638

PR #3153, opened 2018-09-21
tests: add an RGW node on osd0 for ooo-collocation

guits:
get more coverage by adding an RGW daemon collocated on osd0.
We've missed a bug in the past which could have been caught earlier in
the CI.
Let's add this additional daemon in order to have a better coverage.


gfidente:
thanks, I'll see if I am able to review the ooo params again in the near future

Issue #3154, opened 2018-09-21
restart_osd_daemon.sh.j2 - RETRIES never reset to original value


mcv21:
**Bug Report**

Despite fixing this script to accept all active+clean states (cf #3137 ), we were still having problems with our production servers not managing to restart all their OSDs correctly.

The problem is that RETRIES is initialised once (l3), and then only ever decremented (l17); our production boxes have 60 OSDs, and `handler_health_osd_check_retries` is 40 by default, so while the script will wait up to 40x30s after the restart of the first OSD, you only have to need to wait once per OSD-restart to run out of RETRIES before you get to the last OSD.

The fix is fairly straightforward - set RETRIES before each time check_pgs is called:

```diff
diff --git a/roles/ceph-defaults/templates/restart_osd_daemon.sh.j2 b/roles/ceph
index 0781c34..1d9db15 100644
--- a/roles/ceph-defaults/templates/restart_osd_daemon.sh.j2
+++ b/roles/ceph-defaults/templates/restart_osd_daemon.sh.j2
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-RETRIES="{{ handler_health_osd_check_retries }}"
 DELAY="{{ handler_health_osd_check_delay }}"
 CEPH_CLI="--name client.bootstrap-osd --keyring /var/lib/ceph/bootstrap-osd/{{ 
 
@@ -78,6 +77,7 @@ for unit in $(systemctl list-units | grep -E "loaded * active"
   {% endif %}
   SOCKET=/var/run/ceph/{{ cluster }}-osd.${osd_id}.asok
   while [ $COUNT -ne 0 ]; do
+    RETRIES="{{ handler_health_osd_check_retries }}"
     $docker_exec test -S "$SOCKET" && check_pgs && continue 2
     sleep $DELAY
     let COUNT=COUNT-1
```

Would you take a PR to fix this? I'd rather produce one once #3137 is merged, for obvious reasons :)

**Environment**:
* OS (e.g. from /etc/os-release): Ubuntu Xenial
* ceph-ansible version (e.g. `git head or tag or stable branch`): stable-3.0
* Ceph version (e.g. `ceph -v`): Jewel

mcv21:
[as an aside, the fact that the script waits one more `$DELAY` before failing and calling `ceph -s` made this much harder to track down - often the final `$DELAY` would be enough for all the pgs to be `active+clean` again]

PR #3155, opened 2018-09-21
restart_osd_daemon.sh.j2 - Reset RETRIES between calls of check_pgs

mcv21:
Previously RETRIES was set (by default to 40) once at the start of the
script; this meant that it would only ever wait for up to 40 lots of
30s across *all* the OSDs on a host before bombing out. In fact, we
want to be prepared to wait for the same amount of time after each OSD
restart for the clusters' pgs to be happy again before continuing.

Closes: #3154


guits:
@mcv21 that's correct, nice catch.
Thanks!

mcv21:
@guits Sorry, could this get backported to stable-3.0 also, please? I think the mergify bot can do that, but I don't think I can ask it to do so, IYSWIM...

PR #3179, opened 2018-09-29
[skip ci] use uid instead of username, required for docker deployment

foosinn:
When deploying ceph in docker containers there is no user ceph on the host. Other tasks also use `ceph_uid` to work around this issue.


foosinn:
Updated commit message and signed it. Can you check if everything correct?

guits:
In any case we should use these facts here anyway.

foosinn:
Updated to master branch.

foosinn:
I'm not a 100% sure about that. I have migrated to a none Docker deployment, since i needed to use Ceph on LVM volumes. Cant check anymore

PR #3181, opened 2018-10-02
[skip ci] Fix compatibility with Ansible > 2.4

teralype:
This pull request tries to fix the Ansible code so that it works with Ansible > 2.4.


guits:
@teralype stable-2.1 isn't maintained anymore.

By the way, we don't push directly to stable branches we only do backport from master for bugs and security fixes.

You should consider upgrading to a more recent version.

PR #2866, opened 2018-07-09
ceph-osd: ceph-volume container support

leseb:
Signed-off-by: Sébastien Han <seb@redhat.com>

Closes: https://github.com/ceph/ceph-ansible/issues/2717

todo:

- [x] add code to support ceph-volume in container on the ceph_volume module

- [x] add code to support ceph-volume in container in the playbook

- [x] implement batch command in container

- [x] add doc

- [x] add warning flag explaining as of 3.2 ceph-disk won't be supported anymore (code will remain though), the ceph-disk code will go away on 3.3

- [x] add new functionnal tests for ceph-volume container

Remaining TODO in another PR:

- [ ] add support in shrink-osd.yml container for ceph-volume

- [x] add support in purge-docker-cluster.yml https://github.com/ceph/ceph-ansible/pull/2866

PR #3190, opened 2018-10-03
ceph-volume: makes the lvm scenario idempotent when using the simple configuration

andrewschoen:
This depends on https://github.com/ceph/ceph/pull/24404

As well as making the ``ceph_volume`` module idempotent when the ``action`` is ``batch``, this adds support for ``journal_size`` and ``block_db_size`` which allows us to get an accurate number of OSDs that will be created in ``ceph-config`` before a config file is written. The new ``ceph-volume`` flags ``--journal-size`` and ``--block-db-size`` was added in https://github.com/ceph/ceph/pull/24201 and is still unreleased.

### DIFF
--- a/roles/ceph-config/tasks/main.yml
+++ b/roles/ceph-config/tasks/main.yml
@@ -38,6 +38,7 @@
         report: true
         action: "batch"
       register: lvm_batch_report
+      failed_when: false
       environment:
         CEPH_VOLUME_DEBUG: 1
       when:
@@ -50,6 +51,26 @@
       when:
         - devices | default([]) | length > 0
         - osd_scenario == 'lvm'
+        - (lvm_batch_report.stdout | from_json).changed
+
+    - name: run 'ceph-volume lvm list' to see how many osds have already been created
+      ceph_volume:
+        action: "list"
+      register: lvm_list
+      environment:
+        CEPH_VOLUME_DEBUG: 1
+      when:
+        - devices | default([]) | length > 0
+        - osd_scenario == 'lvm'
+        - not (lvm_batch_report.stdout | from_json).changed
+
+    - name: set_fact num_osds from the output of 'ceph-volume lvm list'
+      set_fact:
+        num_osds: "{{ lvm_list.stdout | from_json | length | int }}"
+      when:
+        - devices | default([]) | length > 0
+        - osd_scenario == 'lvm'
+        - not (lvm_batch_report.stdout | from_json).changed
 
     when:
       - inventory_hostname in groups.get(osd_group_name, [])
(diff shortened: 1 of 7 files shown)


roles/ceph-config/tasks/main.yml
```diff
@@ -27,16 +27,50 @@
...
+        journal_size: "{{ journal_size }}"
+        block_db_size: "{{ block_db_size }}"
+        report: true
+        action: "batch"
+      register: lvm_batch_report
+      environment:
+        CEPH_VOLUME_DEBUG: 1
+      when:
+        - devices | default([]) | length > 0
+        - osd_scenario == 'lvm'
```
leseb: shouldn't we run this on an osd node?

roles/ceph-config/tasks/main.yml
```diff
@@ -27,16 +27,50 @@
...
+
+    - name: set_fact num_osds from the output of 'ceph-volume lvm batch --report'
+      set_fact:
+        num_osds: "{{ (lvm_batch_report.stdout | from_json).osds | length | int }}"
+      when:
+        - devices | default([]) | length > 0
+        - osd_scenario == 'lvm'
+        - (lvm_batch_report.stdout | from_json).changed
+
+    - name: run 'ceph-volume lvm list' to see how many osds have already been created
```
leseb: ditto

leseb:
@andrewschoen can we test it on master with the batch scenario?
Is this expected for 3.2?

andrewschoen:
> @andrewschoen can we test it on master with the batch scenario?
> Is this expected for 3.2?

I tested this yesterday with the branch from ceph/ceph#24404 here https://2.jenkins.ceph.com/job/ceph-ansible-scenario/342/

Yes, we wanted ``lvm batch`` to be idempotent for 3.2 if possible.

roles/ceph-config/tasks/main.yml
andrewschoen: Yeah, all these tasks are in a block that only run on osd nodes. https://github.com/ceph/ceph-ansible/pull/3190/files/497b905dd0acdf6a98d90e8953c57145b4a4732d#diff-408c9554f34e961d163f68b645ee6da3R75

leseb: Oops, missed that.

leseb:
@andrewschoen not sure if you saw https://github.com/ceph/ceph-ansible/pull/2866/commits but as per https://github.com/ceph/ceph-ansible/pull/2866/commits/3354376c44c0e3e9dc7d40a296fd0aacece82dfd the `ceph_volume` module is undergoing a heavy lifting, so it would be nice if you could take a look at it. Ideally, you would also rebase your work on top of it.
Let's sync.

andrewschoen:
jenkins test pipeline

PR #3191, opened 2018-10-03
Convert image tags to string before searching

noahdesu:
Applying tests like "tag | search("unbuntu")" fail with numeric tags.
Convert to string first.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1635823

Signed-off-by: Noah Watkins <nwatkins@redhat.com>


leseb:
Thanks @noahdesu, as a follow up can you also take care of the DEPRECATION WARNING, for similar tasks, see:

```
TASK [ceph-defaults : set_fact ceph_uid for debian based system - container] ***
task path: /home/jenkins-build/build/workspace/ceph-ansible-prs-luminous-docker_cluster/roles/ceph-defaults/tasks/facts.yml:207
Thursday 04 October 2018  03:06:32 +0000 (0:00:00.468)       0:01:12.912 ****** 
[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using 
`result|search` use `result is search`. This feature will be removed in version
 2.9. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
skipping: [mon0] => {
    "changed": false, 
}
```
And:

```
TASK [ceph-docker-common : set_fact ceph_release mimic] ************************
task path: /home/jenkins-build/build/workspace/ceph-ansible-prs-luminous-docker_cluster/roles/ceph-docker-common/tasks/release.yml:20
Thursday 04 October 2018  03:10:11 +0000 (0:00:00.071)       0:04:52.603 ****** 
[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using 
`result|version_compare` use `result is version_compare`. This feature will be 
removed in version 2.9. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
skipping: [mon0] => {
    "changed": false, 
}
```

Thanks in advance.

noahdesu:
@guits it looks like the docker image tag is used before ceph-defaults is evaluated. is there perhaps a better place for that global change?

infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
```diff
@@ -25,6 +25,9 @@
            invoking the playbook"
       when: ireallymeanit != 'yes'
 
+- name: stringify docker image tag
```
guits: wrong indentation?

guits:
@noahdesu you probably have to set that fact in `switch-from-non-containerized-to-containerized-ceph-daemons.yml` as well

noahdesu:
@guits you're right... I copied and pasted irresponsibly. should be fixed now. i added the convert to string task to each play in the playbook. my reading of the ansible docs made it sound like that needed to be done on each host, as the updated fact wouldn't be visible across plays. let me know if you think there is still a better way to do that.

guits:
@noahdesu I think you can put it at L48, that way it would be run on all nodes.

noahdesu:
I reverted this patch set back to piping each tag through `| string`. @guits I tried several of the suggestions you had about doing the conversion once, but even within a single play, overriding a fact (in this case with the type converted version of a variable) didn't seem to work: subsequent tasks always saw the command line specified value, which seems to be inline with the docs on the order of precedence.

I'm still in the process of working my way through various parts of the ansible documentation, so it's not clear how fast I will be able to discover whatever is necessary to make this work the way we expect. If you happen to just know what to change that'd be great, otherwise I think this more verbose patch will have to do for a while.

for example, the last task below won't see the "NEW VERSION" prefix, it sees the version of `ceph_docker_image_tag` specified when invoking the playbook.

```
  - name: stringify docker image tag                                                  
      set_fact:                                                                                                                     
        ceph_docker_image_tag: "NEW VERSION {{ ceph_docker_image_tag }}"                          
                                                                                        
    - name: set_fact ceph_uid for ubuntu                                                      
      set_fact:                                                                                                            
        ceph_uid: 64045                                                                                                             
      when: ceph_docker_image_tag is search("ubuntu")                             
                                                                                                            
    - name: should see the new version                     
      debug:                                                                                                          
        msg: "{{ ceph_docker_image_tag }}" 
```

noahdesu:
hey @guits thanks for looking at this. i am able to get correct behavior like you show above when I define the initial value of the image tag variable in the playbook. the problem comes when the variable is specified from the command line:

```
ansible-playbook test.yml --extra-vars "ceph_docker_image_tag=foobar"
```

which below shows that the new variable value isn't seen. i think this is related to what appears to be the CLI args always taking precedence per the ansible documentation.

```
[nwatkins@localhost ~]$ ansible-playbook -i inven x.yml --extra-vars "ceph_docker_image_tag=foobar"

PLAY [all] *********************************************************************

TASK [Gathering Facts] *********************************************************
ok: [localhost]

TASK [stringify docker image tag] **********************************************
ok: [localhost]

TASK [set_fact ceph_uid for ubuntu] ********************************************
skipping: [localhost]

TASK [should see the new version] **********************************************
ok: [localhost] => {
    "msg": "foobar"
}

PLAY RECAP *********************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=0   

[nwatkins@localhost ~]$ 
```

so i think this issue is still here and it's not clear how to make the patch simpler than pipiing all variable uses through `| string`.

BTW, the current version of this PR is updated to pipe all the variables through `| string` so it is ready anytime.

guits:
@noahdesu that's correct, I can reproduce this behaviour. Let's wait for the CI and it should be good to be merged

PR #3194, opened 2018-10-04
ceph: remove custom cluster name support

leseb:
The dev team decided not to support custom cluster names in the future.
So now, if your cluster variable is different than 'ceph' the playbook
will fail. For existing cluster with custom name you need to set
'use_custom_cluster_name_yes_i_am_sure' to True. This means you know
what you're doing.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1635924

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -33,17 +33,6 @@ dummy:
 # Directory to fetch cluster fsid, keys etc...
 #fetch_directory: fetch/
 
-# The 'cluster' variable determines the name of the cluster.
-# Changing the default value to something else means that you will
-# need to change all the command line calls as well, for example if
-# your cluster name is 'foo':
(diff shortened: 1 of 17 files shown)


leseb:
@gfidente @fultonj I believe OSP will be impacted by this, so I want you to be aware of that. Let me know if you have any concerns. Thanks.

gfidente:
the main reason to support this functionality is to be able to deploy RBDMirror daemons in two different clusters and configure it with the necessary informations about the other instance, I believe this to be important in ceph-ansible even outside openstack context

another reason, more relevant in the openstack scenario though, is that some services might need to connect to multiple ceph clusters and consume each as a different backend, this is especially important for glance in the edge

what are the reasons to remove this functionality instead? and how could the above be resolved then?

PR #3202, opened 2018-10-08
site: use ceph-default to show cluster state

leseb:
We need to play cpeh-defaults before showing that task so if the clsuter
name is 'ceph' then we get the default value from 'cluster' based on the
content of cpeh-defaults.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1636962


guits:
jenkins test luminous-lvm_osds

PR #3205, opened 2018-10-09
ceph-validate: remove versions checks for bluestore and lvm scenario

andrewschoen:
These checks will never pass unless ceph_stable_release is passed and
ceph-defaults is run before ceph-validate. Additionally, we don't want
to support deploying jewel upstream at ceph-ansible master.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1637537


leseb:
CI error un-related.

PR #3220, opened 2018-10-11
ceph-defaults: do not set UID 167 if image tag contains "latest"

berendt:
For an Ubuntu image with "latest" in the tag, the previously set
UID 64045 will be overwritten with the UID 167. This should not be the case.

### DIFF
--- a/roles/ceph-defaults/tasks/facts.yml
+++ b/roles/ceph-defaults/tasks/facts.yml
@@ -216,7 +216,7 @@
     ceph_uid: 167
   when:
     - containerized_deployment
-    - ceph_docker_image_tag | string is search("latest") or ceph_docker_image_tag | string is search("centos") or ceph_docker_image_tag | string is search("fedora")
+    - ceph_docker_image_tag | string is search("centos") or ceph_docker_image_tag | string is search("fedora")
 
 - name: set_fact ceph_uid for red hat
   set_fact:


leseb:
But `latest` is centOS, does that mean you're using a different image?

berendt:
@leseb We simply do not use the images provided on Dockerhub. We use an other image name with an other image tag. I think it makes no sense to assume that a image with the tag "latest" is a CentOS image.

roles/ceph-defaults/tasks/facts.yml
```diff
@@ -216,7 +216,7 @@
     ceph_uid: 167
   when:
     - containerized_deployment
-    - ceph_docker_image_tag | string is search("latest") or ceph_docker_image_tag | string is search("centos") or ceph_docker_image_tag | string is search("fedora")
+    - ceph_docker_image_tag | string is search("centos") or ceph_docker_image_tag | string is search("fedora")
```
leseb:
```suggestion
    - ceph_docker_image_tag | string is search("latest-") or ceph_docker_image_tag | string is search("centos") or ceph_docker_image_tag | string is search("fedora")
```

leseb: Will, that solve your problem?

berendt: @leseb Works for me. We are using ``luminous-latest`` and ``latest`` as our tags.

PR #3223, opened 2018-10-11
[skip ci] ceph-defaults: set ceph_stable_openstack_release_uca to queens

berendt:
Liberty is no longer available in the UCA. The last available release there
is currently Queens.

Signed-off-by: Christian Berendt <berendt@betacloud-solutions.de>


berendt:
I'm not sure if anyone uses it. But Liberty does not make any sense as a default value because it does not exist anymore.

PR #3253, opened 2018-10-19
[skip ci] [DNM]: Create snapshot-aware tox environments

noahdesu:
This is an attempt at speeding up testing for developers by integrating vm snapshotting into the tox test suites. This patch adds the ability to create and snapshot an initial deployment, and then optionally run tests against these snapshots without reproducing the initial deployment.

Usage:

* All previous tox environments should behave identically to before this patch
* Three new factors are added for each of the previous test environments
* {environment}-mksnap: builds the snapshot of the initial deployment
* {environment}-snap: uses the previously built snapshot and runs infra etc..
* {environment}-snap-fast: uses snapshot, but skips all the infra tests

Todo:

* add support for, or disable, for non-libvirt providers
* I haven't figured out how the `cluster_ceph` `shrink_osd` names get injected into the vagrant environment that creates the virtual machine names, so this is currently hard coded for the `cluster` environment. but ideally we also want to expand this name so snapshots for all possible environments can be persisted.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,4 +1,5 @@
 # These are Python requirements needed to run the functional tests
+tox==3.5.0
 testinfra
 pytest-xdist
 pytest==3.6.1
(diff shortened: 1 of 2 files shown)


guits:
No update for a while, closing this.
Feel free to reopen if needed.

PR #3255, opened 2018-10-22
ceph_mgr : systemd reload

c35sys:
Force systemd to reload configuration before trying to start ceph-mgr service, as it doesn't start without it in a node with ceph-mon and ceph-mgr only.


c35sys:
jenkins test pipeline

PR #3260, opened 2018-10-24
[skip ci] ceph_mgr : systemd reload

c35sys:
Force systemd to reload configuration before trying to start ceph-mgr service, as it doesn't start without it in a node with ceph-mon and ceph-mgr only.

### DIFF
--- a/roles/ceph-mgr/tasks/pre_requisite.yml
+++ b/roles/ceph-mgr/tasks/pre_requisite.yml
@@ -32,6 +32,12 @@
     - ceph_mgr_systemd_overrides is defined
     - ansible_service_mgr == 'systemd'
 
+- name: ceph-mgr systemd reload
+  systemd: daemon_reload=yes
+  changed_when: false
+  when:
+- ansible_service_mgr == 'systemd'
+
 - name: start and add that the mgr service to the init sequence
   service:
     name: "ceph-mgr@{{ ansible_hostname }}"


roles/ceph-mgr/tasks/pre_requisite.yml
```diff
@@ -32,6 +32,12 @@
...
 
+- name: ceph-mgr systemd reload
+  systemd: daemon_reload=yes
+  changed_when: false
+  when:
+- ansible_service_mgr == 'systemd'
+
 - name: start and add that the mgr service to the init sequence
   service:
     name: "ceph-mgr@{{ ansible_hostname }}"
```
leseb: add `daemon_reload: yes` here.

roles/ceph-mgr/tasks/pre_requisite.yml
```diff
@@ -32,6 +32,12 @@
     - ceph_mgr_systemd_overrides is defined
     - ansible_service_mgr == 'systemd'
 
+- name: ceph-mgr systemd reload
```
leseb: remove this block

roles/ceph-mgr/tasks/pre_requisite.yml
c35sys: I don't think "daemon_reload" is an option for ansible "service" module, that's why I used a different block using the "systemd" module.
What do you think ?

leseb: Oh right. Can you share your error also because it's not clear that issue this commit fixes. Thanks.

c35sys:
OS : Ubuntu 16.04
Ceph version : luminous

I'm installing ceph on 6 nodes:
- 3 physical nodes for OSDs
- 3 lxc nodes for ceph-mon and ceph-mgr, which are collocated on the same lxc container

When I'm installing ceph without the patch, ceph-mgr can't start, I tried several times and it fails each time. Doing a systemd reload before starting the daemon fixes the problem.

leseb:
@c35sys merci !

PR #3279, opened 2018-10-30
igw fix iscsi/api port setup

mikechristie:
This patchset fixes a couple bugs related to the ceph iscsi gw port setup.

The default port used in the code was 5000 but in some docs and setup files like inf ansible it was 5001. These patches sync it to 5000 so it matches the code in ceph-iscsi-config.

The iscsi target port was also not getting setup in the firewall and if the api_port was something other than the default it was also not getting setup.


mikechristie:
@leseb 

We already had a documentation addition for this issue, so it is not very high priority. If it is too late for 3.2 or you do not have time then just put it in the next release when you have time.

leseb:
@mikechristie no we still have time, I just need a BZ to do the backport :)

roles/ceph-infra/tasks/configure_firewall.yml
```diff
@@ -152,18 +152,35 @@
...
+  firewalld:
+    port: "{{ api_port }}/tcp"
+    zone: "{{ ceph_iscsi_firewall_zone }}"
+    source: "{{ public_network }}"
+    permanent: true
+    immediate: true
+    state: enabled
+  notify: restart firewalld
+  when:
+    - api_port is defined
```
leseb: `api_port` is always defined so we don't need this check.

mikechristie: Are the defaults initialized before the firewall above?

If I do not have that check I get the error:


TASK [ceph-infra : open iscsi api ports] ***************************************
Tuesday 06 November 2018  21:37:15 -0600 (0:00:00.924)       0:00:18.323 ****** 
fatal: [ans6]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'api_port' is undefined\n\nThe error appears to have been in '/root/src/ceph-ansible/roles/ceph-infra/tasks/configure_firewall.yml': line 171, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: open iscsi api ports\n  ^ here\n"}

mikechristie: Ah shoot, have we talked about this before? Do I need to do

port: "{{ api_port | default(5000) }}/tcp"

leseb: My bad `api_port` actually comes from the ceph-iscsi role, I think the `port: "{{ api_port | default(5000) }}/tcp"` is what we want then.

mikechristie: Fix up and repushed.

PR #3287, opened 2018-10-31
validate: ceph_origin and containerized_deployment can not both be set

andrewschoen:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1630975

### DIFF
--- a/plugins/actions/validate.py
+++ b/plugins/actions/validate.py
@@ -48,7 +48,7 @@ class ActionModule(ActionBase):
             notario_store["containerized_deployment"] = host_vars["containerized_deployment"]
             notario.validate(host_vars, install_options, defined_keys=True)
 
-            if host_vars["ceph_origin"] == "repository":
+            if host_vars["ceph_origin"] == "repository" and not host_vars["containerized_deployment"]:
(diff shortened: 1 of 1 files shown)


andrewschoen:
@leseb I realize they are mutually exclusive, but in the case where you're deploying containers you use a different playbook and I believe ``containerized_deployement`` will take precedence right?

I'd really hesitate to add another special downstream all.yml file. This patch lets a downstream user just set ``containerized_deployment: True`` and deploy, not having to worry about using different group_vars. The amount of change needed for option 1 I don't think provides a bunch of value either.

leseb:
@andrewschoen I'm just to trying to avoid confusion and educate the user on how options should be configured. Can't instead fail if both are `ceph_origin` and `containerized_deployement` configured?

leseb:
@andrewschoen looks good! Thanks for the followup :)

andrewschoen:
@leseb ok, I've got this changed. Thanks.

leseb:
@andrewschoen thanks!

leseb:
jenkins test pipeline

leseb:
jenkins test pipeline

PR #3289, opened 2018-10-31
Fix for invalid "changed" property on lvm batch.

Roukoswarf:
It seems like lvm batch bluestore deployment checks incorrectly for the .changed property on ceph command output.

This fix seems to work as intended, but testing/comments required.

### DIFF
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -589,7 +589,6 @@ def run_module():
 
         if not report:
             # if not asking for a report, let's just run the batch command
-            changed = report_result['changed']
             if changed:
                 # Batch prepare the OSD
                 rc, cmd, out, err = exec_command(
--- a/roles/ceph-config/tasks/main.yml
+++ b/roles/ceph-config/tasks/main.yml
@@ -50,7 +50,7 @@
       when:
         - devices | default([]) | length > 0
         - osd_scenario == 'lvm'
-        - (lvm_batch_report.stdout | from_json).changed
+        - lvm_batch_report.changed
 
     - name: run 'ceph-volume lvm list' to see how many osds have already been created
       ceph_volume:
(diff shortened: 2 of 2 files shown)


andrewschoen:
What version of ceph were you trying to deploy? I think the update for ``ceph-volume`` that includes the correct json is not yet released in luminous or mimic.

Roukoswarf:
I am using latest mimic, 13.2.2, which added the batch commands.
I just tested with my changes running the playbook twice, and it seems to try and run the report on the device created last run, resulting in:

`RuntimeError: Unable to use device, already a member of LVM`

Not sure if my changes cause this state issue.

Does does the master branch assume you're running the latest ceph master?

andrewschoen:
Yes, the master branch of ceph-ansible does assume you want to install ceph@master. It has support for luminous and mimic as well, but the commits you need for batch do not exist in mimic yet. This PR that was merged a week or so ago is what you need and it's unreleased. https://github.com/ceph/ceph/pull/24588

leseb:
@Roukoswarf @andrewschoen what's the status on this? The new package should be released by now right?

Roukoswarf:
Well, if master is tracking the ceph master, then this issue is likely invalid and should be closed.
Which stable branch of ceph-ansible is exactly matched to the current mimic stable? I couldn't find one matched specifically.

leseb:
stable-3.* branches use stable ceph releases, so 3.2 or 3.1.

Roukoswarf:
When checking the stable-3.2 branch, the same code is present and will fail in the same way on mimic 13.2.2.

leseb:
@andrewschoen what's your recommendation?

leseb:
@andrewschoen up?

andrewschoen:
@leseb This PR will not work, there's really no issue here. The ``changed`` property has just not made it to any official ceph release yet. If master is installed from dev packages this will all work as expected. I expect this to fail in ``stable-3.2`` as well as it's very close to master, I think ``stable-3.1`` is going to be the best bet for installing a mimic release.

leseb:
@andrewschoen thanks, I'm closing this based on Andrew's comment.

PR #3296, opened 2018-11-01
ceph-common: update_cache whenever a new deb repo is added

alfredodeza:
The use of a handler meant that the cache would be updated at the very
end of the play, which doesn't work when adding a development repo and
trying to install right after it. This mostly reverts
53cdddf88699263763b36643565e5f846d9d13a8 without an actual `git revert`
because that caused other conflicts.

PR that caused the breakage: https://github.com/ceph/ceph-ansible/pull/3277

### DIFF
--- a/roles/ceph-common/handlers/main.yml
+++ b/roles/ceph-common/handlers/main.yml
@@ -1,8 +1,0 @@
----
-- name: update apt cache if a repo was added
-  apt:
-    update_cache: yes
-  register: update_apt_cache
-  retries: 5
-  delay: 2
-  until: update_apt_cache | success
--- a/roles/ceph-common/tasks/installs/debian_community_repository.yml
+++ b/roles/ceph-common/tasks/installs/debian_community_repository.yml
@@ -8,5 +8,4 @@
   apt_repository:
     repo: "deb {{ ceph_stable_repo }} {{ ceph_stable_distro_source | default(ansible_lsb.codename) }} main"
     state: present
-    update_cache: no
-  notify: update apt cache if a repo was added
+    update_cache: yes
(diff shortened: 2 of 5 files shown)


alfredodeza:
Given that there are no Debian tests for PRs I don't know why we should wait for these tests to finish since they are Debian only

rishabh-d-dave:
Perhaps flushing the handler would be a simpler change?

alfredodeza:
@rishabh-d-dave why would flushing the handler be better?

rishabh-d-dave:
It would be better because that would not update cache at every task. Using the handler would allow us to update the cache at the end or when we flush the handlers. This way I think it would be a bit more efficient.

alfredodeza:
@rishabh-d-dave this change wouldn't update the cache at every task, only when there is a change. The docs state that:

> Run the equivalent of apt-get update when a change occurs. Cache updates are run after making changes.

https://docs.ansible.com/ansible/latest/modules/apt_repository_module.html

This is exactly the behavior that we want

andrewschoen:
@leseb 

> Also, you don't need to update the cache each time you run through that task, only the task has changed. So forcing update_cache: yes is wrong.

This isn't my understanding of how ``update_cache: yes`` works, it doesn't update each time you run the task. The docs state it only updates the cache if the repository changes.

https://docs.ansible.com/ansible/latest/modules/apt_repository_module.html

Flushing the handler could work, but I'm wondering why we need the handler at all if ``update_cache`` works as it's documented.

leseb:
@andrewschoen right, if the `update_cache: yes` works as documented we don't need the handler. I read too fast the doc link, I guess that what happens when checking your work while on PTO :).

PR #3303, opened 2018-11-07
igw: stop tcmu-runner on iscsi purge

mikechristie:
When the iscsi purge playbook is run we stop the gw and api daemons but
not tcmu-runner which I forgot on the previous PR.

Fixes Red Hat BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1621255


mikechristie:
@leseb 

I repushed the PR with your comment handled and I also merged it with both the rbd-target-api and rbd-target-gw tasks since they the api one was also a duplicate.

PR #3319, opened 2018-11-09
[skip ci] Extends check_devices tasks to non-collocated an lvm-batch scenarios

VasishtaShastry:
Extends check_devices tasks to non-collocated an lvm-batch scenarios
Tuned name of a task and error message to make it more user understandable

Fixes BZ 1648168 - ceph-validate : devices are not validated in non-collocated and lvm_batch scenario
Signed-off-by: VasishtaShastry <vipin.indiasmg@gmail.com>


roles/ceph-validate/tasks/main.yml
```diff
@@ -52,7 +52,7 @@
   when:
     - osd_group_name in group_names
     - not osd_auto_discovery | default(False)
-    - osd_scenario != "lvm" and devices is not defined
+    - osd_scenario != "lvm" or (osd_scenario == "lvm" and devices is defined)
```
andrewschoen: I think it'd be safe to just say ``- devices is defined`` here.

roles/ceph-validate/tasks/check_devices.yml
```diff
@@ -30,9 +30,9 @@
...
-- name: fail if dedicated_device is not the size length as devices
+- name: fail if  number of dedicated_devices is not equal to number of devices
   fail:
-    msg: "dedicated_device lengtth must be identical to devices's length"
+    msg: "Number of dedicated_devices must be equal to number of devices"
   when:
     - osd_scenario == 'non-collocated'
-    - dedicated_devices|length != devices|length
\ No newline at end of file
+    - dedicated_devices|length != devices|length
```
andrewschoen: If ``dedicated_devices`` isn't defined will this error? Or will that first condition of ``osd_scenario == 'non-collocated'`` keep that from happening?

roles/ceph-validate/tasks/main.yml
VasishtaShastry: Ya, We can simplify as you have mentioned, just that when user chooses to use user created lvm_volumes, these tasks will be redundant
It should be fine

roles/ceph-validate/tasks/check_devices.yml
VasishtaShastry: If dedicated_devices are not defined, previous task fails -  validate dedicated_device is/are actually device(s)

andrewschoen: I suspected as much. I think you'll need to add a `` - dedicated_devices is defined`` to the ``when`` clauses in all the tasks that use it.

roles/ceph-validate/tasks/main.yml
andrewschoen: In that case ``devices`` wouldn't be defined though, or shouldn't be. This works though.

roles/ceph-validate/tasks/check_devices.yml
andrewschoen: That wouldn't work for all cases because the ``lvm`` scenario could have ``devices`` defined and not ``dedicated_devices``. Actually, I had thought that we'd removed all ceph-disk methods of creating OSDs from master. If that's the case we shouldn't need ``dedicated_devices`` at all anymore.

VasishtaShastry: > That wouldn't work for all cases because the `lvm` scenario could have `devices` defined and not `dedicated_devices`.

I think it might work as we are invoking `non_collocated_osd_scenario` only when osd scenario is non-collocated -
https://github.com/ceph/ceph-ansible/blob/master/plugins/actions/validate.py#L87-L88

VasishtaShastry: > I think the problem will be that even if you change that validator the tasks in `check_devices.yml` will still fail when `osd_scenario` is `lvm` and `dedicated_devices` isn't provided. You need to only run the tasks in `check_devices.yml` that check `dedicated_devices` only when they are defined.
check_devices tasks are not failing as we have included the condition in when - `- osd_scenario == 'non-collocated'`

> I don't believe that `dedicated_devices` is even required for `non-collocated` as `dedicated_devices` will be defaulted to `devices` if it's not given. http://docs.ceph.com/ceph-ansible/master/osds/scenarios.html#osd-scenario-non-collocated

If `dedicated_devices` are not defined, playbook is failing in ceph-osd task `set_fact build final dedicated_devices list` saying `'dedicated_devices_tmp' is undefined` (Maybe a bug) -
https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/build_devices.yml#L32

When I had tried non-collocated scenario without defining dedicated_devices long ago, yes it was defaulted to collocated scenario.
I think it is nice if we validate `dedicated_devices` as long as we support luminous, please let me know your views.

roles/ceph-validate/tasks/main.yml
VasishtaShastry: > I think it'd be safe to just say `- devices is defined` here.

Done - 4d17f8b573433d24b00e5d71b6d1ca2306cb416d

roles/ceph-validate/tasks/check_devices.yml
VasishtaShastry: > I suspected as much. I think you'll need to add a ` - dedicated_devices is defined` to the `when` clauses in all the tasks that use it.

Done - 4d17f8b573433d24b00e5d71b6d1ca2306cb416d

VasishtaShastry:
As there were too many commits, squashed and created new PR - 3354

PR #3320, opened 2018-11-09
ceph-nfs: use mkdir instead of /usr/bin/mkdir

berendt:
On Ubuntu mkdir lives in /bin and not in /usr/bin

### DIFF
--- a/roles/ceph-nfs/templates/ceph-nfs.service.j2
+++ b/roles/ceph-nfs/templates/ceph-nfs.service.j2
@@ -6,7 +6,7 @@ After=docker.service
 [Service]
 EnvironmentFile=-/etc/environment
 ExecStartPre=-/usr/bin/docker rm ceph-nfs-%i
-ExecStartPre=/usr/bin/mkdir -p /etc/ceph /etc/ganesha /var/lib/nfs/ganesha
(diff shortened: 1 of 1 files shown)


roles/ceph-nfs/templates/ceph-nfs.service.j2
```diff
@@ -6,7 +6,7 @@ After=docker.service
 [Service]
 EnvironmentFile=-/etc/environment
 ExecStartPre=-/usr/bin/docker rm ceph-nfs-%i
-ExecStartPre=/usr/bin/mkdir -p /etc/ceph /etc/ganesha /var/lib/nfs/ganesha
+ExecStartPre=mkdir -p /etc/ceph /etc/ganesha /var/lib/nfs/ganesha
```
leseb: Typically systemd is not happy if you don't pass full binary name, not sure why though.
Perhaps you could use jinja here to apply the right binary path, whether we run on Debian system or Red Hat system?

berendt: Will fix this soon.

guits:
Closed in favor of #3675 (because of inactivity)

berendt:
@guits Thank you for the takeover and solution. At the moment I am not able to work through any open points.

PR #3324, opened 2018-11-13
[skip ci] switch: do not mask systemd unit

leseb:
If we mask it we won't be able to start the OSD container since now the
osd container use the osd ID as a name such as: ceph-osd@0

Fixes the error:  Failed to execute operation: Cannot send after transport endpoint shutdown

### DIFF
--- a/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
+++ b/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
@@ -239,7 +239,6 @@
         name: "{{ item }}"
         state: stopped
         enabled: no
-        masked: yes
       with_items: "{{ running_osds.stdout_lines | default([])}}"
       when: running_osds != []
 
--- a/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
+++ b/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
@@ -218,23 +218,7 @@
       changed_when: false
       failed_when: false
 
-    - name: collect osd devices
-      shell: |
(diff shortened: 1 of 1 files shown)


leseb:
Need BZ

leseb:
jenkins test pipeline

leseb:
jenkins test pipeline

leseb:
jenkins test pipeline

infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
```diff
@@ -281,24 +274,22 @@
...
+    - name: get osd directories
+      command: ls -1 /var/lib/ceph/osd
+      register: osd_dirs
       changed_when: false
-      with_items: "{{ devices }}"
-      register: devices_canonicalize
 
     - name: unmount all the osd directories
-      command: umount "{{ item.0.stdout }}"1
+      command: umount "{{ item.0.stdout }}"
```
guits: taking my comment in consideration for L292, I think you just need `{{ item.0 }}`
+ you will need to prefix your `umount` command with `/var/lib/ceph/osd/` I think

infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
```diff
@@ -281,24 +274,22 @@
...
       changed_when: false
       failed_when: false
       with_together:
-        - "{{ devices_canonicalize.results }}"
+        - "{{ osd_dirs.stdout }}"
```
guits: `{{ osd_dirs.stdout }}` is a string like :

`ceph-osd1\nceph-osd2\nceph-osd3` right?

don't you want `osd_dirs.stdout_lines` which is an array like `['ceph-osd1', 'ceph-osd2', 'ceph-osd3']` instead ?

PR #3326, opened 2018-11-14
validate: expand all jinja2 templates before validating them

rishabh-d-dave:
Signed-off-by: Rishabh Dave <ridave@redhat.com>


plugins/actions/validate.py
```diff
@@ -121,6 +122,29 @@ def run(self, tmp=None, task_vars=None):
...
+            if expanded_var == var:
+                if not isinstance(expanded_var, string_types):
+                    raise AnsibleUndefinedVariable
+                expanded_var = self._templar.template("{{" + expanded_var + \
+                                                      "}}", convert_bare=True,
+                                                      fail_on_undefined=True)
+        except AnsibleUndefinedVariable as e:
+            expanded_var = u"VARIABLE IS NOT DEFINED!"
+            if self._display.verbosity > 0:
+                expanded_var += u": %s" % str(e)
```
alfredodeza: same here, this isn't a safe string substitution because of the addition. Unless you are absolutely certain this will never blow up because the values are always ensured to be strings

alfredodeza: This is a bit confusing, is the `AnsibleUndefinedVariable` exception you are catching here meant to be the one being raised a few lines above? or is this expecting to find that exception coming from other parts? If the former, it might be better to address the problem right in that block, vs. raising the exception to catch it three lines later

rishabh-d-dave: Yeah, it can be raised by `self._templar.template()` or `self._templar.is_template()`. I am looking at the code for the debug module of Ansible to write this patch.

Umm... now that reflect on this, I think I should catch the exception in `expand_all_jinja2_template()` since it can also be raised by `self._templar.is_template()`.

plugins/actions/validate.py
```diff
@@ -121,6 +122,29 @@ def run(self, tmp=None, task_vars=None):
...
+                variables[k] = self.expand_jinja2_template(v)
+        return variables
+
+    def expand_jinja2_template(self, var):
+        try:
+            expanded_var = self._templar.template(var, convert_bare=True,
+                                                  fail_on_undefined=True)
+            if expanded_var == var:
+                if not isinstance(expanded_var, string_types):
+                    raise AnsibleUndefinedVariable
```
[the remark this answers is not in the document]
rishabh-d-dave: Sorry, I don't understand what do you mean by 'bare' exception.

It's being done in the same way in debug module of Ansible (at which I've been looking to know how to expand the jinja2 templates in plugins). This approach looks OK to me.

rishabh-d-dave:
@alfredodeza 
Updated the patch -- stopped using `+` for strings, moved the statement for catching AnsibleUndefinedException to a better location and removed a couple of LoCs which I thought were redundant.

leseb:
@alfredodeza anything else from you? Please approve if you're ok with that change, thanks!

alfredodeza:
The Python part looks OK to me. Unsure what the implication of the change means for ceph-ansible

leseb:
jenkins test pipeline

PR #3333, opened 2018-11-15
infra: don't restart firewalld if unit is masked

guits:
if firewalld.service systemd unit is masked, the handler will fail when
trying to restart it.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1650281


guits:
jenkins test pipeline

PR #3340, opened 2018-11-19
osd_memory_target: standardize unit and fix calculation

neha-ojha:
* The default value of osd_memory_target used by ceph is 4294967296 bytes,
so use the same as ceph-ansible default.

* Convert ansible_memtotal_mb to bytes to calculate osd_memory_target

Fixes bug in https://github.com/ceph/ceph-ansible/pull/3113
Signed-off-by: Neha Ojha <nojha@redhat.com>

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -360,7 +360,7 @@ dummy:
 #is_hci: false
 #hci_safety_factor: 0.2
 #non_hci_safety_factor: 0.7
-#osd_memory_target: 4000000000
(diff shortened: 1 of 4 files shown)


leseb:
jenkins test pipeline

leseb:
@neha-ojha we need a BZ for this bug, thanks in advance.

neha-ojha:
@leseb can we use https://bugzilla.redhat.com/show_bug.cgi?id=1595003?

leseb:
@neha-ojha probably

PR #3345, opened 2018-11-19
[skip ci] Discover rbd facts.

ProgVal:
Example "output":

```json
        "rbd_images": {
            "rbd-root-images": {
                "base-debian-stretch": {
                    "info": {
                        "block_name_prefix": "rbd_data.1b14046b8b4567", 
                        "create_timestamp": "Mon Nov 19 20:38:01 2018", 
                        "features": [
                            "layering", 
                            "exclusive-lock", 
                            "object-map", 
                            "fast-diff", 
                            "deep-flatten"
                        ], 
                        "flags": [], 
                        "format": 2, 
                        "name": "base-debian-stretch", 
                        "object_size": 4194304, 
                        "objects": 25600, 
                        "order": 22, 
                        "size": 107374182400
                    }, 
                    "status": {
                        "watchers": []
                    }
                }
            }
        }
```

I only tested this on Ansible 2.2 and with my own config; it may not be very robust.

### DIFF
--- a/library/ceph_facts
+++ b/library/ceph_facts
@@ -127,6 +127,34 @@ def run_ceph_facts(module):
         for (k,v) in rados_df_ds.items():
             setup_options["rados_df_%s" % k] = v
 
+    setup_options['rbd_images'] = {}
+    for pool in setup_options.get('osd_status_pools', []):
+        if 'rbd' in pool['application_metadata']:
+            pool_name = pool['pool_name']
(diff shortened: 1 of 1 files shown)


leseb:
This is not triggered anywhere in the ceph-ansible code. We are not using it we might remove it, but I didn't know people were using it.
I'm going to merge it as is since we don't have any coverage and it might be beneficial to you.

Thanks for the contribution.

PR #3346, opened 2018-11-19
start_osds: Use list instead of keys

b-ranto:
If you use python3 based ansible then keys() returns a dict_keys object,
not a list of keys. This breaks the installation on such a system. Using
the list filter provides a more robust solution that should work on both
python2 and python3 based ansible. You can find some more information
about the issue, here:

https://github.com/ansible/ansible/issues/19514

### DIFF
--- a/roles/ceph-defaults/tasks/facts.yml
+++ b/roles/ceph-defaults/tasks/facts.yml
@@ -228,7 +228,7 @@
 - name: set_fact rgw_hostname
   set_fact:
     rgw_hostname: "{% set _value = ansible_hostname -%}
-    {% for key in ceph_current_status['servicemap']['services']['rgw']['daemons'].keys() -%}
+    {% for key in (ceph_current_status['servicemap']['services']['rgw']['daemons'] | list) -%}
     {% if key == ansible_fqdn -%}
     {% set _value = key -%}
     {% endif -%}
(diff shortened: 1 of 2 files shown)


guits:
@b-ranto 

we might face the same issue [here](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/tasks/facts.yml#L231) right?

b-ranto:
@guits I hit this while running ceph-ansible on my (recently updated) Fedora 29 machine that comes with ansible 2.7.x that uses python3 in the background. The bits that you are referring to did not cause the playbook to fail for me but I suppose, the lookup no longer works as intended on my machine.

leseb:
Weird it's running fine here on python3: https://2.jenkins.ceph.com/job/ceph-ansible-prs-dev-container_podman/28/

b-ranto:
@leseb While python2 is still default on Fedora 29, ansible does use python3, not python2 by default, now. I had to install a couple of python3 dependencies for it to even work (e.g. python3-notario, python3-netaddr). I guess it is continuation of the attempt to move towards python 3 as a default in Fedora.

This was not the case in Fedora 28, the ansible there still used python2 by default. You can find some information on it, here:

https://fedoraproject.org/wiki/Changes/Ansible_python3_default

guits:
I would say I don't see any reason to block this PR.
I think In any case it is still worth it to have something that ensures it will work with both python2 and 3

leseb:
@b-ranto strange, they might have done something different for F29 Atomic then... +1 on merging this.

PR #3358, opened 2018-11-26
rgw: add support for multiple rgw instances on a single host

guihecheng:
With this, we could have multiple rgw instances on a single host
with a single run, don't have to use rgw-standalone.yml which does not
seems able to bind ports separately.
And if you don't want to have multiple rgw instances, just do not
define 'rgw_instances'.
Not compatible with Multi-Site yet.

Signed-off-by: guihecheng <guihecheng@cmiot.chinamobile.com>

### DIFF
--- a/roles/ceph-facts/tasks/facts.yml
+++ b/roles/ceph-facts/tasks/facts.yml
@@ -269,3 +269,10 @@
   import_tasks: set_radosgw_address.yml
   when:
     - inventory_hostname in groups.get(rgw_group_name, [])
+
+- name: set_fact rgw_instances
+  set_fact:
+    rgw_instances: "{{ rgw_instances|default([]) | union([{'instance_name': 'rgw' + item|string, 'radosgw_address': _radosgw_address, 'radosgw_frontend_port': radosgw_frontend_port|int + item|int}]) }}"
+  with_sequence: start=0 end={{ radosgw_num_instances|int - 1 }}
+  when:
+    - inventory_hostname in groups.get(rgw_group_name, [])
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -405,6 +405,7 @@ dummy:
 #radosgw_address: 0.0.0.0
 #radosgw_address_block: subnet
 #radosgw_keystone_ssl: false # activate this when using keystone PKI keys
+#radosgw_num_instances: 1
 # Rados Gateway options
 #email_address: foo@bar.com
 
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -109,7 +109,7 @@
 
   - name: stop ceph rgws with systemd
     service:
-      name: ceph-radosgw@rgw.{{ ansible_hostname }}
+      name: ceph-radosgw@rgw.*
       state: stopped
       enabled: no
     failed_when: false
--- a/infrastructure-playbooks/purge-docker-cluster.yml
+++ b/infrastructure-playbooks/purge-docker-cluster.yml
@@ -180,7 +180,7 @@
 
   - name: disable ceph rgw service (new unit name)
     service:
-      name: "ceph-radosgw@rgw.{{ ansible_hostname }}"
+      name: "ceph-radosgw@*"
       state: stopped
       enabled: no
     ignore_errors: true
@@ -188,7 +188,7 @@
   - name: remove ceph rgw container
     docker_container:
       image: "{{ ceph_docker_registry }}/{{ ceph_docker_image }}:{{ ceph_docker_image_tag }}"
-      name: "ceph-rgw-{{ ansible_hostname }}"
+      name: "ceph-rgw-{{ ansible_hostname }}-*"
       state: absent
     ignore_errors: true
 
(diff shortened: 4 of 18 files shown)


guihecheng:
@leseb thanks for your comments, and I think only use the  'rgw_instances' and set it default to 1 is possible, I'll update with this idea and try to and the test stuff as you pointed.

leseb:
@guihecheng thanks

guihecheng:
@leseb updated a single variable version, s/rgw_instances/radosgw_instances/g, thanks :)

leseb:
jenkins test dev-docker_cluster

leseb:
@guihecheng please rebase, thanks.

guihecheng:
@leseb ok, I'll rebase soon.

guihecheng:
@leseb patch rebased, but I found that the jenkins job failed, but this works in my own environment, is it related to the ansible version? I am using ansible2.7.2.

leseb:
The job ran on 2.7.4. Looks like a jinja error.

guihecheng:
@leseb sorry, but I failed to reproduce the problem reflected in the CI test, my env is very close to the testing env:
CentOS 7.5(1804)
ansible 2.7.4
I googled a lot about the ERROR : 
`coercing to Unicode: need string or buffer, list found`
But I didn't got any workable solutions, It seems to point to the set_fact line above
[https://github.com/ceph/ceph-ansible/blob/b13301ef90df6a68f8d270b478421bdc8fea4555/roles/ceph-facts/tasks/facts.yml#L275](url)
Any ideas on this?

guits:
@guihecheng please, share the full log with verbose mode

guihecheng:
@guits below is the error log from the CI result, I haven't reproduced in my own environment yet.
https://2.jenkins.ceph.com/job/ceph-ansible-prs-dev-purge_cluster_non_container/15/consoleFull#-729725784b5dd38fa-a56e-4233-a5ca-584604e56e3a

roles/ceph-facts/tasks/facts.yml
```diff
@@ -269,3 +269,8 @@
   import_tasks: set_radosgw_address.yml
   when:
     - inventory_hostname in groups.get(rgw_group_name, [])
+
+- name: set_fact rgw_instances
+  set_fact:
+    rgw_instances: "{{ rgw_instances|default([]) | union([{'instance_name': 'rgw' + item|string, 'radosgw_address': _radosgw_address, 'radosgw_frontend_port': radosgw_frontend_port|int + item|int}]) }}"
```
guits: s/rgw_instances/radosgw_instances ?

the default value isn't needed anyway since you have a default in`ceph-facts/defaults/main.yml`

guihecheng: em....the meanings of the 2 variables are different, the `radosgw_instances` is the number of rgw instances, default to 1, and `rgw_instances` is a list of key-value parameters of each rgw instance.
Here I want to induce the list `rgw_instances`.

guihecheng: yes, I shall rename 1 of these 2 to make it clear.

guits: I think you are missing here a `when: inventory_hostname in groups.get(rgw_group_name, [])` on this task.

guits: `_radosgw_address` is set on rgw nodes. In the log I can see it undefined because you are running on a monitor node. I'm not sure it's the cause of your issue but it's still something you need to address.

guihecheng: @guits oh, thanks for your reminding, these are my missing points, will fix these soon :)

guihecheng:
Here the issue above is solved by adopting the advices from @guits, thanks very much!
And it seems I should fix the test functions, since the rgw instances' names are changed by this patch.

guihecheng:
e....so the non-container stuff passed the tests, now I am going to add the containerized support.

guihecheng:
hi,@leseb @guits, here I've got an failed case , and it points to a failed ceph-rgw container.
As I detected, the container failed after startup and will restart again because of the `Restart=always` config.
I'm now checking why the ceph-rgw container failed.

guihecheng:
OK, I shall fix the rgw restart script.

guihecheng:
@leseb @guits hi, the recent commit hit a failure on a test of osds, could you please help with this problem, maybe the testing environment is not so clean? Thanks :)

guihecheng:
@guits @leseb hi, maintainers, the last round failed at the podman scenario with a host unreachable error, I feel that is should be a environment error ?

leseb:
@guihecheng you're right, the problem is not in your code. Let me do a last review. Thanks for your patience :)

roles/ceph-facts/tasks/facts.yml
leseb: @guihecheng has this been addressed?

guihecheng: @leseb yes, this is solved. As guits pointed, the case is the missing `when: inventory_hostname in groups.get(rgw_group_name, [])` that cause an mon host to set rgw_instances, so we got the original error.

guits:
let's merge this anyway
thanks for your patience @guihecheng

guits:
@guihecheng thanks for this contribution

PR #3360, opened 2018-11-26
osd: expose udev into the container

leseb:
In order to be able to retrieve udev information, we must expose its
socket. As per, https://github.com/ceph/ceph/pull/25201 ceph-volume will
start consuming udev output.


leseb:
jenkins test pipeline

PR #3386, opened 2018-11-29
config: write jinja comment with appropriate syntax

guits:
jinja comment should be written using the jinja syntax `{# ... #}`

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1654441


guits:
The condition L17 aims to put those comments, if we make them as jinja comments, the condition becomes useless.
The next stable release will be intended to deploy only nautilus and after, right? If yes, It means we can remove this condition I guess.

guits:
jenkins test pipeline

PR #3389, opened 2018-11-29
rolling_update: do not fail on missing keys

leseb:
We don't want to fail on key that are not present since they will get
created after the mons are updated. They will be created by the task
"create potentially missing keys (rbd and rbd-mirror)".

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1650572


ktdreyer:
What was the stable-3.2 change for this? There's no cherry-picked-from line.

PR #3393, opened 2018-11-29
rgw users/buckets module

Daniel-Pivonka:
module to bulk create rgw users and buckets


Daniel-Pivonka:
sorry i should have read the contribution file. As for unit test i don't have any currently i can work on that, the module depends on a rgw host to do anything so ill need to deal with that. I have an example playbook. where should i put it? the infrastructure-playbooks directory?

PR #3419, opened 2018-12-04
tox: add missing ceph_docker_image_tag in shrink

leseb:
When calling shrink on containerized deployment, we were first doing the
setup with `latest-master` and then when calling the playbook we were
using the default value for `ceph_docker_image_tag` that comes from
ceph-defaults. Now we pass
`ceph_docker_image_tag={env:CEPH_DOCKER_IMAGE_TAG:latest-master}` so the
play will run the right container image.


leseb:
jenkins test dev-shrink_osd_container

Issue #3421, opened 2018-12-06
centos7_cluster job crashes locally with an error


rishabh-d-dave:
**Bug Report**

What happened: Running a  CI job locally for luminous against centos7_scenario leads to an error with the following traceback -

```
The full traceback is:
Traceback (most recent call last):
  File "<stdin>", line 113, in <module>
  File "<stdin>", line 105, in _ansiballz_main
  File "<stdin>", line 48, in invoke_module
  File "/tmp/ansible_ceph_key_payload_SjfPup/__main__.py", line 714, in <module>
  File "/tmp/ansible_ceph_key_payload_SjfPup/__main__.py", line 710, in main
  File "/tmp/ansible_ceph_key_payload_SjfPup/__main__.py", line 656, in run_module
  File "/tmp/ansible_ceph_key_payload_SjfPup/__main__.py", line 475, in lookup_ceph_initial_entities
NameError: global name 'module' is not defined

fatal: [ceph-mon1]: FAILED! => {
    "changed": false, 
    "rc": 1
}

MSG:

MODULE FAILURE
See stdout/stderr for the exact error


MODULE_STDERR:

Traceback (most recent call last):
  File "<stdin>", line 113, in <module>
  File "<stdin>", line 105, in _ansiballz_main
  File "<stdin>", line 48, in invoke_module
  File "/tmp/ansible_ceph_key_payload_SjfPup/__main__.py", line 714, in <module>
  File "/tmp/ansible_ceph_key_payload_SjfPup/__main__.py", line 710, in main
  File "/tmp/ansible_ceph_key_payload_SjfPup/__main__.py", line 656, in run_module
  File "/tmp/ansible_ceph_key_payload_SjfPup/__main__.py", line 475, in lookup_ceph_initial_entities
NameError: global name 'module' is not defined
```

What you expected to happen: The job should've completed running successfully.

How to reproduce it (minimal and precise): Running following command with repo root as CWD should be enough - `sudo ANSIBLE_FORCE_COLOR=TRUE RELEASE=luminous SCENARIO=centos7_cluster WORKSPACE=./ bash -x ./ceph-ansible-prs.sh | tee ~/centos7_cluster.log`

rishabh-d-dave:
[...]

PR #3423, opened 2018-12-06
ceph_keys: pass in module for error messages

noahdesu:
fixes: #3421

it looks like this will need to be backported too, but I'm not sure how that process works. Are there docs for that?

### DIFF
--- a/library/ceph_key.py
+++ b/library/ceph_key.py
@@ -459,7 +459,7 @@ def exec_commands(module, cmd_list):
     return rc, cmd, out, err
 
 
-def lookup_ceph_initial_entities(out):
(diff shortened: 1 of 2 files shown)


leseb:
@noahdesu for backports just assign the right label :)

leseb:
@noahdesu we just need a BZ for the downstream backport.

noahdesu:
donzo

noahdesu:
seems like travis needs to be kicked off again, but that usually happens even with a force push. not sure what's up with that.

noahdesu:
well, apparently i should just try refreshing the page.

PR #3425, opened 2018-12-07
ceph-defaults: do not use podman only on atomic

leseb:
We want to test podman on f29 non-atomic, atomic is not a hard
requirement. However, if you want to get podman then you will have to
install it first before running the playbook.


leseb:
CI failure is unrelated, podman passed independently of the pipeline, which is ok.

PR #3433, opened 2018-12-10
ensure at least one osd is up

dwaiting:
The existing task checks that the number of OSDs is equal to the number of up OSDs before continuing.

The problem is that if none of the OSDs have been discovered yet, the task will exit immediately and subsequent pool creation will fail (num_osds = 0, num_up_osds = 0).

This is related to Bugzilla 1578086.

In this change, we also check that at least one OSD is present. In our testing, this results in the task correctly waiting for all OSDs to come up before continuing.


gfidente:
check-rdo

gfidente:
check-rdo

dwaiting:
Still seeing this problem in 3.2.0 and it's causing our OpenStack builds to fail. Can we accept this PR or get some other fix?

guits:
@dwaiting sorry for the delay. I'll take a look at this in high priority

PR #3436, opened 2018-12-11
purge-cluster: skip tasks that use ceph-volume if it's not installed

andrewschoen:
This will allow the playbook to be idempotent.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1656935


infrastructure-playbooks/purge-cluster.yml
```diff
@@ -303,6 +303,11 @@
     failed_when: false
     register: ceph_lockbox_partition_to_erase_path
 
+  - name: see if ceph-volume is installed
+    shell: "which ceph-volume"
```
leseb: why not using `command` instead? Also using `command -v` is generally preferred.

guits: @andrewschoen I think @leseb is means 
```
- name: see if ceph-volume is installed
  command: command -v ceph-volume
```

eg (in a shell) :

```
# command -v ceph-volume
/sbin/ceph-volume
#
```

guits: I'm still curious why `command -v` would be better than `which ceph-volume` though

leseb: https://github.com/koalaman/shellcheck/wiki/SC2230

@andrewschoen please use `command -v ceph-volume`. Thanks

andrewschoen: @leseb ok, I've made that change.

PR #3437, opened 2018-12-11
[skip ci] mergify: trigger rdo ci on each PR

leseb:
mergify will now put a comment on a each PR with 'check-rdo' so this
will trigger the RDO CI to test the ceph-ansible PR.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -84,3 +84,11 @@ pull_request_rules:
     conditions:
       - label=backport-stable-3.2
     name: backport stable-3.2
+
+# this will trigger the RDO CI to test the ceph-ansible PR
+  - name: check rdo
+    conditions:
+      - head=master
+    actions:
+      comment:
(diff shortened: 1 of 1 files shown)


leseb:
@gfidente @fultonj whenever you're ready approve this

.mergify.yml
```diff
@@ -84,3 +84,11 @@ pull_request_rules:
     conditions:
       - label=backport-stable-3.2
     name: backport stable-3.2
+
+# this will trigger the RDO CI to test the ceph-ansible PR
+  - name: check rdo
+    conditions:
+      - head=master
```
gfidente: this should work on stable-3.2 as well

gfidente:
I don't think we need this yet, it is probably possible to trigger rdo on all PRs from SF but it's certainly good to know how... especially if this can filter the submissions to be checked basing on some random PR properties

PR #3442, opened 2018-12-12
purge-container: move facts gathering after ceph-defaults role import

guits:
This task has to be called after the role `ceph-defaults` has been
played, otherwise, `mon_group_name` will never be known.

### DIFF
--- a/infrastructure-playbooks/purge-docker-cluster.yml
+++ b/infrastructure-playbooks/purge-docker-cluster.yml
@@ -308,18 +308,17 @@
   gather_facts: true
   become: true
 
-  pre_tasks:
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/purge-docker-cluster.yml
```diff
@@ -325,6 +318,12 @@
       name: ceph-facts
       private: false
 
+  - name: gather monitors facts
```
leseb: Don't we need this before the ceph-facts role?
It was the initial issue AFAIR.

guits: correct

Issue #3445, opened 2018-12-13
Aliases in ceph-aliases.sh.j2 need to be renamed or removed


idryomov:
When "rbd map" is invoked on a host with a ceph-mon container on it, it gets redirected to that container.  However, "rbd map" requires privileges: on top of dealing with udev and device nodes, it has to be able to add a key to the keyring, which is locked down by docker's default seccomp profile.

```
# rbd map pool/img
rbd: failed to add secret 'client.admin' to kernel
In some cases useful info is found in syslog - try "dmesg | tail".
rbd: map failed: (1) Operation not permitted

# type rbd
rbd is a function
rbd ()
{
    sudo docker exec ceph-mon-c10-h19-r730xd rbd --cluster ceph ${@}
}
```

The silent aliasing has resulted in some profound confusion and is bound to cause more in the future, not just with "rbd map".  Version mismatches, subtle permission and file access issues, capabilities, seccomp, etc.

The original PR (https://github.com/ceph/ceph-ansible/pull/2426) sited [1] as an example, but these etcdctl wrappers have their own names and don't alias the actual etcdctl.  Neither do they silently exec into a container...

[1] https://github.com/openshift/openshift-ansible/blob/release-3.7/roles/etcd/etcdctl.sh

leseb:
I believe the `ceph` and `radosgw-admin` commands will still be useful, perhaps we should remove `rbd` and `rados` only?

idryomov:
I think they should all be renamed or removed.  Redirecting into a container is just too sneaky:

```
# ceph osd getcrushmap -o /tmp/cmap.bin
47
# stat /tmp/cmap.bin
stat: cannot stat ‘/tmp/cmap.bin’: No such file or directory
```

leseb:
Yes because the file is in the container...
Ok we can remove that I suppose.

PR #3447, opened 2018-12-13
mon: remove ceph aliases for containers

leseb:
These aliases have led to several issues making believe that ceph
binaries are actually present on the host when running the command.
However it wasn't explicit that the commands were only ran inside a
container.
It has brought to much confusion so we decided to remove them.

Closes: https://github.com/ceph/ceph-ansible/issues/3445


guits:
jenkins test pipeline

gfidente:
@guits I think we have an issue with this change for pre-existing deployments where the .sh file was initially deployed but never removed ... especially with the migration to podman the alias don't work at all

can we try removing it via rolling_update ?

guits:
@gfidente see https://github.com/ceph/ceph-ansible/pull/6906

PR #3463, opened 2018-12-21
add 'custom' as valid ceph_repository value

jtriley:
Validation currently fails when `ceph_repository: custom` even though this is a valid choice according to:

https://github.com/ceph/ceph-ansible/blob/561746f75e3913b30e6ae3f14768ebc8a516bf66/group_vars/all.yml.sample#L245

Setting the following works as documented/advertised with this patch:
```
---
ceph_repository: custom
ceph_custom_repo: http://somehost.com/path/to/custom.repo
```


jtriley:
Perhaps I'm not interpreting the logs correctly but seems like the errors aren't related to this merge request. Looks like CI/CD test hosts are unreachable to ansible. So either they're not launching or have some issue with SSH.

PR #3481, opened 2019-01-08
facts: clean code

guits:
clean some leftover and duplicate code.

https://bugzilla.redhat.com/show_bug.cgi?id=1622688

### DIFF
--- a/roles/ceph-config/templates/ceph.conf.j2
+++ b/roles/ceph-config/templates/ceph.conf.j2
@@ -19,9 +19,9 @@ osd crush chooseleaf type = 0
 {% set nb_mon = groups.get(mon_group_name, []) | length | int %}
 {% set nb_client = groups.get(client_group_name, []) | length | int %}
 {% set nb_osd = groups.get(osd_group_name, []) | length | int %}
-{% if inventory_hostname in groups.get(client_group_name, []) and not inventory_hostname == groups.get(client_group_name, []) | first %}}
-{% set ceph_release = hostvars[groups[client_group_name][0]]['ceph_release'] %}}
-{% endif %}}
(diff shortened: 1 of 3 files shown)


roles/ceph-facts/tasks/facts.yml
```diff
@@ -92,54 +97,35 @@
...
-  set_fact:
-    fsid: "{{ ceph_current_status.fsid }}"
+- block:
+  - name: set_fact fsid from ceph_current_status
+    set_fact:
+      fsid: "{{ ceph_current_status.fsid }}"
+
+  - name: reuse cluster fsid when cluster is already running
+    shell: echo {{ fsid }} | tee {{ fetch_directory }}/ceph_cluster_uuid.conf
+    args:
```
leseb: I wonder if we actually need to write this down since the `fsid` is a fact and given it's never re-used.

roles/ceph-facts/tasks/facts.yml
```diff
@@ -92,54 +97,35 @@
...
 
-- name: set_fact fsid from ceph_current_status
-  set_fact:
-    fsid: "{{ ceph_current_status.fsid }}"
+- block:
+  - name: set_fact fsid from ceph_current_status
+    set_fact:
+      fsid: "{{ ceph_current_status.fsid }}"
+
+  - name: reuse cluster fsid when cluster is already running
```
leseb: I wonder if we actually need to write this down since the fsid is a fact and given it's never re-used.

guits: looks like we don't need it

leseb:
https://bugzilla.redhat.com/show_bug.cgi?id=1622688

PR #3482, opened 2019-01-08
mon: force peer addition

leseb:
Somewhat something changed with the introduction of msg2 and we have to
add each node as a peer so the monitors can form a quorum. This might be
due to our CI environment, although adding this is completly harmless
and solves monitors not being able to form quorum.

It seems that the initial monitor map wasn't containing the right
information about the peers (addresses like 0.0.0.0/0r1, for each rank.

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -1,4 +1,18 @@
 ---
+- name: wait for the monitor socket to exist
+  stat:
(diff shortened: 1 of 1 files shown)


leseb:
jenkins test dev-centos7_cluster

guits:
the container failures are going to be fixed in a dedicated PR in ceph-container, so merging this anyway

liewegas:
let's revert this once https://github.com/ceph/ceph/pull/25886 merges

PR #3484, opened 2019-01-09
ceph-infra: Remove test of is_atomic

badone:
When configure_firewall runs is_atomic is undefined.

### DIFF
--- a/roles/ceph-infra/tasks/configure_firewall.yml
+++ b/roles/ceph-infra/tasks/configure_firewall.yml
@@ -176,7 +176,6 @@
       - firewall
 
   when:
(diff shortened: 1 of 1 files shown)


badone:
http://qa-proxy.ceph.com/teuthology/yuriw-2019-01-07_20:10:12-ceph-ansible-wip-yuri-testing-2019-01-07-1727-luminous-distro-basic-ovh/3434450/teuthology.log is an example of the error.

guits:
@badone 
your patch means that on atomic, `configure_firewall: true` will never work.
I think we should try to set `is_atomic` in `pre_tasks` instead.

By the way, do not make pull request against stable-3.2 branch directly please.

PR #3488, opened 2019-01-10
site: Make sure is_atomic is defined

badone:
configure_firewall tests the is_atomic variable if the firewalld package
is not present.


guits:
`is_atomic` is already defined in `ceph-facts`, simply including it where @leseb pointed out is enough although it's a pitty to have to play the whole role just for this...

badone:
Done. Tested with stable-3.2 as well and the equivalent change works. Thanks @leseb and @guits

site.yml.sample
```diff
@@ -76,6 +76,8 @@
         name: ceph-defaults
     - import_role:
         name: ceph-validate
+    - import_role:
```
leseb: Nit, usually we prefer to call `ceph-facts` right after `ceph-defaults`. Thanks.

badone: OK, let me reorder that :)

badone:
@leseb I think this is good now?

leseb:
@badone yes, thanks!

PR #3489, opened 2019-01-10
Add module to perform host validation checks

pcuzner:
A new module, ceph_check_role provides a simple
way to perform some initial validation checks
against hosts prior to installation. The idea is
to pick up simple configuration issues that could
resort in a failed installation.

Signed-off-by: Paul Cuzner <pcuzner@redhat.com>


checkrole.yml.sample
```diff
@@ -0,0 +1,30 @@
...
+#   flashusage: journal 
+
+- name: Validate hosts against desired cluster state
+  hosts: all
+  become: true
+  gather_facts: false
+
+  tasks:
+  
+    - name: CEPH_CHECK_ROLE
```
leseb: can you add this task to the ceph-validate role please? Somewhere here should be fine: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-validate/tasks/main.yml#L5

pcuzner: OK

pcuzner: @leseb the module uses parms passed to it by the caller at runtime. For the ceph-installer, options selected in the UI are passed to the playbook dynamically. For ceph-validate, could I simply set the 'role' to mons,osds?

i.e.

`- name: check_role module
  ceph_check_role:
    role: 'mons,osds'`

Would that work?

pcuzner: @leseb it's a bit of a misnomer. The roles in the GUI refer to the group names in ansible hence mons and osds not ceph-mon.

leseb: Do you anything from me to get this moving? Thanks.

checkrole.yml.sample
```diff
@@ -0,0 +1,30 @@
+---
+
+# check host hardware specification against a given set of Ceph roles.
+#
+# Requires the variables to be passed in by the caller or vars defined
+# e.g.
+# vars:
+#   inventory:
+#     host1: osds,mons
```
guits: from what I'm understanding with comments below `osds,mons` would be the groups `host1` would belong to, right ?

checkrole.yml.sample
```diff
@@ -0,0 +1,30 @@
...
+- name: Validate hosts against desired cluster state
+  hosts: all
+  become: true
+  gather_facts: false
+
+  tasks:
+  
+    - name: CEPH_CHECK_ROLE
+      ceph_check_role:
+        role: "{{ inventory[inventory_hostname] }}"
```
guits: with this variable, it feels to me it makes us to rewrite the inventory
I would probably suggest something like following:
```suggestion
        role: "{{ ','.join(hostvars[inventory_hostname]['group_names']) }}"
```
to avoid having to rewrite the inventory
or :
```suggestion
        role: "{{ inventory[inventory_hostname] | default(','.join(hostvars[inventory_hostname]['group_names'])) }}"
```
if you really want to keep the possibility to use that `inventory` variable

I even wonder whether you could manage with this param directly in the module so it wouldn't be a required param eg :
```python
    fields = dict(
        role=dict(
            type='str',
            required=False),
```

library/ceph_check_role.py
```diff
@@ -0,0 +1,817 @@
...
+    fields = dict(
+        role=dict(
+            type='str',
+            required=True),
+        mode=dict(
+            type='str',
+            choices=['prod', 'dev'],
+            default='prod',
+            required=False),
+        deployment=dict(
```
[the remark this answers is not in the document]
guits: not sure if you can get hostvars content from this module? This way we could rely on what we already set in the main playbook with `containerized_deployment`, so you wouldn't have to provide this parameter.

pcuzner:
The module is being included in the installer package, so closing this PR (at least for now)

PR #3500, opened 2019-01-14
switch: do not fail on missing key

leseb:
Some people use the switch playbook to perform upgrade so they end up in
the same situation than https://bugzilla.redhat.com/show_bug.cgi?id=1650572
This is applying the same fix as
729744c6a8c69f5fdf66b67fb28063297996e30a.

We don't want to fail on key that are not present since they will get
created after the mons are updated. They will be created by the task
"create potentially missing keys (rbd and rbd-mirror)".


leseb:
I'm pushing in stable-3.2 directly since the master has diverged too much and this task does not exist anymore.

leseb:
jenkins test dev-switch_to_containers

leseb:
Oops didn't mean to trigger dev-switch_to_containers.

PR #3504, opened 2019-01-15
start_osds: use list instead of keys (re-introduce)

noahdesu:
backport dict keys list fix


noahdesu:
https://github.com/ceph/ceph-ansible/pull/3420

PR #3512, opened 2019-01-18
Preserve rolling_update backward compatibility with ansible < 2.5

gfidente:
Signed-off-by: Giulio Fidente <gfidente@redhat.com>

PR #3514, opened 2019-01-18
rolling_update: enforce default value for client_update_batch

guits:
`ansible_forks` isn't something always set by ansible.

This commit allows to either pass `client_update_batch` as an extra variable
that will be the *number* of node per batch or
`client_update_batch_percentage` as an extra variable as well that will
be the percentage of the total number of client node in a decimal
representation (eg: `0.5` for `50%`).

If none of them are passed, the default value is 0.2 (20%).

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1650184

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -800,7 +800,7 @@
     upgrade_ceph_packages: True
   hosts:
     - "{{ client_group_name|default('clients') }}"
-  serial: "{{ client_update_batch | default(ansible_forks) }}"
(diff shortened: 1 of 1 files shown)


guits:
perhaps it could be nice to allow users setting how many % they want with another extra variable.

guits:
jenkins test dev-update_cluster

gfidente:
personally, I don't think calculating the number of nodes to upgrade in parallel as a percentage of the total number of clients is any useful ... I'd rather shoot for the highest working number hence ansible_forks was more meaningful to me

I think #3512 is better

guits:
closing in favor of #3512

PR #3522, opened 2019-01-22
override ceph_release with ceph_stable_release

guits:
when `ceph_origin` is set to `'repository'` and `ceph_repository` to
`'community'` we need to ensure `ceph_release` reflect `ceph_stable_release`.

### DIFF
--- a/roles/ceph-common/tasks/main.yml
+++ b/roles/ceph-common/tasks/main.yml
@@ -44,6 +44,14 @@
   tags:
     - always
 
+- name: set_fact ceph_release - override ceph_release with ceph_stable_release
+  set_fact:
+    ceph_release: "{{ ceph_stable_release }}"
+  when:
+    - ceph_origin == 'repository'
+  tags:
+    - always
+
 - name: include facts_mon_fsid.yml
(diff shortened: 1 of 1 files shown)


roles/ceph-container-common/tasks/main.yml
```diff
@@ -37,4 +37,13 @@
...
 - name: include release.yml
-  include_tasks: release.yml
\ No newline at end of file
+  include_tasks: release.yml
+
+- name: override ceph_release with ceph_stable_release
+  set_fact:
+    ceph_release: "{{ ceph_stable_release }}"
+  when:
+    - ceph_origin == 'repository'
```
leseb: This has no effect when deploying containers. Why do you set this fact?

roles/ceph-common/tasks/main.yml
```diff
@@ -44,6 +44,14 @@
   tags:
     - always
 
+- name: override ceph_release with ceph_stable_release
```
leseb:
```suggestion
- name: set_fact override ceph_release with ceph_stable_release
```

PR #3524, opened 2019-01-23
Don't add loop devices with osd_auto_discovery

eb4x:
Discovered this as a problem when deploying openstack-ansible,
during container creation, /var/lib/machines.raw gets mounted,
which leads to a loop0 device with sectors != "0"

Signed-off-by: Erik Berg <openstack@slipsprogrammor.no>


guits:
@fultonj I think this is gonna break OSP testing?

guits:
Closing this in favor of #3595

PR #3535, opened 2019-01-24
Fix ceph.conf generation

zmc:
877979c78 in #3486 broke ceph.conf generation entirely. Remove the stray
curly brace.


zmc:
@guits good catch - updated

guits:
failures are unrelated (currently working on fixing those in a separate PR)

PR #3540, opened 2019-01-26
delegate key creation to first mon

batrick:
Otherwise keys get scattered over the mons.

Also, create all keys on the first mon and copy those to the other mons to be
consistent.


batrick:
@leseb I can't recall the exact error and unfortunately I didn't save it for the commit message. It had something to do with the mgr failing to start because it didn't have keys IIRC.

I'll see if I can reproduce the problem again.

batrick:
@leseb, I don't know if it's a problem but it seems unnecessary? Anyway, I just re-verified the fix resolves the issue and included the failure in the commit message. Please check.

guits:
@batrick could you rebase on master please

batrick:
@guits done

PR #3541, opened 2019-01-27
do not set ceph_release to dummy

batrick:
When we're deploying a dev branch.


roles/ceph-common/tasks/main.yml
```diff
@@ -49,6 +49,7 @@
     ceph_release: "{{ ceph_stable_release }}"
   when:
     - ceph_origin == 'repository'
+    - ceph_stable_release != 'dummy'
```
guits: According to your commit message, I think adding `ceph_repository != 'dev'` would be more accurate

batrick: sure, done

Issue #3556, opened 2019-01-31
Allow creation of custom EC profiles during deployment


bencherian:
**Is your feature request related to a problem? Please describe.**
I want to have custom EC profiles setup on a new Ceph cluster
**Describe the solution you'd like**
I would like to be able to setup customized EC profiles when running deployment with ceph-ansible.

**Describe alternatives you've considered**
I could do in a post deployment playbook but since we can already define "simple rules" in ceph-ansible it would make sense to define these here as well.

mistur:
Hello,

will this feature be available in the 4.0.0 release ? Thanks.

Best regards,

Yoann

PR #3557, opened 2019-01-31
Enable creation of EC and replicated rules during deployment

bencherian:
Allows users to create EC profiles and replicated rules with specific device types.
Solves issue #3556

### DIFF
--- a/roles/ceph-mon/tasks/crush_rules.yml
+++ b/roles/ceph-mon/tasks/crush_rules.yml
@@ -11,7 +11,28 @@
     - create_crush_tree
     - hostvars[item]['osd_crush_location'] is defined
 
-- name: create configured crush rules
+- name: create crush ec profiles
+  command: "{{ docker_exec_cmd }} ceph --cluster {{ cluster }} osd erasure-code-profile set {{ item.name }} plugin={{ item.plugin }} {{ item.ec_config | to_json | regex_replace('\\:\\ ','=') | regex_replace('[\\[\\]{}\\\"]') | regex_replace('\\,\\ ',' ') }}"
+  with_items: "{{ ec_profiles | unique }}"
+  changed_when: false
+  when:
+    - inventory_hostname == groups.get(mon_group_name) | last
+
+- name: create configured crush ec rules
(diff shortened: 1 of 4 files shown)


bencherian:
It would be nice if this feature could be backported to stable-3.2 as well. (I can make a PR for that as well if it would be helpful.)

bencherian:
This is failing because the example crush rule cannot be created since there is no hdd bucket in the test configuration. However, if I don't include a bucket in the sample file it won't really be very helpful in terms of pointing out the benefits of the new replicated option. Any suggestions?

bencherian:
Is there interest in this feature? If so, I believe the failing tests can be fixed by setting `crush_device_class` in the test group_vars. I'm not sure if this would cause other issues though? (Another alternative is to simply comment out the rule creation lines in default.yml so they don't get run by the existing tests.)

guits:
I'm not a crush expert but I think one of the issues here is that `crush_rule_config` is a condition for the creation of simple rule only. Perhaps it should be independent of the other type of rule creation you are adding in your patch?

eg: if someone sets `crush_rule_config: True` but doesn't want to create `erasure-coded rule` and/or `replicated rule` and/or `EC profile` it's a problem with the current implementation.

roles/ceph-mon/tasks/crush_rules.yml
```diff
@@ -11,7 +11,28 @@
     - create_crush_tree
     - hostvars[item]['osd_crush_location'] is defined
 
-- name: create configured crush rules
+- name: create crush ec profiles
+  command: "{{ docker_exec_cmd }} ceph --cluster {{ cluster }} osd erasure-code-profile set {{ item.name }} plugin={{ item.plugin }} {{ item.ec_config | to_json | regex_replace('\\:\\ ','=') | regex_replace('[\\[\\]{}\\\"]') | regex_replace('\\,\\ ',' ') }}"
```
guits: I usually try to avoid using such regex imbrication.
I think that when we need to manage with "complex" yaml data structure in ansible, it's a signal we should maybe create a specific ansible module.

bencherian:
The bigger problem is that the crush hierarchy is not created until osd creation. So even if you do have device classes setup in your osd definition, this will fail on first deployment. (This will only work after the cluster has been deployed...I'll close this PR and think about a better way to do this.)

Issue #3561, opened 2019-01-31
net=host not used in some container_binary calls; results in CNI errors with podman


fultonj:
With 'podman version 1.0.0' on RHEL8 beta the 'get ceph version' and 'ceph monitor mkfs' commands fail [1] with "error configuring network namespace for container Missing CNI default network".

When net=host is added these errors are resolved. net=host is used in many other calls (grep -R net=host | wc -l --> 38).

[1]
```
TASK [ceph-container-common : get ceph version] ************************************************************************************************************
task path: /root/ceph-ansible/roles/ceph-container-common/tasks/main.yml:26
Tuesday 29 January 2019  23:18:08 +0000 (0:00:00.042)       0:00:32.462 ******* 
Using module file /usr/lib/python3.6/site-packages/ansible/modules/commands/command.py
<192.168.24.2> ESTABLISH LOCAL CONNECTION FOR USER: root
<192.168.24.2> EXEC /bin/sh -c '/usr/bin/python3 && sleep 0'
fatal: [rhel8]: FAILED! => changed=false 
  cmd:
  - podman
  - run
  - --rm
  - --entrypoint
  - /usr/bin/ceph
  - docker.io/ceph/daemon:latest-master
  - --version
  delta: '0:00:00.143798'
  end: '2019-01-29 23:18:08.443993'
  invocation:
    module_args:
      _raw_params: podman run --rm --entrypoint /usr/bin/ceph docker.io/ceph/daemon:latest-master --version
      _uses_shell: false
      argv: null
      chdir: null
      creates: null
      executable: null
      removes: null
      stdin: null
      warn: true
  msg: non-zero return code
  rc: 127
  start: '2019-01-29 23:18:08.300195'
  stderr: 'error configuring network namespace for container 85bea32c249182c64e151e31e67b3ebc161759c37d5df2dba3b9c7827d68bbd5: Missing CNI default network'
  stderr_lines:
  - 'error configuring network namespace for container 85bea32c249182c64e151e31e67b3ebc161759c37d5df2dba3b9c7827d68bbd5: Missing CNI default network'
  stdout: ''
  stdout_lines: <omitted>
```

PR #3526, opened 2019-01-23
podman: support podman installation on rhel8

guits:
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1667101

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #3562, opened 2019-01-31
Fix CNI error when net=host is not used in some podman calls

fultonj:
With 'podman version 1.0.0' on RHEL8 beta the 'get ceph version' and
'ceph monitor mkfs' commands fail [1] with "error configuring network
namespace for container Missing CNI default network".

When net=host is added these errors are resolved. net=host is used in
many other calls (grep -R net=host | wc -l --> 38).

Fixes: #3561


leseb:
This might need to wait until https://github.com/ceph/ceph-ansible/pull/3526 gets merged.

Issue #3568, opened 2019-02-04
num_osds fact setting fails because create_ceph_initial_dirs.yml must run first


fultonj:
Deployment with podman failed [1] on the `[ceph-config : run 'ceph-volume lvm batch --report' to see how many osds are to be created]` because /var/lib/ceph/ did not exist. 

AFAICT the following should prevent this issue:
```
  - name: include create_ceph_initial_dirs.yml
    include_tasks: create_ceph_initial_dirs.yml
```
but it looks like the above is not getting executed before the failing task which seems to have been introduced by https://github.com/ceph/ceph-ansible/commit/fe1528adb4fa853693ba1c207856dfad58cef270 Perhaps the include of the create initial dirs can happen before this task?

[1] 
```
2019-02-04 16:20:03,997 p=29810 u=root |  TASK [ceph-config : run 'ceph-volume lvm batch --report' to see how many osds are to be created] ***
2019-02-04 16:20:03,997 p=29810 u=root |  task path: /home/stack/ceph-ansible/roles/ceph-config/tasks/main.yml:17
2019-02-04 16:20:03,997 p=29810 u=root |  Monday 04 February 2019  16:20:03 +0000 (0:00:00.045)       0:00:38.306 ******* 
2019-02-04 16:20:04,163 p=29810 u=root |  Using module file /home/stack/ceph-ansible/library/ceph_volume.py
2019-02-04 16:20:04,372 p=29810 u=root |  The full traceback is:
  File "/tmp/ansible_ceph_volume_payload_7qlbmifo/__main__.py", line 613, in run_module
    report_result = json.loads(out)
  File "/usr/lib64/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None

2019-02-04 16:20:04,377 p=29810 u=root |  fatal: [rhel8]: FAILED! => changed=true 
  cmd:
  - podman
  - run
  - --rm
  - --privileged
  - --net=host
  - -v
  - /run/lock/lvm:/run/lock/lvm:z
  - -v
  - /var/run/udev/:/var/run/udev/:z
  - -v
  - /dev:/dev
  - -v
[...]

PR #3569, opened 2019-02-04
config: Create Ceph Initial Dirs earlier

fultonj:
Include tasks from create_ceph_initial_dirs earlier during ceph config role.

Fixes: #3568


fultonj:
> initial dir must be created only for containerized deployment so please add a condition

Thanks for the tip. I added it.

guits:
it needs to be rebased on master once  #3526 gets merged

Issue #3572, opened 2019-02-06
ntp_daemon_type: ntpd for container scenario ends up crashing


rishabh-d-dave:
**Bug Report**

What happened: Running `dev-centos-container-add_osds` on master with `ntp_daemon_type: ntpd` in group_vars/all.yml leads to following error -

```
TASK [ceph-infra : install ntp] ************************************************
task path: /home/rishabh/repos/ceph-ansible/master/roles/ceph-infra/tasks/setup_ntp.yml:29
Wednesday 06 February 2019  10:56:52 +0530 (0:00:02.569)       0:00:35.438 **** 
FAILED - RETRYING: install ntp (3 retries left).
FAILED - RETRYING: install ntp (3 retries left).
FAILED - RETRYING: install ntp (2 retries left).
FAILED - RETRYING: install ntp (2 retries left).
FAILED - RETRYING: install ntp (1 retries left).
FAILED - RETRYING: install ntp (1 retries left).
fatal: [osd0]: FAILED! => changed=false 
  attempts: 3
  msg: 'missing required arguments: image, backend'
fatal: [mon0]: FAILED! => changed=false 
  attempts: 3
```

What you expected to happen: `dev-centos-container-add_osds`  should abort since installation can't happen on Atomic.

Share your group_vars files, inventory
```
$ ls group_vars/*.yml
group_vars/all.yml
$ cat group_vars/all.yml
---
ntp_daemon_type: ntpd
$ 
```

**Environment**:
* Ansible version (e.g. `ansible-playbook --version`): 2.7.4
* ceph-ansible version (e.g. `git head or tag or stable branch`): master
* Ceph version (e.g. `ceph -v`): dev (nautilus)

PR #3573, opened 2019-02-06
don't install NTPd on Atomic

rishabh-d-dave:
Since Atomic doesn't allow any installations and NTPd is not present
on Atomic image we are using, abort when ntp_daemon_type is set to ntpd.

https://github.com/ceph/ceph-ansible/issues/3572

### DIFF
--- a/roles/ceph-infra/tasks/setup_ntp.yml
+++ b/roles/ceph-infra/tasks/setup_ntp.yml
@@ -10,51 +10,62 @@
         ntp_service_name: ntpd
       when: ansible_os_family in ['RedHat', 'Suse']
 
-- name: setup ntp daemon
+# Installation of NTP daemons needs to be a separate task since installations
+# can't happen on Atomic
+- name: install the ntp daemon
+  when: not is_atomic
   block:
-    - name: install and enable timesyncd
+    - name: install ntpd
+      package:
+        name: ntp
+        state: present
+      register: result
+      until: result is succeeded
+      when:
+        - ntp_daemon_type == "ntpd"
+
+    - name: install chrony
+      package:
+        name: chrony
+        state: present
+      register: result
+      until: result is succeeded
+      when:
+        - ntp_daemon_type == "chronyd"
+
+- name: enable the ntp daemon and disable the rest
+  block:
+    - name: enable timesyncing on timesyncd
       command: timedatectl set-ntp on
-      when: ntp_daemon_type == "timesyncd"
       notify:
         - disable ntpd
         - disable chronyd
+      when:
+        - ntp_daemon_type == "timesyncd"
 
     - name: disable time sync using timesyncd if we are not using it
       command: timedatectl set-ntp no
-      when: ntp_daemon_type != "timesyncd"
+      when:
+        - ntp_daemon_type != "timesyncd"
 
-    - name: setup ntpd
-      when: ntp_daemon_type == "ntpd"
-      block:
-        - name: install ntp
-          package:
-            name: ntp
-            state: present
-          register: result
-          until: result is succeeded
-        - name: enable and start ntp
-          service:
-            name: "{{ ntp_service_name }}"
-            enabled: yes
-            state: started
-          notify:
-            - disable chronyd
-            - disable timesyncd
+    - name: enable ntpd
+      service:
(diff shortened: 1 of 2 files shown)


roles/ceph-validate/tasks/main.yml
```diff
@@ -36,6 +36,15 @@
...
     - ntp_daemon_type not in ['chronyd', 'ntpd', 'timesyncd']
 
+# Since NTPd can not be installed on Atomic...
+- name: abort if ntp_daemon_type is ntp on Atomic
+  fail:
+    msg: installation can't happen on Atomic and ntpd needs to be installed
+  when:
+    - is_atomic
+    - ansible_os_family == 'RedHat'
+    - ntp_daemon_type == 'ntpd'
```
leseb:
```suggestion
    - ntp_daemon_type != 'timesyncd'
```

Wouldn't this be more appropriate?

rishabh-d-dave: chrony is present on Atomic, by default. So we can allow it.

leseb: Then `!= 'chrony'`?

rishabh-d-dave: timesyncd comes with systemd, so that too is fine.

rishabh-d-dave:
jenkins test pipeline

roles/ceph-infra/tasks/setup_ntp.yml
```diff
@@ -10,9 +10,30 @@
...
+  when: not is_atomic
   block:
-    - name: install and enable timesyncd
+    - name: install ntpd
+      package:
+        name: ntp
+        state: present
+      register: result
+      until: result is succeeded
+      when: ntp_daemon_type == "ntpd"
```
guits: keep consistency with the rest of the playbook.


change from:
`when: ntp_daemon_type == "ntpd"`

to:
```
when:
  - ntp_daemon_type == "ntpd"
```

roles/ceph-infra/tasks/setup_ntp.yml
```diff
@@ -23,38 +44,22 @@
...
-            enabled: yes
-            state: started
-          notify:
-            - disable chronyd
-            - disable timesyncd
+      service:
+        name: chronyd
+        enabled: yes
+        state: started
+      notify:
```
guits: move the `when` after the `notify`

roles/ceph-infra/tasks/setup_ntp.yml
rishabh-d-dave: I find this way of writing conditions very odd. But I admit that's only because I am used to writing C/C++.

rishabh-d-dave:
* incorporated the change Guillaume mentioned
* rebased the branch

roles/ceph-infra/tasks/setup_ntp.yml
guits: actually it could make sense to have everything in one line when only 1 condition is present and everything split on multiple lines when more than 1 condition but all conditions in the playbook are written this way whatever there is 1 or several conditions, so to keep consistency, let's write all condition the same way it has been done until now.

I would add that sometimes we must adapt to the project policies even if we personally find something is very odd. (Imagine if everyone could do things the way they like...)

rishabh-d-dave: > actually it could make sense to have everything in one line when only 1 condition is present and everything split on multiple lines when more than 1 condition but all conditions in the playbook are written this way whatever there is 1 or several conditions, so to keep consistency, let's write all condition the same way it has been done until now.
> 
> I would add that sometimes we must adapt to the project policies even if we personally find something is very odd. (Imagine if everyone could do things the way they like...)

I totally agree with this.

I hope I am not annoying you with my coding styles. I guess I am.  :P

rishabh-d-dave:
* rebased & resolved the merge conflict in `setup_ntp.yml`

guits:
unrelated failure in `dev-centos-container-lvm_auto_discovery`, it will be addressed in a separate PR

rishabh-d-dave:
For record: this PR is backported to stable-3.2 on PR #4112.

PR #3584, opened 2019-02-07
Fix uses of default(omit) with string concatenation

chneukirchen:
When {{omit}} is concatenated with another string, it expands to something
like __omit_place_holder__63eea0d96dd6ed867b95405e11d87dddf61f448d.
However, in these use-cases we need an empty string.

Regression introduced in d53f55e807e.

Signed-off-by: Leah Neukirchen <leah.neukirchen@mayflower.de>


chneukirchen:
(This issue also applies to master.)

leseb:
@chneukirchen arghhh, I didn't notice you pushed to a stable branch directly, we **never** do that unless the code between master and a stable branch has diverged too much. (https://github.com/ceph/ceph-ansible/blob/master/CONTRIBUTING.md)

That's my fault I should have been more careful.
@chneukirchen Could you please send a PR to master?

Thanks.

PR #3588, opened 2019-02-08
Fix uses of default(omit) with string concatenation

chneukirchen:
Port of #3584 to master.

When {{omit}} is concatenated with another string, it expands to something
like __omit_place_holder__63eea0d96dd6ed867b95405e11d87dddf61f448d.
However, in these use-cases we need an empty string.

Regression introduced in d53f55e807e.

Signed-off-by: Leah Neukirchen <leah.neukirchen@mayflower.de>


leseb:
@chneukirchen Thanks!

Issue #3065, opened 2018-08-27
ceph-defaults role is not deleting unused sockets when more then one socket per daemon type exists


artur-fijalkowski:
**Bug Report**

What happened:

Following task [source](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/tasks/check_socket_non_container.yml#L30):

```
# https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/tasks/check_socket_non_container.yml#L30

- name: check for a ceph osd socket
  shell: |
    stat --printf=%n {{ rbd_client_admin_socket_path }}/{{ cluster }}-osd*.asok
  changed_when: false
  failed_when: false
  check_mode: no
  register: osd_socket_stat
  when:
    - inventory_hostname in groups.get(osd_group_name, [])
```

Is not properly handling multiple sockets matching same pattern. This is caused by the fact that stat with formatting used in shell module returns all filenames concatenated in single string (without any separator).

What you expected to happen:

Proper approach should be something like below:

```
- name: check for a ceph osd socket
  shell: |
    stat --printf="%n\n" {{ rbd_client_admin_socket_path }}/{{ cluster }}-osd*.asok
  changed_when: false
  failed_when: false
  check_mode: no
  register: osd_socket_stat
  when:
    - inventory_hostname in groups.get(osd_group_name, [])
```

The next 2 tasks should then be looped against `osd_socket_stat.stdout_lines` instead of direct usage of `osd_socket_stat.stdout`. Also I do not understand why `stat` command with shell module is used instead of using find with command module, which will improve security (no shell escaping possible!). Example task would look like:

```
- name: check for a ceph osd socket
  command: |
    find {{ rbd_client_admin_socket_path }} -type s -iname {{ cluster }}-osd*.asok
  changed_when: false
  failed_when: false
  check_mode: no
  register: osd_socket_stat
  when:
    - inventory_hostname in groups.get(osd_group_name, [])
```

**This of course applies to ALL daemon types, not only OSDs.**

G-Tarik:
I just have discovered this issue too. If we omit the problem of using find instead of direct shell command then I think it's enough to modify ```stat``` command to this ```stat --printf="%n "```  Just insert space because ```fuser``` command (in next task) can handle multiple files separated by space.

guits:
@artur-fijalkowski / @G-Tarik  Hi! Thanks for reporting this.

Do you mind sending a PR for this if you think you have a solution?
[...]

PR #3071, opened 2018-08-28
Fix in roles/ceph-defaults/tasks/check_socket_non_container.yml to

artur-fijalkowski:
properly handle multiple sockets per type of server role (mon, mgr, osd,
etc.). Change of usage of shell module to command which is more secure
also looping on the list of sockets.

Closes: #3065

Signed-off-by: Artur Fijalkowski <wiki.warx@gmail.com>

PR #3597, opened 2019-02-11
Fixes: #3065 check_socket_non_container.yml: add space in stat outputs

G-Tarik:
Problem: ```stat --printf=%n``` for OSD sockets produces one concatenated string
 of multiple paths, consequently in the next ansible task ```fuser``` command
 fails with non-zero exit code and triggers next task deleting of non-existed
 path.

Solution: add space to separate multiple sockets in stat output by modifying the
 command to ```stat --printf="%n "```. ```fuser``` can handle multiple space
 separated arguments. For consistency modify all usages of stat in current file,
 although this issue can happen only with OSD. It's unlikely that there will be
 multiple sockets on one machine for MON, MDS, RGW or other Ceph services.

Why this solution: simple quick fix, minimal probability to introduce new bug.
 May uncover other hidden bugs :-)
Alternatives: PR #3071, which does a lot of additional improvements.

Signed-off-by: G-Tarik <gtarik@gmail.com>

### DIFF
--- a/roles/ceph-defaults/tasks/check_socket_non_container.yml
+++ b/roles/ceph-defaults/tasks/check_socket_non_container.yml
@@ -1,6 +1,6 @@
 ---
 - name: check for a ceph mon socket
-  shell: stat --printf=%n {{ rbd_client_admin_socket_path }}/{{ cluster }}-mon*.asok
+  shell: stat --printf="%n " {{ rbd_client_admin_socket_path }}/{{ cluster }}-mon*.asok
   changed_when: false
   failed_when: false
   always_run: true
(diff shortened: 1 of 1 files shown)


guits:
@G-Tarik do not commit in stable-3.0 please
patch in stables branches are backported from master only

G-Tarik:
I have found this http://docs.ceph.com/ceph-ansible/master/dev/index.html
So should I do two PRs (one to master and second to stable-3.0)?

guits:
one PR to master, once it gets merged, we can backport it to stable-3.2 first, then stable-3.1 and finally stable-3.0

Thanks!

PR #3600, opened 2019-02-12
fix mistake in task that aborts when ntpd is chosen on Atomic

rishabh-d-dave:
Since it's already confusing whether ntp_daemon_type should be "ntp" or
"ntpd", fix the mistake in the title of the task that aborts if
ntp_daemon_type is set to "ntpd" and OS being used is Atomic.


rishabh-d-dave:
Only `ceph-ansible-prs-dev-centos-container-lvm_auto_discovery` failed. The failure is not relevant since it pops elsewhere too and more importantly, because the change here non-functional one.

PR #3601, opened 2019-02-12
handler: handle multiple socket in check socket

### LINKED ISSUE #3065: ceph-defaults role is not deleting unused sockets when more then one socket per daemon type exists

G-Tarik:
Problem: ```stat --printf=%n``` for OSD sockets produces one concatenated string
 of multiple paths, consequently in the next ansible task ```fuser``` command
 fails with non-zero exit code and triggers next task deleting of non-existed
 path.

Solution: add space to separate multiple sockets in stat output by modifying the
 command to ```stat --printf="%n "```. ```fuser``` can handle multiple space
 separated arguments. For consistency modify all usages of stat in current file,
 although this issue can happen only with OSD. It's unlikely that there will be
 multiple sockets on one machine for MON, MDS, RGW or other Ceph services.

Why this solution: simple quick fix, minimal probability to introduce new bug.
 May uncover other hidden bugs :-)
Alternatives: PR #3071, which does a lot of additional improvements.

Signed-off-by: G-Tarik <gtarik@gmail.com>
(cherry picked from commit 39df7fe7cdee0ec3191041ff09697174a7653d8e)

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -15,7 +15,7 @@ pull_request_rules:
     conditions:
       - label!=DNM
       - title~=\[skip ci\]
-      - '#approved-reviews-by>=1'
+      - '#approved-reviews-by>=2'
     actions:
       merge:
         method: rebase
--- a/Makefile
+++ b/Makefile
@@ -7,16 +7,16 @@ NAME = ceph-ansible
 # Examples:
 #
 #  A "git describe" value of "v2.2.0beta1" would create an NVR
-#  "ceph-ansible-2.2.0-0.beta1.1.el7"
+#  "ceph-ansible-2.2.0-0.beta1.1.el8"
 #
 #  A "git describe" value of "v2.2.0rc1" would create an NVR
-#  "ceph-ansible-2.2.0-0.rc1.1.el7"
+#  "ceph-ansible-2.2.0-0.rc1.1.el8"
 #
 #  A "git describe" value of "v2.2.0rc1-1-gc465f85" would create an NVR
-#  "ceph-ansible-2.2.0-0.rc1.1.gc465f85.el7"
+#  "ceph-ansible-2.2.0-0.rc1.1.gc465f85.el8"
 #
 #  A "git describe" value of "v2.2.0" creates an NVR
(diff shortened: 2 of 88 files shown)


guits:
@G-Tarik could you add the `Fixes: #3065` in your commit message and rebase on master?

Thanks!

PR #3604, opened 2019-02-14
container: fix tmpfiles.d ceph files

guits:
- fix uid/gid in ceph tmpfiles
- move file to `/etc/tmpfiles.d`

### DIFF
--- a/roles/ceph-container-common/tasks/pre_requisites/prerequisites.yml
+++ b/roles/ceph-container-common/tasks/pre_requisites/prerequisites.yml
@@ -92,8 +92,8 @@
 
 - name: ensure tmpfiles.d is present
   lineinfile:
-    path: /usr/lib/tmpfiles.d/ceph-common.conf
-    line: "d /run/ceph 0770 ceph ceph -"
+    path: /etc/tmpfiles.d/ceph-common.conf
+    line: "d /run/ceph 0770 root root -"
     owner: root
     group: root
     mode: 0644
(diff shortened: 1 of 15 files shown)


roles/ceph-container-common/tasks/pre_requisites/prerequisites.yml
```diff
@@ -93,7 +93,7 @@
 - name: ensure tmpfiles.d is present
   lineinfile:
     path: /usr/lib/tmpfiles.d/ceph-common.conf
```
leseb: Your commit said `- move file to /etc/tmpfiles.d`, that doesn't seem to be the case

guits: thanks maggle

roles/ceph-facts/tasks/facts.yml
```diff
@@ -212,7 +212,8 @@
     ceph_uid: 167
   when:
     - containerized_deployment
-    - ceph_docker_image_tag | string is search("latest") or ceph_docker_image_tag | string is search("centos") or ceph_docker_image_tag | string is search("fedora")
+    - (ceph_docker_image_tag | string is search("latest") or ceph_docker_image_tag | string is search("centos") or ceph_docker_image_tag | string is search("fedora")
+      or (ansible_distribution == 'RedHat'))
```
guits: @leseb
Since we provided `v4.0.0beta-nautilus` for OSP, the `ceph_uid` fact doesn't get set because it doesn't match the list of expected tags.
I think we can assume when hosts are running a RHEL system, the containers will be RHEL as well?

leseb: Yes, it's a fair assumption.

PR #3615, opened 2019-02-18
ansible.cfg: Add library path to configuration

dsavineau:
Ceph module path needs to be configured if we want to avoid issues
like:

```console
no action detected in task. This often indicates a misspelled module
name, or incorrect module path
```

Currently the ansible-lint command in Travis CI complains about that.


leseb:
CI failure un-related.

PR #3621, opened 2019-02-20
osd: make the 'wait for all osd to be up' task configurable

guits:
introduce two new variables to make the check that 'wait for all osd to
be up' configurable.
It's possible that for some deployments, OSDs can take longer to be seen
as UP and IN.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1676763

### DIFF
--- a/roles/ceph-osd/defaults/main.yml
+++ b/roles/ceph-osd/defaults/main.yml
@@ -257,3 +257,11 @@ ceph_osd_docker_run_script_path: "/usr/share" # script called by systemd to run 
 #ceph_osd_systemd_overrides:
 #  Service:
 #    PrivateDevices: False
+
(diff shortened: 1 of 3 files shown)


roles/ceph-osd/defaults/main.yml
```diff
@@ -257,3 +257,11 @@ ceph_osd_docker_run_script_path: "/usr/share" # script called by systemd to run
 #ceph_osd_systemd_overrides:
 #  Service:
 #    PrivateDevices: False
+
+
+###########
+#  CHECK  #
+###########
+
+nb_retry_wait_osd_up: 60
```
leseb: prefix with `osd_`?

PR #3626, opened 2019-02-22
fix network interfaces names in conftest.py

guits:
Set network interfaces names according to the OS distribution in conftest.py


guits:
jenkins test pipeline

Issue #3613, opened 2019-02-18
Do not always collocate mon and mgr daemons


b-ranto:
Hi,

currently, we do collocate mon and mgr daemons by default. This makes less and less sense as the mgr moves forward since it is getting more and more resource heavy (we have more than 10 mgr plugins nowadays and they can do a lot of stuff). This would just steal the resources from a random monitor node and forego the whole point of splitting mon and mgr nodes.

This can also easily cause a confussion and a regression in behaviour if you were previously setting e.g. a single node mgrs section. You will no longer be able to access your mgr by a fixed IP since it will also be collocated on the monitor nodes, etc. You can also hardly expect people to know this intuitively (i.e. that mons and mgrs are collocated and your mgr plugins can now run on mons). At least for me, this was a real wtf moment.

If we really do want to default to collocating mons and mgrs, we should only do so if the mgrs section is not defined at all.

zmc:
I agree - we should retain the capability to have mgrs separate from mons, regardless if whether we want them to colocate by default

leseb:
@zmc collocating by default mon and mgr does not mean losing the ability to have dedicated mgr nodes. You still can do that if you want to.

leseb:
I don't think for now they are any evidence that this is causing regression. People won't likely dedicate one machine for the `ceph-mgr` daemon so it's probably best to keep it as is.
Again, you can always change your `site.yml` to change that default behavior.

And even if your cluster has been deployed already you can always change that back and move the mgr daemon to another node quite easily if you think it needs more resources.

b-ranto:
@leseb The whole point of creating the mgr daemons was to improve the performance of the monitor nodes. If we (always) collocate mons and mgrs (by default), we would be undermining the whole point of the split. I guess the further discussion is moot now that we merged #3627 and we can close this.

PR #3627, opened 2019-02-22
site.yml: do not bootstrap mgrs on monitors by default

guits:
Let's bootstrap mgrs on monitors only if there's no mgrs section in
inventory hostfile.

Closes: #3613

### DIFF
--- a/site-container.yml.sample
+++ b/site-container.yml.sample
@@ -126,6 +126,7 @@
         name: ceph-mon
     - import_role:
         name: ceph-mgr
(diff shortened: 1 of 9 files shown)


guits:
@leseb 

if mgr section is empty, it still bootstraps mgrs on monitor nodes.
if mgr section is defined, it will bootstrap mgrs on dedicated nodes.

IMO we should use collocated scenario to cover the scenario where mons and mgrs are collocated and use all_daemons with dedicated nodes for monitors and mgrs.

PR #3637, opened 2019-02-26
rgw: change default frontend on nautilus

dsavineau:
As discussed in ceph/ceph#26599, beast is now the default frontend
for rados gateway with nautilus release.


group_vars/all.yml.sample
```diff
@@ -387,7 +387,7 @@ dummy:
 
 ## Rados Gateway options
 #
-#radosgw_frontend_type: civetweb # For additionnal frontends see: http://docs.ceph.com/docs/mimic/radosgw/frontends/
+#radosgw_frontend_type: beast # For additionnal frontends see: http://docs.ceph.com/docs/nautilus/radosgw/frontends/
```
cbodley: typo while you're here additionnal -> additional

group_vars/all.yml.sample
```diff
@@ -397,7 +397,7 @@ dummy:
 # https://github.com/civetweb/civetweb/blob/master/docs/UserManual.md
 
 #radosgw_frontend_port: "{{ radosgw_civetweb_port if radosgw_frontend_type == 'civetweb' else '8080' }}"
-#radosgw_frontend_options: "{{ radosgw_civetweb_options if radosgw_frontend_type == 'civetweb' }}"
+#radosgw_frontend_options: "{{ radosgw_civetweb_options if radosgw_frontend_type == 'civetweb' else '' }}"
```
cbodley: we do still want the num_threads option from radosgw_civetweb_options

cbodley:
> I guess we need to check if beast/civetweb support the same args.

the num_threads= and port= options that ceph-ansible is currently using are both the same between beast and civetweb. there are some important differences with some features, like specifying ssl ports or binding to specific network addresses. i'm happy to provide details if anyone's interested in implementing those

group_vars/all.yml.sample
dsavineau: according to the documentation num_threads is specific to civetweb not beast right ?

dsavineau:
> the num_threads= and port= options that ceph-ansible is currently using are both the same between beast and civetweb

maybe the doc isn't correct [1] but num_threads seems specific to civetweb (unless beast is a superset of civetweb) also port parameter is not used the same way (ip:port for civetweb and port number only for beast) 

[1] http://docs.ceph.com/docs/nautilus/radosgw/frontends/

cbodley:
> but num_threads seems specific to civetweb

you're right, thanks. beast uses the value of `rgw_thread_pool_size` directly

dsavineau:
@leseb beast and civetweb don't have a lot of options in common.
We could deprecated the radosgw_civetweb* variables and only use radosgw_frontend instead (what beast is currently using). But civetweb is still usable in nautilus right ? or is it deprecated ?

cbodley:
we haven't made a decision to deprecate civetweb, so it will be supported for the N and O releases at least

re: num_threads, it's probably best to deprecate the `radosgw_civetweb_num_threads` option and use the `rgw_thread_pool_size` config value instead for both frontends - if `rgw_frontends` doesn't contain a `num_threads` option for civetweb, it defaults to the value of `rgw_thread_pool_size`. there's already an issue with ceph-ansible here, because the default value of `rgw_thread_pool_size` was raised to 512 in https://github.com/ceph/ceph/pull/22581 but this civetweb option is overriding that change

for the other frontends options, i do think it would be worth adding some abstraction to handle the differences in port vs endpoint, and eventually support for ssl

cbodley:
cc @alimaredia

PR #3639, opened 2019-02-27
mon: do not create unnecessarily mgr keyrings

guits:
there's no need to generate mgr keyrings 'mgr.monX' when mgrs aren't
collocated with monitors.

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -46,8 +46,7 @@
       CEPH_CONTAINER_IMAGE: "{{ ceph_docker_registry + '/' + ceph_docker_image + ':' + ceph_docker_image_tag if containerized_deployment else None }}"
       CEPH_CONTAINER_BINARY: "{{ container_binary }}"
     with_items:
-      - "{{ groups.get(mgr_group_name, []) }}" # this honors the condition where mgrs run on separate machines
-      - "{{ groups.get(mon_group_name, []) if groups.get(mgr_group_name, []) | length == 0 else [] }}" # this honors the new rule where mgrs are always collocated with mons
+      - "{{ groups.get(mon_group_name) if groups.get(mgr_group_name, []) | length == 0 else groups.get(mgr_group_name, []) }}"
     run_once: True
     delegate_to: "{{ groups[mon_group_name][0] }}"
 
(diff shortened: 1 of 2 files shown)


roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -47,7 +47,8 @@
       CEPH_CONTAINER_BINARY: "{{ container_binary }}"
     with_items:
       - "{{ groups.get(mgr_group_name, []) }}" # this honors the condition where mgrs run on separate machines
-      - "{{ groups.get(mon_group_name, []) }}" # this honors the new rule where mgrs are always collocated with mons
+      - "{{ groups.get(mon_group_name, []) if groups.get(mgr_group_name, []) | length == 0 else [] }}" # this honors the new rule where mgrs are always collocated with mons
```
leseb: Isn't `else []` handled by `, [])` already in the `.get`?

guits: see the new approach, it's clearer i think

roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -46,8 +46,8 @@
       CEPH_CONTAINER_IMAGE: "{{ ceph_docker_registry + '/' + ceph_docker_image + ':' + ceph_docker_image_tag if containerized_deployment else None }}"
       CEPH_CONTAINER_BINARY: "{{ container_binary }}"
     with_items:
-      - "{{ groups.get(mgr_group_name, []) }}" # this honors the condition where mgrs run on separate machines
-      - "{{ groups.get(mon_group_name, []) }}" # this honors the new rule where mgrs are always collocated with mons
+      - "{{ groups.get(mon_group_name) if groups.get(mgr_group_name, []) | length == 0 else groups.get(mgr_group_name, []) }}"
```
leseb: I don't think that handles the case where `groups.get(mon_group_name, []) | length > 0` AND `groups.get(mgr_group_name, []) | length > 0`

guits: generally, I don't see a real case where you have `groups.get(mon_group_name, []) | length > 0`

guits: I suspect you were talking about having mgrs collocated on monitors AND on dedicated nodes?

In that case it means you would have something like this in your inventory :
```
[mons]
mon0
mon1
mon2

[mgrs]
mon0
mgr1
```

leseb: Yes that's my concern.

guits: it will be handled anyway in that case

PR #3642, opened 2019-02-27
mon: Add mds permissions to client.admin

dsavineau:
The administrator keyring needs full capabilities on mds like mon,
osd and mgr.
Whithout this, the client.admin key won't be able to run commands
against mds (like ceph tell mds.0 session ls)

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1672878

### DIFF
--- a/roles/ceph-mon/tasks/deploy_monitors.yml
+++ b/roles/ceph-mon/tasks/deploy_monitors.yml
@@ -46,17 +46,6 @@
     mode: "u=rwX,g=rX,o=rX"
     recurse: true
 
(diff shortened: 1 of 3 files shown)


roles/ceph-mon/tasks/deploy_monitors.yml
```diff
@@ -51,7 +51,7 @@
     client_admin_ceph_authtool_cap:
```
guits: I wonder why we should use this set_fact, I think it could be moved somewhere in ceph-defaults.
at least, if we really want to keep it, I think it should be moved into ceph-facts

PR #3652, opened 2019-02-28
Ability to tune kernel parameters

leseb:
ceph-ansible is already capable of tuning kernel parameters but only
parameters that can change on the live system.
This commit allows us to change specific kernel parameters that require
a reboot.
So during the initial phase, we play the ceph-infra role which will
apply some tuning if specified. Then all the machines will be rebooted.
We wait up to 15min for that reboot before failing, assuming the machine
is stuck in its init sequence.

To change a specific parameter just set:

tune_kernel_parameters: "key=value key=value"

This implementation has some assumptions:

* the system is RHEL8
* the 'grubby' command does things correctly

Things we don't do:

* validate the kernel param is valid
* validate that has been set (we trust 'grubby' for this)

We only print the bootloader cmdline so you check if the system has the
correct value.


leseb:
jenkins test ceph-ansible-prs-dev-centos-non_container-all_daemons

leseb:

roles/ceph-infra/tasks/tune_kernel_parameters.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: "tune kernel parameters {{ tune_kernel_parameters }} and merge with existing templates from /etc/grub.d"
+  command: >
+    grubby
+    --update-kernel=ALL
+    --copy-default
+    --args="{{ tune_kernel_parameters | quote }}"
+  changed_when: false
+
+- name: reboot the system to apply new kernel parameters
```
leseb: We might need to serialize this task.

leseb:
Need to add an infra playbook so this can be played on existing clusters too.

roles/ceph-infra/tasks/tune_kernel_parameters.yml
```diff
@@ -0,0 +1,32 @@
...
+  shell: sleep 2 && shutdown -r now
```
dsavineau: You should switch the shell(shutdown)+wait_for+command(uptime) tasks by the reboot module [1]

[1] https://docs.ansible.com/ansible/latest/modules/reboot_module.html

roles/ceph-infra/tasks/tune_kernel_parameters.yml
dsavineau: This will probably also need to use the changed state from the previous task if we want to avoid to reboot the node on every playbook call

roles/ceph-infra/tasks/tune_kernel_parameters.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: "tune kernel parameters {{ tune_kernel_parameters }} and merge with existing templates from /etc/grub.d"
+  command: >
+    grubby
+    --update-kernel=ALL
+    --copy-default
+    --args="{{ tune_kernel_parameters | quote }}"
+  changed_when: false
```
dsavineau: probably need to add a when for RedHat ansible_distribution ?

roles/ceph-infra/tasks/tune_kernel_parameters.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: "tune kernel parameters {{ tune_kernel_parameters }} and merge with existing templates from /etc/grub.d"
+  command: >
+    grubby
+    --update-kernel=ALL
+    --copy-default
```
dsavineau: trying this option on CentOS 7 but it fails

```console
# grubby --update-kernel=ALL --copy-default --args="foo=bar"
grubby: kernel path expected
```

Without the --copy-default it's working fine.
Is that specific to RHEL 8 ?

PR #3662, opened 2019-03-01
purge: fix purge of lvm devices

guits:
using `shell` module seems to be the only way to make this task working
on rhel based distribution AND debian based distributions.

on ubuntu, using `command` ansible module fails like following
(not due to `sudo` usage or not):
```
ok: [osd1] => changed=false
  cmd: command -v ceph-volume
  failed_when_result: false
  msg: '[Errno 2] No such file or directory: ''command'': ''command'''
  rc: 2
```

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1653307

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -118,7 +118,7 @@
 - name: purge ceph rbd-mirror cluster
 
   vars:
-    rbdmirror_group_name: rbd-mirrors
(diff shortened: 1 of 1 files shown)


guits:
perhaps it's better to consider using `which` instead of `command -v` to avoid using `shell` ansible module.
@dsavineau any thought?

guits:
jenkins test pipeline

PR #3677, opened 2019-03-05
ceph-mgr: run mgr_modules.yml only on the first mgr host

wumingqiao:
Signed-off-by: wumingqiao <wumingqiao@beyondcent.com>


wumingqiao:
the task will be delegated to mons[0] for all mgr hosts, so we can just run it on the first host and have the same effect.

PR #3682, opened 2019-03-05
fix pool min_size customization

guits:
b8d580b3f48c69ba9882df773c4d144b73d01c95 introduced a bug when
`min_size` isn't set (default to 0).

Typical error:

```
Error EINVAL: pool min_size must be between 1 and 1
```

### DIFF
--- a/roles/ceph-client/tasks/create_users_keys.yml
+++ b/roles/ceph-client/tasks/create_users_keys.yml
@@ -122,7 +122,7 @@
       changed_when: false
       when:
         - pools | length > 0
-        - item.min_size | default(osd_pool_default_min_size) != ceph_osd_pool_default_min_size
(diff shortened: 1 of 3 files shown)


roles/ceph-client/tasks/create_users_keys.yml
```diff
@@ -122,7 +122,7 @@
       changed_when: false
       when:
         - pools | length > 0
-        - item.min_size | default(osd_pool_default_min_size) != ceph_osd_pool_default_min_size
+        - item.min_size | default(osd_pool_default_min_size) > ceph_osd_pool_default_min_size
```
guits: this condition is valid only if `ceph_osd_pool_default_min_size` is `0`

PR #3688, opened 2019-03-06
Remove trailing forward slash in ceph_docker_registry variable from g…

phuonghnguyenau:
Remove trailing forward slash in ceph_docker_registry variable from group_vars/rhcs.yml.sample file.

This had caused the docker pull command in site-docker.yml.sample to fail due to the double forward slash // when rendered by Ansible to the command line.
For example: timeout 300s docker pull registry.access.redhat.com/rhceph//rhceph-3-rhel7:latest which is incorrect syntax.
Is not affected when using docker.io defaults for registry due to lack of trailing forward slash in the ceph_docker_registry variable.

Also fixed rhcs_edits.txt for variable ceph_docker_registry.

Signed-off-by: Phuong Nguyen <pnguyen@redhat.com>

### DIFF
--- a/rhcs_edits.txt
+++ b/rhcs_edits.txt
@@ -2,7 +2,7 @@ ceph_repository: rhcs
 ceph_origin: repository
 fetch_directory: ~/ceph-ansible-keys
 ceph_rhcs_version: 3
-ceph_docker_image: "rhceph-3-rhel7"
+ceph_docker_image: "rhceph/rhceph-3-rhel7"
 ceph_docker_image_tag: "latest"
-ceph_docker_registry: "registry.access.redhat.com/rhceph/"
+ceph_docker_registry: "registry.access.redhat.com"
 # END OF FILE, DO NOT TOUCH ME!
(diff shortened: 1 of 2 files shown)


phuonghnguyenau:
Updated PR as requested

rhcs_edits.txt
```diff
@@ -2,7 +2,7 @@ ceph_repository: rhcs
...
 fetch_directory: ~/ceph-ansible-keys
 ceph_rhcs_version: 3
-ceph_docker_image: "rhceph-3-rhel7"
+ceph_docker_image: "rhceph/rhceph-3-rhel7"
 ceph_docker_image_tag: "latest"
-ceph_docker_registry: "registry.access.redhat.com/rhceph/"
-# END OF FILE, DO NOT TOUCH ME!
\ No newline at end of file
+ceph_docker_registry: "registry.access.redhat.com"
+# END OF FILE, DO NOT TOUCH ME!
```
dsavineau: Would it be possible to not touch the last line ?

phuonghnguyenau:
Removed the extra line after the end of line comment as requested

phuonghnguyenau:
> On a second thought could you update the PR to have only one commit ? There's no need to get three commits for this.
> Also add a commit message even if it's a short one.

Squashed the 3 commits into one

PR #3697, opened 2019-03-08
Set the default crush rule in ceph.conf

dsavineau:
Currently the default crush rule value is added to the ceph config
on the mon nodes as an extra configuration applied after the template
generation via the ansible ini module.

This implies two behaviors:

1/ On each ceph-ansible run, the ceph.conf will be regenerated via
ceph-config+template and then ceph-mon+ini_file. This leads to a
non necessary daemons restart.

2/ When other ceph daemons are collocated on the monitor nodes
(like mgr or rgw), the default crush rule value will be erased by
the ceph.conf template (mon -> mgr -> rgw).

This patch adds the osd_pool_default_crush_rule config to the ceph
template and only for the monitor nodes (like crush_rules.yml).
The default crush rule id is read (if exist) from the current ceph
configuration.
The default configuration is -1 (ceph default).

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1638092

### DIFF
--- a/roles/ceph-facts/tasks/facts.yml
+++ b/roles/ceph-facts/tasks/facts.yml
@@ -249,6 +249,23 @@
   set_fact:
     osd_pool_default_min_size: "{{ ceph_conf_overrides.get('global', {}).get('osd_pool_default_min_size', ceph_osd_pool_default_min_size) }}"
 
+- name: check if the ceph conf exists
+  stat:
+    path: '/etc/ceph/{{ cluster }}.conf'
+  register: ceph_conf
+
+- name: get default crush rule value from ceph configuration
+  command: grep 'osd pool default crush rule' /etc/ceph/{{ cluster }}.conf
+  register: crush_rule_variable
+  changed_when: false
+  failed_when: false
+  when: ceph_conf.stat.exists
+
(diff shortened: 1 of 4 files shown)


guits:
bien ouéj maggle!

roles/ceph-facts/tasks/facts.yml
```diff
@@ -249,6 +249,22 @@
   set_fact:
     osd_pool_default_min_size: "{{ ceph_conf_overrides.get('global', {}).get('osd_pool_default_min_size', ceph_osd_pool_default_min_size) }}"
 
+- name: check if the ceph conf exists
+  stat:
+    path: '/etc/ceph/{{ cluster }}.conf'
+  register: ceph_conf
+
+- command: grep 'osd pool default crush rule' /etc/ceph/{{ cluster }}.conf
```
guits: add `- name: get 'osd pool default crush rule' value from /etc/ceph/{{ cluster }}.conf` or something similar please

dsavineau: Sure I'll add that

guits:
jenkins test pipeline

PR #3700, opened 2019-03-09
osd: backward compatibility with old disk_list.sh location

guits:
Since all files in container image have moved to `/opt/ceph-container`
this check must look for new AND the old path so it's backward
compatible. Otherwise it could end up by templating an inconsistent
`ceph-osd-run.sh`.

### DIFF
--- a/roles/ceph-osd/tasks/start_osds.yml
+++ b/roles/ceph-osd/tasks/start_osds.yml
@@ -11,8 +11,16 @@
     when:
       - ceph_docker_on_openstack
 
+  - name: test if the container image has directory {{ container_bin_path }}
+    command: "{{ container_binary }} run --rm --net=host --entrypoint=test {{ ceph_docker_registry }}/{{ ceph_docker_image }}:{{ ceph_docker_image_tag }} -d {{ container_bin_path }}"
+    changed_when: false
+    failed_when: false
+    register: test_container_bin_path
+    when:
+      - osd_scenario != 'lvm'
+
   - name: test if the container image has the disk_list function
-    command: "{{ container_binary }} run --rm --net=host --entrypoint=stat {{ ceph_docker_registry }}/{{ ceph_docker_image }}:{{ ceph_docker_image_tag }} disk_list.sh"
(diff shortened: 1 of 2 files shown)


roles/ceph-osd/tasks/start_osds.yml
```diff
@@ -12,13 +12,21 @@
...
   - name: test if the container image has the disk_list function
-    command: "{{ container_binary }} run --rm --net=host --entrypoint=stat {{ ceph_docker_registry }}/{{ ceph_docker_image }}:{{ ceph_docker_image_tag }} disk_list.sh"
+    command: "{{ container_binary }} run --rm --net=host --entrypoint=stat {{ ceph_docker_registry }}/{{ ceph_docker_image }}:{{ ceph_docker_image_tag }} /opt/ceph-container/disk_list.sh"
     changed_when: false
     failed_when: false
     register: disk_list
     when:
       - osd_scenario != 'lvm'
 
+  - name: test if the container image has the old disk_list function (backward compatibility)
```
guits: could be nice to think to a nicer approach

guits:
(wip)

roles/ceph-osd/tasks/start_osds.yml
leseb: perhaps a loop?

dsavineau: We could introduce a new variable for the container bin directory path (default to /opt/ceph-container/bin) and then run a test -d {dir} on the container image. If the directory isn't present we set the path to '' or '/'.
WDYT ?

roles/ceph-osd/tasks/start_osds.yml
```diff
@@ -12,13 +12,21 @@
       - ceph_docker_on_openstack
 
   - name: test if the container image has the disk_list function
-    command: "{{ container_binary }} run --rm --net=host --entrypoint=stat {{ ceph_docker_registry }}/{{ ceph_docker_image }}:{{ ceph_docker_image_tag }} disk_list.sh"
+    command: "{{ container_binary }} run --rm --net=host --entrypoint=stat {{ ceph_docker_registry }}/{{ ceph_docker_image }}:{{ ceph_docker_image_tag }} /opt/ceph-container/disk_list.sh"
```
dsavineau: disk_list.sh is located in /opt/ceph-container/bin directory

roles/ceph-osd/tasks/start_osds.yml
guits: sounds like a good idea

PR #3701, opened 2019-03-09
osd: support numactl options on OSD activate

guits:
This commit adds OSD containers activate with numactl support.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1684146

### DIFF
--- a/roles/ceph-osd/templates/ceph-osd-run.sh.j2
+++ b/roles/ceph-osd/templates/ceph-osd-run.sh.j2
@@ -70,6 +70,10 @@ fi
 # MAIN #
 ########
 
+{% if ceph_osd_numactl_opts != "" %}
+numactl \
+{{ ceph_osd_numactl_opts }} \
+{% endif %}
 /usr/bin/{{ container_binary }} run \
   --rm \
   --net=host \
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -352,8 +352,8 @@ dummy:
 #cephfs_metadata: cephfs_metadata # name of the metadata pool for a given filesystem
 
 #cephfs_pools:
-#  - { name: "{{ cephfs_data }}", pgs: "{{ osd_pool_default_pg_num }}", size: "{{ osd_pool_default_size }}", min_size: "{{ osd_pool_default_size }}" }
-#  - { name: "{{ cephfs_metadata }}", pgs: "{{ osd_pool_default_pg_num }}", size: "{{ osd_pool_default_size }}", min_size: "{{ osd_pool_default_size }}" }
+#  - { name: "{{ cephfs_data }}", pgs: "{{ osd_pool_default_pg_num }}", size: "{{ osd_pool_default_size }}", min_size: "{{ osd_pool_default_min_size }}"}
+#  - { name: "{{ cephfs_metadata }}", pgs: "{{ osd_pool_default_pg_num }}", size: "{{ osd_pool_default_size }}", min_size: "{{ osd_pool_default_min_size }}"}
 
 ## OSD options
 #
@@ -583,7 +583,7 @@ dummy:
 #  expected_num_objects: ""
 #  application: "rbd"
 #  size: "{{ osd_pool_default_size }}"
-#  min_size: "{{ osd_pool_default_size }}"
+#  min_size: "{{ osd_pool_default_min_size }}"
 #openstack_cinder_pool:
 #  name: "volumes"
 #  pg_num: "{{ osd_pool_default_pg_num }}"
@@ -594,7 +594,7 @@ dummy:
 #  expected_num_objects: ""
 #  application: "rbd"
 #  size: "{{ osd_pool_default_size }}"
-#  min_size: "{{ osd_pool_default_size }}"
+#  min_size: "{{ osd_pool_default_min_size }}"
 #openstack_nova_pool:
 #  name: "vms"
 #  pg_num: "{{ osd_pool_default_pg_num }}"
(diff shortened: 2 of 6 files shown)


guits:
@fultonj

fultonj:
@bengland2 fyi

fultonj:
@guits can this get backported to 3.2?

guits:

bengland2:
This is a starting point.  In particular the use of numactl --preferred will prevent swapping or OOM when NUMA node 0 runs out of memory (but other NUMA nodes have free memory).  It will work for certain cases where we want to restrict OSDs to a single socket.

But it doesn't take into account which NUMA socket is connected to the network NIC or NVM devices, and does not allow user to override which NUMA socket is used.   @markhpc is this consistent with changes Sage was making for NUMA?  See [this pull request for NUMA support](https://github.com/ceph/ceph/pull/25792).   

Another option, for non-hyperconverged storage, is to balance the OSDs across NUMA sockets.   This gets advantages of NUMA locality for any OSD X, keeping all threads for OSD X on the same NUMA socket, while allowing OSDs in aggregate to consume more CPU and memory than is available on 1 NUMA socket.    This can be done by evenly distributing OSDs across NUMA nodes.  I used to use a round-robin method [back in /etc/init.d](http://perf1.perf.lab.eng.bos.redhat.com/pub/bengland/public/ceph/hyperconv/numa/etc-init.d-ceph.patch) days but that wouldn't work anymore.

roles/ceph-osd/templates/ceph-osd-run.sh.j2
```diff
@@ -70,6 +70,10 @@ fi
 # MAIN #
 ########
 
+{% if ceph_osd_numactl_opts != "" %}
+numactl \
```
dsavineau: @guits Not sure if numactl is available by default. Maybe it's worth to install the associated package in the ceph-osd role.

fultonj:
> This is a starting point. In particular the use of numactl --preferred will prevent swapping or OOM when NUMA node 0 runs out of memory (but other NUMA nodes have free memory). It will work for certain cases where we want to restrict OSDs to a single socket.
> 
> But it doesn't take into account which NUMA socket is connected to the network NIC or NVM devices, and does not allow user to override which NUMA socket is used. 

Wait, it _does_ allow the user to override the NUMA socket used. By default the new variable ceph_osd_numactl_opts is an empty string and numactl is not used to start an OSD. If you set the ceph_osd_numactl_opts  variable to "-N 0 --preferred=0" (e.g. in all.yml), then it will use preferred and node 0. The user can set this variable to whatever they want. E.g. membind, though I understand this is not desirable, or a different NUMA socket.

> @markhpc is this consistent with changes Sage was making for NUMA? See [this pull request for NUMA support](https://github.com/ceph/ceph/pull/25792).

My understanding is that ceph PR 25792 is only available in Nautilus. This patch let's Lumionus users use numactl start OSDs with docker. I don't expect this option would be used with Nautilus, though technically they could.

> Another option, for non-hyperconverged storage, is to balance the OSDs across NUMA sockets. This gets advantages of NUMA locality for any OSD X, keeping all threads for OSD X on the same NUMA socket, while allowing OSDs in aggregate to consume more CPU and memory than is available on 1 NUMA socket. This can be done by evenly distributing OSDs across NUMA nodes. I used to use a round-robin method [back in /etc/init.d](http://perf1.perf.lab.eng.bos.redhat.com/pub/bengland/public/ceph/hyperconv/numa/etc-init.d-ceph.patch) days but that wouldn't work anymore.

So you want each OSD to alternate between NUMA nodes. This patch actually does let you do that, but it's sneaky. You'd have to set the ceph_osd_numactl_opts as follows in your all.yml (or similar)

`ceph_osd_numactl_opts: "-N $(($1%2)) --preferred=$(($1%2))"`

Because in the run_osd_start.sh $1 is the number of the OSD, the %2 would round robin it between NUMA nodes 0 and 1. If you %N, where N is the number of sockets the pattern will continue to work. Here's an example:
```
[fultonj@skagra ~]$ cat alternate.sh 
#!/bin/bash
export OPT="-N $(($1%2)) --preferred=$(($1%2))"
echo $OPT
[fultonj@skagra ~]$ for i in $(seq 0 9); do ./alternate.sh $i; done
-N 0 --preferred=0
-N 1 --preferred=1
-N 0 --preferred=0
-N 1 --preferred=1
-N 0 --preferred=0
-N 1 --preferred=1
-N 0 --preferred=0
-N 1 --preferred=1
-N 0 --preferred=0
-N 1 --preferred=1
[fultonj@skagra ~]$ 
```

bengland2:
@fultonj  I took a look at stable-3.2 branch of ceph-ansible and you're right, my bad.    It's really not user-friendly, particularly the round-robin thing, and there are too many ways that inexperienced users could mess it up, but yes you can do it.   So I guess then the question becomes:  how do we expose higher-level policies to the user that let them ask for the end-result that they want without having to know how numactl works?

And more troubling to me is: why doesn't the Linux scheduler prefer putting threads within the same process on the same NUMA node, when possible?  And why isn't there a primitive that says "please schedule this process near device X".   If it did this, then none of this probably would have been necessary, right?  But I'll have to take that up with a different group.

roles/ceph-osd/templates/ceph-osd-run.sh.j2
fultonj: @dsavineau I agree and I think that's a good idea for a follow up patch. It's not included in Red Hat based distros by default. FWIW in the OpenStack case we're triggering ceph-ansible with a tag to not install any packages as we prefer to let TripleO's image have the packages pre-installed. Here's the patch to have TripleO do this: 

https://review.openstack.org/#/c/642466

dsavineau: See https://github.com/ceph/ceph-ansible/pull/3705 for the follow up

ppokorny:
Well, numactl does allow "ip:HOST" as a "node number" so if you want the cores "near" a network address you can specify that.  And "-l" aka "--localalloc" is equivalent to --preferred with the same node number so that would probably be the better thing to use.

It also allows "block:PATH" as a "node number" as well, so if you want to schedule the OSD near the storage, you've got that option as well.

And, "netdev:DEV" is also an option if you know the NIC but not the IP address.

PR #3704, opened 2019-03-11
fix how os.path.join() is being used

rishabh-d-dave:
**First issue fixed by fixed by PR -**
In most places os.path.join() is passed single string -
```
os.path.join(pathcomponent1 + pathcomponent2 + pathcomponent3 + ...)
```

The way to use it is -
```
os.path.join(pathcomponent1, pathcomponent2, pathcomponent3 + ...)
```

Passing a single string doesn't lead to any checks.

**Second issue fixed by this PR -**
Instances like the ones below doesn't need any check from `os.path.join()`, since there's only 1 path component.
```
os.path.join('--entrypoint=' + binary),
````
There were 2 of them; I've deleted both.

**Third issue fixed by this PR -**
Appending strings during the call makes it *very probable* that an incorrect way of calling (like above) might be missed. It's difficult to figure in following call whether strings are being appended to form directory/file names or to form the whole path (which might mean missing a slash it does happen in last 2 instances)
```
./library/test_ceph_key.py:107:        fake_file_destination = os.path.join(fake_dest, fake_cluster + "." + fake_name + ".keyring")
./library/test_ceph_key.py:136:        fake_file_destination = os.path.join(fake_dest, fake_cluster + "." + fake_name + ".keyring")
./library/test_ceph_key.py:175:        fake_file_destination = os.path.join(fake_dest, fake_cluster + "." + fake_name + ".keyring")
./library/test_ceph_key.py:198:        fake_file_destination = os.path.join(fake_dest, fake_cluster + "." + fake_name + ".keyring")
./library/test_ceph_key.py:246:        fake_file_destination = os.path.join(fake_dest, fake_cluster + "." + fake_name + ".keyring")
./library/test_ceph_key.py:279:        fake_file_destination = os.path.join(fake_dest, fake_cluster + "." + fake_name + ".keyring")
./library/test_ceph_key.py:437:        fake_file_destination = os.path.join(fake_dest, fake_cluster + "." + fake_name + ".keyring")
./library/test_ceph_key.py:461:        fake_file_destination = os.path.join(fake_dest, fake_cluster + "." + fake_name + ".keyring")
./library/test_ceph_key.py:474:        fake_key = os.path.join("/var/lib/ceph/mon/", fake_cluster + "-" + fake_hostname + "/keyring") # noqa E501
./library/test_ceph_key.py:486:        fake_key = os.path.join("/var/lib/ceph/mon/", fake_cluster + "-" + fake_hostname + "/keyring") # noqa E501
```

### DIFF
--- a/library/ceph_key.py
+++ b/library/ceph_key.py
@@ -332,8 +332,8 @@ def create_key(module, result, cluster, name, secret, caps, import_key, dest, co
 
     if import_key:
         user = "client.admin"
-        user_key = os.path.join(
-            "/etc/ceph/" + cluster + "." + user + ".keyring")
+        user_key = os.path.join("/etc/ceph/", cluster + "." + user + \
+                                ".keyring")
         cmd_list.append(generate_ceph_cmd(
             cluster, args, user, user_key, container_image))
 
@@ -354,8 +354,7 @@ def update_key(cluster, name, caps, container_image=None):
 
     args = generate_caps(args, "ceph", caps)
     user = "client.admin"
-    user_key = os.path.join(
-        "/etc/ceph/" + cluster + "." + user + ".keyring")
+    user_key = os.path.join("/etc/ceph/", cluster + "." + user + ".keyring")
     cmd_list.append(generate_ceph_cmd(
         cluster, args, user, user_key, container_image))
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: library/ceph_key.py [hunk 1]
```diff
@@ -557,13 +556,13 @@ def run_module():
     # will return either the image name or None
     container_image = is_containerized()
 
+    keyring_filename = cluster + '.' + user + '.keyring'
```
dsavineau: user variable is undefined is this case

rishabh-d-dave: Fixed.

rishabh-d-dave:
@guits @dsavineau 
Can we hurry a bit with this since my other PRs depend on this one?  :)

PR #3547, opened 2019-01-28
allow adding a monitor to a deployed cluster

rishabh-d-dave:
This (early) PR aims to add a playbook, `add-mon.yml`, to `infrastructure-playbooks` for adding a monitor to already deployed Ceph cluster.

PR #3716, opened 2019-03-15
a full stop should not be part of the filename

rishabh-d-dave:
Besides, ceph_key.py takes care of adding a full stop since elsewhere
full stop is not made part of the filename.

### DIFF
--- a/library/ceph_key.py
+++ b/library/ceph_key.py
@@ -121,7 +121,7 @@ caps:
 
 - name: create monitor initial keyring
   ceph_key:
-    name: mon.
+    name: mon
     state: present
     secret: AQAin8tUMICVFBAALRHNrV0Z4MXupRw4v9JQ6Q==
     caps:
--- a/roles/ceph-mon/tasks/deploy_monitors.yml
+++ b/roles/ceph-mon/tasks/deploy_monitors.yml
@@ -13,7 +13,7 @@
 
 - name: create monitor initial keyring
   ceph_key:
-    name: mon.
+    name: mon
     state: present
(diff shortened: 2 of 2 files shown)


guits:
@rishabh-d-dave why did you feel the need to write this commit ?

rishabh-d-dave:
While working on PR #3547, I use [this patch](https://paste.fedoraproject.org/paste/YdRFzDQQuAPztJSqeNpHDA) to circumvent the following error -

```
TASK [ceph-mon : create monitor initial keyring] ******************************************************************************************************************************************************************
task path: /home/rishabh/repos/ceph-ansible/add-mon/roles/ceph-mon/tasks/deploy_monitors.yml:12
Monday 18 March 2019  15:01:46 +0530 (0:00:02.190)       0:03:41.708 **********
fatal: [mon0]: FAILED! => changed=false 
  msg: path /var/lib/ceph/tmp/ceph.mon..keyring does not exist
  path: /var/lib/ceph/tmp/ceph.mon..keyring
  state: absent
```

Besides this, from a general perspective, full stops are separators between filename and extension and not a part of filename. We can move them to ceph-key plugin instead.

rishabh-d-dave:
> The change breaks the mon deployment because there's other mon. occurence in the deploy_monitors.yml.
> From a mon name convention, the logical id should always be mon.
> At this step, because we are bootstraping the monitor, it makes sense to use mon.

Ok. I see that we have mon.keyring and mon..keyring. In that case, my interpretation of this error is not correct, I guess.

Issue #3565, opened 2019-02-02
Python3 seems to break TASK [ceph-mon : create monitor initial keyring]


pcfe:
**Bug Report**

# What happened:
Using stable-3.2 to control Fedora ARM 29 nodes, when I use Python3 on those ARM nodes; the firewall gets set up as expected but I get a failure on `TASK [ceph-mon : create monitor initial keyring]`. 

To be able to run a copy of `site.yml.sample`, I have to use the default of Pyton2 on those Fedora ARM 29 nodes and can thus not configure the firewall (It is not ceph-ansible's problem that F29 offers no python2-firewall).

## details with Python3
While `ansible_python_interpreter=/usr/bin/python3` allows me to configure firewall  (`configure_firewall: True`) it fails on `TASK [ceph-mon : create monitor initial keyring]`

```
TASK [ceph-mon : create monitor initial keyring] ****************************************************************************************
Saturday 02 February 2019  13:22:05 +0100 (0:00:00.578)       0:03:51.103 *****
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: rstrip arg must be None or str
fatal: [odroid-hc2-00]: FAILED! => {"changed": false, "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_wt9j1z5d/ansible_module_ceph_key.py\", line 697, in <module>\n    main()\n  File \"/tmp/ansible_wt9j1z5d/ansible_module_ceph_key.py\", line 693, in main\n    run_module()\n  File \"/tmp/ansible_wt9j1z5d/ansible_module_ceph_key.py\", line 681, in run_module\n    stdout=out.rstrip(b\"\\r\\n\"),\nTypeError: rstrip arg must be None or str\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 1}
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: rstrip arg must be None or str
[...]

PR #3725, opened 2019-03-19
ceph_volume: fix rstrip for python 3

Bruceforce:
Since the same effect as described in #3565 also applies to ceph_volume it would be nice if this could be backported to stable-3.2 (in master this was already changed in a96e910114529aba0fe29edb11b314ae742ece45).

### DIFF
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -630,8 +630,8 @@ def run_module():
         end=str(endd),
         delta=str(delta),
         rc=rc,
-        stdout=out.rstrip(b'\r\n'),
-        stderr=err.rstrip(b'\r\n'),
+        stdout=out.rstrip('\r\n'),
+        stderr=err.rstrip('\r\n'),
         changed=changed,
     )
 


Bruceforce:
Done, but it's not in master for https://github.com/ceph/ceph-ansible/blob/master/library/ceph_crush.py#L196

dsavineau:
> Done, but it's not in master for https://github.com/ceph/ceph-ansible/blob/master/library/ceph_crush.py#L196

Oh I didn't check that sorry. In that case, you should keep this PR to ceph_volume only (sorry for the unnecessary work) and you could open another PR on master for ceph_crush (this one will be backported to stable-3.2 automatically).
Is that ok for you ?

dsavineau:
If you could re-add the `ceph_volume:` prefix in the commit message it would be great !

Bruceforce:
Okay, I hope I got everything right now 😁

dsavineau:
Yes thanks !

PR #3726, opened 2019-03-19
ceph_crush: fix rstrip for python 3

Bruceforce:
Removing bytes literals since rstrip only supports type String or None.

Please backport to stable-3.2 as discussed in #3725


dsavineau:
We probably need to remove `encoding=None` from the [run_command call](https://github.com/ceph/ceph-ansible/blob/master/library/ceph_crush.py#L148)

PR #3731, opened 2019-03-20
purge-docker-cluster: Remove ceph-osd service

dsavineau:
The systemd ceph-osd@.service file used for starting the ceph osd
containers is used in all osd_scenarios.
Currently purging a containerized deployment using the lvm scenario
didn't remove the ceph-osd systemd service.
If the next deployment is a non-containerized deployment, the OSDs
won't be online because the file is still present and override the
one from the package.


guits:
jenkins test pipeline

Issue #3728, opened 2019-03-19
containerized deployment: possibly missing lvmetad on host to allow mounting of devices in containers


devster31:
**Bug Report**

What happened: upon rebooting the server the `systemd` services that spin up ceph containers started but was unable to mount `/dev/ceph-<uuid>` as this was missing in the host.

What you expected to happen: reboot spins up ceph containers

How to reproduce it (minimal and precise): unfortunately I don't have spare machines to test with, but I believe the absence of `lvmetad` on the host machine is the root cause.

Share your group_vars files, inventory
```yaml
--- all.yml
osd_objectstore: bluestore
radosgw_civetweb_port: 7480
ceph_conf_overrides:
  mon:
    'mon allow pool delete': 'true'
ceph_docker_image_tag: latest-nautilus
containerized_deployment: true
containerized_deployment_with_kv: false # does not seem to work
mon_containerized_default_ceph_conf_with_kv: false # does not seem to work
--- mgr.yml
ceph_mgr_modules:
  - status
  - balancer
  - dashboard
  - prometheus
  - restful
  - iostat
ceph_mgr_overrides:
  - "set mgr mgr/dashboard/server_addr '::'"
  - "set mgr mgr/dashboard/server_port 8088"
  - "set mgr mgr/dashboard/ssl false"
--- mons.yml
ceph_mon_docker_memory_limit: "{{ (ansible_memtotal_mb / 4) | round | int }}m"
--- osds.yml
devices:
  - /dev/disk/by-id/ata-HGST_HUH728080ALE600_R5G2DRGV
  - /dev/disk/by-id/ata-HGST_HUH728080ALE600_R6GWWBXY
  - /dev/disk/by-id/ata-HGST_HUH728080ALE600_VLG6K3BY
  - /dev/disk/by-id/ata-HGST_HUH728080ALE600_VLHMNLBY
osd_scenario: lvm
```
**Environment**:
* OS (e.g. from /etc/os-release): Ubuntu LTS 18.04.2
* Kernel (e.g. `uname -a`): Linux 4.15.0-46-generic #49-Ubuntu SMP Wed Feb 6 09:33:07 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
* Docker version if applicable (e.g. `docker version`):
```
Client:
 Version:           18.09.3
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        774a1f4
 Built:             Thu Feb 28 06:53:11 2019
[...]

PR #3732, opened 2019-03-20
ceph-osd: Ensure lvm2 is installed

dsavineau:
When using osd_scenario lvm, we never check if the lvm2 package is
present on the host.
When using containerized deployment and docker on CentOS/RedHat this
package will be automatically installed as a dependency but not for
Ubuntu distribution.
OSD deployed via ceph-volume require the lvmetad.socket to be active
and running.

Resolves: #3728

### DIFF
--- a/roles/ceph-osd/tasks/main.yml
+++ b/roles/ceph-osd/tasks/main.yml
@@ -23,6 +23,17 @@
   tags:
     - with_pkg
 
+- name: install lvm2
+  package:
+    name: lvm2
+  register: result
+  until: result is succeeded
+  when:
+    - osd_scenario == 'lvm'
+    - not is_atomic
+  tags:
+    - with_pkg
+
 - name: include_tasks common.yml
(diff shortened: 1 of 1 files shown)


guits:
looks like the dependency is present for Suse in the specfile but not for rhel:
https://github.com/ceph/ceph/blob/master/ceph.spec.in#L634

Even if it shouldn't hurt to ensure the package is installed at ceph-ansible level, I guess @leseb is right.

alfredodeza:
@leseb Ceph already requires LVM2 when installing the ceph-osd package. Where do you think the dependency should be set?

dsavineau:
This isn't a ceph packaging issue. As @alfredodeza mentionned, ceph-osd package already has a dependency on lvm2.
The problem about this PR was on containerized deployment because we don't install any ceph packages on the system.
On CentOS/RHEL, lvm2 was installed by dependency from docker which is not true for Ubuntu.

PR #3740, opened 2019-03-22
Add uca to ceph_repository choices validation

dsavineau:
Ubuntu cloud archive is configurable via ceph_repository variable but
the uca choice isn't accepted.
This commit fixes this issue and also validates the associated uca
repository variables.

Resolves: #3739

Signed-off-by: Dimitri Savineau <dsavinea@redhat.com>

PR #3742, opened 2019-03-26
Add uca as acceptable ceph_repository value to validate.py

gregoryo2014:
When I set:
```
ceph_repository: uca
```
I get:
`ceph_repository must be either 'community', 'rhcs', 'dev', or 'custom'`

http://docs.ceph.com/ceph-ansible/stable-3.2/installation/methods.html says that uca is acceptable, and when I make this change, my entire build works fine. Pushing the change to validate.py to allow it.

### DIFF
--- a/plugins/actions/validate.py
+++ b/plugins/actions/validate.py
@@ -134,7 +134,7 @@ def ceph_origin_choices(value):
 
 
 def ceph_repository_choices(value):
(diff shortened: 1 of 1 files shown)


guits:
duplicate of #3740 
(btw, do not commit directly to stable-3.2 please)

PR #3748, opened 2019-03-26
tox: Fix container purge jobs

dsavineau:
On containerized CI jobs the playbook executed is purge-cluster.yml
but it should be set to purge-docker-cluster.yml


dsavineau:

PR #3753, opened 2019-03-28
move when keyword before block keyword

rishabh-d-dave:


rishabh-d-dave:
@guits  @dsavineau 

At many places we have -
```
  when:
      - the only condition
```

Would you too find it better to change all of them to -
```
    when: the only condition
```

guits:
I agree when having only one condition it should be written as you suggest here since the `-` actually means `and`

dsavineau:
+1 too if there's only one condition we shouldn't have a list on the when

rishabh-d-dave:
Cool. I'll go ahead.

PR #3760, opened 2019-03-29
tox: Set nautilus as default release

dsavineau:
On stable-4.0 branch we don't want to use dev setup but stable
release (nautilus).
Also update the container image tag to reflect this change.


dsavineau:
@guits I think we need to update ceph-build and add nautilus pipeline first right ?

PR #3769, opened 2019-04-02
purge: fix lvm-batch purge osd

guits:
`lvm_volumes` and/or `devices` variable(s) can be undefined depending on
the scenario chosen.

These tasks should be run only if these variable are defined, otherwise
it ends up with undefined variable errors.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1653307


guits:

guits:
jenkins test dev-centos-non_container-purge

PR #3780, opened 2019-04-03
Check ceph_health_raw.stdout value as string during mon bootstrap

fmount:
According to rdo testing https://review.rdoproject.org/r/#/c/18721 a check on the output of the ceph_health value is added to allow the playbook to make several attempts (according to the retry/delay variables) when waiting the cluster quorum or when the container bootstrap is not ended.
It avoids the failure of the command execution when it doesn't receive a valid json object to decode (because cluster is too slow to boostrap compared to ceph-ansible task execution).


dsavineau:
Could you rebase your patch and squash them into a single commit ?

guits:
closing in favor of #3781

Issue #3790, opened 2019-04-04
wrong variable substitution in create_mds_filesystems.yml


joerncgm:
please change in roles/ceph-mds/tasks/create_mds_filesystems.yml

```
--- roles/ceph-mds/tasks/create_mds_filesystems.yml     2019/04/04 13:07:40     1.1
+++ roles/ceph-mds/tasks/create_mds_filesystems.yml     2019/04/04 13:07:57
@@ -39,7 +39,7 @@
     - check_existing_cephfs.rc != 0

 - name: assign application to cephfs pools
-  command: "{{ hostvars[groups[mon_group_name][0]]['docker_exec_cmd'] | default('') }} ceph --cluster {{ cluster }} osd pool application enable {{ item }} {{ cephfs }}"
+  command: "{{ hostvars[groups[mon_group_name][0]]['docker_exec_cmd'] | default('') }} ceph --cluster {{ cluster }} osd pool application enable {{ item }} cephfs"
   with_items:
     - "{{ cephfs_data }}"
     - "{{ cephfs_metadata }}"

```

The substitution of "cephfs" is wrong. This probably passed tests, because the default value for "cephfs" is also "cephfs", i.e. it evaluates accidentally to the correct string.

PR #3791, opened 2019-04-04
ceph-mds: Set application pool to cephfs

dsavineau:
We don't need to use the cephfs variable for the application pool
name because it's always cephfs.
If the cephfs variable is set to something else than the default
value it will break the appplication pool task.

Resolves: #3790


tests/functional/all_daemons/ceph-override.json
```diff
@@ -1,21 +1,26 @@
...
-		  "pgs": 8,
-		  "size": 1
-		},
-		{
-		  "name": "cephfs_data",
-		  "pgs": 8,
-		  "size": 1
-		}
-	],
+  "ceph_conf_overrides": {
```
guits: could you add `--extra-vars @ceph-override.json` in tox.ini (L64)?

thanks maggle!

PR #3830, opened 2019-04-09
ansible: use bool filter on boolean variables

dsavineau:
Starting ansible 2.8 we now have a deprecation warning about bare
variables not using boolean filter.
This will break us in 2.12 if we don't adapt the code.

```console
[DEPRECATION WARNING]: evaluating dev_setup as a bare variable, this
behaviour will go away and you might need to add |bool to the
expression in the future. Also see CONDITIONAL_BARE_VARS configuration
toggle.. This feature will be removed in version 2.12. Deprecation
warnings can be disabled by setting deprecation_warnings=False in
ansible.cfg.
```

### DIFF
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -254,7 +254,7 @@
       - wait for server to boot
       - remove data
     when:
-      - reboot_osd_node
+      - reboot_osd_node | bool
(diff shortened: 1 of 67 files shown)


guits:
this is typically the kind of PR we shouldn't backport but in the other hand if we don't, there might be a lot of future backport PRs conflicting.

Issue #3609, opened 2019-02-17
ceph-container-common forces installation of docker even when present


devster31:
**Bug Report**

#### What happened ####
I'm running the `site-docker.yml` playbook to test what would happen in my host, `docker` is already been installed as the machine has been provisioned using `docker-machine`, however the playbook would force the removal of the current docker daemon and replace it with the ubuntu package.

#### What you expected to happen ####
the playbook to test for the existence of the `docker` command and skip the installation

#### How to reproduce it ####
command to run is simply `ansible-playbook -D -i ../inventories/local.yml site-docker.yml -C`
here the output (trimmed)
```
install docker on ubuntu...
  sagittarius done | stdout: Reading package lists...
Building dependency tree...
Reading state information...
The following package was automatically installed and is no longer required:
   containerd.io (1.2.2-3)
Use 'sudo apt autoremove' to remove it.
The following additional packages will be installed:
   golang-docker-credential-helpers (0.5.0-2)
   libpython-stdlib (2.7.15~rc1-1)
   libpython2.7-minimal (2.7.15~rc1-1ubuntu0.1)
   libpython2.7-stdlib (2.7.15~rc1-1ubuntu0.1)
   libsecret-1-0 (0.18.6-1)
   libsecret-common (0.18.6-1)
   python (2.7.15~rc1-1)
   python-backports.ssl-match-hostname (3.5.0.1-1)
   python-certifi (2018.1.18-2)
   python-chardet (3.0.4-1)
   python-dockerpycreds (0.2.1-1)
   python-idna (2.6-1)
   python-ipaddress (1.0.17-1)
   python-minimal (2.7.15~rc1-1)
   python-pkg-resources (39.0.1-2)
   python-requests (2.18.4-2ubuntu0.1)
   python-six (1.11.0-2)
   python-urllib3 (1.22-1)
   python-websocket (0.44.0-0ubuntu2)
   python2.7 (2.7.15~rc1-1ubuntu0.1)
   python2.7-minimal (2.7.15~rc1-1ubuntu0.1)
Suggested packages:
   aufs-tools (1:4.9+20170918-1ubuntu1)
   btrfs-progs (4.15.1-1build1)
   docker-doc (18.06.1-0ubuntu1.2~18.04.1)
   rinse (3.2)
   python-doc (2.7.15~rc1-1)
   python-tk (2.7.15~rc1-1)
   python-setuptools (39.0.1-2)
[...]

PR #3867, opened 2019-04-16
ceph-container-common: modify requirement flow

dsavineau:
Until now it was not possible to install a specific container package
because it was somehow hardcoded.
This patch allows to override the container package name (docker.io
vs docker-ce) and refacts the package installation. This could be
achieve via the container_package_name variable.
Instead of using one task per distribution we can set the package and
service name in vars. This allows to have a unified package task.
Also refactorize the debian_prerequisites tasks because the content
was outdated.

https://docs.docker.com/install/linux/docker-ce/debian/
https://docs.docker.com/install/linux/docker-ce/ubuntu/

Resolves: #3609


guits:

guits:
unrelated failures, merging anyway

PR #3875, opened 2019-04-17
ceph-iscsi-gw: Remove library directory

dsavineau:
The library directory that contain the custom ceph modules in present
in the ceph-ansible root directory.
All igw_* mocules are already present there so we don't need the one
present in roles/ceph-iscsi-gw/library.
Also remove the associated spec file.


guits:

PR #3887, opened 2019-04-18
doc: update index.rst with current information for stable-4.0

fghaas:
I noticed today that stable-4.0 has been branched, but that the documentation hadn't yet been updated to include that fact. So I put together what I _think_ is the correct combination of supported Ceph versions and Ansible versions for each branch. If I got something wrong, please correct me!


fghaas:
Whoa that was quick, thanks for merging!

PR #3904, opened 2019-04-23
ceph-osd: Increase cpu limit to 4

dsavineau:
In containerized deployment the default osd cpu quota is too low
for production environment using NVMe devices.
This is causing performance degradation compared to bare-metal.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1695880


guits:
jenkins test dev-centos-container-lvm_osds

Issue #3870, opened 2019-04-16
crash in igw_gateway (tgt) | configure iscsi target (gateway)


hoerup:
Using latest dev builds of nautilus + ceph-ansible/stable-4.0 git + ceph-iscsi 3.0 latest

Playbook crashes with the followin stacktrace:

An exception occurred during task execution. To see the full traceback, use -vvv. The error was: KeyError: 'iqn.2003-01.com.redhat.iscsi-gw:th'
fatal: [hpceph03]: FAILED! => changed=false
  module_stderr: |-
    Traceback (most recent call last):
      File "<stdin>", line 113, in <module>
      File "<stdin>", line 105, in _ansiballz_main
      File "<stdin>", line 48, in invoke_module
      File "/tmp/ansible_igw_gateway_payload_FTb7IO/__main__.py", line 135, in <module>
      File "/tmp/ansible_igw_gateway_payload_FTb7IO/__main__.py", line 105, in ansible_main
      File "/usr/lib/python2.7/site-packages/ceph_iscsi_config/target.py", line 542, in manage
        self.update_acl(config)
      File "/usr/lib/python2.7/site-packages/ceph_iscsi_config/target.py", line 233, in update_acl
        target_config = config.config["targets"][self.iqn]
    KeyError: 'iqn.2003-01.com.redhat.iscsi-gw:th'
  module_stdout: ''
  msg: |-
    MODULE FAILURE
    See stdout/stderr for the exact error
  rc: 1

To be honest, i don't know if it is a bug in ceph-ansible or ceph-iscsi ?

dsavineau:
that's probably on ceph-ansible side because we're using custom ansible modules for ceph iscsi (like igw_gateway) and this has been tested only with 2.x so if the ceph iscsi python binding changed in 3.x that could be related.
The question is maybe more about do we support ceph iscsi 3 with stable-4.0

@guits wdyt ?

hoerup:
Doesn't it also depend on which version of ceph-iscsi should be used together with nautilus ?

Maybe @mikechristie has some input on this ?

ricardoasmarques:
@hoerup looks like your `ceph-iscsi` configuration is outdated.

What is the output of `# gwcli / export copy | grep version`? It should be '**"version": 8**'.

If the config is outdated you have to restart the `rbd-target-api` service to see it updated.

hoerup:
[...]

PR #3930, opened 2019-05-02
[skip ci] Use ceph-iscsi 3.x package names

### LINKED ISSUE #3872: ceph-iscsi 3 package names
From ceph-iscsi v3, the packages ceph-iscsi-cli and ceph-iscsi-config is merged into a single rpm called ceph-iscsi (see https://github.com/ceph/ceph-iscsi/issues/64)

The package install in 
https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-iscsi-gw/tasks/non-container/prerequisites.yml needs to be updated accordingly 

Also the dev repos from shaman in same file should be adjusted

hoerup:
In ceph-iscsi 3.x ceph-iscsi-config and ceph-iscsi-cli are merged into a single ceph-iscsi package.

Fixes #3872

### DIFF
--- a/roles/ceph-iscsi-gw/tasks/non-container/prerequisites.yml
+++ b/roles/ceph-iscsi-gw/tasks/non-container/prerequisites.yml
@@ -11,10 +11,9 @@
         - name: set_fact ceph_iscsi_repos
           set_fact:
             ceph_iscsi_repos:
-              - ceph-iscsi-config
               - tcmu-runner
               - python-rtslib
-              - ceph-iscsi-cli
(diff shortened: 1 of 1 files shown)


hoerup:
Should be backported to stable-4.0 as well

hoerup:
@dsavineau yes, there is more work pending on ceph ansible in regards to ceph-ansible and iscsi - but I don't know the status of the patch that Mike Christie mentions in #3870

hoerup:
Fixed in #3977

Issue #3926, opened 2019-04-28
rolling_update.yml fails at restarting upgraded rgw


JoshuaLyle:
**Bug Report**

What happened: `rolling_update.yml` when upgrading from luminous to mimic fails the playbook every time when running the `restart ceph rgw daemon(s) - non container` handler shown below

```
RUNNING HANDLER [ceph-handler : restart ceph rgw daemon(s) - non container] ************************************************************
Sunday 28 April 2019  15:56:07 -0500 (0:00:01.083)       0:03:38.557 ********** 
skipping: [x.x.x.249] => (item=x.x.x.245) 
failed: [x.x.x.249 -> x.x.x.249] (item=x.x.x.249) => {"changed": true, "cmd": ["/usr/bin/env", "bash", "/tmp/restart_rgw_daemon.sh"], "delta": "0:01:40.251344", "end": "2019-04-28 20:57:49.317083", "item": "x.x.x.249", "msg": "non-zero return code", "rc": 1, "start": "2019-04-28 20:56:09.065739", "stderr": "", "stderr_lines": [], "stdout": "Socket file /var/run/ceph/ceph-client.rgw.xxx.asok could not be found, which means Rados Gateway is not running.
...
```

What you expected to happen:
I expected the playbook to wait for the rgw to come back up and for the playbook to proceed with the update. The upgrade process works fine if you just rerun the playbook because by the time the playbook gets to this point, the rgw has had enough time to come back up properly.

How to reproduce it (minimal and precise):
running `ansible-playbook -i hosts infrastructure-playbooks/rolling_update.yml` with `ceph_stable_release: mimic` on a luminous cluster triggers this

**Environment**:
* OS (e.g. from /etc/os-release): Ubuntu 16.04.6
* Ansible version (e.g. `ansible-playbook --version`): `ansible 2.6.0`
* ceph-ansible version (e.g. `git head or tag or stable branch`): `stable-3.2`
* Ceph version (e.g. `ceph -v`): `12.2.12`

dsavineau:
Could you share your ansible.log, inventory and group_vars/*.yml files ?

JoshuaLyle:
hosts
```
[mons]
x.x.x.245
x.x.x.249
x.x.x.253

[osds]
x.x.x.245
x.x.x.249
x.x.x.253
x.x.x.64
x.x.x.65
x.x.x.66
x.x.x.67

[mdss]
x.x.x.245
x.x.x.249
x.x.x.253

[...]

PR #3946, opened 2019-05-07
ceph-handler: Fix rgw socket in restart script

dsavineau:
Since Mimic the radosgw socket has two extra fields in the socket
name (before the .asok suffix): <pid>.<ctid>

Before:
  /var/run/ceph/ceph-client.rgw.cephaio-1.asok
After:
  /var/run/ceph/ceph-client.rgw.cephaio-1.16913.23928832.asok

The radosgw restart script doesn't handle this and could fail during
an upgrade.
If the SOCKETS variable isn't defined in the script then the test
command won't fail because the return code is 0

$ test -S
$ echo $?
0

There multiple issues in that script:
  - The default SOCKETS value isn't defined due to a typo
SOCKET vs SOCKETS.
  - Because the socket name uses the pid then we need to check the
socket name after the service restart.
  - After restarting the radosgw service we need to wait few seconds
otherwise the socket won't be created.
  - Update the wget parameters because the command is doing a loop.
We now use the same option than curl.
  - The check_rest function doesn't test the radosgw at all due to
a wrong test command (test against a string) and always returns 0.
This needs to use the DOCKER_EXECS variable in order to execute the
command.

$ test 'wget http://192.168.100.11:8080'
$ echo $?
0

Also remove the test based on the ansible_fqdn because we only use
the ansible_hostname + rgw instance name.

Finally group all for loop into a single one.

Resolves: #3926

### DIFF
--- a/roles/ceph-handler/templates/restart_rgw_daemon.sh.j2
+++ b/roles/ceph-handler/templates/restart_rgw_daemon.sh.j2
@@ -14,19 +14,14 @@ fi
 declare -a DOCKER_EXECS
 for ((i=0; i<${RGW_NUMS}; i++)); do
   DOCKER_EXECS[i]=""
-{% if containerized_deployment %}
+{% if containerized_deployment | bool %}
   CONTAINER_NAME="ceph-rgw-${HOST_NAME}-rgw${i}"
   DOCKER_EXECS[i]="{{ container_binary }} exec ${CONTAINER_NAME}"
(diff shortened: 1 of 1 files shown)


dsavineau:
@guits I'm wondering if modifying the restart script is the best way to fix this or if we should fix the socket name in the ceph config (if possible) to have the same socket name structure on all ceph daemons.
wdyt ?

PR #3951, opened 2019-05-09
Create CRUSH rules with device classes

gaudenz:
This allows to set an optional device class for CRUSH rules.

### DIFF
--- a/group_vars/mons.yml.sample
+++ b/group_vars/mons.yml.sample
@@ -43,14 +43,16 @@ dummy:
 
 #crush_rule_hdd:
 #  name: HDD
-#  root: HDD
+#  root: default
 #  type: host
+#  device_class: hdd
 #  default: false
 
 #crush_rule_ssd:
 #  name: SSD
-#  root: SSD
+#  root: default
 #  type: host
+#  device_class: ssd
 #  default: false
 
 #crush_rules:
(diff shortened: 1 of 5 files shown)


guits:
I think this is a breaking change @gaudenz

bencherian:
This won't work when the playbook is run for a new cluster deployment because the crush tree with device classes will not be defined prior to running the play. I had tried doing something similar in an earlier PR #3557  which I closed due to this problem.

gaudenz:
@guits I don't see where this is a breaking change. With the same configuration this creates the same rules as before. `create-simple` is just a simplified version of `create-replicated`.

@bencherian I looked into the issue with not yet existing device classes. The problem is that I could not find a way to create a device class without actually assigning an OSD to it. So we can't precreate the device classes. I see several options to solve this:
* Only create the rules if the device class already exists. This is the simplest solution but would make the behavior non idempotent.
* Move the creation of CRUSH rules to the ceph-osd role.
* Hack up something by manually editing the CRUSH map using `ceph osd getcrushmap`, `crushtool` and `ceph osd setcrushmap`.

I'd prefer the second option. What do you think?

guits:
@gaudenz at first glance I would prefer second option too.
I didn't work so much on that part of the playbook so I'm not sure what could be the impacts of moving this into ceph-osd role.

If you want to give it a try that would be great though :+1:

bencherian:
@gaudenz I agree that the second option (doing rule creation in the ceph-osd role) makes the most sense. However, it does seem a bit odd that the group_vars for defining the rule would then be in mons.yml (And moving everything wholesale to osds.yml would presumably not be backwards compatible.)

PR #3954, opened 2019-05-09
igw: Fix rolling update service ordering

mikechristie:
We must stop tcmu-runner after the other rbd-target-* services because
they may need to interact with tcmu-runner during shutdown. There is
also a bug in some kernels where IO can get stuck in the kernel and by
stopping rbd-target-* first we can make sure all IO is flushed.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1659611

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -737,9 +737,9 @@
         masked: yes
       failed_when: false
       with_items:
-        - tcmu-runner
         - rbd-target-api
         - rbd-target-gw
+        - tcmu-runner
       when: not containerized_deployment
 
     - import_role:


dsavineau:
> We must stop tcmu-runner before the other rbd-target-* services
We must stop tcmu-runner after ?

Also is this related to https://bugzilla.redhat.com/show_bug.cgi?id=1659611 ?

dsavineau:
Ok I've got my answer.
@mikechristie Could you ammend your commit message and add the BZ link ? (Closes/Resolves: <bz link> and change the before/after (if I'm right).

mikechristie:
Yeah, the git commit message should have been "after". I will fix that.

One question. What is the difference between closes and resolves? The patch is needed on top of the patches in this bz to fix up rolling update:

https://bugzilla.redhat.com/show_bug.cgi?id=1659611

dsavineau:
Closes and Resolves are similar just pick one :)

> The patch is needed on top of the patches in this bz to fix up rolling update

The patch mentionned in the BZ is already merged so that's not a problem.

mikechristie:
Thanks. I the git commit message should be fixed now.

guits:

guits:

PR #3964, opened 2019-05-10
Updated rbd_client_admin_socket_path to match ceph-common role

kevincoakley:
Updated the owner, group and mode for rbd_client_admin_socket_path in order to match how rbd_client_admin_socket_path is defined in /roles/ceph-common/tasks/create_rbd_client_dir.yml. The default for the mode in the ceph-common role as defined by rbd_client_directory_mode is 0770 while the mode is hard coded as 0755 in the ceph-rgw role. This was causing a mode clash on /var/run/ceph when the ceph-rgw role is enabled.

### DIFF
--- a/roles/ceph-rgw/tasks/common.yml
+++ b/roles/ceph-rgw/tasks/common.yml
@@ -3,9 +3,9 @@
   file:
     path: "{{ item }}"
     state: directory
-    owner: "{{ ceph_uid if containerized_deployment else 'ceph' }}"
-    group: "{{ ceph_uid if containerized_deployment else 'ceph' }}"
-    mode: "0755"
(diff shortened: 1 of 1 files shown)


guits:
@kevincoakley I think this could break existing deployments

kevincoakley:
@guits It seems like a bug that the owner, group, and permissions are defined as a variable in the ceph-common role for rbd_client_admin_socket_path but hard coded in the ceph-rgw role. Do you have a suggestion for the proper way to solve this issue or should I create an issue for this?

guits:
@kevincoakley I think it might cause issues when collocating rgw with an other ceph daemon

kevincoakley:
Ok, I'll just set `rbd_client_directories: False` to fix the mode clash since ceph-rgw will just override the values.

PR #3987, opened 2019-05-15
purge-docker-cluster: remove is_atomic condition

dsavineau:
There's no reason to not stop the docker service on atomic host. If
this tasks isn't executed then the /var/lib/docker directory removal
will fail.


guits:
well done maggle

PR #3989, opened 2019-05-16
purge-docker-cluster: don't remove data on atomic

dsavineau:
Because we don't manage the docker service on atomic (yet) via the
ceph-container-common role then we can't stop docker dans remove
the data.
For now let's do that only for non atomic hosts.


dsavineau:

dsavineau:

Issue #3994, opened 2019-05-17
add gpg package installation if missing (Ubuntu bionic minimal install)


mistur:
**Bug Report**

What happened: The playbook stop because it didn't find gpg binaries 

What you expected to happen:
it should install gpg package if missing

How to reproduce it (minimal and precise):

I installed all my node with ubuntu 18.04 minimal install, then gpg is not installed by default.
it looks like there is a dependence on gpg package installation. : 

```
<!-- Please let us know any circumstances for reproduction of your bug. -->
TASK [ceph-common : configure debian ceph community repository stable key] **********************************************************************************************************************************************************************************************************************************************
Friday 17 May 2019  11:21:20 +0200 (0:00:00.173)       0:00:44.092 ************ 
FAILED - RETRYING: configure debian ceph community repository stable key (3 retries left).
FAILED - RETRYING: configure debian ceph community repository stable key (3 retries left).
FAILED - RETRYING: configure debian ceph community repository stable key (3 retries left).
FAILED - RETRYING: configure debian ceph community repository stable key (2 retries left).
FAILED - RETRYING: configure debian ceph community repository stable key (2 retries left).
FAILED - RETRYING: configure debian ceph community repository stable key (2 retries left).
FAILED - RETRYING: configure debian ceph community repository stable key (1 retries left).
FAILED - RETRYING: configure debian ceph community repository stable key (1 retries left).
FAILED - RETRYING: configure debian ceph community repository stable key (1 retries left).
fatal: [iccluster015.iccluster.epfl.ch]: FAILED! => changed=false 
  attempts: 3
  msg: 'Failed to find required executable gpg in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin'
fatal: [iccluster023.iccluster.epfl.ch]: FAILED! => changed=false 
  attempts: 3
  msg: 'Failed to find required executable gpg in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin'
fatal: [iccluster019.iccluster.epfl.ch]: FAILED! => changed=false 
  attempts: 3
  msg: 'Failed to find required executable gpg in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin'
```

Share your group_vars files, inventory

**Environment**:
[...]

PR #3996, opened 2019-05-17
common: install dependencies for apt modules

dsavineau:
When using a minimal Debian/Ubuntu distribution there's no
ca-certificates and gpg packages installed so the apt modules will
fail:

Failed to find required executable gpg in paths:
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

apt.cache.FetchFailedException:
W:https://download.ceph.com/debian-luminous/dists/bionic/InRelease:
No system certificates available. Try installing ca-certificates.

Resolves: #3994

### DIFF
--- a/roles/ceph-common/tasks/installs/debian_community_repository.yml
+++ b/roles/ceph-common/tasks/installs/debian_community_repository.yml
@@ -1,4 +1,11 @@
 ---
+- name: install dependencies for apt modules
+  package:
+    name: ['apt-transport-https', 'ca-certificates', 'gpg', 'software-properties-common']
+    update_cache: yes
+  register: result
+  until: result is succeeded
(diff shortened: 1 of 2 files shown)


fghaas:
This patch unfortunately breaks installation on Ubuntu Xenial, where no `gpg` package exists. `gpg` was added in Bionic as a package for "minimalist public key operations" with GnuPG; Xenial has only the full `gnupg` package. I suppose the installation overhead for installing the full `gnupg` package isn't *that* terrible, so may I suggest we just install `gnupg` on Ubuntu and Debian?

fghaas:
Adding this just so the two issues get cross-referenced: see also #4013.

Issue #3995, opened 2019-05-17
add python-routes package installation if missing (Ubuntu bionic minimal install)


mistur:
**Bug Report**
dashboard does not start if python-routes is not install on mgr nodes

What happened:
I got "Module 'dashboard' has failed: No module named routes" in syslog
cf https://tracker.ceph.com/issues/24420

What you expected to happen:
python-routes is installed if missing.

How to reproduce it (minimal and precise):
install a ubuntu 18.04 minimal and deploy a ceph cluster in nautilus

**Environment**:
* OS (e.g. from /etc/os-release): Ubuntu 18.04.2 LTS
* Kernel (e.g. `uname -a`): Linux iccluster007 4.15.0-50-generic #54-Ubuntu SMP Mon May 6 18:46:08 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
* Ansible version (e.g. `ansible-playbook --version`): ansible-playbook 2.8.0
* ceph-ansible version (e.g. `git head or tag or stable branch`): v4.0.0rc7
* Ceph version (e.g. `ceph -v`): Nautilus 14.2.1

dsavineau:
@mistur because ubuntu bionic used python 3 by default do you know if the dashboard module is based on python2 or python3 ?
Because we can install either python-routes or python3-routes

PR #3997, opened 2019-05-17
ceph-mgr: install python-routes for dashboard

dsavineau:
The ceph mgr dashboard requires routes python library to be installed
on the system.

Resolves: #3995


dsavineau:
just need to know if we need only python-routes or python3-routes as well for python3

bencherian:
@dsavineau I have a dashboard installation working on Ubuntu bionic with only python-routes installed.

PR #4001, opened 2019-05-17
container-common: remove centos extras repo task

dsavineau:
This task just doesn't work when using the variable
ceph_docker_enable_centos_extra_repo to true.

```console
fatal: [xxx]; FAILED! => {"changed": false, "msg": "Parameter 'baseurl', 'metalink' or 'mirrorlist' is required."}
```

The CentOS extras repository is enabled by default so it's pretty
safe to remove this task and the associated variable.


dsavineau:
included in #4002

Issue #3947, opened 2019-05-08
Podman support on Ubuntu


alitvak69:
**Is your feature request related to a problem? Please describe.**
Podman doesn't rely on single daemon as docker does, whence no single point of failure.  Podman 1.3 is available for Ubuntu, however, the current code is Redhat 8 heavy.  

**Describe the solution you'd like**
I would like to be able to use podman instead of docker as long as a podman binary is present regardless of a distro.  I would like to be able to use it with Mimic or Nautilus if possible.

**Describe alternatives you've considered**
I considered manually changing systemd units, but it is not a workable solution even for testing.

**Additional context**

dsavineau:
> I would like to be able to use podman instead of docker as long as a podman binary is present regardless of a distro. I would like to be able to use it with Mimic or Nautilus if possible

AFAIK podman requires an extra repository like docker-ce so it won't work out of the box. I'm curious to see if the repository and podman are installed before the ceph-ansible execution (and using `container_binary: podman`) will work or not.

If we want to manage the podman installation in ceph-ansible this should be possible with `container_package_name` like we're doing for RHEL 8 (this can be overrided in the group_vars or as an extra var) but it still requires to setup the repository manually.

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-container-common/vars/RedHat-8.yml

> I considered manually changing systemd units, but it is not a workable solution even for testing.

The systemd units use the `container_binary` variable so you shouldn't have to do this because we're using the same template for docker and podman.

alitvak69:
I believe that adding a ppa:projectatomic/ppa shouldn't be a problem.  With Ubuntu it is a well documented here.  My concerns are:
1.  Conditional checks for RedHat 8 where ever podman is shown, would like to have a distro-neutral installation if possible.
2.  Currently, options for podman are slightly different although mostly is compatible.  For example, logging drivers are not supported as much as with docker and I prefer my containers to send logs to syslog only rather than storing it on a local file system.  What would be a solution for ceph?

alitvak69:
qFor some reason, I am failing to start podman mon container on a second node or third node

May  9 06:13:05 storage2n2-la.xcastlabs.net systemd[1]: Starting Ceph Monitor...
May  9 06:13:06 storage2n2-la.xcastlabs.net podman[38307]: Error: no container with name or ID ceph-mon-storage2n2-la found: no such container
[...]

PR #4003, opened 2019-05-17
container-common: support podman on Ubuntu

dsavineau:
Currently we're only able to use podman on ubuntu if podman's
installation is done manually before the ceph-ansible execution
because the deb package is present in an external repository.
We already manage the docker-ce installation via an external
repository so we should be able to allow the podman installation
with the same mechanism too.

https://github.com/containers/libpod/blob/master/install.md#ubuntu

Resolves: #3947


guits:
not sure what's wrong with mergify here, manually merging anyway...

PR #4010, opened 2019-05-20
nfs: support internal ganesha with external ceph cluster

guits:
This commits allows to deploy an internal ganesha with an external ceph cluster.

This requires to define `external_cluster_mon_ips` with a comma
separated list of external monitors.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1710358

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -474,7 +474,7 @@ dummy:
 # Set this to true to enable File access via NFS.  Requires an MDS role.
 #nfs_file_gw: false
 # Set this to true to enable Object access via NFS. Requires an RGW role.
-#nfs_obj_gw: true
+#nfs_obj_gw: "{{ False if groups.get(mon_group_name, []) | length == 0 else True }}"
 
 
 #############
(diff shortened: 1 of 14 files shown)


guits:
container won't start because of this: https://github.com/ceph/ceph-container/blob/master/src/daemon/start_nfs.sh#L20-L23

gfidente:
I think having to provision the Ganesha client keyring on the ceph-nfs nodes makes perfect sense; I think that can be configured using the "ceph_nfs_ceph_user" variable and provisioned on the node using the "keys" structure (from ceph-client role)? What do you think?

guits:
@gfidente the variable `nfs_obj_gw` is `True` by default in ceph-ansible.

the template for ganesha.conf has [this code](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-nfs/templates/ganesha.conf.j2#L77-L113) which requires `ceph_nfs_rgw_access_key` and `ceph_nfs_rgw_secret_key` to be set.
Usually ceph-ansible generates them itself [here](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-nfs/tasks/create_rgw_nfs_user.yml#L26-L40)
but in this use case, it means they have to be manually set. Is that ok for you?

gfidente:
I think that would be okay ... and in TripleO we only want the MDS FSAL so we disable nfs_obj_gw [1]; do you think things will work when it is disabled?

1. https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/ceph-ansible/ceph-nfs.yaml#L59

guits:
@gfidente yes, it will work.

guits:
jenkins test pipeline

PR #4031, opened 2019-05-23
tests: test podman against atomic os instead rhel8

guits:
the rhel8 image used is an outdated beta version, it is not worth it to
maintain this image upstream, since it's possible to test podman with a
newer version of centos/atomic-host image.


guits:
depends on ceph/ceph-build#1303

guits:
jenkins test dev-centos-container-lvm_osds

guits:

guits:
jenkins test dev-centos-container-lvm_osds

guits:

dsavineau:
@guits I found the issue with podman + reboot on atomic host.
Because we don't execute the ceph-container-common prerequisites on atomic host [1] then we don't create the tmpfile rules for the ceph socket directory (/run/ceph) [2] so this directory isn't persistant on reboot (tmpfs).
That's why the container aren't up & running after a reboot.
```console
May 29 20:52:23 atomic.novalocal systemd[1]: Started Ceph Monitor.
May 29 20:52:23 atomic.novalocal podman[2968]: error checking path "/var/run/ceph": stat /var/run/ceph: no such file or directory
May 29 20:52:23 atomic.novalocal systemd[1]: ceph-mon@atomic.service: main process exited, code=exited, status=125/n/a
```
Note that /var/run/ is a symlink to /run.
I've tried to add the rule manually on an atomic host and it solves the issue.
I'm pretty sure we can safely remove the `not is_atomic` ansible condition in [1].

[1] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-container-common/tasks/main.yml#L2-L4
[2] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-container-common/tasks/prerequisites.yml#L10-L18

dsavineau:
@guits I've updated the PR

dsavineau:
@guits only the OSD nodes don't come back after the reboot

dsavineau:
@guits I tried the updated flags 'dz' in tmpfiles but it doesn't seem to work.
The /run/ceph directory isn't created.

PR #4042, opened 2019-05-31
ceph-facts: generate fsid on mon node

dsavineau:
The fsid generation is done via a python command. When the ansible
controller node only have python3 available (like RHEL 8) then the
python command isn't necessarily present causing the fsid generation
to fail.
We already do some resource creation (like ceph keyring secret) with
the python command too but from the mon node so we should do the same
for fsid.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1714631


guits:
failure in podman scenario will be addressed in a dedicated PR

Issue #4041, opened 2019-05-31
add-osd.yml fails replacing osd after the physical device broke down in collocated and non-collocated scenarios


jadebustos:
**Bug Report**

What happened:

After a disk device crashed and after replacing by a new one using add-osd.yml to replace the osd the following error happens:

ASK [ceph-osd : prepare ceph containerized osd disk collocated] **************************************************************************************************************************************************
Tuesday 28 May 2019  23:32:59 +0200 (0:00:00.107)       0:03:09.062 ***********
[...]

PR #4051, opened 2019-06-05
[skip ci] collocated.yml: adding --rm just to remove the container used to prep…

jadebustos:
collocated.yml: adding --rm just to remove the container used to prepare the osd.

When add-osd.yml is used to add a new osd after an osd failure the execution of the playbook fails. The problem is that the container used to prepare the failed osd still exits (ceph-osd-prepare-ceph3-bs-vdb container for osd vdb). After deleting that container running the playbook again adds the new osd to the cluster.

Replacing a failed osd with a new one works fine with lvm but not using collated osds. This fix works with collocated scenario.

Fixes: #4041

Signed-off-by: Jose Angel de Bustos Perez <jadebustos@gmail.com>

### DIFF
--- a/roles/ceph-osd/tasks/scenarios/collocated.yml
+++ b/roles/ceph-osd/tasks/scenarios/collocated.yml
@@ -4,7 +4,7 @@
 # starting the next task
 - name: prepare ceph containerized osd disk collocated
   shell: |
-    docker run --net=host \
+    docker run --rm --net=host \
     --pid=host \
     --privileged=true \
     --name=ceph-osd-prepare-{{ ansible_hostname }}-{{ item.1 | regex_replace('/dev/', '') }} \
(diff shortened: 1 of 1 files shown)


guits:
@jadebustos, as @dsavineau mentioned [here](https://github.com/ceph/ceph-ansible/issues/4041#issuecomment-497803956) this playbook isn't intended to replace an OSD.

by the way, this patch might break upgrades

dLobatog:
@guits How can this break upgrades? The docker container used to add the OSD should only run once, even in the `add` use case (not replace) what is the point of keeping the container?

dsavineau:
@dLobatog probably because the container prepare is used later https://github.com/ceph/ceph-ansible/blob/stable-3.2/roles/ceph-osd/templates/ceph-osd-run.sh.j2#L28-L32

guits:
> @dLobatog probably because the container prepare is used later https://github.com/ceph/ceph-ansible/blob/stable-3.2/roles/ceph-osd/templates/ceph-osd-run.sh.j2#L28-L32

exactly

PR #4064, opened 2019-06-07
Enabled 'chrony' service instead of 'chronyd'.

fbarbeira:
I tried the following configuration in file 'all.yml':

```
ntp_service_enabled: true
ntp_daemon_type: 'chronyd'
```

but it fails with the following message:

```
TASK [ceph-infra : enable chronyd] ************************************************************************************************************************************************
Friday 07 June 2019  09:05:52 +0200 (0:00:00.664)       0:02:01.979 ***********
fatal: [ceph-monitor03]: FAILED! => changed=false
  msg: 'Could not find the requested service chronyd: host'
fatal: [ceph-monitor02]: FAILED! => changed=false
  msg: 'Could not find the requested service chronyd: host'
fatal: [ceph-monitor01]: FAILED! => changed=false
  msg: 'Could not find the requested service chronyd: host'
[...]
```

With the change proposed it works fine.

OS: Ubuntu 18 bionic.
Ansible: 2.8
Branch: stable-4.0

### DIFF
--- a/roles/ceph-infra/tasks/setup_ntp.yml
+++ b/roles/ceph-infra/tasks/setup_ntp.yml
@@ -45,7 +45,7 @@
 
     - name: enable chronyd
       service:
-        name: chronyd
+        name: chrony
         enabled: yes
         state: started
       notify:


guits:
@rishabh-d-dave I know you've worked a bit on this part of the playbook, do you mind reviewing this PR please?
Thanks!

roles/ceph-infra/tasks/setup_ntp.yml
```diff
@@ -45,7 +45,7 @@
 
     - name: enable chronyd
       service:
-        name: chronyd
+        name: chrony
```
rishabh-d-dave: `chronyd` works fine with CentOS. Therefore, I suppose it's best to add a new task that runs only for Ubuntu.

dsavineau: chronyd is the daemon name on RedHat/CentOS so this change will break it.
So this would require some conditional between linux distribution.

dsavineau:
I guess this one is now duplicate of https://github.com/ceph/ceph-ansible/pull/4089

PR #4066, opened 2019-06-07
validate: fail in check_devices at the right task

guits:
see https://bugzilla.redhat.com/show_bug.cgi?id=1648168#c17 for details.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1648168#c17

### DIFF
--- a/roles/ceph-validate/tasks/check_devices.yml
+++ b/roles/ceph-validate/tasks/check_devices.yml
@@ -2,15 +2,16 @@
 - name: devices validation
   when: devices is defined
   block:
-    - name: validate devices is actually a device
+    - name: get devices information
       parted:
         device: "{{ item }}"
         unit: MiB
       register: devices_parted
+      failed_when: False
       with_items: "{{ devices }}"
 
     - name: fail if one of the devices is not a device
       fail:
         msg: "{{ item }} is not a block special file!"
-      when: item.failed
+      when: item.rc is defined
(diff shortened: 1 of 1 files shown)


guits:
NOTE to myself:
backport in 3.2 will have to apply this change on this part of the code 
```yaml
- name: validate dedicated_device is/are actually device(s)
  parted:
    device: "{{ item }}"
    unit: MiB
  register: dedicated_device_parted
  with_items: "{{ dedicated_devices }}"
  when:
    - osd_scenario == 'non-collocated'

- name: fail if one of the dedicated_device is not a device
  fail:
    msg: "{{ item }} is not a block special file!"
  with_items: "{{ dedicated_device_parted.results }}"
  when:
    - osd_scenario == 'non-collocated'
    - item.failed
```

roles/ceph-validate/tasks/check_devices.yml
```diff
@@ -2,15 +2,16 @@
 - name: devices validation
   when: devices is defined
   block:
-    - name: validate devices is actually a device
+    - name: get devices information
       parted:
         device: "{{ item }}"
         unit: MiB
       register: devices_parted
+      failed_when: False
```
dsavineau: wondering what's the best way to do that compared to using `ignore_errors: true` instead of `failed_when: false`.
With `ignore_errors: true` you don't have to change the next task.
Any thoughts ?

Issue #4072, opened 2019-06-11
Modprobe binary is called in node_exporter with incorrect path in ubuntu.


fbarbeira:
**Bug Report**

Modprobe binary is called with absolute path and it breaks in ubuntu:

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-node-exporter/templates/node_exporter.service.j2#L23

```
Jun 11 07:39:38 host systemd[547614]: node_exporter.service: Failed to execute command: No such file or directory
Jun 11 07:39:38 host systemd[547614]: node_exporter.service: Failed at step EXEC spawning /usr/sbin/modprobe: No such file or directory
Jun 11 07:39:38 host systemd[1]: node_exporter.service: Control process exited, code=exited status=203
Jun 11 07:39:38 host systemd[1]: node_exporter.service: Failed with result 'exit-code'.
Jun 11 07:39:38 host systemd[1]: Failed to start Node Exporter.
```

```
# which modprobe
/sbin/modprobe
# ls -la /usr/sbin/modprobe
ls: cannot access '/usr/sbin/modprobe': No such file or directory
#
```

**Environment**:
* OS (e.g. from /etc/os-release): Ubuntu 18.04.2 LTS

PR #4074, opened 2019-06-11
ceph-node-exporter: use modprobe ansible module

dsavineau:
Instead of using the modprobe command from the path in the systemd
unit script, we can use the modprobe ansible module.
That way we don't have to manage the binary path based on the linux
distribution.

Resolves: #4072


dsavineau:

PR #4076, opened 2019-06-11
ceph-node-exporter: Fix systemd template

dsavineau:
069076b introduced a bug in the systemd unit script template. This
commit fixes the options used by the node-exporter container.


dsavineau:

PR #4103, opened 2019-06-13
bugfix: syntax error when enabling crush_rule_config

peteryj:
With 'unique' filter, it will return error like 

  The task includes an option with an undefined variable. The error was: 
  'ansible.utils.unsafe_proxy.AnsibleUnsafeText object' has no attribute 'name'

And the following task also has no 'unique' filter.
So remove it.

Signed-off-by: Yang Jun <yangjun@qiyi.com>

### DIFF
--- a/roles/ceph-mon/tasks/crush_rules.yml
+++ b/roles/ceph-mon/tasks/crush_rules.yml
@@ -13,7 +13,7 @@
 
 - name: create configured crush rules
   command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} osd crush rule create-simple {{ item.name }} {{ item.root }} {{ item.type }}"
-  with_items: "{{ crush_rules | unique }}"
+  with_items: "{{ crush_rules }}"
   changed_when: false
   when: inventory_hostname == groups.get(mon_group_name) | last
 


dsavineau:
How did you defined the `crush_rules` variable ? because it should be a list of dict with the key `name` present

https://github.com/ceph/ceph-ansible/blob/master/group_vars/mons.yml.sample#L44-L58

peteryj:
@dsavineau 

I defined 'crush_rules' in the inventory file.
`
[mons:vars]
crush_rule_config=true
create_crush_tree=true
crush_rules=[{"name":"hdd-zone1-rule-replicate","root":"hdd-zone1","type":"rack","default":true},{"name":"ssd-zone1-rule-replicate","root":"ssd-zone1","type":"host","default":false}]
`

I have tried 
`
crush_rules="[{"name":"hdd-zone1-rule-replicate","root":"hdd-zone1","type":"rack","default":true},{"name":"ssd-zone1-rule-replicate","root":"ssd-zone1","type":"host","default":false}]"
`
and 
`
crush_rules=['{"name":"hdd-zone1-rule-replicate","root":"hdd-zone1","type":"rack","default":true}','{"name":"ssd-zone1-rule-replicate","root":"ssd-zone1","type":"host","default":false}]'"
`
but all failed.

Using inventory file to define these variables because we have different configurations for different clusters. Do you have other suggestions? Thanks a lot.

dsavineau:
it seems complicated to manage list/dict variables in the inventory file.
As an alternative you can use the extra-vars option with the ansible-playbook command.
Either the full variable with the CLI or via a json file

with the full variable
```console
$ ansible-playbook .... -e '{"crush_rules": [{"name":"hdd-zone1-rule-replicate","root":"hdd-zone1","type":"rack","default":true},{"name":"ssd-zone1-rule-replicate","root":"ssd-zone1","type":"host","default":false}]}'
```

or with file:
```console
$ cat crush_rules.json 
crush_rules:
  - {"name":"hdd-zone1-rule-replicate","root":"hdd-zone1","type":"rack","default":true}
  - {"name":"ssd-zone1-rule-replicate","root":"ssd-zone1","type":"host","default":false}
$ ansible-playbook .... -e @crush_rules.json
```

peteryj:
Thanks a lot  @dsavineau .   It works for me.

PR #4104, opened 2019-06-13
Generate ceph.conf locally for containerized

kplantjr:
Currently we do not generate ceph.conf locally for containerized deployments
when 'ceph_conf_local' is true but, we do generate ceph.conf locally for
non-containerized deployments.

Added a task to the ceph-config role to generate ceph.conf locally for
containerized deployments when 'ceph_conf_local' is true.

roles/ceph-config/tasks/main.yml: added tasks to generate ceph.conf locally

Signed-off-by: Keith Plant <keith.plant@charter.com>


roles/ceph-config/tasks/main.yml
```diff
@@ -130,6 +130,20 @@
       - ((inventory_hostname in groups.get(mon_group_name, [])) or
         (groups.get(nfs_group_name, []) | length > 0) and inventory_hostname == groups.get(nfs_group_name, [])[0])
 
+  - name: "generate {{ cluster }}.conf configuration file locally"
```
dsavineau: does it require to create the local directory first ?

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/tasks/main.yml#L93-L101

kplantjr: the task that runs right before the task i added to create the configuration file should take care of the directory creation

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/tasks/main.yml#L120-L131

dsavineau: it only creates the `fetch_directory` but not the underlying `fsid/etc/ceph` sub directories.
Unless the config_template action is doing this creation I'm not sure if it will work or not.
I'll need to test this

dsavineau: confirmed, it doesn't work

```console
TASK [create a local fetch directory if it does not exist] ***********************************************************************************************************************************************************************************
Thursday 13 June 2019  14:45:12 -0400 (0:00:01.889)       0:00:01.971 ********* 
ok: [conf -> localhost] => changed=false 
  gid: 1000
  group: ds
  mode: '0755'
  owner: ds
  path: fetch
  size: 4096
  state: directory
  uid: 1000

TASK [generate ceph.conf configuration file locally] *****************************************************************************************************************************************************************************************
Thursday 13 June 2019  14:45:12 -0400 (0:00:00.329)       0:00:02.300 ********* 
fatal: [conf -> localhost]: FAILED! => changed=false 
  msg: Destination directory fetch/b491a91e-f5b8-4d9a-865e-da700ca464c8/etc/ceph does not exist

```

kplantjr: You're right, total oversight on my part since I already had those local directories created. I'll revise.

dsavineau:
Could you also rebase your PR ?

Thanks !

PR #4109, opened 2019-06-14
container: add docker2podman tags in main playbook

guits:
This aims to ease the rhel7 to rhel8 upgrade.                                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                                       
Since rhel8 will be podman only, the idea is to prepare all systemd                                                                                                                                                                                                                                                                                                                                                                    
units to use podman before the system is upgraded.                                                                                                                                                                                                                                                                                                                                                                                     
Once the OS is upgraded to rhel8, it will be able to start ceph daemons                                                                                                                                                                                                                                                                                                                                                                
via podman.                                                                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                       
Calling the playbook with `--tags docker2podman` will reuse all existing                                                                                                                                                                                                                                                                                                                                                               
roles in order to render systemd units. However it could be very long to                                                                                                                                                                                                                                                                                                                                                               
run the whole playbook just for few tasks, the idea here is to play with                                                                                                                                                                                                                                                                                                                                                               
tags so we play only required tasks.                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                       
The `docker2podman` variable must be set to `True` so handlers won't be                                                                                                                                                                                                                                                                                                                                                                
trigger to avoid a restart of all services.                                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                       
### DIFF
--- a/infrastructure-playbooks/docker2podman.yml
+++ b/infrastructure-playbooks/docker2podman.yml
@@ -1,0 +1,156 @@
+- hosts:
+  - mons
+  - osds
+  - mdss
+  - rgws
+  - nfss
+  - rbdmirrors
+  - clients
+  - iscsigws
+  - mgrs
+  - grafana-server
+
+  gather_facts: false
+  become: True
+  any_errors_fatal: true
+
+  vars:
+    delegate_facts_host: True
+
+  pre_tasks:
+    - import_tasks: "{{ playbook_dir }}/../raw_install_python.yml"
+
+  tasks:
+    # pre-tasks for following import -
+    - name: gather facts
+      setup:
+      when: not delegate_facts_host | bool
+
+    - name: gather and delegate facts
+      setup:
+      delegate_to: "{{ item }}"
+      delegate_facts: True
+      with_items: "{{ groups['all'] }}"
+      run_once: true
+      when: delegate_facts_host | bool
+
+- hosts:
+    - "{{ mon_group_name }}"
+    - "{{ osd_group_name }}"
+    - "{{ mds_group_name }}"
+    - "{{ rgw_group_name }}"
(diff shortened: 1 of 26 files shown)


roles/ceph-mon/tasks/start_monitor.yml
```diff
@@ -7,6 +7,8 @@
     - not containerized_deployment | bool
     - ceph_mon_systemd_overrides is defined
     - ansible_service_mgr == 'systemd'
+  tags:
```
guits:
```suggestion
  tags: docker2podman
```

roles/ceph-mon/tasks/start_monitor.yml
```diff
@@ -7,6 +7,8 @@
...
+    - docker2podman
```
guits:
```suggestion
```

infrastructure-playbooks/docker2podman.yml
```diff
@@ -0,0 +1,204 @@
+---
```
guits: note sure whether having a dedicated playbook is really useful in the end

I thought I could have a section like following
```yaml
- hosts: all
  tasks:
    - import_role:
        name: ceph-defaults
      tags: ['docker2podman']
    - import_role:
        name: ceph-facts
      tags: ['docker2podman']
```

so I could run those roles on all nodes at a time instead of playing them for each hosts section but I must have been confused at some point

PR #4110, opened 2019-06-14
tests: Update ansible ssh_args variable

dsavineau:
Because we're using vagrant, a ssh config file will be created for
each nodes with options like user, host, port, identity, etc...
But via tox we're override ANSIBLE_SSH_ARGS to use this file. This
remove the default value set in ansible.cfg.

Also adding PreferredAuthentications=publickey because CentOS/RHEL
servers are configured with GSSAPIAuthenticationis enabled for ssh
server forcing the client to make a PTR DNS query.


guits:
jenkins test pipeline

guits:
jenkins test pipeline

guits:
jenkins test pipeline

PR #4128, opened 2019-06-19
Set grafana_server_addr fact for ipv6 scenarios.

fmount:
As the [bz1721914](https://bugzilla.redhat.com/show_bug.cgi?id=1721914) describes, the grafana_server_addr
fact is not defined if ip_version used is ipv6.
This commit adds the ip_version condition to set
correctly this fact.

### DIFF
--- a/roles/ceph-facts/tasks/facts.yml
+++ b/roles/ceph-facts/tasks/facts.yml
@@ -303,9 +303,18 @@
         chrony_daemon_name: chrony
       when: ansible_os_family == "Debian"
 
-- name: set grafana_server_addr fact
+- name: set grafana_server_addr fact - ipv4
   set_fact:
     grafana_server_addr: "{{ (hostvars[groups[grafana_server_group_name][0] if groups.get(grafana_server_group_name, []) | length > 0 else groups[mgr_group_name][0]])['ansible_all_ipv4_addresses'] | ipaddr(public_network) | first }}"
   when:
     - (groups.get(grafana_server_group_name, []) | length > 0 or groups.get(mgr_group_name, []) | length > 0)
+    - ip_version == 'ipv4'
+    - dashboard_enabled | bool
+
+- name: set grafana_server_addr fact - ipv6
+  set_fact:
+    grafana_server_addr: "{{ (hostvars[groups[grafana_server_group_name][0] if groups.get(grafana_server_group_name, []) | length > 0 else groups[mgr_group_name][0]])['ansible_all_ipv6_addresses'] | ipaddr(public_network) | last | ipwrap }}"
+  when:
+    - (groups.get(grafana_server_group_name, []) | length > 0 or groups.get(mgr_group_name, []) | length > 0)
+    - ip_version == 'ipv6'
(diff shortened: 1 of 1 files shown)


roles/ceph-facts/tasks/facts.yml
```diff
@@ -302,7 +302,16 @@
...
   set_fact:
     grafana_server_addr: "{{ (hostvars[groups[grafana_server_group_name][0] | default(groups[mgr_group_name][0])])['ansible_all_ipv4_addresses'] | ipaddr(public_network) | first }}"
-  when:  groups.get(grafana_server_group_name, []) | length > 0 or groups.get(mgr_group_name, []) | length > 0
+  when:
+    - groups.get(grafana_server_group_name, []) | length > 0 or groups.get(mgr_group_name, []) | length > 0
+    - ip_version == 'ipv4'
+
+- name: set grafana_server_addr fact - ipv6
+  set_fact:
+    grafana_server_addr: "{{ (hostvars[groups[grafana_server_group_name][0] | default(groups[mgr_group_name][0])])['ansible_all_ipv6_addresses'] | ipaddr(public_network) | first }}"
```
dsavineau: I don't know if it's the best place here or where the `grafana_server_addr` variable is used but this needs to use brackets for ipv6 by adding the `ipwrap` filter.

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-facts/tasks/set_monitor_address.yml#L14

Also the first filter should be replace by last with ipv6

dsavineau:
Also could you add `Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1721914` in the commit body ?

roles/ceph-facts/tasks/facts.yml
fmount: Thanks for the suggestion, I made the requested changes.
This variable is used in many templates (grafana datasource, prometheus, alertmanager), for this reason I think it's good setting a fact that could be consumed from many roles.

roles/ceph-facts/tasks/facts.yml
```diff
@@ -303,9 +303,18 @@
         chrony_daemon_name: chrony
       when: ansible_os_family == "Debian"
 
-- name: set grafana_server_addr fact
+- name: set grafana_server_addr fact - ipv4
   set_fact:
-    grafana_server_addr: "{{ (hostvars[groups[grafana_server_group_name][0] if groups.get(grafana_server_group_name, []) | length > 0 else groups[mgr_group_name][0]])['ansible_all_ipv4_addresses'] | ipaddr(public_network) | first }}"
+    grafana_server_addr: "{{ (hostvars[groups[grafana_server_group_name][0] | default(groups[mgr_group_name][0])])['ansible_all_ipv4_addresses'] | ipaddr(public_network) | first }}"
```
guits: this reintroduce a bug that was just fixed

fmount: sorry, I correctly rebased the PR now.

guits:
unrelated failure, merging anyway

PR #4131, opened 2019-06-19
Add higher retry/delay defaults to check the quorum status.

fmount:
As per [bz1718981](https://bugzilla.redhat.com/show_bug.cgi?id=1718981), this commit adds higher values to check the quorum status.
This is helpful for several OSP deployments that fail during the scale up.

Signed-off-by: fpantano <fpantano@redhat.com>


guits:
jenkins test pipeline

dsavineau:
jenkins test pipeline

PR #4132, opened 2019-06-20
ceph-volume: Set max open files limit on container

dsavineau:
The ceph-volume lvm list command takes ages to complete when having
a lot of LV devices on containerized deployment.
For instance, with 25 OSDs on a node it takes 3 mins 44s to list the
OSD.
Adding the max open files limit to the container engine cli when
executing the ceph-volume command seems to improve a lot thee
execution time ~30s.

This was impacting the OSDs creation with ceph-volume (both filestore
and bluestore) when using multiple LV devices.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1702285

### DIFF
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -190,6 +190,7 @@ def container_exec(binary, container_image):
     container_binary = os.getenv('CEPH_CONTAINER_BINARY')
     command_exec = [container_binary, 'run',
                     '--rm', '--privileged', '--net=host', '--ipc=host',
+                    '--ulimit', 'nofile=1024:1024',
                     '-v', '/run/lock/lvm:/run/lock/lvm:z',
                     '-v', '/var/run/udev/:/var/run/udev/:z',
                     '-v', '/dev:/dev', '-v', '/etc/ceph:/etc/ceph:z',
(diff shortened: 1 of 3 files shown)


dsavineau:
@andrewschoen @alfredodeza FYI not sure why we need to do this but it seems to improve a lot the ceph-volume execution time in containerized deployment.

dsavineau:
jenkins test pipeline

alfredodeza:
30 seconds to list 25 OSDs is still pretty terrible performance :( Would love to find out what else do we need to enable here, this is certainly not a behaviour we see in baremetal or VMs

dsavineau:
@alfredodeza I'm agree with you and that's only a workaround to improve the current situation but we still need to investigate why we have this behaviour.
BTW we also saw the same issue with ceph-disk.

If you're interesting : https://bugzilla.redhat.com/show_bug.cgi?id=1722562

dsavineau:
@alfredodeza FYI 30 seconds to list 25 OSDs is also the execution time on baremetal.

alfredodeza:
I wasn't expecting that issue on baremetal! I guess we need a bug report specific to this problem on the cpeh-volume side

Thank you for taking the time to explain and point to the other issues

PR #4135, opened 2019-06-20
Change ansible_lsb by ansible_distribution_release

dsavineau:
The ansible_lsb fact is based on the lsb package (lsb-base,
lsb-release or redhat-lsb-core).
If the package isn't installed on the remote host then the fact isn't
populated.

```
"ansible_lsb": {},
```

Switching to the ansible_distribution_release fact instead.


dsavineau:
mergify fails the automatic merge

PR #4172, opened 2019-06-28
dashboard: move code into a dedicated playbook

dsavineau:
Move dashboard, grafana/prometheus and node-exporter plays into a
dedicated playbook in infrastructure-playbook directory.
To avoid using 'dashboard_enabled | bool' condition multiple time
in the main playbook we can just import the dashboard playbook or
not.
This patch also allows to use an unique dashboard playbook for
both baremetal and container playbooks.


dsavineau:
This has been cherry pix in #4268

PR #4176, opened 2019-06-28
tests: check cluster health after reboot

dsavineau:
When we reboot the whole ceph cluster, we don't check the ceph health
before running testinfra.
This could lead to testinfra failures because some ceph daemons could
not be fully up & running.

### DIFF
--- a/tests/functional/reboot.yml
+++ b/tests/functional/reboot.yml
@@ -1,9 +1,36 @@
 ---
 - hosts: all
-  gather_facts: true
+  gather_facts: false
   tasks:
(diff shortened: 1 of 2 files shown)


guits:
@dsavineau I'm -1 on merging this as it is.
In case of valid failure after a reboot, I would prefer testinfra to report what's wrong so we have more details.
We used to have a `sleep 30` before running testinfra. Looks like I removed it (probably a mistake) in d5be83e5042a5e22ace6250234ccd81acaffb0a2

PR #4191, opened 2019-07-03
Scale mons

Ilyashestopalov:
The ability to add nodes with the monitor role to an existing cluster whose name differs from the default name is fixed.


dsavineau:
You shouldn't send a PR against stable branch for this kind of issue. Instead you need to create a PR agains master and we will backport the change if necessary.
Also I think you messed up with your PR because you shouldn't have 217 commits.

dsavineau:
The change has been added in https://github.com/ceph/ceph-ansible/pull/4196

PR #4193, opened 2019-07-04
add tag on ceph-grafana-dashboards package.

fmount:
According to the OSP pattern, we need the package-install tag
to control what is installed on the host. This commit just add
the missing tag to meet the TripleO requirements.

See: /issues/4197 for details

Fixes: #4197

### DIFF
--- a/roles/ceph-grafana/tasks/configure_grafana.yml
+++ b/roles/ceph-grafana/tasks/configure_grafana.yml
@@ -8,6 +8,7 @@
   when:
     - not containerized_deployment
     - ansible_os_family in ['RedHat', 'Suse']
+  tags: package-install
 
 - name: make sure grafana is down
   service:


fultonj:
@fmount in a containerized scenario we prefer to NOT install packages on the container host. E.g. we want it to be able to run on atomic.

fultonj:
I think it's a bug that this task isn't tagged with skip-tags so I opened a bug report for it

https://github.com/ceph/ceph-ansible/issues/4197

fmount:
@fultonj  agree for the package-install issue. 
I wasn't totally aware about the current pattern, but I definitively agree with that.
First thing I added the missing tag on the task, according to [1] on TripleO side we'll be able to skip the tag and handle this requirement differently.

[1] https://github.com/openstack/tripleo-ansible/blob/master/tripleo_ansible/roles/tripleo-ceph-common/defaults/main.yml#L135

@guits @dsavineau do you have considerations to add on this?

roles/ceph-grafana/tasks/configure_grafana.yml
```diff
@@ -6,8 +6,8 @@
   register: result
   until: result is succeeded
   when:
-    - not containerized_deployment
     - ansible_os_family in ['RedHat', 'Suse']
+  tags: package-install
```
fultonj: I like adding this tag because it fixes https://github.com/ceph/ceph-ansible/issues/4197

roles/ceph-grafana/tasks/configure_grafana.yml
```diff
@@ -6,8 +6,8 @@
   register: result
   until: result is succeeded
   when:
-    - not containerized_deployment
```
fultonj: Shouldn't we keep this because doing a contianerized deployment, to me, means not installing packages?

fmount: The _ceph-grafana-dashboards_ package provides a set of yaml files that Grafana uses as the default dashboard layout [1].
Removing this boolean we need to update the grafana container config [2] or, if the package is not present, the container will fail with the following:

```
lvl=eror msg="server shutdown" logger=server reason="service init failed: failed to provision dashboards: failed to provision config ceph dashboard: stat /etc/grafana/dashboards/ceph-dashboard: no such file or directory"
```

This happens because, as the systemd unit shows [3], we mount the whole /etc/grafana to properly configure the container: without this volume it will result in a classic grafana-server without the dashboard layouts  but with the prometheus datasource configured, (in this case, the operator should  configure the cluster).
I suggest to decouple the grafana bootstrapping phase with the grafana config phase, so we can better handle how this package is provided and what kind of config we need to pass to the grafana server.
Before update the PR, i would like to collect some considerations.

@fultonj , @guits , @dsavineau , what do you think?


[1] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/defaults/main.yml#L724
[2] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-grafana/templates/dashboards-ceph-dashboard.yml.j2#L12
[3] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-grafana/templates/grafana-server.service.j2#L16

roles/ceph-grafana/tasks/configure_grafana.yml
```diff
@@ -6,8 +6,9 @@
...
     - ansible_os_family in ['RedHat', 'Suse']
+    - not containerized_deployment
```
guits: do not modify this line please

roles/ceph-grafana/tasks/configure_grafana.yml
guits: @fmount see https://github.com/ceph/ceph/pull/26415

roles/ceph-grafana/tasks/configure_grafana.yml
fmount: +1

guits:
unrelated failure

PR #4204, opened 2019-07-08
mon: create crush rule from osd class

mslovy:
osd class can distinguwish ssd from hdd now so that we can leverage the features. 

Resolves: #4400
Signed-off-by: Ning Yao <yaoning@unitedstack.com>

### DIFF
--- a/group_vars/mons.yml.sample
+++ b/group_vars/mons.yml.sample
@@ -43,14 +43,16 @@ dummy:
 
 #crush_rule_hdd:
 #  name: HDD
-#  root: HDD
+#  root: default
 #  type: host
+#  class: hdd
 #  default: false
 
 #crush_rule_ssd:
 #  name: SSD
-#  root: SSD
+#  root: default
 #  type: host
+#  class: ssd
 #  default: false
 
 #crush_rules:
--- a/roles/ceph-mon/defaults/main.yml
+++ b/roles/ceph-mon/defaults/main.yml
@@ -35,14 +35,16 @@ crush_rule_config: false
 
 crush_rule_hdd:
   name: HDD
-  root: HDD
+  root: default
   type: host
+  class: hdd
   default: false
 
 crush_rule_ssd:
   name: SSD
-  root: SSD
+  root: default
   type: host
+  class: ssd
   default: false
 
 crush_rules:
--- a/roles/ceph-mon/tasks/crush_rules.yml
+++ b/roles/ceph-mon/tasks/crush_rules.yml
@@ -12,7 +12,7 @@
     - hostvars[item]['osd_crush_location'] is defined
 
 - name: create configured crush rules
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} osd crush rule create-simple {{ item.name }} {{ item.root }} {{ item.type }}"
+  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} osd crush rule create-replicated {{ item.name }} {{ item.root }} {{ item.type }} {{ item.class | default('') }}"
   with_items: "{{ crush_rules | unique }}"
   changed_when: false
   when: inventory_hostname == groups.get(mon_group_name) | last


group_vars/mons.yml.sample
```diff
@@ -43,14 +43,16 @@ dummy:
...
-#  root: HDD
+#  root: default
 #  type: host
+#  class: hdd
 #  default: false
 
 #crush_rule_ssd:
 #  name: SSD
-#  root: SSD
+#  root: deufalt
```
guits: typo:
```suggestion
#  root: default
```
in anycase you mustn't edit sample files directly, please read https://github.com/ceph/ceph-ansible/blob/master/CONTRIBUTING.md#sample-files

dsavineau:
I think the change could break backward compatibility since users couldn't have define the `class` key in the crush rule dict.
@guits thoughts ?

mslovy:
> I think the change could break backward compatibility since users couldn't have define the `class` key in the crush rule dict.
> @guits thoughts ?

so we can rely on the number of args.  4 args means previous without class.  5 args means current with class. In this way, we can solve compatibility issues?

mslovy:
@guits,  this will compatible with the previous no class configurations.

mslovy:
@alfredodeza , can you help to take a look, thx.

dsavineau:
Could you rebase your PR, add a commit message (not just title) and add a 'Resolves: #4400' (because you opened an issue related to that PR) ?

dsavineau:
@guits wondering if it's worth it to have two tasks instead of using a single one with condition statement

something like:

```yaml
   command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} osd crush rule {{ 'create-replicated' if item.class is defined else 'create-simple' }} {{ item.name }} {{ item.root }} {{ item.type }} {{ item.class | default('') }}"
```

mslovy:
> condition

good suggestion,

mslovy:
> ```yaml
> command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} osd crush rule {{ 'create-replicated' if item.class is defined else 'create-simple' }} {{ item.name }} {{ item.root }} {{ item.type }} {{ item.class | default('') }}"
> ```

create-simple cannnot pass more args, execute command bellow would be failed:
ceph osd crush rule create-simple SSD default rack ''

So I think we would directly use create-replicated to replace the create-simple. There would be the same if create-replicated does not pass item.class args. like below:

command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} osd crush rule create-replicated {{ item.name }} {{ item.root }} {{ item.type }} {{ item.class | default('') }}"

dsavineau:
> create-simple cannnot pass more args, execute command bellow would be failed:
> ceph osd crush rule create-simple SSD default rack ''

Are you sure ? using default('') will create an empty string but the actual quotes won't be present so we shouldn't have the extra argument.

mslovy:
> > create-simple cannnot pass more args, execute command bellow would be failed:
> > ceph osd crush rule create-simple SSD default rack ''
> 
> Are you sure ? using default('') will create an empty string but the actual quotes won't be present so we shouldn't have the extra argument.

yeah, I have tried in my test environment

dsavineau:
> yeah, I have tried in my test environment

It works for me

```console
TASK [ceph-mon : create configured crush rules] **********************************************************************************************************************************************************************************************
task path: crush/roles/ceph-mon/tasks/crush_rules.yml:14
Wednesday 04 September 2019  09:49:55 -0400 (0:00:00.030)       0:01:11.988 *** 
ok: [crush] => (item={'name': 'HDD', 'root': 'default', 'type': 'host', 'default': False}) => changed=false 
  ansible_loop_var: item
  cmd:
  - docker
  - exec
  - ceph-mon-crush
  - ceph
  - --cluster
  - ceph
  - osd
  - crush
  - rule
  - create-simple
  - HDD
  - default
  - host
  delta: '0:00:00.595676'
  end: '2019-09-04 13:49:56.084427'
  item:
    default: false
    name: HDD
    root: default
    type: host
  rc: 0
  start: '2019-09-04 13:49:55.488751'
  stderr: ''
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```

Can you share logs about your failure ?

mihaiplasoianu:
Can we reopen and merge this? It's a useful feature.

dsavineau:
@mihaiplasoianu this has already be handled by https://github.com/ceph/ceph-ansible/pull/4703

PR #4207, opened 2019-07-09
shrink-osd: (ceph-disk only) ease ceph-volume migration

guits:
- shrink-osd should remove the prepare container to avoid a leftover preventing from redeploying an OSD after it has been removed.
- remove the gpt header on devices so if osds are redeployed with ceph-volume it will no complain because of this.

NOTE: this hasn't been backported from master because ceph-disk doesn't exist anymore as of stable-4.0

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1613735


dsavineau:
mergify failed to automerge, doing it manually

PR #4225, opened 2019-07-10
ceph-infra: Open prometheus port

dsavineau:
The Prometheus porrt 9090 isn't open in the firewall configuration.
Also the dashboard task on the grafana node was not required because
it's already present on the mgr node.


guits:
jenkins test pipeline

PR #4227, opened 2019-07-11
dashboard: use variables for port value

dsavineau:
The current port value for alertmanager, grafana, node-exporter and
prometheus is hardcoded in the roles so it's not possible to change the
port binding of those services.


dsavineau:

PR #4229, opened 2019-07-11
ceph-dashboard: enable rgw options conditionally

dsavineau:
The dashboard rgw frontend options only need to be applied when there's
some nodes present in the rgw ansible group.


dsavineau:

PR #4230, opened 2019-07-11
improve dashboard integration

dsavineau:
* The current port value for alertmanager, grafana, node-exporter and
prometheus is hardcoded in the roles so it's not possible to change the
port binding of those services.
* The Vagrant dashboard scenario creates a dedicated grafana node but
was not use in the ansible inventory.
* The dashboard rgw frontend options only need to be applied when there's
some nodes present in the rgw ansible group.


dsavineau:

guits:
@fmount FI

guits:
@dsavineau @fmount it seems you are working both on the dashboard and might be impacting each other when pushing commits, I think you should ask for reviews each other so you are aware who's doing what exactly and avoid potential duplicated work.
Just saying... :relieved:

fmount:
@guits  thanks,
this change looks good to me and since the defaults don't change I'm ok with that.

dsavineau:

PR #4231, opened 2019-07-12
Avoid to setup provisioners in a fully containerized environment

fmount:
This commit adds a when clause to avoid the setup of grafana
provisioners in a fully containerized scenario.
This is needed when the ceph-grafana-dashboards package is not
installed and this task could result in a wrong grafana
configuration that let the container crash.


fmount:
According to the https://github.com/ceph/ceph/pull/26415 in a fully containerized scenario we need to populate the grafana provisioners in a different way (injecting the layouts using the api), so we need to skip the task.

roles/ceph-grafana/tasks/configure_grafana.yml
```diff
@@ -46,6 +47,7 @@
     src: dashboards-ceph-dashboard.yml.j2
     dest: /etc/grafana/provisioning/dashboards/ceph-dashboard.yml
     mode: 0640
+  when: not containerized_deployment
```
dsavineau: could you add the bool filter there ? ( `| bool` )

fmount: done

guits:
jenkins test pipeline

gfidente:
nice, I guess we want the backport-4.0 label on this one?

guits:
jenkins test pipeline

guits:
nfs-ganesha dev repositorie failures, merging anyway

Issue #4071, opened 2019-06-10
Error if python module six is not installed


adrianlzt:
**Bug Report**

What happened:
Failed container install

What you expected to happen:
No error

How to reproduce it (minimal and precise):
Run playbook site-container.yml against a server without python-six

Share your group_vars files, inventory

**Environment**:
* OS (e.g. from /etc/os-release):
* Kernel (e.g. `uname -a`):
* Docker version if applicable (e.g. `docker version`): 
* Ansible version (e.g. `ansible-playbook --version`):
* ceph-ansible version (e.g. `git head or tag or stable branch`): stable-4.0
* Ceph version (e.g. `ceph -v`): nautilus

```
TASK [ceph-config : run 'ceph-volume lvm batch --report' to see how many osds are to be created] ***
Monday 10 June 2019  16:51:43 +0200 (0:00:00.071)       0:01:45.639 ***********
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ImportError: No module named six
fatal: [servera]: FAILED! => changed=false
  module_stderr: |-
    Traceback (most recent call last):
      File "<stdin>", line 114, in <module>
      File "<stdin>", line 106, in _ansiballz_main
      File "<stdin>", line 49, in invoke_module
      File "/tmp/ansible_ceph_volume_payload_LUMbtu/__main__.py", line 6, in <module>
    ImportError: No module named six
  module_stdout: ''
  msg: |-
    MODULE FAILURE
    See stdout/stderr for the exact error
  rc: 1
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ImportError: No module named six
fatal: [servera]: FAILED! => changed=false
  module_stderr: |-
    Traceback (most recent call last):
      File "<stdin>", line 114, in <module>
      File "<stdin>", line 106, in _ansiballz_main
      File "<stdin>", line 49, in invoke_module
      File "/tmp/ansible_ceph_volume_payload_rJ6A7l/__main__.py", line 6, in <module>
    ImportError: No module named six
  module_stdout: ''
  msg: |-
    MODULE FAILURE
    See stdout/stderr for the exact error
  rc: 1
```

PR #4234, opened 2019-07-12
library/ceph_volume.py: remove six dependency

dsavineau:
The ceph nodes couldn't have the python six library installed which
could lead to error during the ceph_volume custom module execution.

  ImportError: No module named six

The six library isn't useful in this module if we're sure that all
action variables passed to the build_ceph_volume_cmd function are a
list and not a string.

Resolves: #4071


guits:
jenkins test pipeline

PR #4236, opened 2019-07-15
tests: remove useless setting

guits:
this setting is not needed here since we explicitely set it for
container and non container context.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -370,8 +370,6 @@ setenv=
   # only available for ansible >= 2.5
   ANSIBLE_STDOUT_CALLBACK = yaml
   non_container: DEV_SETUP = True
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -360,7 +360,6 @@ setenv=
   # only available for ansible >= 2.5
   ANSIBLE_STDOUT_CALLBACK = yaml
   non_container: DEV_SETUP = True
-  centos: CEPH_ANSIBLE_VAGRANT_BOX = centos/7
   fedora: CEPH_ANSIBLE_VAGRANT_BOX = fedora/29-atomic-host
```
dsavineau: wondering if we should remove fedora as well ?

PR #4245, opened 2019-07-18
Add all_group_name var in all.yml , fix group variable in prometheus …

mluchkin:
…role , fix group names in sample yml


group_vars/all.yml.sample
```diff
@@ -45,6 +45,7 @@ dummy:
 #cluster: ceph
 
 # Inventory host group variables
+#all_group_name: all
```
dsavineau: I don't think we need to define this because the `'all'` group name is an ansible builtin.

PR #4251, opened 2019-07-19
Fix handler logic if radosgw user https.

mluchkin:
If use https you catch error with restart daemon handler: 

curl https://10.3.3.104
curl: (51) Unable to communicate securely with peer: requested domain name does not match the server's certificate.

Signed-off-by: Mluchkin <6980570@gmail.com>


guits:
Due to inactivity, closing in favor of #4350

PR #4252, opened 2019-07-19
Fix backward compat with old cephfs_pools format

gfidente:
Previously cephfs_pools items used to have a pgs: key but not
pgp_num: nor pg_num:


dsavineau:
Could you amend your commit and add a signoff ?

dsavineau:
Unrelated CI failure due to missing nfs-ganesha packages upstream.

PR #4261, opened 2019-07-23
tests: bump nfs-ganesha version

guits:
in stable-4.0, nfs-ganesha 2.8 should be used.


dsavineau:
is it just for CI testing purpose ? If not, we should update nfs_ganesha_stable_branch in ceph-defaults role instead

PR #4276, opened 2019-07-27
openSUSE: support newer Leap releases

johanneskastl:
Current behaviour will fail on openSUSE Leap 15.0 and 15.1, that are way newer than openSUSE Leap 42.x, as unfortunately openSUSE switched from 42.x to 15.x to match SLES15 development.

Successfully tested on Leap 15.1 using this `group_vars/all.yml`:
```
---
ceph_origin: distro
ceph_stable_release: nautilus

#
public_network: 0.0.0.0/0
monitor_address_block: 10.11.12.0/24
cluster_network: 10.11.13.0/24

```

### DIFF
--- a/roles/ceph-validate/tasks/check_system.yml
+++ b/roles/ceph-validate/tasks/check_system.yml
@@ -46,12 +46,12 @@
     - ceph_repository == 'uca'
     - ansible_distribution != 'Ubuntu'
 
-- name: fail on unsupported openSUSE distribution
+- name: "fail on unsupported openSUSE distribution (only 15.x supported)"
   fail:
     msg: "Distribution not supported: {{ ansible_distribution }}"
   when:
     - ansible_distribution == 'openSUSE Leap'
-    - ansible_distribution_version is version_compare('42.3', '<')
+    - ansible_distribution_major_version != '15'
 
 - name: fail on unsupported ansible version (1.X)
   fail:


roles/ceph-validate/tasks/check_system.yml
```diff
@@ -46,12 +46,12 @@
     - ceph_repository == 'uca'
     - ansible_distribution != 'Ubuntu'
 
-- name: fail on unsupported openSUSE distribution
+- name: "fail on unsupported openSUSE distribution (42.x)"
```
dsavineau: Maybe a better way to do it is to compare to the major distribution version fact

```console
$ ansible -i hosts all -m setup -a 'filter=ansible_distribution_major_version'
xxxxxxxxx | SUCCESS => {
    "ansible_facts": {
        "ansible_distribution_major_version": "15",
        "discovered_interpreter_python": "/usr/bin/python3.6"
    },
    "changed": false
}
```

So something like `ansible_distribution_major_version != '15'` should be enough to support Leap 15.x

johanneskastl: Do you want to me explicitly whitelist Leap 15.x only, so Leap 16.x (once released) will be unsupported? Or do you want me to only blacklist 42.x?

dsavineau: For now we can focus on Leap 15 because we don't really now what will be the next openSuSe release since they changed a lot the release number during last years.

johanneskastl:
@dsavineau I have adapted the logic and rephrased the failure message accordingly.

@guits I have tried to follow the guidelines, thanks for pointing this out. Sorry for the trouble. I hope it is fine now.

johanneskastl:
@guits Are you fine with this in its current state? Anything I have missed in the guidelines?

Issue #4288, opened 2019-07-30
TASK 'wait for grafana to start' fails


SalsaBr:
ceph-grafana is not starting and ceph-ansible fails.

Tried to run ceph-ansible in a Ubuntu and a CentOS machine, and both had the same results.

```console
TASK [ceph-grafana : wait for grafana to start] *******************************************************************************************************************************************************************
task path: /home/salsa/repo/ceph-ansible/roles/ceph-grafana/tasks/configure_grafana.yml:94
Tuesday 30 July 2019  13:55:30 -0300 (0:00:01.008)       0:06:58.355 **********
Using module file /usr/lib/python2.7/dist-packages/ansible/modules/utilities/logic/wait_for.py
Pipelining is enabled.
<ceph-mon-01> ESTABLISH SSH CONNECTION FOR USER: None
<ceph-mon-01> SSH: EXEC ssh -vvv -o ControlMaster=auto -o ControlPersist=600s -o PreferredAuthentications=publickey -o StrictHostKeyChecking=no -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=60 -o ControlPath=/root/.ansible/cp/%h-%r-%p ceph-mon-01 '/bin/sh -c '"'"'sudo -H -S -n  -u root /bin/sh -c '"'"'"'"'"'"'"'"'echo BECOME-SUCCESS-omyipiutbdwqxilzorxszkcxjuorwerf ; /usr/bin/python'"'"'"'"'"'"'"'"' && sleep 0'"'"''
Escalation succeeded
[...]

PR #4289, opened 2019-07-30
ceph-grafana: Add grafana addr to wait_for task

dsavineau:
The ansible wait_for task is checking the grafana port (3000) against
the 127.0.0.1 host, which is the default value of the module.
Instead we should use the grafana address because the process is
binding on a specific address.

Resolves: #4288


dsavineau:
This PR is indeed not fixing #4288 because the issue is completly different.
Also grafana is binding on all IP addresses so it doesn't change anything

PR #4290, opened 2019-07-30
ceph-grafana: Set grafana uid/gid on files

dsavineau:
We don't need to create a grafana system user (in fact we even don't
set the righ uid to this user) because we're using a container setup.
Instead we just need to be sure to set the owner/group to 472 (grafana
user/group from the container) like we do for ceph/167.
We don't need to set the user/group recursively on /etc/grafana
directory in a dedicated task.
Also on Ubuntu system, the ceph-grafana-dashboards isn't present so on
non containerized deployment we won't have the
/etc/grafana/dashboards/ceph-dashboard directory present (coming with
the package) so we need to be sure it exists.

### DIFF
--- a/roles/ceph-grafana/tasks/configure_grafana.yml
+++ b/roles/ceph-grafana/tasks/configure_grafana.yml
@@ -6,7 +6,7 @@
   register: result
   until: result is succeeded
   when:
-    - not containerized_deployment
+    - not containerized_deployment | bool
     - ansible_os_family in ['RedHat', 'Suse']
   tags: package-install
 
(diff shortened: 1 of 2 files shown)


roles/ceph-grafana/tasks/configure_grafana.yml
```diff
@@ -25,7 +25,10 @@
     path: "{{ item }}"
     state: directory
     recurse: yes
+    owner: '{{ grafana_uid }}'
```
guits: just cosmetic change,

why don't we have double quotes everywhere?
```suggestion
    owner: "{{ grafana_uid }}"
```

just to keep consistency

dsavineau: :+1:

PR #4293, opened 2019-07-31
ceph-infra: Apply firewall rules with container

dsavineau:
We don't have a reason to not apply firewall rules on the host when
using a containerized deployment.
The TripleO environments already manage the ceph firewall rules outside
ceph-ansible and set the configure_firewall variable to false.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1733251


guits:
jenkins test pipeline

PR #4303, opened 2019-08-02
ceph-dashboard: Add run_once on delegate tasks

dsavineau:
Because we need to execute commands from a monitor node (the first one 
in the mons list) we are using delegate_to option.
If there's multiple nodes running the ceph-dashboard role then the 
delegated task will be executed multiple times.
Also remove a mgr config-key option not present for nautilus+ releases.

### DIFF
--- a/infrastructure-playbooks/dashboard.yml
+++ b/infrastructure-playbooks/dashboard.yml
@@ -67,7 +67,7 @@
             status: "Complete"
             end: "{{ lookup('pipe', 'date +%Y%m%d%H%M%SZ') }}"
 
-- hosts: "{{ grafana_server_group_name }}"
+- hosts: "{{ groups[mgr_group_name] | default(groups[mon_group_name]) }}"
   become: true
   pre_tasks:
     - name: set ceph dashboard install 'In Progress'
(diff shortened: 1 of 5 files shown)


dsavineau:
jenkins test pipeline

guits:
@fmount https://github.com/ceph/ceph-ansible/pull/4303/files#diff-d07a382d4015296d40bcae72077e925eR70

infrastructure-playbooks/dashboard.yml
```diff
@@ -67,7 +67,7 @@
             status: "Complete"
             end: "{{ lookup('pipe', 'date +%Y%m%d%H%M%SZ') }}"
 
-- hosts: "{{ grafana_server_group_name }}"
+- hosts: '{{ groups["mgrs"] | default(groups["mons"]) }}'
```
guits:
```suggestion
- hosts: "{{ groups[mgr_group_name] | default(groups[mon_group_name]) }}"
```
but I'm not sure this is what we really wanna do here. I think we should stick to have the whole stack deployed in 1 node which would be a member of `grafana_server_group_name` (that could be either an mgr, a mon or anything).

Why would we absolutely want to enforce this to be deployed either on mgr or mon only?

@dsavineau let's discuss this as soon as you get online today, please.

fmount: @dsavineau I think we can keep `grafana_server_group_name` since we use delegate_to to execute the commands from the (mon|mgr)[0]; it can help us to logically keep the dashboard roles separated. 
+1 for the `run_once: true` option.

guits:
mergify failure, merging anyway

PR #4305, opened 2019-08-02
Execute common roles once on all nodes

dsavineau:
The common roles don't need to be executed again on each group plays
(like mons, osds, etc..).
We only need to execute them during the first play. That wat, we will
apply the changes on all nodes in parallel instead of doing it once per
group.


guits:
jenkins test pipeline

guits:
jenkins test pipeline

dsavineau:
jenkins test pipeline

dsavineau:

PR #4315, opened 2019-08-06
mgr: refact 'wait for all mgr to be up' task

guits:
There's no need to use `shell` module here.
Instead of using `| python -c`, let's use `from_json` filter.


guits:

PR #4316, opened 2019-08-06
ceph-handler: Use /proc/net/unix for rgw socket

dsavineau:
If for some reason, there's an old rgw socket file present in the
/var/run/ceph/ directory then the test command could fail with
```console
test: xxxxxxxxx.asok: binary operator expected
```
```console    
$ ls -hl /var/run/ceph/
total 0
srwxr-xr-x. ceph-client.rgw.rgw0.rgw0.68.94153614631472.asok
srwxr-xr-x. ceph-client.rgw.rgw0.rgw0.68.94240997655088.asok
```
We can check the radosgw socket in /proc/net/unix to avoid using wildcard
in the socket name.

### DIFF
--- a/roles/ceph-handler/templates/restart_rgw_daemon.sh.j2
+++ b/roles/ceph-handler/templates/restart_rgw_daemon.sh.j2
@@ -19,7 +19,6 @@ for ((i=0; i<${RGW_NUMS}; i++)); do
   DOCKER_EXECS[i]="{{ container_binary }} exec ${CONTAINER_NAME}"
 {% endif %}
 done
-declare -a SOCKETS
 RGW_IP={{ hostvars[inventory_hostname]['_radosgw_address'] }}
 SOCKET_PREFIX="/var/run/ceph/{{ cluster }}-client.rgw.${HOST_NAME}.rgw"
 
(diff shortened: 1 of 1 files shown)


roles/ceph-handler/templates/restart_rgw_daemon.sh.j2
```diff
@@ -29,7 +29,8 @@ check_socket() {
   local count=10
   # Wait and ensure the socket exists after restarting the daemon
   while [ $count -ne 0 ]; do
-    ${DOCKER_EXECS[i]} test -S ${SOCKET_PREFIX}${i}.*.*.asok && SOCKETS[i]=$(stat --printf=%n ${SOCKET_PREFIX}${i}.*.*.asok)
+    RGW_PID=$(pidof -s -n radosgw)
```
guits: what if there are multiple radosgw instances?

guits:
@dsavineau  should we remove the DNM here?

dsavineau:
finally reproduce the issue \o/

So we could have multiple rgw socket for the same instance. And for some reason, after a restart of the service the PID is still the same.

```console
# ls -hl /var/run/ceph/
total 0
srwxr-xr-x. 1 167 167 0 Oct 31 19:45 ceph-client.rgw.rgw0.rgw0.68.94240997655088.asok
srwxr-xr-x. 1 167 167 0 Oct 31 19:51 ceph-client.rgw.rgw0.rgw0.68.94378871513648.asok
```

dsavineau:
> And for some reason, after a restart of the service the PID is still the same.

@guits I'm wondering if we should use `--pid=host` in this situation because the PID present in the socket name is the one from the container wich is different from the host.

guits:
@dsavineau is there a BZ for that?

PR #4317, opened 2019-08-06
ceph-osd: Add ulimit nofile on container start

dsavineau:
On containerized deployment, the OSD entrypoint runs some ceph-volume
commands (lvm/simple scan and/or activate) which perform badly without
the ulimit option.
This option was added for all previous ceph-volume commands but not on
the ceph-osd container startup.
Also updating hard limit value to 4096 to reflect default baremetal
value.


guits:

guits:
unrelated failure

PR #4320, opened 2019-08-08
Fix debian custom repo

antrusd:
Currently custom repo doesn't have apt or rpm key defined,
so apt update gonna be failed.

This PR should fix that.

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -235,6 +235,7 @@ dummy:
 # a URL to the .repo file to be installed on the targets.  For deb,
 # ceph_custom_repo should be the URL to the repo base.
 #
+#ceph_custom_key: https://server.domain.com/ceph-custom-repo-key.asc
 #ceph_custom_repo: https://server.domain.com/ceph-custom-repo
 
 
--- a/roles/ceph-common/tasks/installs/debian_custom_repository.yml
+++ b/roles/ceph-common/tasks/installs/debian_custom_repository.yml
@@ -1,7 +1,14 @@
 ---
+- name: configure debian custom apt key
+  apt_key:
+    url: "{{ ceph_custom_key }}"
+    state: present
+  register: result
+  until: result is succeeded
+  when: ceph_custom_key is defined
+
 - name: configure debian custom repository
   apt_repository:
     repo: "deb {{ ceph_custom_repo }} {{ ansible_distribution_release }} main"
     state: present
-    update_cache: no
-  notify: update apt cache if a repo was added
+    update_cache: yes
--- a/roles/ceph-common/tasks/installs/redhat_custom_repository.yml
+++ b/roles/ceph-common/tasks/installs/redhat_custom_repository.yml
@@ -1,4 +1,12 @@
 ---
+- name: configure red hat custom rpm key
+  rpm_key:
+    key: "{{ ceph_custom_key }}"
+    state: present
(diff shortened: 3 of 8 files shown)


antrusd:
jenkins test pipeline

antrusd:
jenkins test pipeline

roles/ceph-common/tasks/installs/redhat_custom_repository.yml
```diff
@@ -1,4 +1,11 @@
 ---
+- name: configure red hat custom rpm key
+  rpm_key:
+    key: "{{ ceph_custom_key }}"
+    state: present
+  register: result
+  until: result is succeeded
```
guits: ditto

roles/ceph-common/tasks/installs/debian_custom_repository.yml
```diff
@@ -1,7 +1,13 @@
 ---
+- name: configure debian custom apt key
+  apt_key:
+    url: "{{ ceph_custom_key }}"
```
guits: so this becomes a mandatory option when `ceph_custom_repo` is used?

otherwise, we probably need this:

```suggestion
    url: "{{ ceph_custom_key }}"
    when: ceph_custom_key is defined
```

roles/ceph-common/tasks/installs/redhat_custom_repository.yml
guits: ditto

group_vars/all.yml.sample
```diff
@@ -235,6 +235,7 @@ dummy:
 # a URL to the .repo file to be installed on the targets.  For deb,
 # ceph_custom_repo should be the URL to the repo base.
 #
+#ceph_custom_key: https://server.domain.com/ceph-custom-repo-key.asc
```
guits: you shouldn't edit this file directly, you must edit `roles/ceph-defaults/defaults/main.yml` and run the `generate_group_vars_sample.sh` script

roles/ceph-common/tasks/installs/debian_custom_repository.yml
antrusd: OK, I'll go with ` when: ceph_custom_key is defined`

roles/ceph-common/tasks/installs/redhat_custom_repository.yml
antrusd: OK

group_vars/all.yml.sample
antrusd: OK

antrusd:
jenkins test pipeline

antrusd:
jenkins test pipeline

antrusd:
jenkins test pipeline

antrusd:
jenkins test pipeline

antrusd:
Ahhh, finally, jenkins test succeeded :dancing_men:  

Hi @guits, could you please take a look ?

Thanks

clwluvw:
@dsavineau @guits Can it be backport to `stable-4.0` please?

clwluvw:
This really needs in using custom repo because it doesn't update cache after adding custom repo and it will fails and key should be import before. @guits @dsavineau

dsavineau:
@clwluvw the backport is already opened https://github.com/ceph/ceph-ansible/pull/5689

PR #4322, opened 2019-08-08
ceph-config: Set changed_when to false on fact gathering statements

kevincoakley:
The "run 'ceph-volume lvm batch --report' to see how many osds are to be
created" and "run 'ceph-volume lvm list' to see how many osds have already been
created" statements only register the lvm_batch_report and lvm_list variables.
Running those ceph-volume commands should never produce a change on the system.
Adding changed_when: false prevents irrelevant change messages from Ansible.


guits:
unrelated failure

PR #4324, opened 2019-08-11
Only install python-xml on SUSE/openSUSE if python2 is being used

johanneskastl:
On SUSE/openSUSE 15 or newer, the default python is python3, which does not need `python-xml` to be installed. 

Especially as this is a python2 package, that pulls in python2 and confuses ansible (unless you set the ansible_python_interpreter parameter to point to python3).

**Todo**:
- [X] ~~Check if installing python via [raw_install_python.yml](https://github.com/ceph/ceph-ansible/blob/master/raw_install_python.yml) can be safeguarded in the same way.~~ This has been fixed in master in the meantime.

### DIFF
--- a/roles/ceph-common/tasks/installs/install_on_suse.yml
+++ b/roles/ceph-common/tasks/installs/install_on_suse.yml
@@ -19,6 +19,7 @@
   with_items: "{{ suse_package_dependencies }}"
   register: result
   until: result is succeeded
+  when: ansible_python_version is version( '3', '<')
 
 - name: include install_suse_packages.yml
   include_tasks: install_suse_packages.yml
--- a/roles/ceph-common/tasks/installs/install_suse_packages.yml
+++ b/roles/ceph-common/tasks/installs/install_suse_packages.yml
@@ -5,6 +5,7 @@
     state: present
   register: result
   until: result is succeeded
+  when: ansible_python_version is version( '3', '<')
 
 - name: install suse ceph packages
   package:


johanneskastl:
Fix for `raw_install_python.yml` is prepared in #4326

roles/ceph-common/tasks/installs/install_on_suse.yml
```diff
@@ -19,6 +19,7 @@
   with_items: "{{ suse_package_dependencies }}"
   register: result
   until: result is succeeded
+  when: ansible_python_version is version( '3', '<')
```
guits: Looks like this task is a duplicate

roles/ceph-common/tasks/installs/install_suse_packages.yml
```diff
@@ -5,10 +5,11 @@
     state: present
   register: result
   until: result is succeeded
+  when: ansible_python_version is version( '3', '<')
```
guits: The variable `suse_package_dependencies` is not intended to install `python-xml` only, someone might want install packages which could be needed regardless the installed python version.

johanneskastl: Then I would suggest to leave this variable empty (for people to use if they need it) and add a task installing python-xml explicitely, but only if python2 was found as python interpreter.

Would that be a valid approach in your opinion?

roles/ceph-common/tasks/installs/install_on_suse.yml
johanneskastl: Should I remove this task completely (in another PR to not mix things up)?

johanneskastl: See #4349

roles/ceph-common/tasks/installs/install_suse_packages.yml
guits: @johanneskastl I don't even know why this package is needed when deploying on Suse, any idea?

johanneskastl: For python2, this one is needed to have ansible's zypper module work correctly, as far as I remember. Python3 includes this by default, so no need to install anything.

I think the cleanest solution would be to move the installation of python-xml to the raw_install_python.yml, as this would be the correct place to install it, before any of ansible's modules are being used.

This would mean my changes in this PR would be void.

johanneskastl: See #4351 if this is fine with you, then we could close this PR.

guits: in `raw_install_python.yml`, it will install `python-xml` only if python wasn't detected, what if you have python but you don't have `python-xml` installed?

guits: given that this `python-xml` is a requirement when python2 only, I think, your first approach :

> Then I would suggest to leave this variable empty (for people to use if they need it) and add a task installing python-xml explicitely, but only if python2 was found as python interpreter.
> 
> Would that be a valid approach in your opinion?

is the best option.

johanneskastl: As raw_install_python.yml has all the logic for finding the python installation(s), I would suggest to add a separate python-xml task there.

guits: You will need to install this package after facts are gathered which happens after `raw_install_python.yml` is called.

johanneskastl: I tried to tackle this in [commit 7e354de](https://github.com/ceph/ceph-ansible/pull/4351/commits/7e354deeb0be70ee4af6f40d0e843b4e954d5965) in PR #4351

guits:
closing this in favor of #4351

PR #4296, opened 2019-08-01
various fixes and improvements

guits:
This PR gathers various fixes and improvements:
- refacts the raw installation of python,
- replaces some `shell` module with `command`
- get rid of complex tests using a combination of bash commands redirected in python in order to parse json data where we can use the `from_json` filter,
- update some functional tests given the json data structure returned by `ceph -s` has changed in octopus.

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #4326, opened 2019-08-11
raw_install_python.yml: also check for /usr/bin/python3

johanneskastl:
openSUSE Leap 15.x do no longer have python2 installed, and the
installed python3 does not provide `/usr/bin/python`.

The new logic will only install python2 if python3 has not been found, either.

Signed-off-by: Johannes Kastl <kastl@b1-systems.de>

### DIFF
--- a/raw_install_python.yml
+++ b/raw_install_python.yml
@@ -5,6 +5,12 @@
   ignore_errors: yes
   register: systempython
 
+- name: check for python3
+  stat:
+    path: /usr/bin/python3
+  ignore_errors: yes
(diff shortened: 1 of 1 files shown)


guits:
Hi @johanneskastl 
see https://github.com/ceph/ceph-ansible/pull/4296/commits/de20df03b4f807fc2f1cd9d65df6658da2161db7

johanneskastl:
Hi @guits 
Nice one. I actually like yours a lot better, thanks!

PR #4329, opened 2019-08-12
Adjust openstack_keys

noonedeadpunk:
Since glance do not need write access to volumes pool and
set permission is static (do not rely on openstack_cinder_pool)
we can easily drop it.

Also as glance do not need write access to volumes pool and
set permission is static (do not rely on openstack_cinder_pool)
we can easily drop it.


noonedeadpunk:
retest this please

noonedeadpunk:
retest this please

noonedeadpunk:
retest this please

guits:
@gfidente @fultonj @fmount

group_vars/all.yml.sample
```diff
@@ -690,8 +690,8 @@ dummy:
 # By default, keys will be auto-generated.
 #
 #openstack_keys:
-#  - { name: client.glance, caps: { mon: "profile rbd", osd: "profile rbd pool=volumes, profile rbd pool={{ openstack_glance_pool.name }}, osd blacklist"}, mode: "0600" }
```
gfidente: the glance key does not need r/w on the volume pools, this is correct

group_vars/all.yml.sample
```diff
@@ -690,8 +690,8 @@ dummy:
...
-#  - { name: client.cinder, caps: { mon: "profile rbd", osd: "profile rbd pool={{ openstack_cinder_pool.name }}, profile rbd pool={{ openstack_nova_pool.name }}, profile rbd pool={{ openstack_glance_pool.name }}, osd blacklist"}, mode: "0600" }
+#  - { name: client.glance, caps: { mon: "profile rbd", osd: "allow class-read object_prefix rbd_children, profile rbd pool={{ openstack_glance_pool.name }}, osd blacklist"}, mode: "0600" }
+#  - { name: client.cinder, caps: { mon: "profile rbd", osd: "allow class-read object_prefix rbd_children, profile rbd pool={{ openstack_cinder_pool.name }}, profile rbd pool={{ openstack_nova_pool.name }}, profile rbd pool={{ openstack_glance_pool.name }}, osd blacklist"}, mode: "0600" }
```
gfidente: 'class-read object_prefix rbd_children' is implicitly allowed by 'profile rbd' so it is not necessary

group_vars/all.yml.sample
gfidente: I might have added this comment in the wrong place, I meant that it is correct to remove permissions on the volumes from the glance key

group_vars/all.yml.sample
noonedeadpunk: Are you sure about that? Since I didn't find anything related to class-read for `profile rbd` in [docs](https://docs.ceph.com/docs/nautilus/rados/operations/user-management/):

> When used as an OSD cap, it provides read-write access to an RBD client application.

So according to this I'd say that it's regular `rwx` privileges. And without class-read I was having issues with creating volumes from images.

gfidente: pretty sure it includes class-read yes [1] [2]

1. https://github.com/ceph/ceph/blob/nautilus/src/osd/OSDCap.cc#L335
2. https://github.com/ceph/ceph/blob/nautilus/src/osd/OSDCap.h#L42

noonedeadpunk:
retest this please

group_vars/all.yml.sample
noonedeadpunk: Yep, seems that you're right... Ok, I've dropped specification of object class out of commit.

logan2211: @gfidente actually setting `allow class-read object_prefix rbd_children` is very necessary on all openstack keys.

Let me use `client.glance` as an example. When you create a glance image and then build CoW disks from it in nova, the image is protected in ceph. Once you decide to delete an image in glance, glance will first try to unprotect the image, which results in ceph scanning all pools for CoW children of the glance image. This process fails without `allow class-read object_prefix rbd_children` since glance does not have the ability to introspect all pools on the cluster to find child disks, and the unprotect fails, so the image cannot be deleted.

This is also true in nova (for nova live snapshots), and in cinder for volume snapshots.

All openstack keys should have `allow class-read object_prefix rbd_children` set.

gfidente: sure it is, what I am saying is that it's a subset of the permissions granted by the 'profile rbd' setting

logan2211: The 'profile rbd' is scoped to a specific pool

logan2211: The problem is the pools that the key does not have `profile rbd` access to

gfidente: I think I understand but I am not sure this is what happens; for example nova does not have permissions to read images in the glance pool either.

Instead, I think nova will download the glance image via API and populate the guests' disk with its cached copy. It does not need permissions to read into the glance backend and the image snapshots are created by glance and stay in the glance pool.

To be able to use cinder volumes instead, libvirt needs to be configured with the cinder secret so that nova can attach guests to images found in the cinder pool. Similarily to what happens for glance images, volume snapshots stay in the cinder pool.

logan2211: Framing this another way, forget about OpenStack for a minute. You've got OpenStack running and all is well.

Now you decide to create another pool `foobar` on your ceph cluster for non-OpenStack uses. Maybe iSCSI shares or something. OpenStack services don't know about or use this pool, but suddenly glance can no longer delete disks from its own `images` pool because when it attempts to unprotect a disk, it tries to scan `foobar` (along with all other pools on the cluster) for children, but fails due to the missing `class-read object_prefix rbd_children` permission.

logan2211: > I think I understand but I am not sure this is what happens; for example nova does not have permissions to read images in the glance pool either.
> 

Actually it does. That is how you can build rbd children from the images pool. See https://github.com/ceph/ceph-ansible/blob/70cf2a5846e38e98b32c47a176cb38488af3dc12/group_vars/all.yml.sample#L694 specifically `profile rbd pool={{ openstack_glance_pool.name }}`

The issue here is not image creates. The issue is unprotecting parent images on clusters where non-OpenStack pools exist (RGW, etc.).

When a service (glance, cinder, nova, whatever) tries to unprotect an image, *all* pools on the cluster are scanned for children of the image you are unprotecting. If some pools cannot be scanned (ie. because the `class-read object_prefix rbd_children` permission is scoped to specific pools only, rather than cluster-wide), then the unprotect operation fails and the rbd cannot be deleted.

guits:
@logan2211 from what I read here, I understand @noonedeadpunk should set back the `class-read object_prefix rbd_children` permission on this patch?

noonedeadpunk:
@logan2211 thanks for stepping in and describing the problem in details (while I was really missing them)
@guits I've returned `object_prefix rbd_children`

guits:
@noonedeadpunk sounds good to me
@logan2211 thanks, I think it's pretty clear for everyone unless @gfidente has something different to share?

group_vars/all.yml.sample
gfidente: adding class-read on rbd-children for the glance keyring is fine and certainly the least disruptive change but I wanted to share my rationale on why it might make more sense to use a single keyring for cinder/glance/nova then

1) to be able to create clones into another pool, the other consumer (for example nova) must first have read permissions on the glance pool

2) nova not only reads from the glance pool to create COW clones but it also tries to write in the glance pool when shelving guests

3) nova (libvirt) also needs to know the cinder secret to be able to start guests from cinder volumes

4) cinder behaves like nova

ceph-ansible is currently not creating a keyring for nova, probably assuming nova will be configurd to use the cinder keyring for its ephemeral storage as well? if so it is already sharing its keyring with glance, what point having a separate keyring for glance only?

noonedeadpunk:
retest this please

group_vars/all.yml.sample
logan2211: @gfidente if you want to use the same key for all 3 services, it looks like `client.openstack` does what you're proposing.

@noonedeadpunk we probably want to add `class-read object_prefix rbd_children` to that `client.openstack` key also I'm guessing.

gfidente: @logan2211 I am used to deploy a single keyring indeed; maybe I see little benefit in having a shared keyring for nova/cinder with all the permissions to write in the glance pool too ... and then a different one for glance only with permissions to scan all pools for children objects. It seems to have the only benefit of preventing glance (and glance only) from using the cinder and nova pools.

Adding class-read for children to the client.openstack keyring surely makes sense in a scenario where you assume other tools have read permissions on the openstack pools and could, hence, create clones in their own pools. But this seems to me more a specific deployment decision, not something I would assume in a default setting.

gfidente: Noticed we ended up adding class-read on children for the cinder key as well, even though it does have enough permissions on the glance and nova pools because of the profile.

This goes a bit into the scenario where there could be other consumers with read permissions on the cinder pool creating clones elsewhere.

gfidente:
Considering the previous implementation matched what was suggested in ceph docs at [1], maybe this should trigger a change in the docs as well?

1. https://docs.ceph.com/docs/master/rbd/rbd-openstack/

group_vars/all.yml.sample
noonedeadpunk: Dropped class-read for cinder.

noonedeadpunk:
> Considering the previous implementation matched what was suggested in ceph docs at [1], maybe this should trigger a change in the docs as well?
> 
> 1. https://docs.ceph.com/docs/master/rbd/rbd-openstack/

Placed https://github.com/ceph/ceph/pull/29865 to update docs

gfidente:
@guits I thought we wanted this one in?

noonedeadpunk:
Rebased on top of the master

dsavineau:
@noonedeadpunk Could you amend your commit message ? because it's mentioned cinder & glance but the code is only modifying the cinder caps.

noonedeadpunk:
@dsavineau Good catch, as I've lost volumes permission drop during rebase.

noonedeadpunk:
retest this please

noonedeadpunk:
@guits @gfidente So lot's of time passed since PR submission. Considering that caph-ansible will be sunseted in favor of the new deployment system - does this patch still has any sense?

dsavineau:

dsavineau:

dsavineau:

dsavineau:
> retest this please

https://docs.ceph.com/ceph-ansible/master/dev/index.html#pull-request-testing

> @guits @gfidente So lot's of time passed since PR submission. Considering that caph-ansible will be sunseted in favor of the new deployment system - does this patch still has any sense?

What new deployment system are you talking about ?

noonedeadpunk:
@dsavineau I was talking about ceph orchestrator, which is going to replace ceph-deploy, ceph-ansible, etc in octopus
https://docs.google.com/presentation/d/1JpcETNXpuB1JEuhX_c8xtgNnv0gJhSaQUffM7aRjUek

dsavineau:
AFAIK ceph-ansible will still be there for octopus

guits:
> AFAIK ceph-ansible will still be there for octopus

correct

Issue #4330, opened 2019-08-12
rolling update should set "nodeep-scrub" flag instead of "norebalance"


racpatel:
When we use rolling_update.yml to update/upgrade cluster it sets 2 flags "noout" and "norebalance".
IMHO, during rolling_update we should set "nodeep-scrub" flag rather than "no rebalance"
(more on flags - https://docs.ceph.com/docs/mimic/rados/operations/health-checks/#osdmap-flags)

Issue with "norebalance"
After OSD upgrade, it waits for "active+clean" state (no of retry defined by user).
when data and no of retry count is large, it  will be stuck there for a longer period.
e.g.
In one of our cluster no of retry was= 10000 and upgrade was stuck for 2 days due to status of cluster.

FAILED - RETRYING: waiting for clean pgs... (93858 retries left) 
FAILED - RETRYING: waiting for clean pgs... (93857 retries left) 
FAILED - RETRYING: waiting for clean pgs... (93856 retries left) 
and so on

pg status always as "norebalance" flag was set

 pgs:     40437/247052355 objects misplaced (0.016%)
             4284 active+clean
             2    active+undersized+remapped+backfilling
             1    active+remapped+backfilling
             1    active+remapped+backfill_wait

PR #4122, opened 2019-06-18
update: reset flags before and after each osd node upgrade

guits:
It might be possible at some point even with osd flags `noout` and
`norebalance` set the PGs states can change depending on the amount of data
written meantime. It means the check for PGs state will fail.

This commit changes the way we set those flags:
we set them before an OSD node upgrade and unset them before the PGs
state check so they can recover.

Fixes: #3961

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #4331, opened 2019-08-12
Avoid deep-scrub and allow rebalance during upgrade

racpatel:
1) Don't set "rebalance" flag during upgrade
2) Set "nodeep-scrub" flag during upgrade

Fixes: #4330
Signed-off-by: Rachana Patel <racpatel@redhat.com>


guits:
@racpatel Looks like how flags act on the cluster is a bit blurry depending on what is the cluster state, what is the traffic on the cluster...

Could you explain why we should have to remove `norebalance` here?

racpatel:
> @racpatel Looks like how flags act on the cluster is a bit blurry depending on what is the cluster state, what is the traffic on the cluster...
> 
> Could you explain why we should have to remove `norebalance` here?
> 
> BTW, I remember this patch: [08d9432](https://github.com/ceph/ceph-ansible/commit/08d94324545b3c4e0f6a1caf6224f37d1c2b36db)


When we use rolling_update.yml to update/upgrade cluster it sets 2 flags "noout" and "norebalance".
IMHO, during rolling_update we should set "nodeep-scrub" flag rather than "no rebalance"
(more on flags - https://docs.ceph.com/docs/mimic/rados/operations/health-checks/#osdmap-flags)

Issue with "norebalance"
After OSD upgrade, it waits for "active+clean" state (no of retry defined by user).
Let's take case where cluster was not in "active+clean" state when you start upgrade of cluster(or upgrade of OSD)
When data and no of retry count is large, it will be stuck there for a longer period.
e.g.
In one of our cluster 

pg status always as "norebalance" flag was set (Data will not be rebalanced)

pgs: 40437/247052355 objects misplaced (0.016%)
4284 active+clean
2 active+undersized+remapped+backfilling
1 active+remapped+backfilling
1 active+remapped+backfill_wait

It keep trying and was failed.

If we do not set "norebance" and set "nodeep-scrub", data rebalance would be faster

Let me know if you need more info

guits:
> Issue with "norebalance"
> After OSD upgrade, it waits for "active+clean" state (no of retry defined by user).
> Let's take case where cluster was not in "active+clean" state when you start upgrade of cluster(or upgrade of OSD)

I think this has been addressed by #4122

guits:

Issue #4333, opened 2019-08-13
set_fact _monitor_address - 'ipaddr: unknown filter type: 172.120.3.0/24,172.117.3.0/24,172.118.3.0/24,172.119.3.0/24'


hjensas:
**Bug Report**

What happened:
The public_network is a L3 routed network with multiple subnet CIDR's. When configuring the `monitor_address_block` using the same list of CIDR's as used for the `public_network` the deployment fails with the following error:

```
TASK [ceph-facts : set_fact _monitor_address to monitor_address_block ipv4]
task path: /usr/share/ceph-ansible/roles/ceph-facts/tasks/set_monitor_address.yml:2
Friday 09 August 2019  23:43:45 +0000 (0:00:00.921)       0:01:17.061 ********* 
fatal: [hostname]: FAILED! => 
msg: 'ipaddr: unknown filter type: 172.120.3.0/24,172.117.3.0/24,172.118.3.0/24,172.119.3.0/24'
```

What you expected to happen:
`set_fact _monitor_address to monitor_address_block ipv4` should support a comma separated list of address blocks in the same way public_network and cluster_network does. So that monitor address for monitors placed in different subnets can be installed without setting the monitor_address or monitor_interface variables.

How to reproduce it (minimal and precise):
Set monitor_address_block to include a comma separated list of address blocks. For example:
`monitor_address_block: '172.120.3.0/24,172.117.3.0/24,172.118.3.0/24,172.119.3.0/24'`

Share your group_vars files, inventory

**Environment**:
* OS (e.g. from /etc/os-release): `Red Hat Enterprise Linux 8.0 (Ootpa)`
* Kernel (e.g. `uname -a`): `Linux overcloud-controller-0 4.18.0-80.7.2.el8_0.x86_64 #1 SMP Fri Jul 26 10:48:21 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`
* Ansible version (e.g. `ansible-playbook --version`): `ansible-playbook 2.8.3`
* ceph-ansible version (e.g. `git head or tag or stable branch`): `ceph-ansible-4.0.0-0.1.rc13.el8cp.noarch`
* Ceph version (e.g. `ceph -v`): Ceph 4.0

PR #4339, opened 2019-08-14
Replace ipaddr() with ips_in_ranges()

hjensas:
Replace ipaddr() with ips_in_ranges()
    
This change implements a filter_plugin that is used in the
ceph-facts, ceph-validate roles and infrastucture-playbooks.
The new filter plugin will return a list of all IP address
that reside in any one of the given IP ranges. The new filter
replaces the use of the ipaddr filter.
    
ceph.conf already support a comma separated list of CIDRs
for the public_network and cluster_network options.
    
Changes: [1] and [2] introduced a regression in ceph-ansible
where public_network can no longer be a comma separated list
of cidrs.
    
With this change a comma separated list of subnet CIDRs can
also be used for monitor_address_block and radosgw_address_block.
    
NOTE:
* This change also updates the RPM spec file so that
  the filter_plugins are copied to the buildroot.
* Add a comment include 'ansible' in tests/requirements.txt
  so that netaddr is installed as requiement for the
  ceph-ansible-pr-syntax-check.
    
[1] commit: d67230b2a26b40651c1c1dbee68a92b0e851f3d5
[2] commit: 20e4852888ecc76d8d0fa194a438fa2a90e1cde3
    
Related-To: https://bugs.launchpad.net/tripleo/+bug/1840030
Related-To: https://bugzilla.redhat.com/show_bug.cgi?id=1740283
    
Closes: #4333
Please backport to stable-4.0
    
Signed-off-by: Harald Jensås <hjensas@redhat.com>

### DIFF
--- a/roles/ceph-facts/tasks/set_radosgw_address.yml
+++ b/roles/ceph-facts/tasks/set_radosgw_address.yml
@@ -1,7 +1,7 @@
 ---
 - name: set_fact _radosgw_address to radosgw_address_block ipv4
   set_fact:
-    _radosgw_address: "{{ hostvars[inventory_hostname]['ansible_all_ipv4_addresses'] | ipaddr(radosgw_address_block) | first }}"
+    _radosgw_address: "{{ hostvars[inventory_hostname]['ansible_all_ipv4_addresses'] | ips_in_ranges(hostvars[item]['radosgw_address_block'].split(',')) | first }}"
   when:
     - radosgw_address_block is defined
     - radosgw_address_block != 'subnet'
@@ -9,7 +9,7 @@
 
 - name: set_fact _radosgw_address to radosgw_address_block ipv6
   set_fact:
-    _radosgw_address: "{{ hostvars[inventory_hostname]['ansible_all_ipv6_addresses'] | ipaddr(radosgw_address_block) | last | ipwrap }}"
+    _radosgw_address: "{{ hostvars[inventory_hostname]['ansible_all_ipv6_addresses'] | ips_in_ranges(hostvars[item]['radosgw_address_block'].split(',')) | last | ipwrap }}"
   when:
     - radosgw_address_block is defined
     - radosgw_address_block != 'subnet'
--- a/tox.ini
+++ b/tox.ini
@@ -22,6 +22,7 @@ setenv=
   ANSIBLE_ACTION_PLUGINS = {toxinidir}/plugins/actions
   ANSIBLE_CALLBACK_PLUGINS = {toxinidir}/plugins/callback
   ANSIBLE_CALLBACK_WHITELIST = profile_tasks
+  ANSIBLE_FILTER_PLUGINS = {toxinidir}/plugins/filter
   CEPH_STABLE_RELEASE = luminous
   # only available for ansible >= 2.5
   ANSIBLE_STDOUT_CALLBACK = yaml
@@ -370,8 +371,6 @@ sitepackages=True
 setenv=
   ANSIBLE_SSH_ARGS = -F {changedir}/vagrant_ssh_config -o ControlMaster=auto -o ControlPersist=600s -o PreferredAuthentications=publickey
   ANSIBLE_CONFIG = {toxinidir}/ansible.cfg
-  ANSIBLE_ACTION_PLUGINS = {toxinidir}/plugins/actions
-  ANSIBLE_CALLBACK_PLUGINS = {toxinidir}/plugins/callback
   ANSIBLE_CALLBACK_WHITELIST = profile_tasks
   ANSIBLE_KEEP_REMOTE_FILES = 1
   ANSIBLE_CACHE_PLUGIN = memory
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,5 @@ install:
   - pip install ansible-lint pytest
 script:
   - if [[ -n $(grep --exclude-dir=.git -P "\xa0" -r .) ]]; then echo 'NBSP characters found'; exit 1; fi
-  - pytest -vvvv library/
+  - pytest -vvvv library/ plugins/
   - for i in $(ls -1 roles/); do ansible-lint -x 204 -v roles/$i/ ; done
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -6,6 +6,7 @@ ansible_managed = Please do not change this file directly since it is managed by
 library = ./library
 action_plugins = plugins/actions
 callback_plugins = plugins/callback
+filter_plugins = plugins/filter
 roles_path = ./roles
 # Be sure the user running Ansible has permissions on the logfile
 log_path = $HOME/ansible/ansible.log
--- a/infrastructure-playbooks/purge-cluster.yml
+++ b/infrastructure-playbooks/purge-cluster.yml
@@ -67,7 +67,7 @@
           run_once: true
 
         - name: get all nfs-ganesha mount points
-          command: grep "{{ hostvars[item]['ansible_all_ipv4_addresses'] | ipaddr(public_network) | first }}" /proc/mounts
+          command: grep "{{ hostvars[item]['ansible_all_ipv4_addresses'] | ips_in_ranges(public_network.split(',')) | first }}" /proc/mounts
           register: nfs_ganesha_mount_points
           failed_when: false
           with_items: "{{ groups[nfs_group_name] }}"
--- a/infrastructure-playbooks/purge-docker-cluster.yml
+++ b/infrastructure-playbooks/purge-docker-cluster.yml
@@ -51,7 +51,7 @@
           run_once: true
 
         - name: get all nfs-ganesha mount points
-          command: grep "{{ hostvars[item]['ansible_all_ipv4_addresses'] | ipaddr(public_network) | first }}" /proc/mounts
+          command: grep "{{ hostvars[item]['ansible_all_ipv4_addresses'] | ips_in_ranges(public_network.split(',')) | first }}" /proc/mounts
           register: nfs_ganesha_mount_points
           failed_when: false
           with_items: "{{ groups[nfs_group_name] }}"
--- a/plugins/filter/ipaddrs_in_ranges.py
+++ b/plugins/filter/ipaddrs_in_ranges.py
@@ -1,0 +1,30 @@
+from ansible import errors
+
+try:
+    import netaddr
+except ImportError:
+    # in this case, we'll make the filter return an error message (see bottom)
+    netaddr = None
+
+
+class FilterModule(object):
+   ''' IP addresses within IP ranges '''
+
+   def ips_in_ranges(self, ip_addresses, ip_ranges):
+       ips_in_ranges = list()
+       for ip_addr in ip_addresses:
+           for ip_range in ip_ranges:
+               if netaddr.IPAddress(ip_addr) in netaddr.IPNetwork(ip_range):
+                   ips_in_ranges.append(ip_addr)
+       return ips_in_ranges
+
+   def filters(self):
+       if netaddr:
+           return {
+               'ips_in_ranges': self.ips_in_ranges
+           }
+       else:
+           # Need to install python's netaddr for these filters to work
+           raise errors.AnsibleFilterError(
+               "The ips_in_ranges filter requires python's netaddr be "
+               "installed on the ansible controller.")
--- a/plugins/filter/test_ipaddrs_in_ranges.py
+++ b/plugins/filter/test_ipaddrs_in_ranges.py
@@ -1,0 +1,43 @@
+from . import ipaddrs_in_ranges
+
+filter_plugin = ipaddrs_in_ranges.FilterModule()
+
+class TestIpaddrsInRanges(object):
+
+    def test_one_ip_one_range(self):
+        ips = ['10.10.10.1']
+        ranges = ['10.10.10.1/24']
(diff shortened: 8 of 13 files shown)


hjensas:
I'm not sure what to do with this CI error:
https://2.jenkins.ceph.com/job/ceph-ansible-prs-dev-ubuntu-non_container-all_daemons/578/consoleFull

It's failing because netaddr is'nt on the master. How can that be when netaddr is in both requirement.txt and tests/requirements.txt?

guits:
setting `monitor_address_block` at the inventory host-level should allow deploying monitors across different subnets.

hjensas:
> setting `monitor_address_block` at the inventory host-level should allow deploying monitors across different subnets.

I'm investigating the possibility to set this at the inventory host-level in TripleO. I am checking with peers who have more experience with ansible services in TripleO. AFICT TripleO does uses host-level vars very sparsely, and they are all set (in tripleo_common/inventory.py) outside of tripleo-heat-templates where service configuration is maintained. All host-level vars used are vary node specific, such as the hostnames on each network of a node, ips, uniqe server id's etc. See [1] for the nitty gritty details. There is no per-service vars defined at all.

[1] https://github.com/openstack/tripleo-common/blob/61f60b2b735d76be34be0336405164b71e89ee2a/tripleo_common/inventory.py#L219-L275

hjensas:
@guits @fultonj @gfidente 

I have been testing a WIP patch to fix this in tripleo[1], however now the deployment fails in a different place. With the above patch the deployment no fail with:

    "TASK [ceph-facts : set grafana_server_addr fact] *******************************",
    "task path: /usr/share/ceph-ansible/roles/ceph-facts/tasks/facts.yml:293",
    "Wednesday 04 September 2019  15:19:21 +0000 (0:00:00.857)       0:01:36.505 *** ",
    "fatal: [overcloud-controller-0]: FAILED! => ",
    "  msg: 'ipaddr: unknown filter type: 172.120.3.0/24,172.117.3.0/24,172.118.3.0/24,172.119.3.0/24'",

This happens since this change[2] started to use the ansible ipaddr module to get the first ip address in 'public_network'. I.e we can no longer pass a comma separated list of IP's as the public_network. 

This is a **regression**! We have documented using a comma separated list for the public_network and cluster_network in Red Hat's OSP documentation starting with OSP13. See the *CephAnsibleExtraConfig* in example[3], [step 3 in the procedure](https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/13/html-single/spine_leaf_networking/index#assigning-routes-for-roles).

We used to be able to do:
```
CephAnsibleExtraConfig:
  public_network: '172.16.0.0/24,172.16.1.0/24,172.16.2.0/24'
  cluster_network: '172.17.0.0/24,172.17.1.0/24,172.17.2.0/24'
```

Doing so, now fails with the `ipaddr: unknown filter type` error seen above.
The filter intoduced in this patch can be used by `set grafana_server_addr fact` as well.

[1] https://review.opendev.org/#/c/679060/
[2] https://github.com/ceph/ceph-ansible/commit/d67230b2a26b40651c1c1dbee68a92b0e851f3d5
[3] https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/13/html-single/spine_leaf_networking/index#assigning-routes-for-roles

fultonj:
Talked to @dsavineau and we would probably need to update the following to deal with multiple IPs:

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-infra/tasks/configure_firewall.yml#L25

Also, we should add a test for this.

tests/requirements.txt
```diff
@@ -5,8 +5,8 @@ pytest-xdist==1.28.0
 pytest>=4.4,<4.5
 notario>=0.0.13
 ansible>=2.8,<2.9
-netaddr
+netaddr # For the ansible filter: ips_in_ranges
```
dsavineau: I don't think you need to update this file.

hjensas: Without this the CI won't install netaddr. Because a grep is used here:
https://github.com/ceph/ceph-build/blob/08b569d032e8d67643fdbbb33fb6006b3f8d635e/ceph-ansible-pr-syntax-check/build/build#L11

Adding a comment containing the "ansible" string is more like a workaround to ensure that CI job passes. (Adding the "ansible" string include that line in the grep result and thus the netaddr requirement is installed.)

dsavineau:
> I was'nt aware of the firewall issue. I would suggest addressing that in a separate commit?

Yes we can.

> Regarding the mon container, it looks like the code[1] already expect a comma separated list as it replaces whitespace? (`-e CEPH_PUBLIC_NETWORK={{ public_network | regex_replace(' ', '') }}`) The other pice of code[2] would be good to investigate, but looking at history most of it is 2 years old. If TripleO is using that code then it did'nt fail during OSP13/OSP14 testing.

Yes this seems only to be useful for network auto detect which isn't used in ceph-ansible.

tests/requirements.txt
dsavineau: No, we install the deps via tox https://github.com/ceph/ceph-ansible/blob/master/tox.ini#L418 otherwise the ipaddr filter would have never work.

hjensas: Not for the ceph-ansible-pr-syntax-check job, it installs deps via tox like this:
`"$VENV"/pip install -r <(grep ansible "$WORKSPACE"/ceph-ansible/tests/requirements.txt)`

Effectively passing only *tests/requirements.txt#L7* containing `ansible>=2.8,<2.9`.

The *ceph-ansible-pr-syntax-check* never exercice any code that use the ipaddr filter, and since ansible does conditional loading it will work until the filter is actually used.

I do however agree that me adding that comment in requirements.txt is a workaround. I will remove it and then we can solve the problem in ceph/ceph-build instead.

hjensas: Here is the CI log showing that dependencies are not properly installed:
https://2.jenkins.ceph.com/job/ceph-ansible-pr-syntax-check/1204/console

hjensas: I pushed this PR https://github.com/ceph/ceph-build/pull/1391 which I belive should fix the problem with netaddr not being pulled in as a dependency in the *ceph-ansible-pr-syntax-check* job.

dsavineau:
jenkins test pipeline

dsavineau:
Could you modify the `.travis.yml` file and add the plugins directory in the pytest command ? otherwise the filter tests won't be executed.
Also trying to run the tests on my side and wasn't able until I created the `__init__.py` file in the same directory.

```console
$ pytest -vvvv plugins/
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.7.3, pytest-5.1.2, py-1.8.0, pluggy-0.12.0 -- /home/ds/Workspace/.venv/pytest/bin/python3
cachedir: .pytest_cache
rootdir: /home/ds/Documents/repo/ceph/ceph-ansible
collected 0 items / 1 errors                                                                                                                                                                                                                 

=================================================================================================================== ERRORS ===================================================================================================================
_________________________________________________________________________________________ ERROR collecting plugins/filter/test_ipaddrs_in_ranges.py __________________________________________________________________________________________
ImportError while importing test module '/home/ds/Documents/repo/ceph/ceph-ansible/plugins/filter/test_ipaddrs_in_ranges.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/home/ds/Workspace/.venv/pytest/lib/python3.7/site-packages/_pytest/python.py:501: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
/home/ds/Workspace/.venv/pytest/lib/python3.7/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
/home/ds/Workspace/.venv/pytest/lib/python3.7/site-packages/_pytest/assertion/rewrite.py:140: in exec_module
    exec(co, module.__dict__)
plugins/filter/test_ipaddrs_in_ranges.py:1: in <module>
    from . import ipaddrs_in_ranges
E   ImportError: attempted relative import with no known parent package
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
============================================================================================================== 1 error in 0.10s ==============================================================================================================
```

Also looking at the CI failure I think we need to modify all tox files and add the `ANSIBLE_FILTER_PLUGINS` env variable otherwise the playbooks located in the infrastructure-playbook directory won't find the custom filter. You can take a look at the `ANSIBLE_ACTION_PLUGINS` configuration as an exemple.

dsavineau:
> Also trying to run the tests on my side and wasn't able until I created the `__init__.py` file in the same directory.

Soory I didn't see that was already handled in the PR.

hjensas:
Thanks @dsavineau, let's hope it passes now.

hjensas:
> Thanks @dsavineau, let's hope it passes now.

Not quite, I'll work no fixing the tests.

hjensas:
afict the CI error is not related to this change:

> Path: /home/jenkins-build/build/workspace/ceph-ansible-prs-dev-ubuntu-container-all_daemons/Vagrantfile
> Line number: 0
> Message: Errno::ENOENT: No such file or directory @ rb_sysopen - /home/jenkins-build/build/workspace/ceph-ansible-prs-dev-ubuntu-container-all_daemons/vagrant_variables.yml

dsavineau:
jenkins test pipeline

roles/ceph-facts/tasks/set_radosgw_address.yml
```diff
@@ -1,15 +1,15 @@
 ---
 - name: set_fact _radosgw_address to radosgw_address_block ipv4
   set_fact:
-    _radosgw_address: "{{ hostvars[inventory_hostname]['ansible_all_ipv4_addresses'] | ipaddr(radosgw_address_block) | first }}"
+    _radosgw_address: "{{ hostvars[inventory_hostname]['ansible_all_ipv4_addresses'] | ips_in_ranges(hostvars[item]['monitor_address_block'].split(',')) | first }}"
```
dsavineau: didn't catch first the first time but I think it's a wrong copy/paste because you changed the radosgw_address_block to monitor_address_block

tox.ini
```diff
@@ -22,6 +22,7 @@ setenv=
   ANSIBLE_ACTION_PLUGINS = {toxinidir}/plugins/actions
   ANSIBLE_CALLBACK_PLUGINS = {toxinidir}/plugins/callback
   ANSIBLE_CALLBACK_WHITELIST = profile_tasks
+  ANSIBLE_FILTER_PLUGINS = {toxinidir}/plugins/filter
```
dsavineau: There's also tox-update.ini and tox-podman.ini files to update even if I'm not sure this is useful since we already define the ansible.cfg via ANSIBLE_CONFIG environment variable

roles/ceph-facts/tasks/set_radosgw_address.yml
hjensas: oh, *facepalm*. Good thing you noticed this!

tox.ini
hjensas: Done.

dsavineau:

dsavineau:

dsavineau:
The failure in the CI seems to come from broken ceph rpm dependencies so unrelated to the PR

dsavineau:

hjensas:
Last push was just a rebase to master.

hjensas:
jenkins test pipeline

hjensas:
@guits Why is the DNM label back on this? Are you working on an alternative solution?

I have worked with @dsavineau and some related fixes have already been merged. (https://github.com/ceph/ceph-ansible/pull/4433 and https://github.com/ceph/ceph-build/pull/1391).

We attempted a fix in TripleO, see [[1]](https://review.opendev.org/679410), [[2]](https://review.opendev.org/679412) and [[3]](https://review.opendev.org/679060). It would be possible for the `monitor_address_range`. **However** when we tested that fix ceph-ansible failed on the next fact: `ceph-facts : set grafana_server_addr fact` with basically the same error. `grafana_server_addr` fact is derived from the `public_network`. The value for `public_network` must be the comma separated list of IP subnets in `ceph.conf`. I.e setting `public_network` to a single IP subnet at the inventory host-level will not work. 

See https://github.com/ceph/ceph-ansible/pull/4339#issuecomment-528066535 which have more details.


[1] https://review.opendev.org/679410
[2] https://review.opendev.org/679412
[3] https://review.opendev.org/679060

hjensas:
jenkins test pipeline

hjensas:

hjensas:

hjensas:
jenkins test pipeline

hjensas:
jenkins test pipeline

hjensas:
jenkins test pipeline

fultonj:
My understanding is that @guits wants to review this further with @dsavineau and then he'll probably request other changes or remove the DNM. He just doesn't want it to merge without his review.

guits:
jenkins test pipeline

guits:
> My understanding is that @guits wants to review this further with @dsavineau and then he'll probably request other changes or remove the DNM. He just doesn't want it to merge without his review.

yup, that's correct

PR #4344, opened 2019-08-19
osd: remove useless condition

guits:
just like `ceph_osd_pool_default_size`, a pool size might change after initial deployment.
Having this condition prevents from customizing the pool in that case.
This is not needed so let's remove it.


guits:

PR #4351, opened 2019-08-20
move python-xml to raw_install_python.yml

johanneskastl:
The package python-xml is needed for ansible's zypper module to interact with
the zypper package management tool.

roles/ceph-defaults/defaults/main.yml:
Remove python-xml from variable suse_package_dependencies to only
install python-xml on SUSE/openSUSE if python is not found.
raw_install_python.yml already contains all the logic needed to check
if there is a valid python installation, so this is better suited there.

openSUSE Leap 15.x / SLES 15.x do no longer have /usr/bin/python,
only /usr/bin/python3, which already contains the xml module, so
nothing needs to be installed in that case.

### DIFF
--- a/raw_install_python.yml
+++ b/raw_install_python.yml
@@ -56,3 +56,14 @@
       until: result is succeeded
       when: stat_zypper.rc == 0
   when: not True in (systempython.results | selectattr('stat', 'defined') | map(attribute='stat.exists') | list | unique)
+
+- name: install python-xml for opensuse only if python2 is installed already
+  raw: zypper -n install python-xml
+  register: result
+  until: result is succeeded
+  with_items: "{{ systempython.results }}"
+  when:
+    - stat_zypper.rc is defined
+    - stat_zypper.rc == 0
+    - item.stat.exists | bool
+    - item.stat.path == '/usr/bin/python'
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -89,8 +89,7 @@ dummy:
 
 #redhat_package_dependencies: []
 
-#suse_package_dependencies:
-#  - python-xml
+#suse_package_dependencies: []
 
 # Whether or not to install the ceph-test package.
 #ceph_test: false
--- a/group_vars/rhcs.yml.sample
+++ b/group_vars/rhcs.yml.sample
@@ -89,8 +89,7 @@ fetch_directory: ~/ceph-ansible-keys
 
 #redhat_package_dependencies: []
 
-#suse_package_dependencies:
-#  - python-xml
+#suse_package_dependencies: []
 
 # Whether or not to install the ceph-test package.
 #ceph_test: false
--- a/roles/ceph-defaults/defaults/main.yml
+++ b/roles/ceph-defaults/defaults/main.yml
@@ -81,8 +81,7 @@ centos_package_dependencies:
 
 redhat_package_dependencies: []
 
-suse_package_dependencies:
-  - python-xml
+suse_package_dependencies: []
 
 # Whether or not to install the ceph-test package.
 ceph_test: false


raw_install_python.yml
```diff
@@ -50,8 +50,25 @@
...
-      raw: zypper -n install python-base
+      raw: zypper -n install python-base python-xml
       register: result
       until: result is succeeded
       when: stat_zypper.rc == 0
-  when: not True in (systempython.results | selectattr('stat', 'defined') | map(attribute='stat.exists') | list | unique)
\ No newline at end of file
+  when: not True in (systempython.results | selectattr('stat', 'defined') | map(attribute='stat.exists') | list | unique)
+
+- name: check for python2 on SUSE/openSUSE
```
guits: this task is a duplicate of the first one

raw_install_python.yml
```diff
@@ -50,8 +50,25 @@
...
+  stat:
+    path: /usr/bin/python
```
guits: how do you know whether this is python2 or python3 ?

guits:
i'm -1 on merging this,
IMO, this is definitely not the best approach.
I think you should stick to handle this `python-xml` installation in ceph-common since you can use a condition like `ansible_python_version is version( '3', '<')`

raw_install_python.yml
johanneskastl: On openSUSE Leap 15.x / SLES 15.x, only python2 provides `/usr/bin/python`.

raw_install_python.yml
johanneskastl: Yes, but I was not sure if I could get only the detection of `/usr/bin/python` out of the first task.

johanneskastl:
> i'm -1 on merging this,
> IMO, this is definitely not the best approach.
> I think you should stick to handle this `python-xml` installation in ceph-common since you can use a condition like `ansible_python_version is version( '3', '<')`

AFAIK python-xml is needed to install packages via the `zypper` module, that ansible uses as soon as you use a `package` task. So this needs to be installed using the raw module before anything else uses the `package` module.

Of course we could somehow add a raw install task later one, but IMHO this is the right place for raw tasks.

guits:
> AFAIK python-xml is needed to install packages via the `zypper` module, that ansible uses as soon as you use a `package` task. 

ah yes, in fact, you already told this in the other PR, I forgot.
ok, in that case, yes, it's kind of egg or chicken issue, indeed.

well, in that case, you could at least reuse the first task.

raw_install_python.yml
```diff
@@ -50,8 +50,25 @@
...
+      raw: zypper -n install python-base python-xml
       register: result
       until: result is succeeded
       when: stat_zypper.rc == 0
-  when: not True in (systempython.results | selectattr('stat', 'defined') | map(attribute='stat.exists') | list | unique)
\ No newline at end of file
+  when: not True in (systempython.results | selectattr('stat', 'defined') | map(attribute='stat.exists') | list | unique)
+
+- name: check for python2 on SUSE/openSUSE
+  stat:
```
guits:
```suggestion
```

raw_install_python.yml
```diff
@@ -50,8 +50,25 @@
...
+  failed_when: false
+  register: systempython2
+  when: 
+    - stat_zypper.rc == 0
+
+- name: install python-xml for opensuse only if python2 is installed already
+  raw: zypper -n install python-xml
+  register: result
+  until: result is succeeded
+  when:
```
guits:
```suggestion
  with_items: "{{ systempython.results }}"
  when: 
```

raw_install_python.yml
```diff
@@ -50,8 +50,25 @@
...
+  when: 
+    - stat_zypper.rc == 0
+
+- name: install python-xml for opensuse only if python2 is installed already
+  raw: zypper -n install python-xml
+  register: result
+  until: result is succeeded
+  when: 
+    - stat_zypper.rc == 0
+    - systempython2.stat.exists
```
guits:
```suggestion
    - item.stat.exists | bool
    - item.stat.path == '/usr/bin/python'
```

raw_install_python.yml
guits: remove this task

raw_install_python.yml
johanneskastl: Skipped, as I deleted the tasks as requested below

raw_install_python.yml
johanneskastl: Done.

raw_install_python.yml
johanneskastl: Done

raw_install_python.yml
johanneskastl: Done

johanneskastl:
Hi Guillaume, sorry it took so long to complete this. I followed your suggestions, please review again.

raw_install_python.yml
guits: you can remove this task now

johanneskastl:
> almost there, just run ./generate_group_vars_sample.sh

Done.

raw_install_python.yml
```diff
@@ -56,3 +56,13 @@
...
       when: stat_zypper.rc == 0
   when: not True in (systempython.results | selectattr('stat', 'defined') | map(attribute='stat.exists') | list | unique)
+
+- name: install python-xml for opensuse only if python2 is installed already
+  raw: zypper -n install python-xml
+  register: result
+  until: result is succeeded
+  with_items: "{{ systempython.results }}"
+  when:
+    - stat_zypper.rc == 0
```
guits:
```suggestion
    - stat_zypper.rc is defined
    - stat_zypper.rc == 0
```

johanneskastl: Fixed.

PR #4359, opened 2019-08-21
fix SUSE/openSUSE naming

johanneskastl:
As SUSE 15.x and openSUSE Leap 15.x share the same base, make it clear that both are targeted by the respective tasks


johanneskastl:
> please, squash your commits

- [X] Done.

PR #4364, opened 2019-08-21
roles/ceph-validate/tasks/check_system.yml: fail on unsupported SUSE versions

johanneskastl:
Fail if SUSE distributions other than 15.x are found, similar to what we have for openSUSE

### DIFF
--- a/roles/ceph-validate/tasks/check_system.yml
+++ b/roles/ceph-validate/tasks/check_system.yml
@@ -46,11 +46,11 @@
     - ceph_repository == 'uca'
     - ansible_distribution != 'Ubuntu'
 
-- name: "fail on unsupported openSUSE distribution (only 15.x supported)"
+- name: "fail on unsupported SUSE/openSUSE distribution (only 15.x supported)"
   fail:
-    msg: "Distribution not supported: {{ ansible_distribution }}"
+    msg: "Distribution not supported: {{ ansible_distribution }} {{ ansible_distribution_major_version }}"
   when:
-    - ansible_distribution == 'openSUSE Leap'
+    - ansible_distribution == 'openSUSE Leap' or ansible_distribution == 'SUSE'
     - ansible_distribution_major_version != '15'
 
 - name: fail on unsupported ansible version (1.X)


roles/ceph-validate/tasks/check_system.yml
```diff
@@ -46,11 +46,11 @@
...
     - ansible_distribution != 'Ubuntu'
 
-- name: "fail on unsupported openSUSE distribution (only 15.x supported)"
+- name: "fail on unsupported SUSE/openSUSE distribution (only 15.x supported)"
   fail:
-    msg: "Distribution not supported: {{ ansible_distribution }}"
+    msg: "Distribution not supported: {{ ansible_distribution }} {{ ansible_distribution_major_version }}"
   when:
-    - ansible_distribution == 'openSUSE Leap'
+    - ansible_distribution == 'openSUSE Leap' or ansible_distribution == 'SUSE'
```
guits: I'm not very familiar with openSUSE/SUSE releases.
just to be sure: regardless we are deploying on `SUSE` or `openSUSE Leap`, `15.x` is the only supported release for both, right?

johanneskastl: I have tested ceph-ansible (using the various tweaks, see the lot of other PRs) successfully on both SLES 15.x and openSUSE Leap 15.x.

I am not sure if anyone tested ceph-ansible on SLES12, I have this on my list. Once I am sure this works, I would adapt this condition once again.

We can of course wait until my tests are done, but that might take a while.

guits: > I am not sure if anyone tested ceph-ansible on SLES12, I have this on my list.

yeh, I think that's fine merging this as it is then.

johanneskastl: For the record:

> I am not sure if anyone tested ceph-ansible on SLES12, I have this on my list. Once I am sure this works, I would adapt this condition once again.

SLES12 SP1 fails to install chrony, as no chrony package is available, so it would need some more modifications for that to work. Not sure if SLES12 SP4 already has that package available, I will check and prepare something working in another pull request.

Issue #4361, opened 2019-08-21
obs not a valid value for ceph_repository


johanneskastl:
**Bug Report**

What happened:
According to the logic in [roles/ceph-common/tasks/installs/install_on_suse.yml](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-common/tasks/installs/install_on_suse.yml#L2), using `ceph_origin = repository`  and `ceph_repository = obs` is a valid combination on openSUSE/SUSE systems.

But the validation throws errors as it seems `obs` is missing as a valid value:
```
TASK [ceph-validate : validate provided configuration] **************************************************************************************************************************************
Wednesday 21 August 2019  18:40:00 +0000 (0:00:00.300)       0:00:31.437 ****** 
 [ERROR]: [kastl-leap151-ceph-04] Validation failed for variable: ceph_repository

 [ERROR]: [kastl-leap151-ceph-04] Given value for ceph_repository: obs

 [ERROR]: [kastl-leap151-ceph-04] Reason: ceph_repository must be either 'community', 'rhcs', 'dev', 'custom' or 'uca'

fatal: [kastl-leap151-ceph-04]: FAILED! => changed=false 
  msg: |-
    [kastl-leap151-ceph-04] Validation failed for variable: ceph_repository
    [kastl-leap151-ceph-04] Reason: ceph_repository must be either 'community', 'rhcs', 'dev', 'custom' or 'uca'
    [kastl-leap151-ceph-04] Given value for ceph_repository: obs
  stderr_lines:
  - '[kastl-leap151-ceph-04] Validation failed for variable: ceph_repository'
  - '[kastl-leap151-ceph-04] Reason: ceph_repository must be either ''community'', ''rhcs'', ''dev'', ''custom'' or ''uca'''
  - '[kastl-leap151-ceph-04] Given value for ceph_repository: obs'

```

What you expected to happen:
Either the description in `roles/ceph-common/tasks/installs/install_on_suse.yml` is outdated, or the validation is faulty.

How to reproduce it (minimal and precise):

**group_vars/all.yml**:
```
---
ceph_origin: repository
ceph_repository: obs
```

**Environment**:
* OS (e.g. from /etc/os-release):
```
NAME="openSUSE Leap"
VERSION="15.1 "
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.1"
PRETTY_NAME="openSUSE Leap 15.1"
ANSI_COLOR="0;32"
[...]

PR #4365, opened 2019-08-21
openSUSE OBS repository using ceph_stable_release

johanneskastl:
Instead of hardcoding `luminous`, use the `ceph_stable_release` variable to point to the correct repository.

Uncomment this in `roles/ceph-defaults/defaults/main.yml` to be avaible, as it is only used if `ceph_repository` is set to `obs`.

Could not be tested due to #4361.

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -209,7 +209,7 @@ dummy:
 # usually has newer Ceph releases than the normal distro repository.
 #
 #
-#ceph_obs_repo: "https://download.opensuse.org/repositories/filesystems:/ceph:/luminous/openSUSE_Leap_{{ ansible_distribution_version }}/"
+#ceph_obs_repo: "https://download.opensuse.org/repositories/filesystems:/ceph:/{{ ceph_stable_release }}/openSUSE_Leap_{{ ansible_distribution_version }}/"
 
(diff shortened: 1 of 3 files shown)


johanneskastl:
> please, squash your commits

- [X] Done.

guits:
> Uncomment this in `roles/ceph-defaults/defaults/main.yml` to be avaible, as it is only used if `ceph_repository` is set to `obs`.

this file isn't intended to be modified.

if you wanna override this, you must do it either in inventory, group_vars, ... but definitely not in `ceph-defaults/defaults/main.yml`

could you please amend your commit message?

thanks!

johanneskastl:
Sorry for the confusion. I meant: I have uncommented this in `ceph-defaults/defaults/main.yml` so this variable is available. This variable is only used on SUSE/openSUSE and only in case one sets ceph_repository to `obs`, so I figure there is no harm in having that variable set.

johanneskastl:
> do not edit sample files directly, use `./generate_group_vars_sample.sh`
> see https://github.com/ceph/ceph-ansible/blob/master/CONTRIBUTING.md#sample-files

I ran that script which made another modification to `group_vars/rhcs.yml.sample`.

I fixed a typo in my commit message and added, that I have recreated the sample files using the script.

PR #4366, opened 2019-08-21
Allow 'obs' as a value for ceph_repository 

johanneskastl:
- Allow `obs` as a value for `ceph_repository`
- Validate that ceph_obs_repo is defined if `ceph_repository` is set to `obs`
- Validate that `ansible_distribution` is set to `openSUSE Leap` if `ceph_repository` is set to `obs`

Todo:
- [X] Check that this actually works and correctly validates...

### DIFF
--- a/plugins/actions/validate.py
+++ b/plugins/actions/validate.py
@@ -47,6 +47,7 @@ class ActionModule(ActionBase):
 
         try:
             notario_store["groups"] = host_vars["groups"]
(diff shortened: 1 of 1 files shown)


johanneskastl:
Tested on openSUSE Leap 15.1 using this group_vars/all.yml:
```
---
ceph_origin: repository
ceph_repository: obs
```

johanneskastl:
> please, squash your commits

- [X] Done.

PR #4374, opened 2019-08-22
facts: fix a typo

johanneskastl:


roles/ceph-facts/tasks/facts.yml
```diff
@@ -321,4 +321,4 @@
 
 - name: set_fact use_new_ceph_iscsi package or old ceph-iscsi-config/cli
   set_fact:
-    use_new_ceph_iscsi: "{{ (gateway_ip_list  == '0.0.0.0' and gateway_iqn | length == 0 and client_connections | length == 0 and rbd_devices | length == 0) | bool | ternary(true, false) }}"
\ No newline at end of file
+    use_new_ceph_iscsi: "{{ (gateway_ip_list  == '0.0.0.0' and gateway_iqn | length == 0 and client_connections | length == 0 and rbd_devices | length == 0) | bool | ternary(true, false) }}"
```
guits: I've noticed all your recent patches modify the last line, not sure whether it comes from your editor?
could you revert this change?

johanneskastl: Yes, but should there not be a newline at the end of the last line? IMHO this is a lot cleaner. Try to `cat` a file without a newline at the end... :-)

But of course I can revert this, if you wish.

guits:
please, always add a commit message even if the change is obvious

we usually prefix the title by the role/component being modified by the patch

suggestion:
title:  `facts: fix a typo`
commit message: `this commit fixes a typo in roles/ceph-facts/tasks/facts.yml`

Thanks! :+1:

johanneskastl:
> please, always add a commit message even if the change is obvious
> 
> we usually prefix the title by the role/component being modified by the patch
> 
> suggestion:
> title: `facts: fix a typo`
> commit message: `this commit fixes a typo in roles/ceph-facts/tasks/facts.yml`

- [X] Done.

roles/ceph-facts/tasks/facts.yml
guits: > Yes, but should there not be a newline at the end of the last line? IMHO this is a lot cleaner. Try to `cat` a file without a newline at the end... :-)

you are probably right

> But of course I can revert this, if you wish.

I would prefer to get a new PR especially for fixing this in that case, like a big commit fixing it everywhere :+1: 

perhaps adding a new job in the CI checking for last line deletion could be an idea as well.

PR #4375, opened 2019-08-22
generating cluster fsid failed if ansible_python_interpreter is set

johanneskastl:
If the user has set the `ansible_python_interpreter`, ansible will not try to discover python, so `discovered_python_interpreter` will not be set.

Solution: Use `ansible_python_interpreter` as fallback, if `discovered_python_interpreter` is unset.

### DIFF
--- a/roles/ceph-facts/tasks/facts.yml
+++ b/roles/ceph-facts/tasks/facts.yml
@@ -21,6 +21,14 @@
   set_fact:
     container_binary: "{{ 'podman' if (podman_binary.stat.exists and ansible_distribution == 'Fedora') or (ansible_os_family == 'RedHat' and ansible_distribution_major_version == '8') else 'docker' }}"
 
+# In case ansible_python_interpreter is set by the user,
+# ansible will not discover python and discovered_python_interpreter
+# will not be set
+- name: set_fact discovered_python_interpreter
+  set_fact:
+    discovered_python_interpreter: ansible_python_interpreter
+  when: ansible_python_interpreter is defined
+
 # Set ceph_release to ceph_stable by default
 - name: set_fact ceph_release ceph_stable_release
   set_fact:


guits:
another option could be to set `discovered_python_interpreter` early in the playbook when `ansible_python_interpreter` is defined?
otherwise falling back is fine but there are a couple of other places where it needs to be done:

```bash
roles/ceph-facts/tasks/facts.yml:109:    command: "{{ discovered_interpreter_python }} -c 'import uuid; print(str(uuid.uuid4()))'"
roles/ceph-handler/templates/restart_mon_daemon.sh.j2:22:  $DOCKER_EXEC ceph --cluster {{ cluster }} -s --format json | "{{ discovered_interpreter_python }}" -c 'import sys, json; exit(0) if "{{ monitor_name }}" in json.load(sys.stdin)["quorum_names"] else exit(1)' && exit 0
roles/ceph-handler/templates/restart_osd_daemon.sh.j2:7:  num_pgs=$($container_exec ceph $CEPH_CLI -s -f json | "{{ discovered_interpreter_python }}" -c 'import sys, json; print(json.load(sys.stdin)["pgmap"]["num_pgs"])')
roles/ceph-handler/templates/restart_osd_daemon.sh.j2:12:    test "$($container_exec ceph $CEPH_CLI -s -f json | "{{ discovered_interpreter_python }}" -c 'import sys, json; print(json.load(sys.stdin)["pgmap"]["num_pgs"])')" -eq "$($container_exec ceph $CEPH_CLI -s -f json | "{{ discovered_interpreter_python }}" -c 'import sys, json; print(sum ( [ i["count"] for i in json.load(sys.stdin)["pgmap"]["pgs_by_state"] if "active+clean" in i["state_name"]]))')"
roles/ceph-mon/tasks/deploy_monitors.yml:14:    {{ discovered_interpreter_python }} -c "import os ; import struct ;
```

johanneskastl:
> another option could be to set `discovered_python_interpreter` early in the playbook when `ansible_python_interpreter` is defined?

Sure, sounds like this needs to be fixed properly.

In which place should I put this? I'll try to do this tonight...

guits:
I think early in `ceph-facts` role is the right place

johanneskastl:
Please have a look if this seems like a valid solution.

roles/ceph-facts/tasks/facts.yml
```diff
@@ -21,6 +21,14 @@
   set_fact:
     container_binary: "{{ 'podman' if (podman_binary.stat.exists and ansible_distribution == 'Fedora') or (ansible_os_family == 'RedHat' and ansible_distribution_major_version == '8') else 'docker' }}"
 
+# In case ansible_python_interpreter is set by the user,
+# ansible will not discover python and discovered_python_interpreter
+# will not be set
+- name: set discovered_python_interpreter if ansible_python_interpreter is defined
```
guits:
```suggestion
- name: set_fact discovered_python_interpreter
```

johanneskastl:
I guess this should be backported to stable-4.0?

PR #4378, opened 2019-08-22
fix openSUSE OBS repo creation

johanneskastl:
roles/ceph-common/tasks/installs/suse_obs_repository.yml:
ansible's zypper_repository module does not know a parameter 'uri', this is called 'repo' instead


guits:
@johanneskastl considering our CI takes a couple of hours to run completely, I think you should make 1 PR for fixing support of suse/opensuse

guits:
@johanneskastl I think you want this backported? just like all the other PRs in fact?

johanneskastl:
Hi Guillaume,
as this is needed for openSUSE with OBS, but is also only used in that case, it is safe to backport.

PR #4383, opened 2019-08-23
Add http_addr option to grafana config

fmount:
We have no reason to make grafana container
listen on *, so this change adds the
http_addr option to the grafana config file
and adds the related option on the wait_for
tasks.

### DIFF
--- a/roles/ceph-grafana/tasks/configure_grafana.yml
+++ b/roles/ceph-grafana/tasks/configure_grafana.yml
@@ -17,6 +17,7 @@
 
 - name: wait for grafana to be stopped
   wait_for:
+    host: '{{ grafana_server_addr }}'
     port: '{{ grafana_port }}'
     state: stopped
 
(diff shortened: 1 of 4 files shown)


fmount:
Since we  moved the hostgroup section to "grafana-server' only (with no defaults), I think we can actually remove the default(_current_monitor_address) section. 
We also have a set_fact for grafana_server_addr variable (https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-facts/tasks/facts.yml#L314), so I think we're pretty sure that we don't need to rely on monitors. 
Are you ok to remove this default?

fmount:
@dsavineau agree, I'll make more tests and then I'll update the PR.

fmount:
@dsavineau @guits I think we can remove the _current_monitor_addr default since if the dashboard boolean is true, we have this [validation](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-validate/tasks/main.yml#L128-138) block that make us sure we need at least one node in the grafana group.

fmount:
Done

dsavineau:
Thanks Francesco !

PR #4413, opened 2019-08-30
tests: update dedidated mgr node all_daemons

dsavineau:
5b29144 change the mgr node to a dedicated node instead of the first
monitor node.
But the change didn't update the switch-to-containers inventory which
cause this playbook to fail.
Also update the ubuntu inventory to have the same configuration.


dsavineau:
jenkins test luminous-centos-non_container-switch_to_containers

PR #4284, opened 2019-07-30
Drop custom group name support

guits:
This feature doesn't work, this brings confusion and issues.

Adding default value to keep backward compatibility for the few places
where people would have used.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1509201

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #4417, opened 2019-09-03
site.yml: use inventory host group variables

samchoraria:
Use the inventory host group variables as defined in group_vars/all.yml.
This allows use of custom inventory groups to be specified; defaults of
existing group names leaves behaviour unchanged if variables not set.

Signed-off-by: Sam Choraria <sam.choraria@bbc.co.uk>


samc-bbc:
Hi, I'd like to get this backported to stable-4.0 also please...

dsavineau:
I think @guits already started to work on that in #4284
Also you only updated the site.yml playbook but you need to do the same for site-container.yml and all other playbooks present in the infrastructure-playbooks directory

samc-bbc:
As requested, for site.yml.sample, site-container.yml.sample & infrastructure-playbooks/

> I think @guits already started to work on that in #4284
> Also you only updated the site.yml playbook but you need to do the same for site-container.yml and all other playbooks present in the infrastructure-playbooks directory

samc-bbc:
Hi just seeing if this needs further re-work or is good to go. Thanks...

samc-bbc:
Hi,  is any further action needed on this PR?

dsavineau:
Well we are considering dropping the support of xxxx_group_name variables because you actually can't change the default values (if you want to give it a try too).

When using `hosts: '{{ xxxx_group_name|default("zzzz") }}'` it will always use `zzzz` even if you defined `xxxx_group_name` in the group_vars with a different value.

So if we drop this feature, your PR no longer makes sense.

samc-bbc:
> Well we are considering dropping the support of xxxx_group_name variables because you actually can't change the default values (if you want to give it a try too).
> 
> When using `hosts: '{{ xxxx_group_name|default("zzzz") }}'` it will always use `zzzz` even if you defined `xxxx_group_name` in the group_vars with a different value.
> 
> So if we drop this feature, your PR no longer makes sense.

I was under the impression openstack-ansible are using this feature also?

dsavineau:
> We've used this approach to successfully upgrade an existing cluster, without any of the hardcoded group names in our inventory. I believe the openstack-ansible project are using this feature also.

I'm curious to see how this could work. Because I tested this 2 days ago with ansible 2.6 and 2.8 and it never worked.

Do you have any logs or sample to share ?

samc-bbc:
> > We've used this approach to successfully upgrade an existing cluster, without any of the hardcoded group names in our inventory. I believe the openstack-ansible project are using this feature also.
> 
> I'm curious to see how this could work. Because I tested this 2 days ago with ansible 2.6 and 2.8 and it never worked.
> 
> Do you have any logs or sample to share ?

I'll try unearth some logs, but essentially we used `rolling_update.yml` (as per PR) against a fork of `stable-4.0` including changes in this PR plus https://github.com/ceph/ceph-ansible/pull/4452
Ran in an `ansible 2.8.4` virtualenv, group names supplied to `ansible-playbook` in the form: 
```
ansible-playbook -vvv -i {inventory_path} {playbook_path} \
                -e mgr_group_name=20_ceph-monitors.os \
                -e rgw_group_name=20_ceph-rgw-hosts \
                -e mon_group_name=20_ceph-monitors.os \
                -e osd_group_name=10_ceph-storage-hosts
```
Our ceph hosts belong to groups as above rather than `mons,mgrs,osds,rgws` etc. We don't use an `all` group exclusively for ceph hosts, hence commit 184a0cc to aggregate `{{ ceph_hosts_all }} `

samc-bbc:
forgot to mention upgrading mimic to nautilus...

dsavineau:
Ok so it's only working when passing the group name value via the extra vars but not with group/host vars which sucks.

samc-bbc:
openstack-ansible project usage of xxx_group_name:
https://opendev.org/openstack/openstack-ansible/src/branch/master/inventory/group_vars/all/ceph.yml

jrosser:
> openstack-ansible project usage of xxx_group_name:
> https://opendev.org/openstack/openstack-ansible/src/branch/master/inventory/group_vars/all/ceph.yml

Yes this is very important to openstack-ansible to be able to override the group names, note that OSA just uses the ceph-ansible roles, not the playbooks, and we are setting the group name overrides via group_vars not extra vars so it would seem that this does work correctly for that use case.

Please do not remove the ability to change the group names :)

samc-bbc:
Hi, is there any further update on this PR? We'd appreciate if it wasn't closed even if requires further time to be considered / changes etc.

dsavineau:
> Hi, is there any further update on this PR? 

Last update on my side is that it's only working when using extra vars so I'm not sure we want to still support group name in the playbook samples (in the roles it's fine).

> We'd appreciate if it wasn't closed even if requires further time to be considered / changes etc.

I set the `do-not-close` flag for that.

bluikko:
It would be very disappointing if the custom inventory host groups were dropped. I am fine using extra vars for it as long as it is documented.

PR #4423, opened 2019-09-04
rhcs: Pin downstream containers

b-ranto:
We should pin down the versions of downstream container for dashboard
instead of using upstream containers.

Please let me know if there is a better/any other way of overriding the downstream versions of these containers.

### DIFF
--- a/group_vars/rhcs.yml.sample
+++ b/group_vars/rhcs.yml.sample
@@ -718,14 +718,14 @@ ceph_docker_registry: "registry.access.redhat.com"
 #dashboard_rgw_api_scheme: ''
 #dashboard_rgw_api_admin_resource: ''
 #dashboard_rgw_api_no_ssl_verify: False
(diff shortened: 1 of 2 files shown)


b-ranto:
@dsavineau I have updated the rhcs_edits.txt and re-run the script. Unfortunately, we can't pin to the latest tag, we are likely to face problems there. The grafana (v6.x) in ocp v4.2 is not fully backwards compatible with dashboard (we currently need v5.x). I have actually had QE reach out to me today because their deployment was not working with v6.x of upstream grafana (grafana failed to start), switching to upstream v5.x helped. To make matters worse, also node-exporter has a history of changing its options as well as metrics that it generates in new releases so we definitely need to pin down that one as well. We should not need to change this all that often though, maybe once a couple of months.

dsavineau:
unrelated CI failure

Issue #4425, opened 2019-09-06
configure_firewall.yml does not support multiple public_network subnets


hjensas:
ceph.conf supports a comma separated list of subnet CIDR's for the public_network ref below quote from[1].

> public network
> Description:	The IP address and netmask of the public (front-side) network
> (e.g., > 192.168.0.0/24). Set in [global]. **You may specify comma-delimited subnets.**
> Type:	{ip-address}/{netmask} [, {ip-address}/{netmask}]
> Required:	No
> Default:	N/A

The ansble code in `configure_firewall.yml` does not support this. It should handle a value such as: `172.16.0.0/24,172.16.1.0/24,172.16.2.0/24`.

[1] https://docs.ceph.com/docs/jewel/rados/configuration/network-config-ref/#network-config-settings

PR #4426, opened 2019-09-06
Support comma-delimited subnets in firewall

hjensas:
ceph.conf supports a comma separated list of
subnet CIDR's for the public_network and the
cluster network. ceph-ansible should support
setting up the firewall for this configuration.

Closes: #4425
Related: #4333
https://docs.ceph.com/docs/nautilus/rados/configuration/network-config-ref/#network-config-settings

Signed-off-by: Harald Jensås <hjensas@redhat.com>

### DIFF
--- a/roles/ceph-infra/tasks/configure_firewall.yml
+++ b/roles/ceph-infra/tasks/configure_firewall.yml
@@ -20,15 +20,16 @@
 
   - name: open monitor and manager ports
     firewalld:
-      service: "{{ item.service }}"
-      zone: "{{ item.zone }}"
-      source: "{{ public_network }}"
+      service: "{{ item[1].service }}"
+      zone: "{{ item[1].zone }}"
+      source: "{{ item[0] }}"
       permanent: true
       immediate: true
       state: enabled
-    with_items:
-      - { 'service': 'ceph-mon', 'zone': "{{ ceph_mon_firewall_zone }}" }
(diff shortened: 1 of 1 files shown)


roles/ceph-infra/tasks/configure_firewall.yml
```diff
@@ -18,17 +18,15 @@
       state: started
       enabled: yes
 
-  - name: open monitor and manager ports
+  - name: open monitor ports
     firewalld:
-      service: "{{ item.service }}"
-      zone: "{{ item.zone }}"
-      source: "{{ public_network }}"
+      service: "{{ ceph-mon }}"
```
dsavineau: `ceph-mon` is a static value not a variable.
Also you shouldn't remove the `ceph` service from this task otherwise the ports for the mgr won't be opened when not using the dedicated mgr ansible group (they are collocated on the mon nodes in that case).

hjensas: Thanks for the guidance @dsavineau.

I hope I got the `with_nested` thing right.

I also changed the `open osd ports` task to use the list `union` filter so that we have just one task for both `cluster_network` and `public_network`.

dsavineau:
jenkins test pipeline

dsavineau:
unrelated CI failure

PR #4431, opened 2019-09-09
Look for additional names when checking ceph-nfs container status

gfidente:
Ganesha cannot be operated active/active, in those deployments
where it is managed by pacemaker the container name can be
different than the default.

This change uses "ceph_nfs_service_suffix" where previously
missing to ensure tasks will work with customized names.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1750005
Signed-off-by: Giulio Fidente <gfidente@redhat.com>


gfidente:
@guits @dsavineau can you guys also label this for backport into 3.x

PR #4432, opened 2019-09-09
rbd-mirror: Use the rbd mirror client keyring

dsavineau:
The admin keyring isn't present by default on the rbd mirror nodes so
the rbd commands related to the mirroring confguration will fail.
Instead we can use the rbd mirror client keyring.


dsavineau:
jenkins test pipeline

PR #4437, opened 2019-09-10
Inject ceph grafana dashboard layouts

fmount:
This change just adds the task to inject from the
ceph dashboard mgr module the required layouts
to show all the cluster metrics on the grafana
instance.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -184,7 +184,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       grf.vm.hostname = "#{LABEL_PREFIX}grafana#{i}"
       if ASSIGN_STATIC_IP
         grf.vm.network :private_network,
-          ip: "#{PUBLIC_SUBNET}.3#{i}"
+          ip: "#{PUBLIC_SUBNET}.2#{i}"
       end
       # Virtualbox
       grf.vm.provider :virtualbox do |vb|
--- a/roles/ceph-facts/tasks/facts.yml
+++ b/roles/ceph-facts/tasks/facts.yml
@@ -318,6 +318,7 @@
     - groups.get(grafana_server_group_name, []) | length > 0
     - ip_version == 'ipv4'
     - dashboard_enabled | bool
+    - inventory_hostname in groups[grafana_server_group_name]
 
 - name: set grafana_server_addr fact - ipv6
   set_fact:
@@ -326,6 +327,25 @@
     - groups.get(grafana_server_group_name, []) | length > 0
     - ip_version == 'ipv6'
     - dashboard_enabled | bool
+    - inventory_hostname in groups[grafana_server_group_name]
+
+- name: set grafana_server_addr fact - ipv4 - (external instance)
+  set_fact:
+    grafana_server_addr: "{{ hostvars[groups.get(grafana_server_group_name, []) | first]['ansible_all_ipv4_addresses'] | ipaddr(public_network) | first }}"
+  when:
+    - groups.get(grafana_server_group_name, []) | length > 0
+    - ip_version == 'ipv4'
+    - dashboard_enabled | bool
+    - inventory_hostname not in groups[grafana_server_group_name]
+
+- name: set grafana_server_addr fact - ipv6 - (external instance)
+  set_fact:
+    grafana_server_addr: "{{ hostvars[groups.get(grafana_server_group_name, []) | first]['ansible_all_ipv6_addresses'] | ipaddr(public_network) | last | ipwrap }}"
+  when:
+    - groups.get(grafana_server_group_name, []) | length > 0
+    - ip_version == 'ipv6'
+    - dashboard_enabled | bool
+    - inventory_hostname not in groups[grafana_server_group_name]
 
 - name: set_fact use_new_ceph_iscsi package or old ceph-iscsi-config/cli
   set_fact:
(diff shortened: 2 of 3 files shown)


dsavineau:
Do we still need to install the ceph-grafana-dashboard packages on non containerized deployment ?

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-grafana/tasks/configure_grafana.yml#L2-L11

fmount:
Not sure about non containerized deployments: I'll conduct more tests and then I'll update the PR accordingly.
I think this task can handle both cases (endpoint and api should be available at that point), hence I suggest removing the `package install task` after some tests.

fmount:
@dsavineau since we're now able to inject grafana layouts using the mgr dashboard module command, do you think we can remove this task [1] and the related template [2] ?

[1] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-grafana/tasks/configure_grafana.yml#L53-60
[2] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-grafana/templates/dashboards-ceph-dashboard.yml.j2

dsavineau:
I honestly don't know if we still need this.

But currently the dashboard update command is failing
```console
  stderr: |-
    Error EINVAL: Traceback (most recent call last):
      File "/usr/share/ceph/mgr/mgr_module.py", line 915, in _handle_command
        return self.handle_command(inbuf, cmd)
      File "/usr/share/ceph/mgr/dashboard/module.py", line 414, in handle_command
        push_local_dashboards()
      File "/usr/share/ceph/mgr/dashboard/grafana.py", line 132, in push_local_dashboards
        retry()
      File "/usr/share/ceph/mgr/dashboard/grafana.py", line 89, in __call__
        result = self.func(*self.args, **self.kwargs)
      File "/usr/share/ceph/mgr/dashboard/grafana.py", line 127, in push
        grafana.push_dashboard(body)
      File "/usr/share/ceph/mgr/dashboard/grafana.py", line 53, in push_dashboard
        raise GrafanaError("Could not connect to Grafana server")
    GrafanaError: Could not connect to Grafana server
```

fmount:
I see this issue and now I'm able to reproduce it.
I partially solved in the last commit fixing the Vagrantfile static ip address assignment that was the same for both mgr and grafana instance (preventing to connect to the real grafana container).
So basically now I see:
```
[vagrant@mgr0 ~]$ ip -o -4 a | grep eth1
3: eth1    inet 192.168.42.30/24

[vagrant@grafana0 ~]$ ip -o -4 a | grep eth1
3: eth1    inet 192.168.42.40/24 
```
and we're ok.
Unfortunately, during the last role execution (ceph-dashboard) the fact related to the `grafana_server_addr` is overwritten and set to the mgr ip address, resulting in the following grafana-api-url configuration:
```
[root@mon0 /]# ceph dashboard get-grafana-api-url
http://192.168.42.30:3000
```
that is wrong and prevent to execute the `ceph dashboard grafana dashboards update` command successfully.

Vagrantfile
```diff
@@ -184,7 +184,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       grf.vm.hostname = "#{LABEL_PREFIX}grafana#{i}"
       if ASSIGN_STATIC_IP
         grf.vm.network :private_network,
-          ip: "#{PUBLIC_SUBNET}.3#{i}"
+          ip: "#{PUBLIC_SUBNET}.4#{i}"
```
dsavineau: `.4x` is also already used by clients nodes

```console
$ grep 'ip: "#{PUBLIC_SUBNET}\.' Vagrantfile 
          ip: "#{PUBLIC_SUBNET}.3#{i}"
          ip: "#{PUBLIC_SUBNET}.3#{i}"
          ip: "#{PUBLIC_SUBNET}.4#{i}"
          ip: "#{PUBLIC_SUBNET}.5#{i}"
          ip: "#{PUBLIC_SUBNET}.6#{i}"
          ip: "#{PUBLIC_SUBNET}.7#{i}"
          ip: "#{PUBLIC_SUBNET}.8#{i}"
          ip: "#{PUBLIC_SUBNET}.9#{i}"
          ip: "#{PUBLIC_SUBNET}.1#{i}"
          ip: "#{PUBLIC_SUBNET}.10#{i}"
```

I don't see any .2x so we could change to this. wdyt ?

fmount: Sure,  I think `.2x` is fine.

roles/ceph-facts/tasks/facts.yml
```diff
@@ -318,6 +318,7 @@
     - groups.get(grafana_server_group_name, []) | length > 0
     - ip_version == 'ipv4'
     - dashboard_enabled | bool
+    - inventory_hostname in groups[grafana_server_group_name]
```
dsavineau: in fact you don't need this condition (and the other one with `not in`)
In both scenarios (collocated or external) you need the grafana address from the grafana node. So the only way is to use `groups[grafana_server_group_name][0]` like you did in the external code but we can directly integrate that in the current code no need to duplicate the code.

roles/ceph-grafana/tasks/configure_grafana.yml
```diff
@@ -1,15 +1,4 @@
 ---
-- name: install ceph-grafana-dashboards package on RedHat or SUSE
```
dsavineau: in fact I'm wondering how this will work in non containerized deployment because the mgr node doesn't have the ceph dashboard package installed so you won't be able to run the dashboard update command.

So I guess there's 2 solutions:
  - keep the `ceph-grafana-dashboards` package installation on the grafana node and not running the dashboard update command.
  - install `ceph-grafana-dashboards` package on the mgr node and run the dashboard update command.

wdyt ? Maybe I'm missing something here

guits:
jenkins test pipeline

guits:
jenkins test pipeline

guits:

PR #4438, opened 2019-09-10
tox-update: set the ansible.cfg path before update

dsavineau:
During an upgrade we're installation the platform with the stable-3.2
branch. But the ansible configuration is still using the file from the
current branch which could have some differences.
Instead we can override the ANSIBLE_CONFIG environment variable with
the stable-3.2 commands.


dsavineau:

dsavineau:

PR #4444, opened 2019-09-10
container: Allow to use registry authentication

dsavineau:
The registry.redhat.io regsitry requires authentication so before pulling
the RHCS 4 container images from the registry we need to do the login
step.
This is done via the new ceph_docker_registry_auth variable. The
default value is false but true for RHCS setup.
When set to true, you need to provide the username and password
for the registry via the associated variables.
This patch also updates the ceph_docker_registry value for RHCS setup.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1748911

### DIFF
--- a/roles/ceph-validate/tasks/main.yml
+++ b/roles/ceph-validate/tasks/main.yml
@@ -142,4 +142,5 @@
     msg: 'ceph_docker_registry_username and/or ceph_docker_registry_password variables need to be set'
   when:
     - ceph_docker_registry_auth | bool
-    - ceph_docker_registry_username is not defined or ceph_docker_registry_password is not defined
+    - (ceph_docker_registry_username is not defined or ceph_docker_registry_password is not defined)
+    - (ceph_docker_registry_username | length == 0 or ceph_docker_registry_password | length == 0)
--- a/roles/ceph-validate/tasks/main.yml
+++ b/roles/ceph-validate/tasks/main.yml
@@ -136,3 +136,10 @@
         msg: "you must add at least one node in the [grafana-server] hosts group"
       when: groups[grafana_server_group_name] | length < 1
   when: dashboard_enabled | bool
+
+- name: validate container registry credentials
+  fail:
+    msg: 'ceph_docker_registry_username and/or ceph_docker_registry_password variables need to be set'
+  when:
+    - ceph_docker_registry_auth | bool
+    - ceph_docker_registry_username is not defined or ceph_docker_registry_password is not defined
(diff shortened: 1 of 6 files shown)


dsavineau:
@fultonj @gfidente @fmount I don't know if it could impact TripleO

dsavineau:
> Isn't the rhcs4 container in registry.redhat.io, not registry.access.redhat.com? Also, for the record it's the use of registry.redhat.io that requires auth, not RHEL8 or podman.

Thanks I wasn't aware about the difference between registry.access.redhat.com and registry.redhat.io registries.
I'll update this PR to change the ceph container registry to registry.redhat.io

> Wouldn't this change mean that the user has to store their credentials "in the clear"?
We already store some secrets in the group_vars files in plain text.
Do you have a solution ?

> Would it make sense to declare these user/password variables in the rhcs.yml.sample too?
From my point of view we don't need to declare user/password variables by default in rhcs.yml file.
I've added a validation test if those variables aren't set [1]

[1] https://github.com/ceph/ceph-ansible/pull/4444/files#diff-d4f8a0c1dabf72e117f5066953ff8eb6R140-R145

0xf2:
Guillaume, we need help on this one... ty.

guits:
@dsavineau This forces to set the password plain text in group_vars why not using the token file downloaded from access.redhat.com instead?

guits:
jenkins test pipeline

guits:
jenkins test pipeline

guits:

guits:
as discussed with @pcuzner let's improve this in a second step.

gfidente:
would it make sense to check if _username and _password are empty to avoid the new task fail if these are set to empty string?

roles/ceph-validate/tasks/main.yml
```diff
@@ -136,3 +136,10 @@
         msg: "you must add at least one node in the [grafana-server] hosts group"
       when: groups[grafana_server_group_name] | length < 1
   when: dashboard_enabled | bool
+
+- name: validate container registry credentials
+  fail:
+    msg: 'ceph_docker_registry_username and/or ceph_docker_registry_password variables need to be set'
+  when:
+    - ceph_docker_registry_auth | bool
+    - ceph_docker_registry_username is not defined or ceph_docker_registry_password is not defined
```
gfidente: I guess here we might look at if they are empty as well?

guits: I guess it wouldn't hurt to check if these variables are empty although I don't really see why you would define them with an empty string.
From a consistency point of view, It would mean we would have to start checking all string variables aren't empty in the whole playbook in that case.

gfidente: but in this specific case, where those variables represent a user/pass set, empty string would cause the login task to fail despite the validation check passing

guits:
unrelated failure

PR #4445, opened 2019-09-10
rgw: extend automatic rgw pool creation capability

alimaredia:

### DIFF
--- a/group_vars/rgws.yml.sample
+++ b/group_vars/rgws.yml.sample
@@ -32,21 +32,48 @@ dummy:
 # limit will require admin intervention.
 #rgw_bucket_default_quota_max_objects: 1638400 # i.e., 100K * 16
 
-# This dictionary will create pools with the given number of pgs.
-# This is important because they would be created with the default
-# of 8.
-# New pools and their corresponding pg_nums can be created
-# by adding to the rgw_create_pools dictionary (see foo).
+# Declaring rgw_create_pools will create pools with the given number of pgs, 
+# size, and pool_type. The following are some important notes on this automatic 
+# pool creation:
+#   - The pools and associated pg_num's below are merely examples of pools that
+#     could be automatically created when rgws are deployed.
+#   - The default pg_num is 8 (from osd_pool_default_pg_num) for pool created 
+#     if rgw_create_pools isn't declared and configured.
+#   - A pgcalc tool should be used to determine the optimal sizes for
+#     the rgw.buckets.data, rgw.buckets.index pools as well as any other
+#     pools declared in this dictionary.
+#     https://ceph.io/pgcalc is the upstream pgcalc tool
+#     https://access.redhat.com/labsinfo/cephpgc is a pgcalc tool offered by
+#     Red Hat if you are using RHCS.
+#   - The default value of {{ rgw_zone }} is 'default'.
+#   - The pool_type must be set as either 'replicated' or 'ec' for
+#     each pool. 
+#   - If a pool's pool_type is 'ec', k and m values must be set via
+#     the ec_k, and ec_m variables.
+
 #rgw_create_pools:
-#  defaults.rgw.buckets.data:
+#  "{{ rgw_zone }}.rgw.buckets.data":
+#    pg_num: 64
+#    pool_type: ec
+#    ec_profile: myecprofile
+#    ec_k: 5
+#    ec_m: 3
+#  "{{ rgw_zone }}.rgw.buckets.index":
 #    pg_num: 16
-#    size: ""
-#  defaults.rgw.buckets.index:
-#    pg_num: 32
-#    size: ""
-#  foo:
-#    pg_num: 4
-#    size: ""
+#    size: 3
+#    pool_type: replicated
+#  "{{ rgw_zone }}.rgw.meta":
+#    pg_num: 8
+#    size: 3
+#    pool_type: replicated
+#  "{{ rgw_zone }}.rgw.log":
+#    pg_num: 8
+#    size: 3
+#    pool_type: replicated
+#  "{{ rgw_zone }}.rgw.control":
+#    pg_num: 8
+#    size: 3
+#    pool_type: replicated
 
 
 ##########
--- a/roles/ceph-rgw/defaults/main.yml
+++ b/roles/ceph-rgw/defaults/main.yml
@@ -24,21 +24,48 @@ copy_admin_key: false
 # limit will require admin intervention.
 #rgw_bucket_default_quota_max_objects: 1638400 # i.e., 100K * 16
 
-# This dictionary will create pools with the given number of pgs.
-# This is important because they would be created with the default
-# of 8.
-# New pools and their corresponding pg_nums can be created
-# by adding to the rgw_create_pools dictionary (see foo).
+# Declaring rgw_create_pools will create pools with the given number of pgs, 
+# size, and pool_type. The following are some important notes on this automatic 
+# pool creation:
+#   - The pools and associated pg_num's below are merely examples of pools that
+#     could be automatically created when rgws are deployed.
+#   - The default pg_num is 8 (from osd_pool_default_pg_num) for pool created 
+#     if rgw_create_pools isn't declared and configured.
+#   - A pgcalc tool should be used to determine the optimal sizes for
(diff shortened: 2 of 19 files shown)


group_vars/rgws.yml.sample
```diff
@@ -33,20 +33,46 @@ dummy:
...
+# pools declared in this dictionary.
+# The default value of {{ rgw_zone }} is 'default'.
+# The pool_type must be set as either 'replicated' or 'ec' for
+# each pool. 
+# If a pool's pool_type is 'ec', k and m values must be set via
+# the ec_k, and ec_m variables.
 #rgw_create_pools:
-#  defaults.rgw.buckets.data:
-#    pg_num: 16
+#  {{ rgw_zone }}.rgw.buckets.data:
```
dsavineau: this needs to be quoted (and all the other keys using a variable)

roles/ceph-rgw/defaults/main.yml
```diff
@@ -25,20 +25,46 @@ copy_admin_key: false
...
+# pools declared in this dictionary.
+# The default value of {{ rgw_zone }} is 'default'.
+# The pool_type must be set as either 'replicated' or 'ec' for
+# each pool. 
+# If a pool's pool_type is 'ec', k and m values must be set via
+# the ec_k, and ec_m variables.
 #rgw_create_pools:
-#  defaults.rgw.buckets.data:
-#    pg_num: 16
+#  {{ rgw_zone }}.rgw.buckets.data:
```
dsavineau: same here

roles/ceph-rgw/tasks/rgw_create_pools.yml
```diff
@@ -0,0 +1,71 @@
+---
+  - name: remove ec profile
+    command: "{{ container_exec_cmd }} ceph --connect-timeout 5 --cluster {{ cluster }} osd erasure-code-profile rm {{ item.ec_profile }}"
```
dsavineau: because we're using a dict lookup the values need to use `item.value.<value>`
So it should be `item.value.ec_profile`

roles/ceph-rgw/tasks/rgw_create_pools.yml
```diff
@@ -0,0 +1,71 @@
...
+    delegate_to: "{{ groups[mon_group_name][0] }}"
+    changed_when: false
+    run_once: true
+    register: result
+    until: result is succeeded
+    when: item.pool_type == 'ec'
+    ignore_errors: yes
+
+  - name: set ec profile
+    command: "{{ container_exec_cmd }} ceph --connect-timeout 5 --cluster {{ cluster }} osd erasure-code-profile set {{ item.ec_profile }} k={{ item.ec_k }} m={{ item.ec_m }}"
```
dsavineau: same here

guits:
@alimaredia when you have a moment :slightly_smiling_face:

roles/ceph-rgw/tasks/rgw_create_pools.yml
```diff
@@ -0,0 +1,74 @@
...
+  - name: set ec profile
+    command: "{{ container_exec_cmd }} ceph --connect-timeout 5 --cluster {{ cluster }} osd erasure-code-profile set {{ item.value.ec_profile }} k={{ item.value.ec_k }} m={{ item.value.ec_m }}"
+    with_dict: "{{ rgw_create_pools }}"
+    delegate_to: "{{ groups[mon_group_name][0] }}"
+    changed_when: false
+    run_once: true
+    register: result
+    when: 
+      - item.value.pool_type is defined  
+      - item.value.pool_type == 'ec'
```
alimaredia: @guits should I check to make sure `ec_k` and `ec_m` are defined before this job is run?

alimaredia:
@guits @dsavineau: This PR should be ready for final review. I made some changes in the `tests` directory, let me know if there's anything else that needs to be done there.

roles/ceph-rgw/tasks/rgw_create_pools.yml
```diff
@@ -0,0 +1,74 @@
+---
+  - name: remove ec profile
+    command: "{{ container_exec_cmd }} ceph --connect-timeout 5 --cluster {{ cluster }} osd erasure-code-profile rm {{ item.value.ec_profile }}"
+    with_dict: "{{ rgw_create_pools }}"
+    delegate_to: "{{ groups[mon_group_name][0] }}"
+    changed_when: false
+    run_once: true
+    register: result
```
guits: all register here seem to be unused

roles/ceph-rgw/tasks/rgw_create_pools.yml
```diff
@@ -0,0 +1,74 @@
...
+    when: 
+      - item.value.pool_type is defined  
+      - item.value.pool_type == 'ec'
+    ignore_errors: yes
```
guits: why?

roles/ceph-rgw/tasks/rgw_create_pools.yml
guits: > @guits should I check to make sure `ec_k` and `ec_m` are defined before this job is run?

That wouldn't hurt :slightly_smiling_face: 

> Also is there a way to fail the job if those variables are not defined but `pool_type` is set to `ec`?

like an early check in ceph-validate?

roles/ceph-rgw/tasks/rgw_create_pools.yml
alimaredia: The reason we had this was if the playbook gets run again. When the playbook gets run again, it will try to rm the ec_profile and we didn't want it to fail if the ec_profile was already created and being used.

roles/ceph-rgw/tasks/rgw_create_pools.yml
alimaredia: Done. Take a look at the changes I just made to the ceph-validate task.

roles/ceph-rgw/tasks/main.yml
```diff
@@ -6,6 +6,10 @@
   include_tasks: pre_requisite.yml
   when: not containerized_deployment | bool
 
+- name: rgw pool creation tasks
```
dsavineau: Why did you move the pools creation here ?

Shouldn't this be done after the multisite tasks ?

roles/ceph-rgw/tasks/rgw_create_pools.yml
```diff
@@ -0,0 +1,67 @@
...
+    with_dict: "{{ rgw_create_pools }}"
+    delegate_to: "{{ groups[mon_group_name][0] }}"
+    changed_when: false
+    run_once: true
+    when: 
+      - item.value.pool_type is defined  
+      - item.value.pool_type == 'ec'
+
+  - name: set crush rule
+    command: "{{ container_exec_cmd }} ceph --connect-timeout 5 --cluster {{ cluster }} osd crush rule create-erasure {{ item.key }} {{ item.value.ec_profile }}"
```
guits: what about this command?
If the playbook is run again, will it return 0 even though it's already created? Otherwise, we might need a `failed_when: false` here as well

roles/ceph-validate/tasks/check_rgw_pools.yml
```diff
@@ -0,0 +1,44 @@
+---
+- name: fail if pool_type is not set
```
dsavineau: I think this is too restrictive and some tasks are breaking the backward compatibility.

As I mentioned in a previous comment, users that wouldn't have defined pool_type in the rgw_create_pools dict will have the playbook failing.

Some checks make sense, like checking ec_m/ec_k when the pool type is ec but others like size and pg_num have default value so you don't actually need to always define these keys in the dict (ie: `{{ item.value.pg_num | default(osd_pool_default_pg_num) }}`)

roles/ceph-rgw/tasks/main.yml
alimaredia: @cbodley and I realized that pool creation should be happening before the rgw is starting up or any multisite tasks.

roles/ceph-rgw/tasks/rgw_create_pools.yml
guits: @alimaredia ?

dsavineau:
Closing in flavor of #4688

PR #4446, opened 2019-09-10
[DNM] rgw multisite: add connectivity check for primary rgws when deploying…

alimaredia:
… secondary clusters

### DIFF
--- a/README-MULTISITE.md
+++ b/README-MULTISITE.md
@@ -79,6 +79,10 @@ system_secret_key: MGecsMrWtKZgngOHZdrd6d3JxGO5CPWgT2lcnpSt
 rgw_pull_proto: http
 rgw_pull_port: 8080
 rgw_pullhost: cluster0-rgw0
+rgw_primary_endpoints:
+  - http://cluster0-rgw0:8080
+  - http://cluster0-rgw1:8080
+  - http://cluster1-rgw0:8080
 ```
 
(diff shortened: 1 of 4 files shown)


roles/ceph-rgw/tasks/multisite/secondary.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- name: check connectivity to other cluster's rgw endpoints
```
dsavineau: Because we're not sure that curl command is available on the nodes.
```yaml
- name: check connectivity to other cluster's rgw endpoints
  uri:
    url: '{{ item }}'
  with_items: "{{ rgw_primary_endpoints }}"
  when: rgw_primary_endpoints is defined
```

README-MULTISITE.md
```diff
@@ -79,6 +79,10 @@ system_secret_key: MGecsMrWtKZgngOHZdrd6d3JxGO5CPWgT2lcnpSt
 rgw_pull_proto: http
 rgw_pull_port: 8080
 rgw_pullhost: cluster0-rgw0
+rgw_primary_endpoints:
```
dsavineau: Does this need to be a static variables ?
I'm just wondering if we could generate this list dynamically from the rgw deployed. WDYT ?

alimaredia:
Closed for this one: https://github.com/ceph/ceph-ansible/pull/4524

PR #4462, opened 2019-09-24
Support running with --limit

mnasiadka:
Currently ceph-ansible doesn't support running plays with --limit,
nor does it support Ansible inventories with non-Ceph hosts included.
This is due to the fact that ceph-ansible delegates facts to 'all' group.

This change adds support for both of those cases.

Signed-off-by: Michal Nasiadka <mnasiadka@gmail.com>


guits:
@mnasiadka what about :
```bash
infrastructure-playbooks/rolling_update.yml:68:      with_items: "{{ groups['all'] }}"
infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml:46:      with_items: "{{ groups['all'] }}"
``` 
do you mind? :slightly_smiling_face:

guits:
@mnasiadka it is needed to run this task against *all* nodes present in the inventory, from what I'm understanding, with this change it would be run only with node targeted by the --limit

mnasiadka:
It's not correct - that variable contains all hosts that are taking part in that play.

As you can see here: https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html we could use two variables:
ansible_play_batch: List of active hosts in the current play run limited by the serial, aka ‘batch’. Failed/Unreachable hosts are not considered ‘active’.
ansible_play_hosts_all: List of all the hosts that were targeted by the play

In the fact delegation case I think the second one is more suitable.

I wrote a simple playbook to visualise this:
---
- hosts: all
  gather_facts: False
  tasks:
    - name: Print ansible_play_hosts_all
      debug:
        var: ansible_play_hosts_all
      run_once: true

And then when you run it without limit - you'll get all hosts from the inventory, and when with limit - only those that are matching limit:

(ceph-ansible)[stack@jumpbox01 ~]$ ansible-playbook -i ~/ion-openstack-poc/src/kayobe-config/etc/kayobe/inventory -e @~/ion-openstack-poc/src/kayobe-config/etc/kayobe/secrets.yml -e ansible_python_interpreter=/usr/bin/python test.yml

PLAY [all] ***************************************************************************************************************************************************

TASK [Print ansible_play_hosts_all] **************************************************************************************************************************
ok: [localhost] => {
    "ansible_play_hosts_all": [
        "localhost",
        "2ck10u19n04",
        "2ck10u19n03",
        "jumpbox01",
        "poc01-seed",
        "2ck11u32s01",
        "2ck09u38s01",
        "2ck11u12n01",
        "2ck11u12n03",
        "2ck11u12n07",
        "2ck11u12n02",
        "2ck11u12n05",
        "2ck11u12n08",
        "2ck11u12n06",
        "2ck10u01s01"
    ]
}

PLAY RECAP ***************************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0

(ceph-ansible)[stack@jumpbox01 ~]$ ansible-playbook -i ~/ion-openstack-poc/src/kayobe-config/etc/kayobe/inventory -e @~/ion-openstack-poc/src/kayobe-config/etc/kayobe/secrets.yml -e ansible_python_interpreter=/usr/bin/python --limit controllers test.yml

PLAY [all] ***************************************************************************************************************************************************

TASK [Print ansible_play_hosts_all] **************************************************************************************************************************
ok: [2ck10u19n03] => {
    "ansible_play_hosts_all": [
        "2ck10u19n03"
    ]
}

PLAY RECAP ***************************************************************************************************************************************************
2ck10u19n03                : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0

mnasiadka:
If you agree to untag DNM and go forward - I'll update the infrastructure-playbooks.

guits:
@mnasiadka in your second run with the `--limit` I expect to get all nodes

guits:
> It's not correct - that variable contains all hosts that are taking part in that play.

I'm not sure to understand, the second run you show here indicates what I said is correct, `ansible_play_hosts_all` + `--limit` only show nodes targeted by the `--limit`.

Either there's a misunderstanding or I'm missing something...

mnasiadka:
So that leaves you with a requirement to have only ceph-ansible related hosts in the inventory.
I just wanted to run --limit with [hosts_that_I_want_ceph_ansible_to_be_run_on], not just limit to one host.
Maybe a better idea (to support both cases), is to delegate to *_group_name? e.g. mon_group_name, osd_group_name, etc? I'll test that in some free time.

guits:
> So that leaves you with a requirement to have only ceph-ansible related hosts in the inventory.

That is correct.

guits:
> Maybe a better idea (to support both cases), is to delegate to *_group_name? e.g. mon_group_name, osd_group_name, etc? I'll test that in some free time.

we are dropping *_group_name because it doesn't work as expected, there are a couple of specific cases where it can't work. Eg: having a variable in `hosts:` section doesn't work.

mnasiadka:
> > So that leaves you with a requirement to have only ceph-ansible related hosts in the inventory.
> 
> That is correct.

So then I'm abandoning that PR - but it would be good to put a note in the docs that it's a requirement to not have any other hosts in the inventory.

guits:
@mnasiadka By the way, if I remember correctly, this is why we did it this way here: https://github.com/ceph/ceph-ansible/blob/master/site.yml.sample#L4-L16

mnasiadka:
Yeah, I understand that - another solution could be to generate site.yml - and still everybody can have their own version of site.yml, so that's not really bad. But if you count on user experience and using ,,stock'' site.yml - then yeah, it's a problem.

guits:
@mnasiadka site.yml.sample and site-container.yml.sample are, as indicated by the name, sample files, so it's intended to be modified indeed.
However, I agree that from user experience point of view it's nice to have something "ready to use" so it is still worth it to submit a patch for this.

mnasiadka:
Well, if you want to remove *_group_name - then I might use the default group names still and delegate to them - to allow proper facts gathering and delegation.
From another perspective - I'm now working on integrating ceph-ansible with kolla-ansible (an OpenStack deployer) - and not supporting *_group_name at all will complicate the task a bit (we use ceph-mon, ceph-osd, ceph-mgr names instead of the default ones).

PR #4470, opened 2019-09-25
ceph-osd: handle loop devices with containers

dsavineau:
Since we change the way to run the OSD containers with the ID instead
of the device name, we lost the ability to use loop devices.
Loop devices are like nvme or cciss devices because the partitions are
referenced with an extra 'p' before the partition number.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1749097


dsavineau:
jenkins test luminous-centos-container-lvm_osds

dsavineau:
jenkins test luminous-centos-non_container-lvm_osds

Issue #4453, opened 2019-09-18
"set grafana url", "set alertmanager host" task is wrong, "set prometheus host" is missing


radut:
**Bug Report**

What happened:

These tasks currently have an ending '/' : https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-dashboard/tasks/configure_dashboard.yml

```
- name: set grafana url
  command: "{{ container_exec_cmd }} ceph dashboard set-grafana-api-url {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ grafana_port }}/"
  delegate_to: "{{ groups[mon_group_name][0] }}"
  run_once: true
  changed_when: false

- name: set alertmanager host
  command: "{{ container_exec_cmd }} ceph dashboard set-alertmanager-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ alertmanager_port }}/"
  delegate_to: "{{ groups[mon_group_name][0] }}"
  run_once: true
  changed_when: false
```
Also there is no "set prometheus host" task

As the documentation states those do not need an '/'. 
**I encountered some 404 / and unnecessary redirects from '//..' to '/..'**
 link https://docs.ceph.com/docs/nautilus/mgr/dashboard/#enabling-the-embedding-of-grafana-dashboards, https://docs.ceph.com/docs/master/mgr/dashboard/#enabling-prometheus-alerting

```
To use it, specify the host and port of the Alertmanager server:

$ ceph dashboard set-alertmanager-api-host <alertmanager-host:port>  # default: ''
For example:

$ ceph dashboard set-alertmanager-api-host 'http://localhost:9093'
To be able to show what a silence will match beforehand, you have to add the host and port of the Prometheus server:

$ ceph dashboard set-prometheus-api-host <prometheus-host:port>  # default: ''
For example:

$ ceph dashboard set-prometheus-api-host 'http://localhost:9090'
```

**Environment**:

* ceph-ansible version (e.g. `git head or tag or stable branch`):
commit 5b1c15653fcb4772f0839f3a57f7e36ba1b86f49 (HEAD -> master, origin/master, origin/HEAD)
Author: Dimitri Savineau <dsavinea@redhat.com>
Date:   Mon Sep 9 11:23:47 2019 -0400

    ceph-handler: Fix osd restart condition

PR #4479, opened 2019-09-26
ceph-dashboard: Add prometheus api host

dsavineau:
The set-prometheus-api-host ceph dashboard subcommand was missing in
ceph-dashboard role. Only grafana and alermanager were present.
This commit also remove the trailing slash at the end of the host/url
values.

Closes: #4453


dsavineau:
@fmount FYI --^

fmount:
@dsavineau @guits I tried setting/unsetting this parameter in my old PR and I didn't put it because I've seen no difference (I suppose this happens because the module uses the rest API exposed by grafana), but at least it doesn't hurt.

PR #4484, opened 2019-09-27
dashboard: fix downstream deployment of the dashboard

b-ranto:
We are hitting several issues when deploying downstream (and with downstream images). We need to change the floating tags for the ose containers since these changed when we switched from unauthenticated registry to registry.redhat.io.

The downstream grafana container is also a lot less user-friendly and you need to handle all the options and plugin installation manually instead. We also need to make sure the container is run with uid 472 and that all the configurations paths are correct.

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -759,7 +759,7 @@ dummy:
 #prometheus_data_dir: /var/lib/prometheus
 #prometheus_conf_dir: /etc/prometheus
 #prometheus_user_id: '65534'  # This is the UID used by the prom/prometheus container image
-#prometheus_port: 9090
+#prometheus_port: 9092
 #alertmanager_container_image: prom/alertmanager:latest
 #alertmanager_container_cpu_period: 100000
 #alertmanager_container_cpu_cores: 2
--- a/group_vars/rhcs.yml.sample
+++ b/group_vars/rhcs.yml.sample
@@ -759,7 +759,7 @@ prometheus_container_image: registry.redhat.io/openshift4/ose-prometheus:4.1
 #prometheus_data_dir: /var/lib/prometheus
 #prometheus_conf_dir: /etc/prometheus
 #prometheus_user_id: '65534'  # This is the UID used by the prom/prometheus container image
-#prometheus_port: 9090
+#prometheus_port: 9092
 alertmanager_container_image: registry.redhat.io/openshift4/ose-prometheus-alertmanager:4.1
 #alertmanager_container_cpu_period: 100000
 #alertmanager_container_cpu_cores: 2
--- a/group_vars/rhcs.yml.sample
+++ b/group_vars/rhcs.yml.sample
@@ -725,7 +725,7 @@ node_exporter_container_image: registry.redhat.io/openshift4/ose-prometheus-node
 # We only need this for SSL (https) connections
 #grafana_crt: ''
 #grafana_key: ''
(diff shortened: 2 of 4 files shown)


roles/ceph-grafana/tasks/setup_container.yml
```diff
@@ -9,6 +9,24 @@
   with_items:
     - /etc/grafana
     - /var/lib/grafana
+    - /var/lib/grafana/plugins
+
+- name: Install grafana plugins
```
guits:
```suggestion
- name: install grafana plugins
```

roles/ceph-grafana/tasks/setup_container.yml
```diff
@@ -9,6 +9,24 @@
...
+  shell: |
+    wget -nv "https://grafana.com/api/plugins/{{ item }}/versions/latest/download" -O "/tmp/{{ item }}.zip"
```
guits: you might want to use unarchive module ?
https://docs.ansible.com/ansible/latest/modules/unarchive_module.html

roles/ceph-grafana/tasks/setup_container.yml
```diff
@@ -9,6 +9,24 @@
...
+    TMPDIR=$(mktemp -d)
+    unzip -q "/tmp/{{ item }}.zip" -d $TMPDIR/
+    mv $TMPDIR/* "/var/lib/grafana/plugins/{{ item }}"
+    rm -rf $TMPDIR "/tmp/{{ item }}.zip"
+  with_items: "{{ grafana_plugins }}"
+
+- name: Fix grafana plugins owner/group
```
guits:
```suggestion
- name: fix grafana plugins owner/group
```

roles/ceph-grafana/tasks/setup_container.yml
```diff
@@ -9,6 +9,24 @@
   with_items:
     - /etc/grafana
     - /var/lib/grafana
+    - /var/lib/grafana/plugins
+
+- name: Install grafana plugins
+  shell: |
+    wget -nv "https://grafana.com/api/plugins/{{ item }}/versions/latest/download" -O "/tmp/{{ item }}.zip"
+    TMPDIR=$(mktemp -d)
```
guits: what about tempfile module ?
https://docs.ansible.com/ansible/latest/modules/tempfile_module.html

dsavineau:
@b-ranto not really related to the PR but we're pinning the grafana container tag to OCS 4.1 because it's using grafana 5.x
For upstream, we're using docker.io/grafana/grafana:latest which uses 6.x. Should we explicitly set the upstream tag to use the 5.x release ?

b-ranto:
We have discussed this on irc with Guillaume and we have decided to use the grafana container from rhcs 3 at least for now, instead (re-pushed). It is more feature complete (aligns with upstream container) and it is the simplest way to enable QE to test.

btw: I have also added a commit that fixes a port clash between prometheus and cockpit in rhel 8.

@dsavineau It would probably be a good idea to do that for upstream as well. The version that got the most testing would probably be v5.2.4 (the same as the grafana container in rhcs 3).

fmount:
@guits @dsavineau I'm aware ceph-ansible provides parameters to configure ports for grafana/prometheus/alertmanager and all the dashboard related services, for this reason I added this [1] to enforce specific ports and don't rely on ceph-ansible defaults: this should avoid breaking firewall rules on TripleO.


[1] https://review.opendev.org/#/c/685465/

PR #4500, opened 2019-09-30
ceph-config: don't notify all handlers

dsavineau:
Because ceph-config role is used to by multiple plays (mons, osds,
etc...) we need to notify the handlers when the ceph configuration
change.
Currently we are notifying all handlers : mons, osds, mdss rgws, rgws,
mgrs and rbdmirrors. The handler code isn't executed because we're
using some condition on the group_name but the ansible handler
notification process could be slow when evaluation the tasks.
To avoid this we could notify only the handler related to the ansible
group name (like mons).

Before:
```console
NOTIFIED HANDLER ceph-handler : mons handler for osd-45
NOTIFIED HANDLER ceph-handler : osds handler for osd-45
NOTIFIED HANDLER ceph-handler : mdss handler for osd-45
NOTIFIED HANDLER ceph-handler : rgws handler for osd-45
NOTIFIED HANDLER ceph-handler : mgrs handler for osd-45
NOTIFIED HANDLER ceph-handler : rbdmirrors handler for osd-45
NOTIFIED HANDLER ceph-handler : rbd-target-api and rbd-target-gw handler for osd-45
```

After:
```console
NOTIFIED HANDLER ceph-handler : restart ceph osds for osd-45
```
If multiple daemons are collocated on the same host then we will notify
each handler
```console
NOTIFIED HANDLER ceph-handler : restart ceph mgrs for mon-2
NOTIFIED HANDLER ceph-handler : restart ceph mons for mon-2
```
This patch removes the handler listen because we now have only one task
to notify and because the listen doens't support variables we can 
fallback to the task name which does.

### DIFF
--- a/roles/ceph-config/tasks/main.yml
+++ b/roles/ceph-config/tasks/main.yml
@@ -62,6 +62,14 @@
       - devices | default([]) | length > 0
       - not (lvm_batch_report.stdout | from_json).changed
 
+- name: generate the ceph notify list
+  set_fact:
+    ceph_notify: "{{ (ceph_notify | default([]) + ['restart ceph ' + item]) | unique }}"
+  with_items: "{{ group_names }}"
+  when:
+    - not rolling_update | bool
+    - item not in [client_group_name, grafana_server_group_name, rgwloadbalancer_group_name]
+
 # ceph-common
 - name: config file operation for non-containerized scenarios
   when: not containerized_deployment | bool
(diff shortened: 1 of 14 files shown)


dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

roles/ceph-config/tasks/main.yml
```diff
@@ -142,10 +142,4 @@
...
       config_overrides: "{{ ceph_conf_overrides }}"
       config_type: ini
-    notify:
-      - restart ceph mons
-      - restart ceph osds
-      - restart ceph mdss
-      - restart ceph rgws
-      - restart ceph mgrs
-      - restart ceph rbdmirrors
+    notify: "{{ ceph_notify | default(omit) }}"
```
guits: not sure `omit` can work in a `notify` ?

guits: according to the ci it seems to work

dsavineau: héhé `omit` FTW :1st_place_medal:

guits:

dsavineau:
@guits I need to take a look to the update playbook because there's the new active_mds and standby_mds groups that breaks the feature.

PR #4504, opened 2019-10-01
Set proper ownership command performance improvement

guits:
By changing the set ownership command from using the file module in combination with a with_items loop to a raw chown command, we can achieve a 98% performance increase here.

On a ceph cluster with a significant amount of directories and files in /var/lib/ceph, the file module has to run checks on ownership of all those directories and files to determine whether a change is needed.

In this case, we just want to explicitly set the ownership of all these directories and files to the ceph_uid

Added context note to all set proper ownership tasks

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1757400

Signed-off-by: Kevin Jones <kevinjones@redhat.com>
(cherry picked from commit 47bf47c9d87fe057bc1402f7a1aa0a84d13b5fd0)


guits:
jenkins test luminous-centos-non_container-switch_to_containers

Issue #4492, opened 2019-09-28
''osd_memory_target'' in section ''osd'' redefined Error during FSID fetch


KingJ:
**Bug Report**

What happened: When running a rolling update, ceph-facts : get current fsid task fails as the executed command returns a non-zero return code, with the stderr containing a warning that the osd_memory_target setting has been redefined.

`stderr: 'Can''t get admin socket path: unable to get conf option admin_socket for mon.host: warning: line 17: ''osd_memory_target'' in section ''osd'' redefined '`

I suspect may be due to the `bluestore_cache_autotune` setting in Ceph. ceph-ansible sets the `osd memory target` setting (with spaces rather than understores) [here](https://github.com/ceph/ceph-ansible/blob/v4.0.0rc16/roles/ceph-config/templates/ceph.conf.j2#L106). However, ceph itself appears to set the `osd_memory_target` setting instead if `bluestore_cache_autotune` is enabled, which is the default setting ([ref](https://docs.ceph.com/docs/nautilus/rados/configuration/bluestore-config-ref/#automatic-cache-sizing)). As a result, the same variable is defined twice - once with spaces, once with underscores. This does not appear to affect regular cluster functionality, however it does cause some commands (such as the one to retrieve the fsid) to output a warning to stderr and a return code of 22.

What you expected to happen: The rolling update playbook to function as normal, including correct retrieval of the FSID.

How to reproduce it (minimal and precise):
Create a `ceph.conf` file containing a single setting defined both using underscores and spaces, for example;

```
[osd]
bluestore compression algorithm = lz4
bluestore compression mode = aggressive
osd max backfills = 8
osd memory target = 4294967296
osd recovery max active = 8

osd_memory_target = 4242538496
osd_memory_base = 2147483648
osd_memory_cache_min = 3195011072
```

Run the rolling update playbook. The ceph-facts : get current fsid task will fail with the task's stderr containing the message `Can''t get admin socket path: unable to get conf option admin_socket for mon.host: warning: line 17: ''osd_memory_target'' in section ''osd'' redefined`.

Remove one of the identical configuration entries and re-run the playbook. The task will no longer fail, however the value will be silently added back.

**Environment**:
* OS (e.g. from /etc/os-release): Ubuntu 18.04.3
[...]

PR #4507, opened 2019-10-01
ceph-facts: use --admin-daemon to get fsid

dsavineau:
During the rolling_update scenario, the fsid value is retrieve from the
current ceph cluster configuration via the ceph daemon config command.
This command tries first to resolve the admin socket path via the
ceph-conf command.
Unfortunately this command won't work if you have a duplicate key in the
ceph configuration even if it only produces a warning. As a result the
task will fail.

Can't get admin socket path: unable to get conf option admin_socket for
mon.xxx: warning: line 13: 'osd_memory_target' in section 'osd' redefined

Instead of using ceph daemon we can use the --admin-daemon option
because we already know what the socket admin path value based on the
ceph cluster and mon hostname values.

Closes: #4492

### DIFF
--- a/roles/ceph-facts/tasks/facts.yml
+++ b/roles/ceph-facts/tasks/facts.yml
@@ -84,7 +84,7 @@
   when: cephx | bool or generate_fsid | bool
 
 - name: get current fsid
-  command: "{{ timeout_command }} {{ container_exec_cmd }} ceph --cluster {{ cluster }} daemon mon.{{ hostvars[mon_host | default(groups[mon_group_name][0])]['ansible_hostname'] }} config get fsid"
+  command: "{{ timeout_command }} {{ container_exec_cmd }} ceph --admin-daemon /var/run/ceph/{{ cluster }}-mon.{{ hostvars[mon_host | default(groups[mon_group_name][0])]['ansible_hostname'] }}.asok config get fsid"
   register: rolling_update_fsid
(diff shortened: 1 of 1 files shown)


dsavineau:

dsavineau:

roles/ceph-facts/tasks/facts.yml
```diff
@@ -84,7 +84,7 @@
   when: cephx | bool or generate_fsid | bool
 
 - name: get current fsid
-  command: "{{ timeout_command }} {{ container_exec_cmd }} ceph --cluster {{ cluster }} daemon mon.{{ hostvars[mon_host | default(groups[mon_group_name][0])]['ansible_hostname'] }} config get fsid"
+  command: "{{ timeout_command }} {{ container_exec_cmd }} ceph --admin-daemon /var/run/ceph/{{ cluster }}-mon.{{ hostvars[mon_host | default(groups[mon_group_name][0])]['ansible_hostname'] }}.asok config get fsid"
```
guits: `/var/run` is a symlink to `/run`

```suggestion
  command: "{{ timeout_command }} {{ container_exec_cmd }} ceph --admin-daemon /run/ceph/{{ cluster }}-mon.{{ hostvars[mon_host | default(groups[mon_group_name][0])]['ansible_hostname'] }}.asok config get fsid"
```

PR #4510, opened 2019-10-02
Vagrantfile: support more than 9 nodes per daemon type

guits:
because of the current ip address assignation, it's not possible to
deploy more than 9 nodes per daemon type.
This commit refact a bit and allows us to get around this limitation.


guits:

PR #4515, opened 2019-10-02
ceph-handler: don't restart all OSDs with limit

dsavineau:
When using the ansible --limit option on one or few OSD nodes and if the
handler is triggered then we will restart the OSD service on all OSDs
nodes instead of the hosts limited by the limit value.
Even if the play is limited by the --limit value we are using all OSD
nodes from the OSD group.

```yaml
  with_items: '{{ groups[osd_group_name] }}'
```

Instead we should iterate only on the nodes present in both OSD group and
limit list.


dsavineau:
unrelated CI failure

PR #4516, opened 2019-10-02
dashboard: remove useless block section

dsavineau:
The block section were used with the dashboard_enabled condition when
the code was included in the main playbooks.
Because this condition isn't present in the dashboard playbook anymore
we can remove the block section.


guits:
dev freeze is passed and we released v4.0.0, not sure it's worth to backport this one @dsavineau

PR #4535, opened 2019-10-04
Fixes 'ssh: host unreachable' error in Copy ceph key task in ceph-mgr role

ashutoshgngwr:
## Context

In 'Copy ceph key(s) task, if 'mgr' instance was given an IP in host inventory,
ceph ansible was passing it as `ip-10-20-30-40` to `delegate_to` field in the task which results
in 'ssh: host unreachable' error. This is one of the possible fixes!


ashutoshgngwr:
Turns out this is an AWS specific thing.

PR #4524, opened 2019-10-03
rgw multisite secondary endpoint checks

alimaredia:
Add two checks in secondary.yml:

1) a check that ensures a connection to the
endpoint the realm is being fetched from.

2) a check that ensures connections from the
host rgw to all the rgws endpoints in a zone and
all the rgw endpoints for all the zonegroups
in a realm.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1731158

Signed-off-by: Ali Maredia <amaredia@redhat.com>

PR #4545, opened 2019-10-04
rgw multisite: enable more than 1 realm per cluster

alimaredia:


roles/ceph-rgw/tasks/multisite/secondary.yml
```diff
@@ -18,21 +18,19 @@
...
+  command: "{{ container_exec_cmd }} radosgw-admin zonegroup default --rgw-realm={{ rgw_realm }} --rgw-zonegroup={{ rgw_zonegroup }}"
   changed_when: false
   delegate_to: "{{ groups[mon_group_name][0] }}"
   run_once: true
 
 - name: create the zone
-  command: "{{ container_exec_cmd }} radosgw-admin zone create --rgw-zonegroup={{ rgw_zonegroup }} --rgw-zone={{ rgw_zone }} --endpoints={{ rgw_multisite_proto }}://{{ rgw_multisite_endpoint_addr }}:{{ radosgw_frontend_port }} --access-key={{ system_access_key }} --secret={{ system_secret_key }} --default"
+  command: "{{ container_exec_cmd }} radosgw-admin zone create --rgw-realm={{ rgw_realm }} --rgw-zonegroup={{ rgw_zonegroup }} --rgw-zone={{ rgw_zone }} --endpoints={{ rgw_multisite_proto }}://{{ rgw_multisite_endpoint_addr }}:{{ radosgw_frontend_port }} --access-key={{ system_access_key }} --secret={{ system_secret_key }} --default"
   delegate_to: "{{ groups[mon_group_name][0] }}"
-  run_once: true
```
dsavineau: What's the reason to remove the `run_once: true`

roles/ceph-rgw/tasks/multisite/secondary.yml
```diff
@@ -18,21 +18,19 @@
...
   when: "'No such file or directory' in zonecheck.stderr"
   notify: update period
 
 - name: add other endpoints to the zone
-  command: "{{ container_exec_cmd }} radosgw-admin zone modify --rgw-zone={{ rgw_zone }} --endpoints {{ rgw_multisite_endpoints_list }}"
+  command: "{{ container_exec_cmd }} radosgw-admin zone modify --rgw-realm={{ rgw_realm }} --rgw-zonegroup={{ rgw_zonegroup }} --rgw-zone={{ rgw_zone }} --endpoints {{ rgw_multisite_endpoints_list }}"
   delegate_to: "{{ groups[mon_group_name][0] }}"
-  run_once: true
```
dsavineau: ditto

dsavineau:
jenkins test dev-centos-non_container-rgw_multisite

dsavineau:
jenkins test dev-centos-container-rgw_multisite

roles/ceph-rgw/tasks/multisite/checks.yml
```diff
@@ -8,15 +8,15 @@
...
+  command: "{{ container_exec_cmd }} radosgw-admin zonegroup get --rgw-realm={{ rgw_realm }} --rgw-zonegroup={{ rgw_zonegroup }}"
   delegate_to: "{{ groups[mon_group_name][0] }}"
   register: zonegroupcheck
   failed_when: False
   changed_when: False
   check_mode: no
 
 - name: check if the zone already exists
-  command: "{{ container_exec_cmd }} radosgw-admin zone get --rgw-zone={{ rgw_zone }}"
+  command: "{{ container_exec_cmd }} radosgw-admin zone get --rgw-realm={{ rgw_realm }} --rgw-zonegreoup={{ rgw_zonegreoup }} --rgw-zone={{ rgw_zone }}"
```
dsavineau: typo s/zonegreoup/zonegroup/g

dsavineau:
@alimaredia could you resolve the conflicts ? otherwise we won't be able to run the rgw multisite jobs in the CI

alimaredia:

tests/functional/rgw-multisite/container/group_vars/rgws
```diff
@@ -4,10 +4,14 @@ copy_admin_key: true
...
 rgw_zone: jupiter
 rgw_zonemaster: true
+rgw_zonedefault: true
 rgw_zonesecondary: false
 rgw_multisite_proto: http
 rgw_zonegroup: solarsystem
+rgw_zonegroupmaster=True 
+rgw_zonegroupdefault=True 
 rgw_zone_user: zone.user
+rgw_zone_user_disply_name="Zone User"
```
dsavineau: You should replace `=` by `:`

tests/functional/rgw-multisite/group_vars/rgws
```diff
@@ -4,10 +4,14 @@ copy_admin_key: true
...
 rgw_zone: jupiter
 rgw_zonemaster: true
+rgw_zonedefault: true
 rgw_zonesecondary: false
 rgw_multisite_proto: http
 rgw_zonegroup: solarsystem
+rgw_zonegroupmaster=True 
+rgw_zonegroupdefault=True 
 rgw_zone_user: zone.user
+rgw_zone_user_disply_name="Zone User"
```
dsavineau: ditto

dsavineau:

dsavineau:

dsavineau:
@alimaredia could you fix the conflict ?

roles/ceph-rgw/tasks/multisite/create_realm_system_user.yml
```diff
@@ -0,0 +1,29 @@
...
+  delegate_to: "{{ groups[mon_group_name][0] }}"
+  register: usercheck
+  failed_when: False
+  changed_when: False
+  check_mode: no
+
+- name: create list realm_system_users
+  set_fact:
+    realm_system_users: "{{ realm_system_users | default([]) + [{ 'realm': hostvars[item]['rgw_realm'], 'zonegroup': hostvars[item]['rgw_zonegroup'], 'zone': hostvars[item]['rgw_zone'], 'access_key': hostvars[item]['system_access_key'], 'secret_key': hostvars[item]['system_secret_key'], 'user': hostvars[item]['rgw_realm_system_user'], 'display_name': hostvars[item]['rgw_realm_system_user_display_name'] }] }}"
+  with_items: "{{ groups.get('rgws', []) }}"
```
dsavineau: rgws group is hardcoded

roles/ceph-rgw/tasks/multisite/create_realm_system_user.yml
```diff
@@ -0,0 +1,29 @@
...
+  run_once: true
+  when: 
+    - hostvars[item]['rgw_zonemaster'] == true
```
dsavineau: Don't use string comparison with boolean.
Use `| bool` instead (or `not ... | bool`)

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
+- name: create list realms
+  set_fact:
+    realms: "{{ realms | default([]) + [{ 'realm': hostvars[item]['rgw_realm'] }] }}"
+  with_items: "{{ groups.get('rgws', []) }}"
```
dsavineau: rgw group hardcoded

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
...
+- name: make all items in realms unique
+  set_fact:
+    realms: "{{ realms | unique }}"
+  run_once: true
+  when: realms is defined
+
+- name: create list secondary_realms
+  set_fact:
+    secondary_realms: "{{ secondary_realms | default([]) + [{ 'realm': hostvars[item]['rgw_realm'], 'is_master': hostvars[item]['rgw_zonemaster'], 'endpoint': hostvars[item]['rgw_pull_proto'] + '://' + hostvars[item]['rgw_pullhost'] + ':' + hostvars[item]['rgw_pull_port']|string, 'access_key': hostvars[item]['system_access_key'], 'secret_key': hostvars[item]['system_secret_key'] }] }}"
+  with_items: "{{ groups.get('rgws', []) }}"
```
dsavineau: ditto

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
...
+  run_once: true
+  when: 
+    - hostvars[item]['rgw_zonemaster'] == false
```
dsavineau: same about string vs boolean

roles/ceph-rgw/tasks/multisite/main.yml
```diff
@@ -2,27 +2,57 @@
...
+  when: rgw_instances is defined
+
+- name: add zonegroup to rgw stanza in ceph.conf
+  ini_file:
+    dest: "/etc/ceph/{{ cluster }}.conf"
+    section: "client.rgw.{{ ansible_hostname }}.{{ item.instance_name }}"
+    option: "rgw_zone"
+    value: "{{ rgw_zone }}"
+  with_items: "{{ rgw_instances }}"
+  when: rgw_instances is defined
```
dsavineau: I'm not sure to understand why you're using the `ini_file` module here instead of modify the ceph.conf template instead.

This means that the configuration will be erase by the ceph-config role and then restore during the multisite tasks. So this will trigger the rgw handlers for nothing.

roles/ceph-rgw/tasks/multisite/main.yml
```diff
@@ -2,27 +2,57 @@
...
-    option: "rgw_zone"
-    value: "{{ rgw_zone }}"
-  with_items: "{{ rgw_instances }}"
+- name: include_tasks start_radosgw.yml for zonesecondary rgws
+  include_tasks: ../start_radosgw.yml
   when:
-    - rgw_instances is defined
-  notify: restart rgw
+    - not rgw_zonemaster | bool
+    - rgw_zonesecondary | bool
```
dsavineau: Why are you including the start_radosgw.yml file again for both master and secondary task when this should have been done during the "classic" ceph-rgw role ?

dsavineau:

dsavineau:

roles/ceph-rgw/tasks/multisite/create_realm_system_user.yml
alimaredia: so it would be `hostvars[item]['rgw_zonemaster`]|bool == true`?

roles/ceph-rgw/tasks/multisite/main.yml
alimaredia: I'll go ahead and do that. This was the way it was before just in a different part of the file.

roles/ceph-rgw/tasks/multisite/main.yml
alimaredia: So @cbodley pointed out to be that no rgws should be turned on until after the period has been updated. I've done that for both the secondary and master zone rgws now. Also the classic ceph-rgw role still applies because that call to start_radosgw.yml applies when `rgw_multisite: false`.

roles/ceph-rgw/tasks/multisite/secondary.yml
alimaredia: I'm not sure I follow. I still see it here on the line right above you comment.

roles/ceph-rgw/tasks/multisite/create_realm_system_user.yml
dsavineau: no just `hostvars[item]['rgw_zonemaster'] | bool`

alimaredia:
@dsavineau The reason I wanted the realm.yml and zone.yml sample files is just so that the user could easily copy those files into new_realm_name.yml and new_zone_name.yml files and just change the variables inside them. I also moved all of the multisite variables out of all.yml because I'd like to have deploying with those two files be the preferred way of deploying multi-site. 

I could re-add the variables back into all.yml if needed.

As for backporting, the commit that removes rgw_instances, I could move that into a separate PR and discuss why it's necessary there.

roles/ceph-rgw/tasks/multisite/secondary.yml
dsavineau: I think it's coming from an old version of the PR because this was true one month and a half ago

guits:
@dsavineau @alimaredia 

I see most of the comments were addressed, however, there are still a couple of questions:

regarding the rgw num instances removal, that's indeed not elegant to remove that like this, I tend to be -1 on that, we can't remove feature this way without marking them as deprecated upstream.

> So @cbodley pointed out to be that no rgws should be turned on until after the period has been updated. I've done that for both the secondary and master zone rgws now. Also the classic ceph-rgw role still applies because that call to start_radosgw.yml applies when rgw_multisite: false.

@dsavineau what do you think on the above?

> If I kept zone.yml.sample and realm.yml.sample, could I have templates for the in ceph-defaults or ceph-rgw and make changes to generate_group_vars.sh?

I would say yes. @dsavineau what do you think?

one last thing, I would probably squash some commits.

dsavineau:
> @dsavineau @alimaredia
> 
> I see most of the comments were addressed, however, there are still a couple of questions:
> 
> regarding the rgw num instances removal, that's indeed not elegant to remove that like this, I tend to be -1 on that, we can't remove feature this way without marking them as deprecated upstream.

+1 also it will mean that we won't be able to backport this into stable branches otherwise we will break backward compatibility.

@alimaredia  Have you consider to add support to radosgw multi instances with realm, zonegoup and zone instead of using a dirty hack (I don't think the implementation follows the ansible philosophy) like using the same host multiple time in the rgw group inventory.

> > So @cbodley pointed out to be that no rgws should be turned on until after the period has been updated. I've done that for both the secondary and master zone rgws now. Also the classic ceph-rgw role still applies because that call to start_radosgw.yml applies when rgw_multisite: false.
> 
> @dsavineau what do you think on the above?

This looks great.

> > If I kept zone.yml.sample and realm.yml.sample, could I have templates for the in ceph-defaults or ceph-rgw and make changes to generate_group_vars.sh?
> 
> I would say yes. @dsavineau what do you think?

We could do that but I'm not a big fan of the  current realm/zone implementation.

@alimaredia What's the relationship between this PR and #4524 ? Because I can see the same code than #4524 in that one ?

https://github.com/ceph/ceph-ansible/pull/4545/commits/4a27ef4ec8152836b657b8208e663b9fbbfbd2a9 vs https://github.com/ceph/ceph-ansible/pull/4524/commits/612ea2f97648c44c1b901b04f55baea93cae8b38

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:
@alimaredia do we need to validate `rgw_realm_system_user_display_name` on the primary zone too ?
Because right now it's only available on the secondary site in the tests scenario.

dsavineau:
> @alimaredia do we need to validate `rgw_realm_system_user_display_name` on the primary zone too ?
> Because right now it's only available on the secondary site in the tests scenario.

Answering to myself, there's a typo in the rgw group_vars on the primary cluster

```yaml
-rgw_realm_system_user_disply_name: "Zone User"
+rgw_realm_system_user_display_name: "Zone User"
```

dsavineau:

dsavineau:

dsavineau:

dsavineau:

roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -129,6 +129,11 @@ rgw frontends = {{ frontend_line(radosgw_frontend_type) }} {{ radosgw_frontend_o
 {% if 'num_threads' not in radosgw_frontend_options %}
 rgw thread pool size = {{ radosgw_thread_pool_size }}
 {% endif %}
+{% if rgw_multisite %}
```
dsavineau: we should consider adding the bool filter here

roles/ceph-facts/tasks/facts.yml
```diff
@@ -279,8 +279,7 @@
 
 - name: set_fact rgw_instances
   set_fact:
-    rgw_instances: "{{ rgw_instances|default([]) | union([{'instance_name': 'rgw' + item|string, 'radosgw_address': _radosgw_address, 'radosgw_frontend_port': radosgw_frontend_port|int + item|int}]) }}"
-  with_sequence: start=0 end={{ radosgw_num_instances|int - 1 }}
+    rgw_instances: "{{ rgw_instances|default([]) | union([{ 'instance_name': inventory_hostname, 'radosgw_address': _radosgw_address, 'radosgw_frontend_port': radosgw_frontend_port, 'rgw_realm': rgw_realm, 'rgw_zonegroup': rgw_zonegroup, 'rgw_zone': rgw_zone }]) }}"
```
dsavineau: there's no sequence lookup here so you need to change the way the list is built

roles/ceph-handler/templates/restart_rgw_daemon.sh.j2
```diff
@@ -3,51 +3,48 @@
 RETRIES="{{ handler_health_rgw_check_retries }}"
 DELAY="{{ handler_health_rgw_check_delay }}"
```
dsavineau: How is this suppose to work on when you're using the same host with multiple instance name ?

Because with the `radosgw_num_instances` we were able to loop over the list but now there will be only one rgw restart script for multiple rgw instance with different value.

At the end the last instance will override the restart script for all the other instances.

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
...
+
+- name: make all items in zones unique
+  set_fact:
+    zones: "{{ zones | unique }}"
+  run_once: true
+  when: zones is defined
+
+- name: create a list of dicts with each rgw endpoint and it's zone
+  set_fact:
+    zone_endpoint_pairs: "{{ zone_endpoint_pairs | default([]) + [{ 'endpoint': 'http://' + hostvars[item]['radosgw_address'] + ':' + hostvars[item]['radosgw_frontend_port']|string, 'rgw_zone': hostvars[item]['rgw_zone'] }] }}"
```
dsavineau: you should use `_radosgw_address` instead of `radosgw_address`.

`_radosgw_address` uses either `radosgw_address`, `radosgw_address_block` and r`adosgw_interface` variables.

guits:
> We could do that but I'm not a big fan of the current realm/zone implementation.

looks like this PRs isn't ready at all to be merged, we aren't all on the same page.
Perhaps we should all have a discussion on what are the expectations here, what should be done, the way it should be done.

@alimaredia @dsavineau please, let me know

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
+- name: create list realms
+  set_fact:
+    realms: "{{ realms | default([]) + [{ 'realm': hostvars[item]['rgw_realm'] }] }}"
+  with_items: "{{ groups.get(rgw_group_name, []) }}"
+  run_once: true
+  when: "'No such file or directory' in hostvars[item]['realmcheck'].stderr"
+
+- name: make all items in realms unique
```
guits: I think this task can be avoided

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
+- name: create list realms
+  set_fact:
+    realms: "{{ realms | default([]) + [{ 'realm': hostvars[item]['rgw_realm'] }] }}"
```
guits:
```suggestion
    realms: "{{ (realms | default([]) + [{ 'realm': hostvars[item]['rgw_realm'] }]) | unique }}"
```

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
...
+
+- name: create list secondary_realms
+  set_fact:
+    secondary_realms: "{{ secondary_realms | default([]) + [{ 'realm': hostvars[item]['rgw_realm'], 'is_master': hostvars[item]['rgw_zonemaster'], 'endpoint': hostvars[item]['rgw_pull_proto'] + '://' + hostvars[item]['rgw_pullhost'] + ':' + hostvars[item]['rgw_pull_port']|string, 'access_key': hostvars[item]['system_access_key'], 'secret_key': hostvars[item]['system_secret_key'] }] }}"
+  with_items: "{{ groups.get(rgw_group_name, []) }}"
+  run_once: true
+  when:
+    - not hostvars[item]['rgw_zonemaster'] | bool
+
+- name: make all items in secondary_realms unique
```
guits: probably possible to avoid this task too

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
...
+- name: create list zonegroups
+  set_fact:
+    zonegroups: "{{ zonegroups | default([]) + [{ 'realm': hostvars[item]['rgw_realm'], 'zonegroup': hostvars[item]['rgw_zonegroup'], 'is_master': hostvars[item]['rgw_zonegroupmaster'] }] }}"
+  with_items: "{{ groups.get(rgw_group_name, []) }}"
+  run_once: true
+  when:
+    - hostvars[item]['rgw_zonemaster'] | bool
+    - "'No such file or directory' in hostvars[item]['zonegroupcheck'].stderr"
+
+- name: make all items in zonegroups unique
```
guits: ditto

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
...
+  when: zonegroups is defined
+
+- name: create list zones
+  set_fact:
+    zones: "{{ zones | default([]) + [{ 'realm': hostvars[item]['rgw_realm'], 'zonegroup': hostvars[item]['rgw_zonegroup'], 'zone': hostvars[item]['rgw_zone'], 'is_master': hostvars[item]['rgw_zonemaster'], 'access_key': hostvars[item]['system_access_key'], 'secret_key': hostvars[item]['system_secret_key'] }] }}"
+  with_items: "{{ groups.get(rgw_group_name, []) }}"
+  run_once: true
+  when: "'No such file or directory' in hostvars[item]['zonecheck'].stderr"
+
+- name: make all items in zones unique
```
guits: ditto

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,86 @@
...
+    - endpoints_string[-1] == ','
+
+- name: create a list of zones and all their endpoints
+  set_fact:
+    zone_endpoints_list: "{{ zone_endpoints_list | default([]) + [{ 'endpoints': hostvars[item]['zone_endpoints_string'], 'zone': hostvars[item]['rgw_zone'], 'zonegroup': hostvars[item]['rgw_zonegroup'], 'realm': hostvars[item]['rgw_realm'], 'is_master': hostvars[item]['rgw_zonemaster'] }] }}"
+  with_items: "{{ groups.get(rgw_group_name, []) }}"
+  run_once: true
+  when: hostvars[item]['zone_endpoints_string'] is defined
+
+- name: make all items in zone_endpoints_list unique
```
guits: ditto

alimaredia:
@guits @dsavineau: I removed the `radosgw_num_instances` commit and opened up https://github.com/ceph/ceph-ansible/pull/4858 for discussion on it. 

I think I've addressed all of your comments you had in your last round of review.

I think the only 2 things that need to be done are:
1. How should I go about adding `.yml.sample` files for the zone and realm configs? 
2. A test case with 1 ceph cluster but more than 1 realm started up on it.

guits:
@alimaredia depending on the discussion we should have about the multi rgw instances removal, this patch might need to be reworked.

alimaredia:
@guits can you explain why? I see the two PRs as being independent. Even if this PR didn't exist, multisite doesn't work as is when `radsogw_num_instances` is more than ` for a host.

The only thing multisite related in https://github.com/ceph/ceph-ansible/pull/4858 is adding realm, zonegroup, zone into `ceph.conf.j2` and `rgw_instances`.

alimaredia:
unstale

dsavineau:
@alimaredia Could you fix the conflict and rebase the PR ?

dsavineau:

dsavineau:

dsavineau:

dsavineau:

roles/ceph-rgw/tasks/multisite/create_realm_zonegroup_zone_lists.yml
```diff
@@ -0,0 +1,58 @@
...
+  set_fact:
+    zones: "{{ zones | default([]) + [{ 'realm': hostvars[item]['rgw_realm'], 'zonegroup': hostvars[item]['rgw_zonegroup'], 'zone': hostvars[item]['rgw_zone'], 'is_master': hostvars[item]['rgw_zonemaster'], 'access_key': hostvars[item]['system_access_key'], 'secret_key': hostvars[item]['system_secret_key'] }] | unique }}"
+  with_items: "{{ groups.get(rgw_group_name, []) }}"
+  run_once: true
+  when: "'No such file or directory' in hostvars[item]['zonecheck'].stderr"
+
+- name: create a list of dicts with each rgw endpoint and it's zone
+  set_fact:
+    zone_endpoint_pairs: "{{ zone_endpoint_pairs | default([]) + [{ 'endpoint': hostvars[item]['rgw_multisite_proto'] + '://' + hostvars[item]['_radosgw_address'] + ':' + hostvars[item]['radosgw_frontend_port']|string, 'rgw_zone': hostvars[item]['rgw_zone'] }] }}"
+  with_items: "{{ groups.get(rgw_group_name, []) }}"
```
dsavineau: This doesn't support multi rgw instances ans `radosgw_frontend_port` is undefined as a hostvars

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

tests/functional/rgw-multisite/container/secondary/hosts
```diff
@@ -5,4 +5,4 @@ mon0
 osd0
 
 [rgws]
-osd0 copy_admin_key=True rgw_multisite=True rgw_zone=mars rgw_zonemaster=False rgw_zonesecondary=True rgw_zonegroup=solarsystem rgw_zone_user=zone.user rgw_realm=milkyway rgw_multisite_proto=http rgw_multisite_endpoint_addr=192.168.107.11 system_access_key=6kWkikvapSnHyE22P7nO system_secret_key=MGecsMrWtKZgngOHZdrd6d3JxGO5CPWgT2lcnpSt rgw_pull_proto=http rgw_pull_port=8080 rgw_pullhost=192.168.105.11
```
dsavineau: `rgw_multisite_endpoint_addr` variable is removed from the multisite tests so how are we supposed to determine the rgw endpoint to use for s3cmd commands ?

dsavineau:
@alimaredia any update on this ?

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

PR #4546, opened 2019-10-04
ceph-rgw: add cluster parameter to multisite tasks

dsavineau:
The rgw multisite tasks didn't use the cluster option so the
radosgw-admin commands would fail if the cluster name isn't the default
value.


dsavineau:
@alimaredia you should need to do the same on your current PR as well when this once will be merged

dsavineau:
jenkins test dev-centos-non_container-rgw_multisite

dsavineau:
jenkins test dev-centos-container-rgw_multisite

guits:
> @alimaredia you should need to do the same on your current PR as well when this once will be merged

can't we make a single PR to fix this everywhere instead?

dsavineau:
closing in flavor of #4688

PR #4551, opened 2019-10-07
nfs: stop nfs server service

guits:
stop nfs server service for containerized AND non containerized deployments.

(this is also needed for containerized deployment, there's a requirement where we must stop nfs service service on the node hosting the containers as well)

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1508506


guits:

PR #4560, opened 2019-10-07
switch_to_containers: umount osd lockbox partition

dsavineau:
When switching from a baremetal deployment to a containerized deployment
we only umount the OSD data partition.
If the OSD is encrypted (dmcrypt: true) then there's an additional
partition (part number 5) used for the lockbox and mount in the
/var/lib/ceph/osd-lockbox/ directory.
Because this partition isn't umount then the containerized OSD aren't
able to start. The partition is still mount by the system and can't be
remount from the container.


dsavineau:
jenkins test dev-centos-non_container-switch_to_containers

dsavineau:
jenkins test dev-centos-non_container-switch_to_containers

Issue #4553, opened 2019-10-07
ceph-mds get keys from monitor - gets them from non-existent mds


mnasiadka:
**Bug Report**

What happened:

ceph-mds get keys from monitor - gets them from non-existent mds

What you expected to happen:

ceph-mds to be deployed

How to reproduce it (minimal and precise):

Share your group_vars files, inventory

Ansible log:
https://zuul.opendev.org/t/openstack/build/3dca7f430c614e41bd6cffc283604895/log/primary/logs/ansible/deploy-ceph#112971

group_vars:
https://zuul.opendev.org/t/openstack/build/3dca7f430c614e41bd6cffc283604895/log/primary/logs/kolla_configs/group_vars/all/ceph-ansible

inventory:
https://zuul.opendev.org/t/openstack/build/3dca7f430c614e41bd6cffc283604895/log/primary/logs/kolla_configs/inventory

**Environment**: 
* OS (e.g. from /etc/os-release): CentOS 7
* Kernel (e.g. `uname -a`): 3.10.0-1062.1.2.el7.x86_64
* Docker version if applicable (e.g. `docker version`): 
* Ansible version (e.g. `ansible-playbook --version`): 2.8.5
* ceph-ansible version (e.g. `git head or tag or stable branch`): stable-4.0
* Ceph version (e.g. `ceph -v`):

PR #4564, opened 2019-10-08
common: do not reset `container_exec_cmd`

guits:
This commit removes some legacy tasks.

These tasks aren't needed, they cause the playbook to fail when
collocating daemons.

Closes: #4553


guits:
failures here are related to this patch ceph/ceph/pull/30603

PR #4566, opened 2019-10-08
validate: prevent from installing OSD on same disk as the OS

guits:
This commit adds a validation task to prevent from installing an OSD on
the same disk as the OS.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1623580

### DIFF
--- a/roles/ceph-validate/tasks/check_devices.yml
+++ b/roles/ceph-validate/tasks/check_devices.yml
@@ -1,4 +1,59 @@
 ---
+- name: find device used for operating system
+  command: findmnt -v -n -T / -o SOURCE
+  changed_when: false
+  register: root_device
+
+- name: resolve root_device
+  command: "readlink -f {{ root_device.stdout }}"
+  changed_when: false
+  register: _root_device
+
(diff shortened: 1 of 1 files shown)


roles/ceph-validate/tasks/check_devices.yml
```diff
@@ -1,4 +1,26 @@
 ---
+- name: find device used for operating system
+  command: findmnt -n -T / -o SOURCE
```
dsavineau: Just wondering if the could just use the `ansible_mounts` fact there. wdyt ?

guits: that is possible indeed, but we would have to resolve the device returned by the fact by using `readlink -f` anyway

guits: using `ansible_mounts` makes it a bit more complex since we have to register a variable in a loop to find which one is mounted on `/` ...

guits:

guits:

dsavineau:

guits:

PR #4568, opened 2019-10-08
tests: reduce handler mon and osd delay

dsavineau:
We don't need to have high handler delay in the CI so reducing to
10 seconds.


dsavineau:

PR #4582, opened 2019-10-09
dashboard: disable facts gathering

dsavineau:
This is already done in the main playbooks but absent in the dashboard
playbook.
The facts are already gathered during the first play of the main
playbooks so we don't need to doing twice.


dsavineau:

PR #4585, opened 2019-10-09
ceph-iscsi: notify rbd target services

dsavineau:
When the iscsi gateway configuration file changes then we need to notify
the rbd target api/gw services to be restarted.


dsavineau:
@mikechristie there's not rbd-target-[api|gw] service restart on ceph.conf modification as well. Should we notify those services too ? or only the api ? (I don't know if both service are using the ceph configuration file)

mikechristie:
Patch LGTM.

I think it would be ok to restart the daemon if there are ceph.conf changes.

dsavineau:

Issue #4518, opened 2019-10-03
"copy ceph key(s) if needed" fails when local ansible username != remote username


kevincoakley:
Environment:

Running Ansible (2.8.5) on a management host as user kcoakley, accessing Ceph cluster as centos user (see hosts file excerpt below):

```
[mons]
sdsc-dev-mon-0 ansible_host=192.168.0.6 ansible_port=22 ansible_user=centos ansible_become=yes ansible_become_method=sudo
sdsc-dev-mon-1 ansible_host=192.168.0.26 ansible_port=22 ansible_user=centos ansible_become=yes ansible_become_method=sudo
sdsc-dev-mon-2 ansible_host=192.168.0.24 ansible_port=22 ansible_user=centos ansible_become=yes ansible_become_method=sudo
```

commit e1d06f498cbd239dbe415ffa26824eaf9116161f changed how ceph keys are copied and added the delegate_to setting to the copy function.

Without the delegate_to setting the ssh command is:

```
<192.168.0.6> SSH: EXEC ssh -vvv -o ControlMaster=auto -o ControlPersist=600s -o PreferredAuthentications=publickey -o StrictHostKeyChecking=no -o Port=22 -
o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o 'User="centos
"' -o ConnectTimeout=60 -o ControlPath=/home/kcoakley/.ansible/cp/%h-%r-%p 192.168.0.6 '/bin/sh -c '"'"'rm -f -r /home/centos/.ansible/tmp/ansible-tmp-15700
61746.22-219962892797935/ > /dev/null 2>&1 && sleep 0'"'"''
```

With the delegate_to setting the ssh command is:

```
<ceph-mon-0> SSH: EXEC ssh -vvv -o ControlMaster=auto -o ControlPersist=600s -o PreferredAuthentications=publickey -o StrictHostKeyChecking=no -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=60 -o ControlPath=/home/kcoakley/.ansible/cp/%h-%r-%p ceph-mon-0 '/bin/sh -c '"'"'echo ~centos && sleep 0'"'"''
```

[...]

PR #4589, opened 2019-10-10
Set remote_user on "copy ceph key(s) if needed"

kevincoakley:
The ssh username is omitted from the ssh command when using copy and delegate_to with a hostname. remote_user was added to include the ssh username for users who use ansible_user in their inventory to specify the remote user.  See issue https://github.com/ceph/ceph-ansible/issues/4518 .


guits:
Closing this in favor of #4591

PR #4607, opened 2019-10-15
playbook: remove duplicate facts

dsavineau:
The is_atomic and container_binary facts are already defined in the
ceph-facts role so we don't need to have dedicated tasks for that
before the ceph-facts role exectution.


guits:

PR #4657, opened 2019-10-23
Allow setting dist and mock configuration in Makefile

javierpena:
Curently, the dist and mock configurations are hardcoded in the
Makefile to be el8 and epel-7-x86_64, respectively. This commit
allows the user to override those settings using the DIST and
MOCK_CONFIG environment variables, falling back to the current
defaults if not set.

This provides additional flexibility when building the RPM directly
from the repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ NAME = ceph-ansible
 #  A "git describe" value of "v2.2.0" creates an NVR
 #  "ceph-ansible-2.2.0-1.el8"
 
+DIST ?= "el8"
+MOCK_CONFIG ?= "epel-7-x86_64"
 TAG := $(shell git describe --tags --abbrev=0 --match 'v*')
 VERSION := $(shell echo $(TAG) | sed 's/^v//')
 COMMIT := $(shell git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


guits:
@ktdreyer FYI

dsavineau:

guits:

ktdreyer:
Would you please say more about the specific issue you're trying to solve?

I would like to understand more before we make this code more complicated.

javierpena:
> Would you please say more about the specific issue you're trying to solve?
> 

I want to be able to build the package for different distribution versions and buildroots, e.g. RHEL 7 and RHEL 8, without having to change the Makefile.

Currently, the makefile is using an EPEL-7 buildroot but sets el8 as a distro version, which kind of works but is wrong. With support for the environment variables,  we could have for example a job that builds the right ceph-ansible package for RHEL 8, and then the planned TripleO job could just take that package and use it on its tests. Otherwise, we would have to either patch the Makefile in the jobs or keep a separate spec file, which would cause unnecessary issues.

ktdreyer:
Thanks, that sounds good. I think we should have a separate PR that changes the "epel-7-x86_64" value to "epel-8-x86_64".

dsavineau:

dsavineau:

dsavineau:

dsavineau:

Issue #4670, opened 2019-10-25
rolling-upgrade: get current fsid: hostvars is undefined


discostur:
**Bug Report**

What happened:

rolling upgraded from mimic -> nautilus failed:

```
ansible-playbook -i ../hosts-dev rolling_update.yml -vv

TASK [ceph-facts : get current fsid] *******************************************************************************************************************************************************************************************************************************************
task path: /root/ansible-playbooks/ceph-ansible-dev/roles/ceph-facts/tasks/facts.yml:86
Friday 25 October 2019  16:29:09 +0200 (0:00:00.198)       0:03:42.386 ******** 
fatal: [ceph-mon01-dev.intern]: FAILED! => 
  msg: |-
    The task includes an option with an undefined variable. The error was: u"hostvars['ceph-mon01-dev']" is undefined
  
    The error appears to be in '/root/ansible-playbooks/ceph-ansible-dev/roles/ceph-facts/tasks/facts.yml': line 86, column 3, but may
    be elsewhere in the file depending on the exact syntax problem.
  
    The offending line appears to be:
  
  
    - name: get current fsid
      ^ here
```

Share your group_vars files, inventory

hosts:
```
[ceph-mgmt-nodes-dev]
ceph-mon01-dev.intern
ceph-mon02-dev.intern
ceph-mon03-dev.intern

[mons:children]
ceph-mgmt-nodes-dev

[mgrs:children]
ceph-mgmt-nodes-dev

[rgws:children]
ceph-mgmt-nodes-dev

[iscsigws]
ceph-osd01-dev.intern
ceph-osd02-dev.intern
ceph-osd03-dev.intern

[osds:]
ceph-osd01-dev.intern
ceph-osd02-dev.intern 
ceph-osd03-dev.intern
```

all.yml
```
---
dummy:
configure_firewall: false 
ntp_daemon_type: ntpd
upgrade_ceph_packages: False 
ceph_origin: repository
ceph_repository: community 
ceph_mirror: http://download.ceph.com
ceph_stable_key: https://download.ceph.com/keys/release.asc
ceph_stable_release: nautilus 
ceph_stable_repo: "{{ ceph_mirror }}/debian-{{ ceph_stable_release }}"
monitor_interface: ens192 
osd_objectstore: bluestore
radosgw_interface: ens192 
ceph_conf_overrides: 
  osd:
    osd_client_watch_timeout: 15
    osd_heartbeat_grace: 20
    osd_heartbeat_interval: 5
[...]

PR #4672, opened 2019-10-25
rolling_update: fix reset mon_host variable

dsavineau:
mon_host should use the inventory hostname and not the node hostname.
Fix creates an issue when the inventory and node hostname are different.

Closes: #4670


dsavineau:

PR #4679, opened 2019-10-26
Add comment to add-mon.yml to avoid issues when adding mons

kmadac:


guits:
@kmadac please, add a commit message and a sign off

PR #4697, opened 2019-10-31
Set grafana-server user and password in ceph-dashboard role

fmount:
This change adds two tasks to set grafana-api user and password
that are required to inject dashboard layouts to the external
grafana instance.
Without these two parameters the ceph-ansible playbook fails
showing an authorization error (HTTPError: 401 Client Error:
Unauthorized").

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1767365


dsavineau:
I'm wondering why we don't see this issue in the CI currently.

@fmount any thoughts ?

PR #4703, opened 2019-10-31
ceph-osd: add device class to crush rules

dsavineau:
This adds device class support to crush rules when using the class key
in the rule dict via the create-replicated sub command.
If the class key isn't specified then we use the create-simple sub
command for backward compatibility.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1636508

### DIFF
--- a/roles/ceph-osd/tasks/crush_rules.yml
+++ b/roles/ceph-osd/tasks/crush_rules.yml
@@ -11,7 +11,7 @@
     - osd_crush_location is defined
 
 - name: create configured crush rules
-  command: "{{ hostvars[groups[mon_group_name][0]]['container_exec_cmd'] | default('') }} ceph --cluster {{ cluster }} osd crush rule create-simple {{ item.name }} {{ item.root }} {{ item.type }}"
+  command: "{{ hostvars[groups[mon_group_name][0]]['container_exec_cmd'] | default('') }} ceph --cluster {{ cluster }} osd crush rule {{ 'create-replicated' if item.class is defined else 'create-simple' }} {{ item.name }} {{ item.root }} {{ item.type }} {{ item.class | default('') }}"
   changed_when: false
   with_items: "{{ hostvars[groups[mon_group_name][0]]['crush_rules'] | unique }}"
   delegate_to: '{{ groups[mon_group_name][0] }}'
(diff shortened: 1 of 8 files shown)


dsavineau:

roles/ceph-osd/tasks/crush_rules.yml
```diff
@@ -2,29 +2,29 @@
...
-    - inventory_hostname == groups.get(mon_group_name) | last
-    - create_crush_tree | bool
-    - hostvars[item]['osd_crush_location'] is defined
+    - hostvars[groups[mon_group_name][0]]['create_crush_tree'] | default(false) | bool
+    - osd_crush_location is defined
 
 - name: create configured crush rules
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} osd crush rule create-simple {{ item.name }} {{ item.root }} {{ item.type }}"
-  with_items: "{{ crush_rules | unique }}"
+  command: "{{ hostvars[groups[mon_group_name][0]]['container_exec_cmd'] | default('') }} ceph --cluster {{ cluster }} osd crush rule {{ 'create-replicated' if item.class is defined else 'create-simple' }} {{ item.name }} {{ item.root }} {{ item.type }} {{ item.class | default('') }}"
```
dsavineau: We probably need to be sure that the device class exists to avoid failures

roles/ceph-osd/tasks/crush_rules.yml
```diff
@@ -33,15 +33,15 @@
   set_fact:
     info_ceph_default_crush_rule_yaml: "{{ item.stdout | from_json() }}"
   with_items: "{{ info_ceph_default_crush_rule.results }}"
-  when:
-    - inventory_hostname == groups.get(mon_group_name) | last
-    - not item.get('skipped', false)
+  run_once: true
+  when: not item.get('skipped', false)
```
guits: what about 
```suggestion
  when: item.skipped is undefined
```

which is more readable IMO

roles/ceph-osd/tasks/crush_rules.yml
guits: that would add more complexity but I agree we should go this way

dsavineau: In fact this solution only moves the issue later.
If we `list the existing device class + register` and use `when + in register.stdout` then the crush rule won't be created if the device class doesn't exist.
At the end we will have the issue on the pool creation task because the crush rule won't exist.

fultonj:
This looks like what I had in mind. Thanks! I had hoped to test it myself but please don't let that block landing this.

PR #4734, opened 2019-11-13
site-docker: add always tag on gather fact tasks

dsavineau:
If we execute the site-docker.yml playbook with specific tags (like
ceph_update_config) then we need to be sure to gather the facts otherwise
we will see error like:

The task includes an option with an undefined variable. The error was:
'ansible_hostname' is undefined

This commit also adds missing 'gather_facts: false' to mons plays.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1754432

### DIFF
--- a/site-docker.yml.sample
+++ b/site-docker.yml.sample
@@ -24,8 +24,8 @@
   pre_tasks:
     - name: gather facts
       setup:
-      when:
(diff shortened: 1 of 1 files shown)


dsavineau:

PR #4736, opened 2019-11-14
ceph_key: restore file mode after a key is fetched

guits:
when `import_key` is enabled, if the key already exists, it will only be
fetched using ceph cli, if the mode specified in the `ceph_key` task is
different from what is applied by the ceph cli, the mode isn't restored because
we don't call `module.set_fs_attributes_if_different()` before
`module.exit_json(**result)`

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1734513

### DIFF
--- a/library/ceph_key.py
+++ b/library/ceph_key.py
@@ -541,6 +541,8 @@ def run_module():
         add_file_common_args=True,
     )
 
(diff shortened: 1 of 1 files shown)


guits:

PR #4747, opened 2019-11-18
Fixes failure of cephfs configuration using --limit while running purge-docker.yml 

VasishtaShastry:
Configuration of cephfs with an existing cluster using --limit used to fail
at different tasks while running with site-docker.yml
This commit addresses both of those tasks

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1773489
Signed-off-by: VasishtaShastry <vipin.indiasmg@gmail.com>


VasishtaShastry:
Verified regression for configuration from scratch also.
It would be nice if we can backport this fix to stable-4.0 also.

guits:

PR #4758, opened 2019-11-20
rolling_update: don't enable ceph-mon unit

dsavineau:
On non containerized deployment the ceph-mon hostname/fqdn systemd
service are stopped at the beginning of the mon upgrade.
But the parameter enabled is set to true for both task so even if we're
not using the fqdn then it will enabled the systemd unit based on it.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1649617

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -105,7 +105,7 @@
       systemd:
         name: ceph-mon@{{ ansible_hostname }}
         state: stopped
-        enabled: yes
+        enabled: no
       ignore_errors: True
       when:
         - not containerized_deployment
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/rolling_update.yml
```diff
@@ -114,7 +114,7 @@
       systemd:
         name: ceph-mon@{{ ansible_fqdn }}
         state: stopped
-        enabled: yes
+        enabled: no
       ignore_errors: True
       when:
```
dsavineau: @guits I'm also wondering why there's no `mon_use_fqdn | bool` in this condition. Any thought ?

guits: if I remember correctly, `mon_use_fqdn` usage was deprecated as of stable-3.2 anyway

PR #4762, opened 2019-11-21
ceph-rgw-loadbalancer: add option for haproxy ssl frontend

slam1987:
Currently RGW HAproxy configuration does not support SSL termination or passthrough. If SSL cert is setup along with HAproxy,  individual RGW instances will be set as SSL endpoints.

This PR will address this issue by setting HAproxy as a SSL endpoint instead of each individual RGW instance.

Signed-off-by: Stanley Lam <stanleylam_604@hotmail.com>

### DIFF
--- a/roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
+++ b/roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
@@ -9,7 +9,11 @@ global
     group       haproxy
     daemon
     stats socket /var/lib/haproxy/stats
-
+{%- if haproxy_frontend_ssl_certificate %}
+    tune.ssl.default-dh-param {{ haproxy_ssl_dh_param }}
+    ssl-default-bind-ciphers {{ haproxy_ssl_ciphers | join(':') }}
+    ssl-default-bind-options {{ haproxy_ssl_options | join(' ') }}
+{% endif %}
 defaults
     mode                    http
     log                     global
@@ -29,7 +33,11 @@ defaults
     maxconn                 8000
 
 frontend rgw-frontend
+{% if haproxy_frontend_ssl_certificate %}
+    bind *:{{ haproxy_frontend_ssl_port }} ssl crt {{ haproxy_frontend_ssl_certificate }}
+{% else %}
     bind *:{{ haproxy_frontend_port }}
+{% endif %}
     default_backend rgw-backend
 
 backend rgw-backend
--- a/group_vars/rgwloadbalancers.yml.sample
+++ b/group_vars/rgwloadbalancers.yml.sample
@@ -14,6 +14,17 @@ dummy:
 ###########
 
 #haproxy_frontend_port: 80
+#haproxy_frontend_ssl_port: 443
+#haproxy_frontend_ssl_certificate:
+#haproxy_ssl_dh_param: 4096
+#haproxy_ssl_ciphers:
+#  - EECDH+AESGCM
+#  - EDH+AESGCM
+#haproxy_ssl_options:
+#  - no-sslv3
+#  - no-tlsv10
+#  - no-tlsv11
+#  - no-tls-tickets
 #
 #virtual_ips:
 #   - 192.168.238.250
--- a/roles/ceph-rgw-loadbalancer/defaults/main.yml
+++ b/roles/ceph-rgw-loadbalancer/defaults/main.yml
@@ -6,6 +6,17 @@
 ###########
 
 haproxy_frontend_port: 80
+haproxy_frontend_ssl_port: 443
+haproxy_frontend_ssl_certificate:
+haproxy_ssl_dh_param: 4096
+haproxy_ssl_ciphers:
+  - EECDH+AESGCM
+  - EDH+AESGCM
+haproxy_ssl_options:
+  - no-sslv3
+  - no-tlsv10
+  - no-tlsv11
+  - no-tls-tickets
 #
 #virtual_ips:
 #   - 192.168.238.250


slam1987:

slam1987:
nekins test centos-container-all_daemons

slam1987:

slam1987:

slam1987:
jenkins test playbook syntax

dsavineau:
Please see https://docs.ceph.com/ceph-ansible/master/dev/index.html#submit-a-patch about the changes on group vars.

dsavineau:
Could you also squash your commits ?

dsavineau:
You need to amend your commit and add a signed off

roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -122,7 +122,7 @@ log file = /var/log/ceph/{{ cluster }}-rgw-{{ hostvars[host]['ansible_hostname']
 {%- if frontend_type == 'civetweb' -%}
 {{ radosgw_frontend_type }} port={{ _rgw_binding_socket }}{{ 's ssl_certificate='+radosgw_frontend_ssl_certificate if radosgw_frontend_ssl_certificate else '' }}
 {%- elif frontend_type == 'beast' -%}
-{{ radosgw_frontend_type }} {{ 'ssl_' if radosgw_frontend_ssl_certificate else '' }}endpoint={{ _rgw_binding_socket }}{{ ' ssl_certificate='+radosgw_frontend_ssl_certificate if radosgw_frontend_ssl_certificate else '' }}
+{{ radosgw_frontend_type }} {{ 'ssl_' if radosgw_frontend_ssl_certificate and not haproxy_frontend_ssl_termination else '' }}endpoint={{ _rgw_binding_socket }}{{ ' ssl_certificate='+radosgw_frontend_ssl_certificate if radosgw_frontend_ssl_certificate and not haproxy_frontend_ssl_termination else '' }}
```
dsavineau: I don't think this is a good idea to mix haproxy variable into the ceph-rgw role.
Also we could have end to end TLS termination on both haproxy AND rgw which is not possible with this condition.

roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
```diff
@@ -29,7 +29,11 @@ defaults
     maxconn                 8000
 
 frontend rgw-frontend
+{% if haproxy_frontend_ssl_termination and radosgw_frontend_ssl_certificate %}
```
dsavineau: if a user wants a different TLS certificate on haproxy and radosgw node ?

group_vars/rhcs.yml.sample
```diff
@@ -32,7 +32,7 @@ dummy:
 #  dev: 99
```
dsavineau: Why are you modifying this file ?

slam1987: This file was generated by `generate_group_vars_sample.sh` as per "Submit a patch" in https://docs.ceph.com/ceph-ansible/master/dev/index.html.

After running the script, are we suppose to commit these changes?

dsavineau: yes you're suppose to commit the changes generated but the `generate_group_vars_sample.sh` but your PR doesn't change anything for RHCS so it doesn't make sense that there's changes in rhcs.yml.sample file.

slam1987: Agreed

nikosmeds: We're thinking there are issues running `./generate_group_vars_sample.sh` on MacOS due to differences in `sed` behavior, as mentioned in https://unix.stackexchange.com/a/131940

Re-ran the generate script on a Ubuntu system and this file is no longer edited. Will update PR plus make further updates as per your comments. Thanks!

roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
slam1987: Removed this logic and set a new variable for haproxy specific certificate.

roles/ceph-config/templates/ceph.conf.j2
slam1987: Condition removed to allow TLS termination on both haproxy and rgw.

roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
```diff
@@ -29,7 +29,11 @@ defaults
     maxconn                 8000
 
 frontend rgw-frontend
+{% if haproxy_frontend_ssl_certificate %}
+    bind *:{{ haproxy_frontend_ssl_port }} ssl crt {{ haproxy_frontend_ssl_certificate }}
```
dsavineau: I know that's not the initial purpose of this PR but it might be a good idea to add strong ciphers and options (disable ssl v3, etc..) to the TLS configuration via new variables.
WDYT ?

slam1987: Yes that sounds great, we were planning on submitting a separate PR for that but if that's fine with you we can add it here.

roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
```diff
@@ -9,7 +9,11 @@ global
     group       haproxy
     daemon
     stats socket /var/lib/haproxy/stats
-
+{%- if haproxy_frontend_ssl_certificate %}
+    tune.ssl.default-dh-param {{ haproxy_ssl_dh_param }}
+    ssl-default-bind-ciphers {% for cipher in haproxy_ssl_ciphers %}{% if loop.last %}{{ cipher }}{% else %}{{ cipher }}:{% endif %}{% endfor %}
+    ssl-default-bind-options{% for option in haproxy_ssl_options %} {{ option }}{% endfor %}
```
dsavineau: To avoid doing a loop and missing the \n you should use the join filter.

```yaml
    ssl-default-bind-ciphers {{ haproxy_ssl_ciphers | join(':') }}
    ssl-default-bind-options {{ haproxy_ssl_options | join(' ') }}
```

Because the current code generates something like:

```console
    tune.ssl.default-dh-param 4096
    ssl-default-bind-ciphers EECDH+AESGCM:EDH+AESGCM    ssl-default-bind-options no-sslv3 no-tlsv10 no-tlsv11 no-tls-tickets
```

slam1987: Thank you, changes have been made :)

dsavineau:

PR #4774, opened 2019-11-25
container: add token registry authentication

dsavineau:
Instead of storing the username and password value for the container
authentication in clear text then one can use the token authentication
system which is basically "username:password" format encoded in base64.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1768584

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -573,8 +573,13 @@ dummy:
 #ceph_docker_image_tag: latest
 #ceph_docker_registry: docker.io
 #ceph_docker_registry_auth: false
+# The container registry authentication could be use with:
+# 1) the container registry username and password
 #ceph_docker_registry_username:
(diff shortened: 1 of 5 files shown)


dsavineau:

dsavineau:

dsavineau:

dsavineau:

guits:

dsavineau:

dsavineau:
> In fact I'd probably use the service account terminology in the variable naming i.e. ceph_docker_service_account_login and ceph_docker_service_account_token to make it more obvious. This is only rhcs anyway, and this is best practice so adopting those names should work for us

If it's only a matter of variable naming then it's not worthwhile.
It won't be backward compatible and even it could be a rhcs only change then you can still use another registry that doesn't support the service account token.

PR #4783, opened 2019-11-27
ceph-osd: wait for all osds once

dsavineau:
cf8c6a3 moves the 'wait for all osds' task from openstack_config to the
main tasks list.
But the openstack_config code was executed only on the last OSD node.
We don't need to do this check on all OSD node so we need to add set
run_once to true on that task.


dsavineau:

dsavineau:

guits:

dsavineau:

dsavineau:

PR #4788, opened 2019-11-27
switch_to_containers: fix umount ceph partitions

dsavineau:
When a container is already running on a non containerized node then the
umount ceph partition task is skipped.
This is due to the container ps command which always returns 0 even if
the filter matches nothing.

We should run the umount task when:
1/ the container command is failing (not installed) : rc != 0
2/ the container command reports running ceph-osd containers : rc == 0

Also we should not fail on the ceph directory listing.

### DIFF
--- a/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
+++ b/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
@@ -250,7 +250,7 @@
 
     - name: check if containerized osds are already running
       command: >
-        {{ container_binary }} ps --filter='name=ceph-osd'
+        {{ container_binary }} ps -q --filter='name=ceph-osd'
       changed_when: false
       failed_when: false
       register: osd_running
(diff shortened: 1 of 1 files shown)


dsavineau:

infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
```diff
@@ -260,14 +260,15 @@
...
+      failed_when: false
 
     - name: unmount all the osd directories
       command: >
         umount {{ item }}
       changed_when: false
       failed_when: false
       with_items: "{{ osd_dirs.stdout_lines }}"
-      when: osd_running.rc != 0
+      when: osd_running.rc != 0 or osd_running.stdout_lines | length == 0
```
guits: do we really need the first part of the condition? given that we know it always returns 0

dsavineau: yes because if docker/podman isn't installed then the rc code won't 0.

BTW the docker/podman installation is done after this part.

guits: makes sense then

PR #4792, opened 2019-11-28
docker2podman: various fixes

guits:
We must import this role in the first play otherwise the first call to
`client_group_name`fails.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1777829


dsavineau:
This is unrelated but we're using the `client_group_name` in the first gather fact task but the hardcoded `clients` name in the second. [1]
Should we consider to update the second task in this PR or a dedicated one ?

[1] https://github.com/ceph/ceph-ansible/blob/master/infrastructure-playbooks/docker-to-podman.yml#L33

Issue #4467, opened 2019-09-25
Dashboard - NFS - No cluster available


skunkr:
**Bug Report**

Hi,
I have installed from stable-4.0,  and everything went well,  but I just can't create NFS exports from dashboard UI.
It says "No cluster available",  although there is already an export configured with rgw backend.
The only command that i've issued is:
"ceph dashboard set-ganesha-clusters-rados-pool-namespace .rgw.root"
On the nfs enabled nodes,  after installation i can see "ganesha.conf", 'rgw_bucket.conf", "rgw.conf" created under /etc/ganesha/
I have uploaded these files as rados objects on the ".rgw.root" pool,  but still the UI says "-- No cluster available --"
I think I am missing something, but I'm also not very sure how to sort this.
Any thoughts ?
Thank you,

Leo

dsavineau:
We probably need to investigate more on the Ganesha configuration in the Ceph Dashboard.

https://docs.ceph.com/docs/master/mgr/dashboard/#nfs-ganesha-management

As you mentioned we need to run one extra `ceph dashboard command` but also update the ganesha configuration.

skunkr:
Thank you Dimitri,
But i think it is also a lack of knowledge on my side. 
What would be the exact extra manual steps that need to be done to have the NFS service added in Dashboard ?  I have read the documentation,  but it seems a bit confusing for me,  regarding to what actually files with what content to place as objects.
Or this should be automatically done by the playbook ?

dsavineau:
From my point of view, this should be done automatically by the playbook.
That said we will need to dig a little bit more more in this area because we also lack of knowledge on this part.

KingJ:
I was about to log a new bug related to this, however I saw this as a related issue and thought it best to group it.

To build up a list of NFS daemons, the dashboard looks in the configured RADOS pool for `conf-*` objects. The objects can, and should, be empty - the dashboard is only looking for them by name. If there are no `conf-*` files in the pool, then the dashboard will report no daemons are available.

**Bug 1**: The ceph-nfs role does not create new `conf-*` objects for hosts in the `[nfss]` group in any RADOS pool. [It only creates a `ganesha-export-index` file in the default CephFS data pool if it doesn't already exist](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-nfs/tasks/start_nfs.yml#L19).

When a daemon is available and an export is created via the dashboard, two things then happen.

[...]

PR #4793, opened 2019-11-30
NFS Ganesha Per-Host Exports

KingJ:
Addresses bugs 1 & 2 of #4467 - the task now creates a `conf-HOSTNAME` RADOS object if one doesn't already exist instead of the default `ganesha-export-index` object. This allows NFS Ganesha to pick up exports per-host rather than globally, and for Ceph's Dashboard to recognise what NFS Ganesha servers exist in the cluster. The NFS Ganesha configuration template has also been updated so that NFS Ganesha uses the host-specific RADOS object for determining exports.

### DIFF
--- a/roles/ceph-nfs/templates/ganesha.conf.j2
+++ b/roles/ceph-nfs/templates/ganesha.conf.j2
@@ -30,7 +30,7 @@ RADOS_URLS {
    ceph_conf = '/etc/ceph/{{ cluster }}.conf';
    userid = "{{ ceph_nfs_ceph_user }}";
 }
-%url rados://{{ cephfs_data_pool.name }}/{{ ceph_nfs_rados_export_index }}
+%url rados://{{ cephfs_data_pool.name }}/conf-{{ ansible_hostname }}
 
 NFSv4 {
 	RecoveryBackend = 'rados_kv';
--- a/roles/ceph-nfs/tasks/start_nfs.yml
+++ b/roles/ceph-nfs/tasks/start_nfs.yml
@@ -6,7 +6,7 @@
     when: containerized_deployment | bool
 
   - name: check if rados index object exists
-    shell: "{{ container_exec_cmd_nfs | default('') }} rados -p {{ cephfs_data_pool.name }} --cluster {{ cluster }} ls|grep {{ ceph_nfs_rados_export_index }}"
+    shell: "{{ container_exec_cmd_nfs | default('') }} rados -p {{ cephfs_data_pool.name }} --cluster {{ cluster }} ls|grep conf-{{ ansible_hostname }}"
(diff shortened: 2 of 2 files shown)


guits:
@KingJ 

It means this could be removed:

```shell
roles/ceph-nfs/defaults/main.yml:51:ceph_nfs_rados_export_index: "ganesha-export-index"
```

roles/ceph-nfs/templates/ganesha.conf.j2
```diff
@@ -30,7 +30,7 @@ RADOS_URLS {
    ceph_conf = '/etc/ceph/{{ cluster }}.conf';
    userid = "{{ ceph_nfs_ceph_user }}";
 }
-%url rados://{{ cephfs_data_pool.name }}/{{ ceph_nfs_rados_export_index }}
+%url rados://{{ cephfs_data_pool.name }}/conf-{{ ansible_hostname }}
 
 NFSv4 {
 	RecoveryBackend = 'rados_kv';
```
epuertat: `watch_url:` is needed pointing to the `conf-<id>` object. See [reference](https://documentation.suse.com/ses/6/html/ses-all/cha-ceph-nfsganesha.html#ceph-nfsganesha-config-service-rados).

KingJ: What's the difference between `%url` outside of a configuration block, and using `watch_url` inside the `RADOS_URLS` section? I've had a look at the configuration reference but I can't quite see how they're used differently. Purely guessing, is `watch_url` used to trigger a reload if the configuration object changes, so that the the configuration loaded by the `%url` section is picked up?

tombarron:
Dumb question here: for Manila the rados object is used deliberately for the index so that it is persisted independent of any particular host.  Are there other users of Ceph-NFS/Ganesha-with-ceph-fsal that don't want this behavior?

fmount:
Just a quick comment on this: I see the issue #4467  and this should be solved by this PR, I just want to enforce the @epuertat idea on having different tasks to address this issue (you can rely on the `dashboard_enabled` bool).
Avoid breaking the manila use case described here [1] is a must, so I would like to move forward w/ the PR maintaining backward compatibility and taking into considerations all the use case involved.

[1] https://docs.openstack.org/manila/rocky/contributor/ganesha.html#nfs-ganesha-configuration

guits:
@KingJ do you want someone to take over this PR?

guits:
@KingJ any news?

KingJ:
@guits Apologies for the lack of activity on this PR. I've not had as much time to work on this as i'd like recently, so if you would be willing to take over this PR that's not a problem by me.

guits:
jenkins test all

PR #4799, opened 2019-12-02
dashboard: use fqdn url for active alert

guits:
When using the shortname, the URL for active alert launches with short
hostname and fails to connect to the server.

This commit changes the template in order to use the fqdn.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1765485


guits:
@b-ranto I'm unsure whether it could bring side-effects?

PR #4801, opened 2019-12-02
tests: fix update scenario (container)

guits:
The path to the inventory isn't correct because we are missing the variable
`CONTAINER_DIR` here.


guits:

PR #4829, opened 2019-12-06
ceph-validate: start with ansible version test

dsavineau:
It doesn't make sense to start validating configuration if the ansible
version isn't the good one.
This commit moves the check_system task as the first task in the
ceph-validate role.
The ansible version test tasks are moved at the top of this file.
Also moving the iscsi kernel tests from check_system to check_iscsi
file.


dsavineau:

PR #4852, opened 2019-12-10
ceph-facts: set use_new_ceph_iscsi on iscsi nodes

dsavineau:
We don't need to set the use_new_ceph_iscsi fact on other nodes than
those present in the iscsigws group.
Also remove the duplicate iscsi_gw_group_name condition already present
on the include_task.
Finally validate the ansible distribution as the first task.


guits:

PR #4856, opened 2019-12-11
improve purge-cluster playbook

dsavineau:
- stop and remove ceph-iscsi
- specify python version on package
- stop ceph-osd.target service
- reset failed systemd unit


dsavineau:

PR #4860, opened 2019-12-13
yum plugin priorities should only for centos

Os-carsun:
Fedora will execute this command
but yum plugin priorities could not be found in fedora, and this package is not needed in fedora
so this command should only for centos


dsavineau:
First, don't send a PR on a stable branch unless it's something specific.
Instead you should create the PR on master and we will then backport the change if necessarily.

Then, the change is not only for CentOS but RHEL too. But the biggest problem here is that the ceph community package doesn't support Fedora.

Os-carsun:
Sorry about that
I'll closethis issue

Os-carsun:
Sorry about that
I'll closethis issue

PR #4861, opened 2019-12-16
Allow ceph-ansible to inject layout with multiple grafana-server instances

fmount:
For ha purposes is common having more than one node to deploy the
monitoring stack provided by grafana, prometheus and alertmanager
containers.
This patch allows ceph-ansible to configure the dashboard info and
inject all the cluster layouts for each monitoring node.

### DIFF
--- a/dashboard.yml
+++ b/dashboard.yml
@@ -80,6 +80,7 @@
 - hosts: "{{ groups[mgr_group_name] | default(groups[mon_group_name]) | default(omit) }}"
   gather_facts: false
   become: true
+  serial: 1
   pre_tasks:
     - name: set ceph dashboard install 'In Progress'
       run_once: true
--- a/roles/ceph-dashboard/tasks/configure_dashboard.yml
+++ b/roles/ceph-dashboard/tasks/configure_dashboard.yml
@@ -94,12 +94,6 @@
   changed_when: false
   until: ac_result.rc == 0
 
-- name: set grafana url
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-grafana-api-url {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ grafana_port }}"
-  delegate_to: "{{ groups[mon_group_name][0] }}"
-  run_once: true
-  changed_when: false
-
 - name: set grafana api user
   command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-grafana-api-username {{ grafana_admin_user }}"
   delegate_to: "{{ groups[mon_group_name][0] }}"
@@ -108,18 +102,6 @@
 
 - name: set grafana api password
   command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-grafana-api-password {{ grafana_admin_password }}"
-  delegate_to: "{{ groups[mon_group_name][0] }}"
-  run_once: true
-  changed_when: false
-
-- name: set alertmanager host
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-alertmanager-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ alertmanager_port }}"
-  delegate_to: "{{ groups[mon_group_name][0] }}"
-  run_once: true
-  changed_when: false
-
-- name: set prometheus host
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-prometheus-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ prometheus_port }}"
   delegate_to: "{{ groups[mon_group_name][0] }}"
   run_once: true
   changed_when: false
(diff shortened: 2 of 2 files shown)


fmount:
@guits @dsavineau This is a patch I'm currently testing to solve the multiple grafana-server instance configuration. 
@gfidente fyi

dashboard.yml
```diff
@@ -73,9 +73,10 @@
 
 # using groups[] here otherwise it can't fallback to the mon if there's no mgr group.
 # adding an additional | default(omit) in case where no monitors are present (external ceph cluster)
-- hosts: "{{ groups[mgr_group_name] | default(groups[mon_group_name]) | default(omit) }}"
+- hosts: "{{ groups[grafana_server_group_name] }}"
```
dsavineau: Don't change this. The Ceph Dashboard is deployed on the mgrs or mons nodes not on the nodes present on the grafana-server group

roles/ceph-dashboard/tasks/configure_dashboard.yml
```diff
@@ -211,11 +193,31 @@
...
+    command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-alertmanager-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ alertmanager_port }}"
+    delegate_to: "{{ groups[mon_group_name][0] }}"
+    run_once: true
+    changed_when: false
+
+  - name: set prometheus host
+    command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-prometheus-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ prometheus_port }}"
+    delegate_to: "{{ groups[mon_group_name][0] }}"
+    run_once: true
+    changed_when: false
```
dsavineau: Why moving this into the a `containerized_deployment` block ?
We need this for non containerized deployment too.

fmount: you're right, we need to set these info for both scenarios. I'm going to update the PR, thanks @dsavineau

dashboard.yml
```diff
@@ -76,6 +76,7 @@
 - hosts: "{{ groups[mgr_group_name] | default(groups[mon_group_name]) | default(omit) }}"
   gather_facts: false
   become: true
+  serial: 1
```
gfidente: I think this is needed because the config keys applied below will be cluster-wide hence we need to set and overwrite them from each node for the "dashboards update" command to work, is this correct?

roles/ceph-dashboard/tasks/configure_dashboard.yml
```diff
@@ -211,12 +193,31 @@
...
+
+  - name: inject grafana dashboard layouts
+    command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard grafana dashboards update"
+    delegate_to: "{{ groups[mon_group_name][0] }}"
+    run_once: true
```
gfidente: you can probably set run_once for the entire block

roles/ceph-dashboard/tasks/configure_dashboard.yml
```diff
@@ -211,12 +193,31 @@
...
 
-- name: inject grafana dashboard layouts
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard grafana dashboards update"
-  delegate_to: "{{ groups[mon_group_name][0] }}"
-  run_once: true
-  changed_when: false
-  when: containerized_deployment | bool
+- name: set dashboard endpoints and inject layouts
+  block:
+  - name: set grafana url
```
gfidente: these vars will just be left set to the value generated on the last node which executed the block though; it'd be interesting to know which other command (in addition to "dashboards update") will consume them

dashboard.yml
fmount: That's correct, I'm trying to translate in Ansible "words" the following:
```
>> for index in $(seq 1 3); do \
echo "set grafana-url to the instance deployed in controller-$index";
echo "set prometheus-host to the instance deployed in controller-$index";
echo "set alaertmanager-host to the instance deployed in controller-$index";
echo "inject the dashboard layouts to the current grafana node (deployed in controller-$index)";
echo "--------------------------------------"
done
```
where `controller` is the node where we're deploying the dashboard. For an external ceph context we can easily make `s/controller/mgr`.

Actually if you look at the templates [1] [2] [3] the endpoints are configured using `grafana_server_addr` fact, hence we can/should have an additional task to `unset` the endpoints in the mgr since they're not used (sounds like the data are coming from node-exporter and ceph CLI), but this is my understanding, maybe @b-ranto can help confirming this.


[1] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-grafana/templates/datasources-ceph-dashboard.yml.j2
[2] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-prometheus/templates/prometheus.yml.j2
[3] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-prometheus/templates/alertmanager.yml.j2

roles/ceph-dashboard/tasks/configure_dashboard.yml
fmount: ++

dashboard.yml
fmount: Restored

roles/ceph-dashboard/tasks/configure_dashboard.yml
fmount: Exactly, because if the purpose of these commands are merely related to the dashboard injection, we should unset these endpoints after pushing them (see my previous comment). 
Each grafana instance points to the local prometheus and alertmanager (we see this from the config files) and all of them have the same view of the cluster and collect data provided by the node-exporter: what I'm trying to say is that I don't see any data coming from a sampling process (passing from the mgr dashboard to the grafana instance).

arxcruz:
check-rdo

gfidente:
@pcuzner can you help us understand how is best to configure the dashboard pieces when we have multiple grafana/prometheus instances?

dsavineau:
@fmount Could you amend the commit to add the associated BZ ?

roles/ceph-dashboard/tasks/configure_dashboard.yml
```diff
@@ -220,12 +202,28 @@
...
+
+  - name: set prometheus host
+    command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-prometheus-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ prometheus_port }}"
+    delegate_to: "{{ groups[mon_group_name][0] }}"
+    changed_when: false
+
+  - name: inject grafana dashboard layouts
+    command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard grafana dashboards update"
+    delegate_to: "{{ groups[mon_group_name][0] }}"
+    changed_when: false
```
dsavineau: In fact this isn't necessary at all and the `when: containerized_deployment | bool` condition on the latest task is missing.

fmount:
@dsavineau I rebased the PR and I'm still testing on the OSP context.

dsavineau:
closing in favor of #4993

PR #4862, opened 2019-12-16
ceph-infra: open dashboard port on monitor

dsavineau:
When there's no mgr group defined in the ansible inventory then the
mgrs are deployed implicitly on the mons nodes.
If the dashboard is enabled then we need to open the dashboard port on
the node that is running the ceph mgr process (mgr or mon).
The current code only allow to open that port on the mgr nodes when they
are present explicitly in the inventory but not implicitly.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1783520

### DIFF
--- a/roles/ceph-infra/tasks/configure_firewall.yml
+++ b/roles/ceph-infra/tasks/configure_firewall.yml
@@ -201,7 +201,8 @@
     when:
       - dashboard_enabled | bool
       - mgr_group_name is defined
-      - mgr_group_name in group_names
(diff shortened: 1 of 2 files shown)


dsavineau:
@guits I think we should drop this feature (mgr on mon by default) on master to avoid this kind of situation in the future.

dsavineau:

guits:
> @guits I think we should drop this feature (mgr on mon by default) on master to avoid this kind of situation in the future.

I agree.

PR #4863, opened 2019-12-16
ceph-defaults: exclude rbd devices from discovery

dsavineau:
The RBD devices aren't excluded from the devices list in the LVM auto
discovery scenario.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1783908

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -384,7 +384,7 @@ dummy:
 #osd_objectstore: bluestore
 
 # Any device containing these patterns in their path will be excluded.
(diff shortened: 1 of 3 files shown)


dsavineau:
jenkins test centos-container-lvm_auto_discovery

dsavineau:
jenkins test centos-non_container-lvm_auto_discovery

dsavineau:
@guits I'm not sure if we need to BP this to 3.2 even if it could be useful since the BZ is only targeted for 4.0

PR #4865, opened 2019-12-16
[stable-3.2] switch_to_containers: set GUID on lockbox part

dsavineau:
The ceph lockbox partition (part number 5) used with non lvm scenarios
and in non containerized deployment don't have a valid PARTUUID.
The value is set to 00000000-0000-0000-0000-000000000000 for each OSD
devices.
```console
$ blkid -t PARTLABEL="ceph lockbox" -o value -s PARTUUID
00000000-0000-0000-0000-000000000000
00000000-0000-0000-0000-000000000000
00000000-0000-0000-0000-000000000000
00000000-0000-0000-0000-000000000000
00000000-0000-0000-0000-000000000000
```
When switching to containerized deployment we manually mount the lockbox
partition by using the PARTUUID.
Unfortunately because we have most of the time multiple OSD on the same
node we can't have the right symlink in /dev/disk/by-partuuid because it
will point to only one partition.

```console
/dev/disk/by-partuuid/00000000-0000-0000-0000-000000000000 -> ../../sdb5
```

After the switch_to_containers playbook then only one OSD will restart
correctly and the other will try to access to the wrong device causing
error like 'xxxx is still in use'.

When deploying with containers and dmcrypt OSDs we force a PARTUUID
value during the ceph-disk prepare task.

### DIFF
--- a/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
+++ b/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
@@ -224,6 +224,33 @@
         - /usr/lib/systemd/system/ceph-volume@.service
         - /etc/systemd/system/ceph.target.wants
 
+    - name: dmcrypt extra operations
+      when:
+        - osd_scenario != 'lvm'
+        - dmcrypt | bool
+      block:
+        - name: remove zero(s) partuuid symlink
+          file:
+            path: /dev/disk/by-partuuid/00000000-0000-0000-0000-000000000000
+            state: absent
+
+        - name: get lockbox partitions
+          command: blkid -t PARTLABEL="ceph lockbox" -o device
+          changed_when: false
+          failed_when: false
+          register: lockbox_partitions
+
+        - name: get lockbox devices without partuuid
+          command: "blkid -t PARTUUID=00000000-0000-0000-0000-000000000000 -o device {{ item }}"
+          changed_when: false
+          failed_when: false
+          register: lockbox_devices
+          with_items: '{{ lockbox_partitions.stdout_lines }}'
+
+        - name: set guid on lockbox partition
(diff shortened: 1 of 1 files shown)


dsavineau:

infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
```diff
@@ -224,6 +224,33 @@
...
+
+        - name: get lockbox devices without partuuid
+          command: "blkid -t PARTUUID=00000000-0000-0000-0000-000000000000 -o device {{ item }}"
+          changed_when: false
+          failed_when: false
+          register: lockbox_devices
+          with_items: '{{ lockbox_partitions.stdout_lines }}'
+
+        - name: set guid on lockbox partition
+          shell: 'sgdisk --partition-guid=5:$(uuidgen) {{ item.stdout[:-2] if item.stdout is match("^/dev/(cciss|nvme|loop).*") else item.stdout[:-1] }}'
```
guits:
```suggestion
          shell: "sgdisk --partition-guid=5:$(uuidgen) {{ item.stdout[:-2] if item.stdout is match("^/dev/(cciss|nvme|loop).*") else item.stdout[:-1] }}"
```

dsavineau:
@alfredodeza any idea why we have lockbox partuuid with `00000000-0000-0000-0000-000000000000` ?
Maybe this has been always like this in Luminous

alfredodeza:
This looks very surprising. `ceph-disk` should use the uuid `fb3aabf9-d25f-47cc-bf5e-721d1816496b` for the lockbox. In `ceph-volume` when we scan for these partitions, we check for both, but prefer the `PARTLABEL` if found: https://github.com/ceph/ceph/blob/master/src/ceph-volume/ceph_volume/util/device.py#L444-L455

I would use that method of trying both. I wouldn't trust anything related to ceph-disk

PR #4880, opened 2020-01-04
purge-cluster.yml: fix purge cluster failed when node-exporter image does not exist

wujie1993:
This commit fix purge cluster failed when node-exporter image does not exist
Signed-off-by: wujie1993 qq594jj@gmail.com


wujie1993:
@Mergifyio refresh

PR #4882, opened 2020-01-05
purge: fix purge cluster failed when container images does not exist

wujie1993:
fix purge cluster failed when container images does not exist

Signed-off-by: wujie1993 <qq594jj@gmail.com>

### DIFF
--- a/infrastructure-playbooks/purge-container-cluster.yml
+++ b/infrastructure-playbooks/purge-container-cluster.yml
@@ -378,26 +378,30 @@
   tasks:
     - import_role:
         name: ceph-defaults
-    - import_role:
-        name: ceph-facts
-        tasks_from: container_binary
 
-    - name: disable node_exporter service
-      service:
-        name: node_exporter
-        state: stopped
-        enabled: no
-      failed_when: false
+    - block:
+        - import_role:
+            name: ceph-facts
+            tasks_from: container_binary
 
-    - name: remove node_exporter service file
-      file:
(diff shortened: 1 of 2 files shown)


wujie1993:
Thanks for you reply @dsavineau .Do you mean we should purge node-exporter and grafana-server only when `dashboard_enabled` is set to `true`?

dsavineau:
> Thanks for you reply @dsavineau .Do you mean we should purge node-exporter and grafana-server only when `dashboard_enabled` is set to `true`?

exactly

dsavineau:

dsavineau:

infrastructure-playbooks/purge-container-cluster.yml
```diff
@@ -532,6 +539,7 @@
 
     - name: remove ceph container image
       command: "{{ container_binary }} rmi {{ ceph_docker_registry }}/{{ ceph_docker_image }}:{{ ceph_docker_image_tag }}"
+      failed_when: false
```
dsavineau: Why are you adding the `failed_when` arg here ?

wujie1993: I am thinking that if the local image of ceph does not exist, an error will occur here.

wujie1993: This part was removed in the latest commit because I previously considered avoiding errors in the absence of container images

dsavineau:

PR #4885, opened 2020-01-06
Adding collocated control plane suuport for ceph on vagrant

shonpaz123:
Hi, 
I am currently using this repository to deploy vagrant machines using vagrant_varibales.yml file. I have noticed that there isn't support for collocated control plane (mon/rgw/mgr) for this kind of deployment. This deployment usually used for test clusters in various versions where people run on their own computer (that usually lacks in memory) and i think deploying those services on different vms is overkill, i a running production ceph cluster and all control plane services are running on the same nodes. 
I have added a booleam variable called "COLLOCATED" which indeed gets true/false, in case the user decided he wants to deploy collocated services, vagrant file directs mgr and rgw services to the mon machine, if the user chose false, deployment works as usual. 
It will be great if you could help me contribute to the community and tell me what do you think and weather this ability useful. 
Vagrant file and vagrant_variables.yml attached.

Thank you very much, 
Shon.

[ceph-ansible-nautilus.zip](https://github.com/ceph/ceph-ansible/files/4027294/ceph-ansible-nautilus.zip)


guits:
you are trying to merge stable-4.0 into master.

Please fix your PR.

PR #4888, opened 2020-01-07
ceph-grafana: Add ceph_grafana_docker_extra_env option

BenoitKnecht:
This can be used to pass extra environment variables to the Grafana
container.

For instance, `HTTP_PROXY` and `HTTPS_PROXY` can be set if Grafana needs
to use an HTTP proxy to download its plugins:

```yaml
ceph_grafana_docker_extra_env:
  -e HTTP_PROXY=http://192.0.2.1 -e HTTPS_PROXY=http://192.0.2.1
```

### DIFF
--- a/roles/ceph-grafana/defaults/main.yml
+++ b/roles/ceph-grafana/defaults/main.yml
@@ -1,0 +1,6 @@
+---
+##########
+# DOCKER #
+##########
+
+ceph_grafana_docker_extra_env:
--- a/roles/ceph-grafana/templates/grafana-server.service.j2
+++ b/roles/ceph-grafana/templates/grafana-server.service.j2
@@ -21,6 +21,7 @@ ExecStart=/usr/bin/{{ container_binary }} run --rm --name=grafana-server \
   --memory={{ grafana_container_memory }}GB \
   --memory-swap={{ grafana_container_memory * 2 }}GB \
   -e GF_INSTALL_PLUGINS={{ grafana_plugins|join(',') }} \
+  {{ ceph_grafana_docker_extra_env }} \
   {{ grafana_container_image }}
 ExecStop=-/usr/bin/{{ container_binary }} stop grafana-server
 Restart=always


roles/ceph-grafana/templates/grafana-server.service.j2
```diff
@@ -21,6 +21,7 @@ ExecStart=/usr/bin/{{ container_binary }} run --rm --name=grafana-server \
   --memory={{ grafana_container_memory }}GB \
   --memory-swap={{ grafana_container_memory * 2 }}GB \
   -e GF_INSTALL_PLUGINS={{ grafana_plugins|join(',') }} \
+  {{ ceph_grafana_docker_extra_env }} \
```
dsavineau: I think we should avoid using docker for new variables name and use a more generic word like container.

Also even if it's not the initial purpose of this PR but I think we should do the same for prometheus right ?

roles/ceph-grafana/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+##########
+# DOCKER #
+##########
+
+ceph_grafana_docker_extra_env:
```
dsavineau: We probably need to move this into ceph-defaults like the other grafana variables (even if it also make sense to put this here) and run the `generate_group_vars_sample.sh` script

roles/ceph-grafana/templates/grafana-server.service.j2
BenoitKnecht: I tried to follow the current convention (`ceph_mon_docker_extra_env`, `ceph_osd_docker_extra_env`, etc.), but I can change it to `ceph_grafana_container_extra_env` if you prefer.

I can do the same for `prometheus` and `alertmanager`.

roles/ceph-grafana/defaults/main.yml
BenoitKnecht: Again, I followed what has been done for `ceph-osd` or `ceph-mon`; in those cases the variables are defined in the role's `defaults` directory, rather than `ceph-defaults`.

I can move this one to `ceph-defaults`, but that would make it a bit inconsistent; do you think it's the right solution nonetheless?

PR #4896, opened 2020-01-08
shrink-mds: do not play ceph-facts entirely

guits:
We only need to set `container_binary`.
Let's use `tasks_from` option.


guits:
jenkins test centos-container-shrink_mds

guits:
jenkins test centos-non_container-shrink_mds

PR #4898, opened 2020-01-08
ceph-iscsi: add ceph-iscsi stable repositories

dsavineau:
This commit adds the support of the ceph-iscsi stable repository when
use ceph_repository community instead of always using the devel
repositories.
We're still using the devel repositories for rtslib and tcmu-runner in
both cases (dev and community).


dsavineau:

PR #4905, opened 2020-01-09
shrink-rgw: refact global workflow

dsavineau:
Instead of running the ceph roles against localhost we should do it
on the first mon.
The ansible and inventory hostname of the rgw nodes could be different.
Ensure that the rgw instance to remove is present in the cluster.
Fix rgw service and directory path.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1677431


dsavineau:
jenkins test centos-container-shrink_rgw

dsavineau:
jenkins test centos-non_container-shrink_rgw

PR #4921, opened 2020-01-10
tests/setup: update mount options on EL 8

dsavineau:
The nobarrier mount flag doesn't exist anymoer on XFS in the EL 8
kernel. That's why the task wasn't working on those systems.
We can still use the other options instead of skipping the task.


dsavineau:

PR #4936, opened 2020-01-14
container: move lvm2 package installation

dsavineau:
Before this patch, the lvm2 package installation was done during the
ceph-osd role.
However we were running ceph-volume command in the ceph-config role
before ceph-osd. If lvm2 wasn't installed then the ceph-volume command
fails:
```console
error checking path "/run/lock/lvm": stat /run/lock/lvm: no such file or
directory
```
This wasn't visible before because lvm2 was automatically installed as
docker dependency but it's not the same for podman on CentOS 8.

(cherry picked from commit de8f2a9f83194e465d10207c7ae0569700345b9c)


dsavineau:

PR #4945, opened 2020-01-15
Ignore failure when attempting to purge node_exporter

jcmdln:
When a deployment fails or is interrupted early, this task prevents
purging the deployment from completing.  By ignoring failures for this
task, purging will continue as expected.


dsavineau:
This is already handle in #4882

PR #4957, opened 2020-01-17
Add OpenStack client.manila user

noonedeadpunk:
According to manila docs [1] it requires it's own user for interacting
with CephFS. This patch adds this user into list of openstack_keys

[1] https://docs.openstack.org/manila/latest/admin/cephfs_driver.html#authorizing-the-driver-to-communicate-with-ceph

Signed-off-by: Dmitriy Rabotyagov <drabotyagov@vexxhost.com>

### DIFF
--- a/roles/ceph-config/templates/ceph.conf.j2
+++ b/roles/ceph-config/templates/ceph.conf.j2
@@ -73,6 +73,12 @@ admin socket = {{ rbd_client_admin_socket_path }}/$cluster-$type.$id.$pid.$cctid
 log file = {{ rbd_client_log_file }} # must be writable by QEMU and allowed by SELinux or AppArmor
 {% endif %}
 
+{% if openstack_config | bool and inventory_hostname in groups.get(client_group_name, []) and openstack_keys | selectattr('name', 'equalto', 'client.manila') | list | length > 0 %}
+[client.manila]
+client mount gid = 0
+client mount uid = 0
+{% endif %}
+
 {% if inventory_hostname in groups.get(osd_group_name, []) %}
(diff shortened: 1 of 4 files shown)


noonedeadpunk:

noonedeadpunk:

noonedeadpunk:

noonedeadpunk:

noonedeadpunk:

noonedeadpunk:
Oh, wait for a while, I think I've forgotten smth - probably need to adjust another variable

roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -73,6 +73,12 @@ admin socket = {{ rbd_client_admin_socket_path }}/$cluster-$type.$id.$pid.$cctid
 log file = {{ rbd_client_log_file }} # must be writable by QEMU and allowed by SELinux or AppArmor
 {% endif %}
 
+{% if openstack_config | bool %}
+[client.manila]
+client mount gid = 0
+client mount uid = 0
+{% endif %}
```
dsavineau: This should be set only for clients node AND when client.manila is present in the `openstack_keys` list (ie: not hardcoding the manila keyring name)

noonedeadpunk: Yeah, indeed. Actually probably better idea would be to include clients conditionally or allow users to override their names but I tried to follow current concept...

noonedeadpunk:

noonedeadpunk:

noonedeadpunk:

PR #4887, opened 2020-01-07
ceph-rgw: Fix custom pool size setting

BenoitKnecht:
RadosGW pools can be created by setting

```yaml
rgw_create_pools:
  .rgw.root:
    pg_num: 512
    size: 2
```

for instance. However, doing so would create pools of size
`osd_pool_default_size` regardless of the `size` value. This was due to
the fact that the Ansible task used

```
{{ item.size | default(osd_pool_default_size) }}
```

as the pool size value, but `item.size` is always undefined; the
correct variable is `item.value.size`.

Signed-off-by: Benoît Knecht <bknecht@protonmail.ch>

PR #4958, opened 2020-01-20
ceph-rgw: Fix customize pool size "when" condition

BenoitKnecht:
In 3c31b19ab39f297635c84edb9e8a5de6c2da7707, I fixed the `customize pool
size` task by replacing `item.size` with `item.value.size`. However, I
missed the same issue in the `when` condition.


BenoitKnecht:
See #4887.

dsavineau:
Thanks for the follow up

PR #4959, opened 2020-01-20
nfs: refact ganesha log directory creation

noonedeadpunk:
Some ganesha packages do not create ganesha log directories
while it's expected to be created while changing it's permissions.
Additionally it's no much sense in doing that as a separate task,
so directory is created as correct permissions are set with creation of
the rest required directories.

Signed-off-by: Dmitriy Rabotyagov <drabotyagov@vexxhost.com>

### DIFF
--- a/roles/ceph-nfs/tasks/pre_requisite_non_container.yml
+++ b/roles/ceph-nfs/tasks/pre_requisite_non_container.yml
@@ -29,8 +29,8 @@
   file:
     path: "{{ item.name }}"
     state: directory
-    owner: "ceph"
-    group: "ceph"
+    owner: "{{ item.owner | default('ceph') }}"
+    group: "{{ item.group | default('ceph') }}"
     mode: "{{ ceph_directories_mode }}"
   with_items:
     - { name: "/var/lib/ceph/bootstrap-rgw", create: "{{ nfs_obj_gw }}" }
@@ -38,6 +38,7 @@
     - { name: "/var/lib/ceph/radosgw/{{ cluster }}-rgw.{{ ansible_hostname }}", create: "{{ nfs_obj_gw }}" }
     - { name: "{{ rbd_client_admin_socket_path }}", create: "{{ nfs_obj_gw }}" }
     - { name: "/var/log/ceph", create: true }
+    - { name: "/var/log/ganesha", create: true, owner: root, group: root }
     - { name: "/var/run/ceph", create: true }
   when: item.create | bool
 
@@ -85,10 +86,3 @@
             owner: "ceph"
             group: "ceph"
             mode: "0600"
-
-- name: change ownership on /var/log/ganesha
-  file:
-    path: /var/log/ganesha
-    owner: "root"
-    group: "root"
-    mode: "0755"


guits:
Hi @noonedeadpunk could you rebase please?

dsavineau:

noonedeadpunk:
> > Some ganesha packages do not create ganesha log directories
> > while it's expected to be created while changing it's permissions.
> 
> So what's the point with this PR since this is already present in the current code ?
> I mean the subject of the PR/commit is wrong since it's already working.

Do not really understand what you mean. In the current code it just tries to change ownership of file/directory, since default state is file (and state is not defined there). So in case /var/log/ganesha won't exist, this task will fail.
Commit changes behaviour by adding `state: directory`, so directory will be created if it do not exist, and it will be set required permissions.

dsavineau:
> Do not really understand what you mean. In the current code it just tries to change ownership of file/directory, since default `state` is `file` (and state is not defined there). So in case /var/log/ganesha won't exist, this task will fail.
> Commit changes behaviour by adding `state: directory`, so directory will be created if it do not exist, and it will be set required permissions.

You're absolutely right, I thought the `state: directory` was present on that task.

> Some ganesha packages do not create ganesha log directories

Do you have an exemple of this ? because at the end it's more a packaging issue rather a ceph-ansible (ie: we don't need to do this work twice)
We don't see this on RHEL/CentOS 7/8 with nfs-ganesha 2.8 and 3.2.
Is it Ubuntu specific ?

```console
# rpm -qa nfs-ganesha
nfs-ganesha-2.8.1.2-0.1.el7.x86_64
# rpm -ql nfs-ganesha | grep /var/log
/var/log/ganesha

# rpm -qa nfs-ganesha
nfs-ganesha-2.8.2-3.el8cp.x86_64
# rpm -ql nfs-ganesha | grep /var/log
/var/log/ganesha

# rpm -qa nfs-ganesha
nfs-ganesha-3.2-1.el8.x86_64
# rpm -ql nfs-ganesha | grep /var/log
/var/log/ganesha
```

noonedeadpunk:
Yes, it is Ubuntu only, since bionic package for Ganesha is not present in http://download.ceph.com/nfs-ganesha/deb-V2.7-stable/nautilus/dists/ and nfs-ganesha provided in Ubuntu upstream repos do not create this directory.
Actually I'd say it won't hurt, as we're not doing job twice - if folder is already there, just ownership will be changed as it was previously, but if it's not here - it will be created with these permissions instead of failed play.

dsavineau:
> Yes, it is Ubuntu (and probably Debian?) only, since bionic package for Ganesha is not present in http://download.ceph.com/nfs-ganesha/deb-V2.7-stable/nautilus/dists/ and nfs-ganesha provided in Ubuntu upstream repos do not create this directory.
> Actually I'd say it won't hurt, as we're not doing job twice - if folder is already there, just ownership will be changed as it was previously, but if it's not here - it will be created with these permissions instead of failed play.

Do you mind open an issue about this on the nfs-ganesha project ?

noonedeadpunk:
@dsavineau I'm not sure about what exactly to claim on tbh - does ganesha-nfs  do packaging for http://download.ceph.com/nfs-ganesha/ ? Or regarding ubuntu packages not creating log directory?

dsavineau:
> @dsavineau I'm not sure about what exactly to claim on tbh - does ganesha-nfs do packaging for http://download.ceph.com/nfs-ganesha/ ? Or regarding ubuntu packages not creating log directory?

AFAIK the Ubuntu nfs-ganesha packages on http://download.ceph.com/nfs-ganesha/ are using https://github.com/nfs-ganesha/nfs-ganesha-debian

PR #4688, opened 2019-10-28
rgw: extend automatic rgw pool creation capability

dsavineau:
Add support for erasure code pools.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1731148

Signed-off-by: Ali Maredia <amaredia@redhat.com>
Co-authored-by: Dimitri Savineau <dsavinea@redhat.com>

PR #4960, opened 2020-01-20
ceph-rgw: Support creation of erasure coded pools

BenoitKnecht:
Add `type` and `profile` attributes to `rgw_create_pools` items; `type`
can be `replicated` (default) or `erasure`, and `profile` can be set to
an existing erasure code profile, if one wants to use something else
than the default.

For instance, the following will create a replicated
`default.rgw.buckets.index` pool, and an erasure-coded
`default.rgw.buckets.data` pool (with the default erasure code profile):

```
rgw_create_pools:
  default.rgw.buckets.index:
    pg_num: 512
  default.rgw.buckets.data:
    type: erasure
    pg_num: 512
```

### DIFF
--- a/roles/ceph-rgw/tasks/main.yml
+++ b/roles/ceph-rgw/tasks/main.yml
@@ -26,7 +26,7 @@
   when: rgw_create_pools is defined
   block:
     - name: create rgw pools if rgw_create_pools is defined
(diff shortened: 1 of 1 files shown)


dsavineau:
Closing in favor of #4688

PR #4967, opened 2020-01-21
common: add a default value for ceph_directories_mode

guits:
Since this variable makes it possible to customize the mode for ceph
directories, let's make it a bit more explicit by adding a default value
in ceph-defaults.


guits:

PR #4985, opened 2020-01-24
filestore-to-bluestore: don't fail when with no PV

dsavineau:
When the PV is already removed from the devices then we should not fail
to avoid errors like:

stderr: No PV found on device /dev/sdb.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1729267


dsavineau:
jenkins test centos-container-filestore_to_bluestore

dsavineau:
jenkins test centos-non_container-filestore_to_bluestore

PR #4991, opened 2020-01-27
node_exporter is only enabled when dashboard is enabled

0X1A:
- purge_cluster failed when dashboard_enabled was false due to
	not having the node-exporter container image


dsavineau:
This is already handle in #4882

0X1A:
@dsavineau 👍

PR #4993, opened 2020-01-27
dashboard: allow configuring multiple grafana host

dsavineau:
When using multiple grafana hosts then we push set the grafana and
prometheus URL and push the dashboard layout to a single node.

grafana_server_addrs is the list of all grafana nodes and used during
the ceph-dashboard role (on mgr/mon nodes).
grafana_server_addr is the current grafana node used during the
ceph-grafana and ceph-prometheus role (on grafana-server nodes).

We don't have the grafana_server_addr fact duplication code between
external vs collocated nodes.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1784011

### DIFF
--- a/roles/ceph-dashboard/tasks/configure_dashboard.yml
+++ b/roles/ceph-dashboard/tasks/configure_dashboard.yml
@@ -94,12 +94,6 @@
   changed_when: false
   until: ac_result.rc == 0
 
-- name: set grafana url
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-grafana-api-url {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ grafana_port }}"
-  delegate_to: "{{ groups[mon_group_name][0] }}"
-  run_once: true
-  changed_when: false
-
 - name: set grafana api user
   command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-grafana-api-username {{ grafana_admin_user }}"
   delegate_to: "{{ groups[mon_group_name][0] }}"
@@ -113,16 +107,21 @@
   changed_when: false
 
 - name: set alertmanager host
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-alertmanager-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ alertmanager_port }}"
+  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-alertmanager-api-host {{ dashboard_protocol }}://{{ grafana_server_addrs | first }}:{{ alertmanager_port }}"
   delegate_to: "{{ groups[mon_group_name][0] }}"
   run_once: true
   changed_when: false
 
 - name: set prometheus host
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-prometheus-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ prometheus_port }}"
+  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-prometheus-api-host {{ dashboard_protocol }}://{{ grafana_server_addrs | first }}:{{ prometheus_port }}"
   delegate_to: "{{ groups[mon_group_name][0] }}"
   run_once: true
   changed_when: false
+
+- include_tasks: configure_grafana_layouts.yml
+  with_items: '{{ grafana_server_addrs }}'
+  vars:
+    grafana_server_addr: '{{ item }}'
 
 - name: dashboard object gateway management frontend
   when: groups.get(rgw_group_name, []) | length > 0
@@ -219,13 +218,6 @@
       with_items: "{{ groups[iscsi_gw_group_name] }}"
       run_once: true
       when: ip_version == 'ipv6'
-
-- name: inject grafana dashboard layouts
-  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard grafana dashboards update"
-  delegate_to: "{{ groups[mon_group_name][0] }}"
-  run_once: true
-  changed_when: false
-  when: containerized_deployment | bool
 
 - name: disable mgr dashboard module (restart)
   command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} mgr module disable dashboard"
--- a/roles/ceph-dashboard/tasks/configure_grafana_layouts.yml
+++ b/roles/ceph-dashboard/tasks/configure_grafana_layouts.yml
@@ -1,0 +1,13 @@
+---
+- name: set grafana url
+  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-grafana-api-url {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ grafana_port }}"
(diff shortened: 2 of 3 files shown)


dsavineau:

dsavineau:
@fmount FYI

fmount:
@dsavineau  thanks for this patch.
I'm currently testing it in the OSP scenario to double check all the layouts are pushed to the grafana instances.

fmount:
@dsavineau  at the end of this process we'll end up on the mgr side w/ just one (the last) grafana instance configured.
Now, speaking from the OSP point of view (but also if you have multiple grafana instances), according to [1] we should have a VIP (the haproxy frontend) for the dashboard, and we want to update the "grafana-api-url" to the VIP.
Do you think we can have a _ceph_grafana_vip_ variable and after pushing the layouts add a new task to update the endpoints?  
```
- name: set grafana url
  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-grafana-api-url {{ dashboard_protocol }}://{{ CEPH_GRAFANA_VIP }}:{{ grafana_port }}"
  run_once: true
  changed_when: false
  when: ceph_grafana_vip is defined
```

[1] https://docs.ceph.com/docs/master/mgr/dashboard/#haproxy-example-configuration

fmount:
@gfidente @epuertat fyi

dsavineau:
@fmount 

> @dsavineau at the end of this process we'll end up on the mgr side w/ just one (the last) grafana instance configured.

Do you mean only one grafana node is configured (grafana/prometheus services) or the mgr nodes are pointing to only one grafana node ?

Because if it's the second option, then it's already like this before the patch (it was configured to the first node). I thought the goal of this BZ/PR was to push the dashboard layouts to all grafana nodes. Are you seeing a different behaviour ?

> [1] we should have a VIP (the haproxy frontend) for the dashboard, and we want to update the "grafana-api-url" to the VIP.
> Do you think we can have a ceph_grafana_vip variable and after pushing the layouts add a new task to update the endpoints?

This should be possible but I think in a dedicated PR (maybe the title of this PR isn't clear enough).

Do we need to set a VIP url for prometheus/alertmanager too ?

fmount:
> @fmount
> 
> > @dsavineau at the end of this process we'll end up on the mgr side w/ just one (the last) grafana instance configured.
> 
> Do you mean only one grafana node is configured (grafana/prometheus services) or the mgr nodes are pointing to only one grafana node ?
> 
> Because if it's the second option, then it's already like this before the patch (it was configured to the first node). I thought the goal of this BZ/PR was to push the dashboard layouts to all grafana nodes. Are you seeing a different behaviour ?
> 
> > [1] we should have a VIP (the haproxy frontend) for the dashboard, and we want to update the "grafana-api-url" to the VIP.
> > Do you think we can have a ceph_grafana_vip variable and after pushing the layouts add a new task to update the endpoints?
> 
> This should be possible but I think in a dedicated PR (maybe the title of this PR isn't clear enough).
> 
> Do we need to set a VIP url for prometheus/alertmanager too ?

Ok, you're right, it's the second option, we end up with the mgr nodes pointing to only one grafana node (for this reason I'm talking about haproxy integration and the vip parameter).
As you said, I think we can address it in a separeted PR, and discuss with more details in the new thread.
Right now, from a first test in the OSP context, your PR LGTM and I see the expected behaviour.

roles/ceph-dashboard/tasks/set_api_url_host.yml
```diff
@@ -0,0 +1,25 @@
...
+  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-alertmanager-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ alertmanager_port }}"
+  delegate_to: "{{ groups[mon_group_name][0] }}"
+  run_once: true
+  changed_when: false
+
+- name: set prometheus host
+  command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} dashboard set-prometheus-api-host {{ dashboard_protocol }}://{{ grafana_server_addr }}:{{ prometheus_port }}"
+  delegate_to: "{{ groups[mon_group_name][0] }}"
+  run_once: true
+  changed_when: false
```
epuertat: You only needed to iterate through grafana servers (as the `inject grafana dashboard layouts` tasks is applied *only* to the Grafana instance pointed by `grafana-api-url`). Doing the same for Prometheus & Alert Manager is not providing any HA (instead, HAProxy should be used), but simply overwriting the previous value set.

fmount: Right, as @epuertat suggested to push layouts you just need to update the grafana ep and execute the `dashboards update` command.
As @dsavineau pointed out, the endpoint configuration for HA using the VIP can be addressed on another thread.

dsavineau: True, I'll update the PR

fmount:
@dsavineau I made a few tests on your PR and on OSP side it looks ok [1].
The second execution [2] shows the same PR execution removing the prometheus/alertmanager host settings from set_api_url_host.yml (I just moved the two tasks in the configure_dashboard.yaml).
I think we can move forward w/ this.

[1] http://pastefile.fr/21a35bf3e9424611d911b755d2e18ee5
[2] http://pastefile.fr/b5f1e73d9d2d782fc3ebb9a8f832b1c0

dsavineau:
> Just a question: when is this planned to land in downstream?

This is targeted for 4.1. We usually use the same milestone in ceph-ansible than the downstream release.

> That will save also the need of iterating, as `set-grafana-api-url` is set up multiple times just for the later `dashboard update` command to take that value. How do you see that?

This will be easier but landing this into stable-4.0 branch would require to detect and compare the ceph version to either run the legacy code or the new code.
I don't know what is the target to this feature (ie: which nautilus release) so I would prefer to have this dashboard update with multiple values in master only (next stable-5.0) to avoid adding complexity.

wdyt ?

dsavineau:

dsavineau:
jenkins test centos-container-all-in-one

dsavineau:

dsavineau:
jenkins test jenkins test centos-container-all_in_one

dsavineau:

epuertat:
@dsavineau yeah, it makes sense, the outcome achieved would be exactly the same as with this change. Nevertheless, the proposal kept the legacy behaviour, so it wouldn't break this either.

@fmount ?

fmount:
> @dsavineau yeah, it makes sense, the outcome achieved would be exactly the same as with this change. Nevertheless, the proposal kept the legacy behaviour, so it wouldn't break this either.
> 
> @fmount ?

IMHO the most important thing is to avoid messing w/ the changes, but, since the new ceph dashboard feature  (allowing multiple grafana addresses) would be backward compatible (which means it doesn't break the existing code), we can:

1. move forward w/ this PR and make sure we have coverage for 4.1;
2. consider improving the ceph-ansible code for stable-5.0 including the new feature (w/o backporting it?)

The result should avoid breaking code and for ceph-ansible perspective the last one could be seen as an improvement of the existing ceph-dashboard role.

dsavineau:
@epuertat @fmount if we're all on the same page, I'm just waiting for your approval on this PR before merge it

fmount:
> @epuertat @fmount if we're all on the same page, I'm just waiting for your approval on this PR before merge it

+1 for me, waiting for @epuertat 's approval.

PR #5001, opened 2020-01-28
Fixed directory reference

Hobadee:
Per the README.md, infrastructure playbooks should be moved to the root directory before running.  To maintain consistency with this policy, we need to fix the assumption that we are running from within the "infrastructure-playbooks" directory when searching for a dependency.

Also apparently this file was missing a newline at the end which github auto-added.

### DIFF
--- a/infrastructure-playbooks/add-osd.yml
+++ b/infrastructure-playbooks/add-osd.yml
@@ -26,7 +26,7 @@
     delegate_facts_host: True
 
   pre_tasks:
-    - import_tasks: "{{ playbook_dir }}/../raw_install_python.yml"
+    - import_tasks: "{{ playbook_dir }}/raw_install_python.yml"
 
     - name: gather facts
       setup:


dsavineau:
> Per the README.md, infrastructure playbooks should be moved to the root directory before running. To maintain consistency with this policy, we need to fix the assumption that we are running from within the "infrastructure-playbooks" directory when searching for a dependency.
> 
> Also apparently this file was missing a newline at the end which github auto-added.

In fact, the README.md should probably be updated because shoudn't have to copy the infrastructure playbooks anymore since [1]

those playbooks need to be executed like :

```console
$ ansible-playbook -vv -i hosts infrastructure-playbooks/add-osd.yml
```
In this case `playbook_dir` refers to the infrastructure-playbooks directory which doesn't contain the `raw_install_python.yml` file.
This file is located in the root ceph-ansible directory.


[1] https://github.com/ceph/ceph-ansible/commit/a1a871cadee5e86d181e1306c985e620b81fccac

PR #5030, opened 2020-02-03
tox: pin ceph_dev_sha1 value for ceph-volume

dsavineau:
Since this commit [1] there's a race condition in ceph-volume when using
multiple LV in the same VG for the OSDs causing multiple scenarios to
fail.
This temporary workaround pin the ceph dev sha1 to the previous shaman
build working.

[1] https://github.com/ceph/ceph/commit/ec2e53f


dsavineau:
iscsigws are using ceph_dev_sha1 variable for shaman setup and because the sha1 provided is only valid for ceph build then it fails on tcmu-runner...

PR #5032, opened 2020-02-03
ceph-defaults: update default dashboard_enabled variable

utp887:
We should not force the deployment of ceph metrics.
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1797765
Signed-off-by: Randy J. Martinez ramartin@redhat.com


guits:
this has been decided by PM a while ago @utp887

dsavineau:
As @guits mentioned, this was a decision from the PM

PR #5033, opened 2020-02-03
ceph-osd: update crush_rule condition

utp887:
During evaluation, when item.default is not defined the following error is thrown: "dict object' has no attribute 'default'"
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1797774
Signed-off-by: Randy J. Martinez ramartin@redhat.com

### DIFF
--- a/roles/ceph-osd/tasks/crush_rules.yml
+++ b/roles/ceph-osd/tasks/crush_rules.yml
@@ -24,7 +24,7 @@
   with_items: "{{ hostvars[groups[mon_group_name][0]]['crush_rules'] | unique }}"
   delegate_to: '{{ groups[mon_group_name][0] }}'
   run_once: true
-  when: item.default | bool
+  when: item.default | default(omit) | bool
 
 # If multiple rules are set as default (should not be) then the last one is taken as actual default.
 # the with_items statement overrides each iteration with the new one.


roles/ceph-osd/tasks/crush_rules.yml
```diff
@@ -24,7 +24,7 @@
   with_items: "{{ hostvars[groups[mon_group_name][0]]['crush_rules'] | unique }}"
   delegate_to: '{{ groups[mon_group_name][0] }}'
   run_once: true
-  when: item.default | bool
+  when: item.default is defined | bool
```
dsavineau: we should always have the `default` key defined in the `crush_rules` dict

utp887:
"""From BZ: 1797774
TASK [ceph-osd : get id for new default crush rule] ****************************************************************************************************************************************************************
Saturday 01 February 2020  21:41:19 -0500 (0:00:03.549)       0:09:40.305 ***** 
fatal: [tower-osd4]: FAILED! => 
  msg: |-
    The conditional check 'item.default | bool' failed. The error was: error while evaluating conditional (item.default | bool): 'dict object' has no attribute 'default'
  
    The error appears to be in '/usr/share/ceph-ansible/roles/ceph-osd/tasks/crush_rules.yml': line 20, column 3, but may
  
    - name: get id for new default crush rule
      ^ here

Additional info:

Reference of my config used: 
---
crush_rule_hdd:
    name: replicated_hdd_rule
    root: root-hdd
    type: host
    class: hdd
    default: False
crush_rules:
  - "{{ crush_rule_hdd }}"
...
"""

roles/ceph-osd/tasks/crush_rules.yml
utp887: I agree. I made the change to use the default(omit) filter, which will make it so and not fail if you do what I did and leave the key commented out by mistake.
 
Also, do you know where this is documented? upstream/or downstream. 

The only reason I configured crush_rules was because of the [condition](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/main.yml#L90). All I really wanted to do was configure crush hierarchy.

Issue with defaults is the expectation of both [classes](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-mon/defaults/main.yml#L47) ... which leads to issues [here](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-osd/tasks/crush_rules.yml#L16)

dsavineau:
> I agree. I made the change to use the default(omit) filter, which will make it so and not fail if you do what I did and leave the key commented out by mistake.

 `default(omit)` isn't the right solution here because omit + bool doesn't make sense.
Instead this should be something like
```yaml
when:
  - item.default is defined
  - item.default | bool
```

> Also, do you know where this is documented? upstream/or downstream.

I don't think so (except the example in group_vars/mons.yml.sample)

Overall I think this is more about a documentation update than the current code. Otherwise why aren't we doing this for all keys in all `crush_rules` list ? Because if we omit the `name`, `type` or `root` key then you will have the same issue.
And this is true for every dict used in variables provided by users.

> Reference of my config used:
> 
> ```
> ---
> crush_rule_hdd:
>     name: replicated_hdd_rule
>     root: root-hdd
>     type: host
>     class: hdd
>     default: False
> crush_rules:
>   - "{{ crush_rule_hdd }}"
> ...
> ```

This is confusing because the config mentioned above is valid and doesn't generate the issue reported.

Issue #4968, opened 2020-01-21
Upgrade error - validate ceph_origin


rlljorge:
Hello there,

I am try upgrade my cluster from mimic to nautilus without success.

Before upgrade I executed site.yaml without errors in mimic release, the cluster is ok and updated.

```console
[root@mon01 ~]# ceph -v
ceph version 13.2.8 (5579a94fafbc1f9cc913a0f5d362953a5d9c3ae0) mimic (stable)
[root@mon01 ~]# ceph -s
  cluster:
    id:     d6fd8a81-377c-47a7-b110-db9800be79cd
    health: HEALTH_OK

  services:
    mon: 3 daemons, quorum mon01,mon02,mon03
    mgr: mon01(active), standbys: mon02, mon03
    osd: 12 osds: 12 up, 12 in
    rgw: 2 daemons active

  data:
    pools:   15 pools, 544 pgs
    objects: 3.05 M objects, 7.4 TiB
    usage:   22 TiB used, 65 TiB / 87 TiB avail
    pgs:     543 active+clean
             1   active+clean+scrubbing+deep

  io:
    client:   9.4 MiB/s rd, 4.2 MiB/s wr, 117 op/s rd, 154 op/s wr
```

SO Info:

cat /etc/redhat-release
CentOS Linux release 7.7.1908 (Core)
Kernel:3.10.0-862.14.4.el7.x86_64
Ansible 2.8.8

[root@ceph-deploy-799889c476-lk27p ceph-ansible]# cat /etc/ansible/hosts
```yaml
[mons]
mon01 monitor_address=10.250.6.1
mon02 monitor_address=10.250.6.2
mon03 monitor_address=10.250.6.3

[osds]
osd01 osd_scenario="lvm" lvm_volumes="[{'data': 'data-lv1', 'data_vg': 'data-vg1', 'wal': 'wal-lv1', 'wal_vg': 'wal-vg1'},{'data': 'data-lv2', 'data_vg': 'data-vg2', 'wal': 'wal-lv2', 'wal_vg': 'wal-vg2'},{'data': 'data-lv3', 'data_vg': 'data-vg3', 'wal': 'wal-lv3', 'wal_vg': 'wal-vg3'}]"
[...]

PR #5035, opened 2020-02-04
Fix rolling update playbook

BenoitKnecht:
Running `ansible-playbook rolling_update.yml` to upgrade a Ceph cluster from 14.2.6 to 14.2.7 failed due to the three issues addressed by the three commits in this PR:

- Wrong `dashboard.yml` playbook import path;
- `norebalance` flag remains set after upgrade;
- `ceph.conf.j2` template fails to render if `mon_host_v1` is disabled.

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -951,7 +951,7 @@
       vars:
         msgr2_migration: True
 
-- import_playbook: ../dashboard.yml
+- import_playbook: dashboard.yml
   when:
     - dashboard_enabled | bool
     - groups.get(grafana_server_group_name, []) | length > 0
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -263,6 +263,26 @@
       delay: "{{ health_mon_check_delay }}"
       when: containerized_deployment | bool
 
+    - name: unset osd flags
+      command: ceph --cluster {{ cluster }} osd unset {{ item }}
+      with_items:
+        - noout
+        - norebalance
+      delegate_to: "{{ mon_host }}"
+      when:
+        - inventory_hostname == groups[mon_group_name] | last
+        - not containerized_deployment | bool
+    - name: unset containerized osd flags
+      command: >
+        {{ container_binary }} exec ceph-mon-{{ hostvars[mon_host]['ansible_hostname'] }} ceph --cluster {{ cluster }} osd unset {{ item }}
+      with_items:
+        - noout
+        - norebalance
+      delegate_to: "{{ mon_host }}"
+      when:
+        - inventory_hostname == groups[mon_group_name] | last
+        - containerized_deployment | bool
+
 
 - name: reset mon_host
   hosts: "{{ mon_group_name|default('mons') }}"
--- a/roles/ceph-config/templates/ceph.conf.j2
+++ b/roles/ceph-config/templates/ceph.conf.j2
@@ -38,7 +38,7 @@ osd pool default crush rule = {{ osd_pool_default_crush_rule | default(ceph_osd_
 fsid = {{ fsid }}
 mon host = {% if nb_mon > 0 %}
 {% for host in _monitor_addresses -%}
-{% if msgr2_migration | default(False) %}
+{% if msgr2_migration | default(False) and mon_host_v1.enabled %}
 [{{ "v2:" + host.addr + mon_host_v2.suffix }},{{ "v1:" + host.addr + mon_host_v1.suffix }}]
 {%- elif not msgr2_migration | default(False) and rolling_update -%}
 {{ host.addr }}


BenoitKnecht:
Hm, the playbook syntax check is unhappy with the new import path, it cannot find the `dashboard.yml` playbook anymore.

It's a bit ugly, but I could create a symlink in `infrastructure-playbooks` pointing to `../dashboard.yml`; that would make the CI happy, and would still allow `rolling_update.yml` to be run from the root of the repository without issues.

BenoitKnecht:
Thanks for the review! I've replied to your comments inline below.

> Two of the three issues mentioned are wrong
> 
> > * Wrong dashboard.yml playbook import path;
> 
> that's because you're running the rolling_update.yml playbook from the infrastructure-playbook directory which is wrong.
> You should use either the full or relative path to the playbook
> 
> ```
> $ ansible-playbook -vv -i hosts infrastructure-playbook/rolling_update.yml
> or
> $ ansible-playbook -vv -i /path/hosts /path/ceph-ansible/infrastructure-playbook/rolling_update.yml
> ```

Actually, I'm running it as described in `infrastructure-playbooks/README.md` (i.e. I copied `rolling_update.yml` to ceph-ansible's root directory, and ran `ansible-playbook rolling_update.yml`), which doesn't work. I had also tried `ansible-playbook infrastructure-playbook/rolling_update.yml`, but it failed too (for a different reason, I think because it didn't find my `group_vars` and `host_vars` directories in that case, but I'm not sure).

So either it's a documentation issue and the README needs to be updated, or the playbook import path does need to be updated.

> > * norebalance flag remains set after upgrade;
> 
> This flag isn't present on master so I assume you're talking about the stable-4.0 branch (due to nautilus).
> So this was already fixed by [1]

You're absolutely right, I was using the `stable-4.0` branch and didn't realize that the issue didn't apply to `master`.

> > * ceph.conf.j2 template fails to render if mon_host_v1 is disabled.
> 
> This looks like a valid issue
> 
> [1] [675b678](https://github.com/ceph/ceph-ansible/commit/675b6788f4d0473fc407f889534d6b0a046e4ba8)

dsavineau:
> Actually, I'm running it as described in `infrastructure-playbooks/README.md` (i.e. I copied `rolling_update.yml` to ceph-ansible's root directory, and ran `ansible-playbook rolling_update.yml`), which doesn't work. I had also tried `ansible-playbook infrastructure-playbook/rolling_update.yml`, but it failed too (for a different reason, I think because it didn't find my `group_vars` and `host_vars` directories in that case, but I'm not sure).
> 
> So either it's a documentation issue and the README needs to be updated, or the playbook import path does need to be updated.

It's a documentation issue as I mentioned in https://github.com/ceph/ceph-ansible/pull/5001#issuecomment-579573413

BenoitKnecht:
@dsavineau I tried again running `ansible-playbook infrastructure-playbook/rolling_update.yml`, and I can confirm that it doesn't work. It fails with

```
TASK [ceph-facts : set_fact _current_monitor_address]
*******************************************************************************
Thursday 06 February 2020  07:19:56 +0000 (0:00:01.105)       0:02:20.912 *****
fatal: [mon-01]: FAILED! => 
  msg: '''_monitor_addresses'' is undefined'
fatal: [mon-02]: FAILED! => 
  msg: '''_monitor_addresses'' is undefined'
fatal: [mon-03]: FAILED! => 
  msg: '''_monitor_addresses'' is undefined'
fatal: [osd-01]: FAILED! => 
  msg: '''_monitor_addresses'' is undefined'
fatal: [osd-02]: FAILED! => 
  msg: '''_monitor_addresses'' is undefined'
fatal: [osd-03]: FAILED! => 
  msg: '''_monitor_addresses'' is undefined'
```

(I have `monitor_interface` defined in `group_vars/all.yml`.)

If instead I copy the playbook to the root of the repository and run `ansible-playbook rolling_update.yml`, it works fine; so I think the README is correct, it is still required to copy those playbooks, otherwise Ansible looks for `group_vars` and `host_vars` in the wrong place.

Are you seeing a different behavior?

dsavineau:
> @dsavineau I tried again running `ansible-playbook infrastructure-playbook/rolling_update.yml`, and I can confirm that it doesn't work. It fails with
> 
> ```
> TASK [ceph-facts : set_fact _current_monitor_address]
> *******************************************************************************
> Thursday 06 February 2020  07:19:56 +0000 (0:00:01.105)       0:02:20.912 *****
> fatal: [mon-01]: FAILED! => 
>   msg: '''_monitor_addresses'' is undefined'
> fatal: [mon-02]: FAILED! => 
>   msg: '''_monitor_addresses'' is undefined'
> fatal: [mon-03]: FAILED! => 
>   msg: '''_monitor_addresses'' is undefined'
> fatal: [osd-01]: FAILED! => 
>   msg: '''_monitor_addresses'' is undefined'
> fatal: [osd-02]: FAILED! => 
>   msg: '''_monitor_addresses'' is undefined'
> fatal: [osd-03]: FAILED! => 
>   msg: '''_monitor_addresses'' is undefined'
> ```
> 
> (I have `monitor_interface` defined in `group_vars/all.yml`.)
> 
> If instead I copy the playbook to the root of the repository and run `ansible-playbook rolling_update.yml`, it works fine; so I think the README is correct, it is still required to copy those playbooks, otherwise Ansible looks for `group_vars` and `host_vars` in the wrong place.
> 
> Are you seeing a different behavior?

I don't see any behavior on my side because the {group,host}_vars directories and the inventory are correctly configured.

Your {group,host}_vars directories aren't loaded because those directories should be either:

- in the playbook directory
- in the inventory directory

See [1] for more information

So I suppose you have your group_vars directory in the main ceph-ansible directory and the inventory file somewhere in /etc/ansible/hosts (as you're not using -i).
This is working on initial deployment because the main playbooks (site.yml and site-container.yml) are already in the root ceph-ansible directory.

See also https://github.com/ceph/ceph-ansible/issues/4968 for a similar issue

[1] https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#organizing-host-and-group-variables

PR #5036, opened 2020-02-04
ceph-mon: use interactive session with aliases

dsavineau:
When using ceph aliases with commands that require manual intervention
to stop then the command will keep running inside the container (like
using Ctrl+c).
For handling this, we should use the interactive session option (-it)
with the docker commands.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1797874


dsavineau:

dsavineau:

dsavineau:
jenkins test playbook syntax

PR #5040, opened 2020-02-06
The _filtered_clients list should intersect with ansible_play_batch

fultonj:
Client configuration with --limit fails without this patch
because certain tasks are only done to the first host in the
_filtered_clients list and it's likely that first host will
not be included in what's sepcified with --limit. To fix this
the _filtered_clients list should be built from all clients
in the inventory that are also in the running play.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1798781

### DIFF
--- a/roles/ceph-client/tasks/create_users_keys.yml
+++ b/roles/ceph-client/tasks/create_users_keys.yml
@@ -19,7 +19,7 @@
   add_host:
     name: "{{ item }}"
     groups: _filtered_clients
-  with_items: "{{ groups[client_group_name] }}"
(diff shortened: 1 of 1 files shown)


dsavineau:
@fultonj could you amend and repush your change. we fixed the CI so we should see less failures (this is to avoid manually triggering each scenario)

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

PR #5044, opened 2020-02-06
ceph-iscsi: don't use ceph_dev_xxx variables

dsavineau:
Using ceph_dev_branch and ceph_dev_sha1 for configuring ceph-iscsi
repositories from shaman doesn't make sense because the ceph devel
branches and sha1 aren't compatible with ceph-iscsi devel.
Instead we could rely on the master branch and the latest sha1.
Currently it's not possible to using a custom ceph branch/sha1 value
with iscsi setup otherwise the repository setup will fail.


dsavineau:

dsavineau:

dsavineau:

PR #5050, opened 2020-02-10
container: use --storage with podman rm

dsavineau:
Until we have podman 1.7.0 that fixes an issue where the container is
still referenced after being stopped and removed (--rm) then we should
use --storage with podman rm commands.


dsavineau:
maybe https://github.com/ceph/ceph-ansible/pull/5051 is a better solution

PR #5059, opened 2020-02-13
ceph-prometheus: add alertmanager HA config

dsavineau:
When using multiple alertmanager nodes (via the grafana-server group)
then we need to specify the other peers in the configuration.

https://prometheus.io/docs/alerting/alertmanager/#high-availability
https://github.com/prometheus/alertmanager#high-availability

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1792225

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -777,6 +777,7 @@ dummy:
 #alertmanager_data_dir: /var/lib/alertmanager
 #alertmanager_conf_dir: /etc/alertmanager
 #alertmanager_port: 9093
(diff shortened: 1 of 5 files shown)


dsavineau:
@fmount You'll probably have to do the firewall modification on TripleO too

fmount:
> @fmount You'll probably have to do the firewall modification on TripleO too

According to [1] we should be ok, but thanks for notifying me about this change!

[1] https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/ceph-ansible/ceph-grafana.yaml#L113

PR #5028, opened 2020-02-03
ceph-rgw-loadbalancer: Fix SSL newline issue

dsavineau:
The ad7a5da commit introduced a regression when using TLS on haproxy
via the haproxy_frontend_ssl_certificate variable.
This cause the "stats socket" and the "tune.ssl.default-dh-param"
parameters to be on the same line resulting haproxy failing to start.

[ALERT] 351/140240 (21388) : parsing [xxxxx] : 'stats socket' : unknown
keyword 'tune.ssl.default-dh-param'. Registered
[ALERT] 351/140240 (21388) : Fatal errors found in configuration.

Fixes: #4869

Signed-off-by: Florian Faltermeier <florian.faltermeier@uibk.ac.at>

PR #5062, opened 2020-02-16
fix mistake in the template/haproxy.cfg.j2

rongzhen-zhan:
There is a mistake in template of haproxy


dsavineau:
This is already handled by https://github.com/ceph/ceph-ansible/pull/5028

PR #5076, opened 2020-02-17
tests: pg num should be a power of two number

dsavineau:
This patch changes the pg_num value of the rgw pools foo and bar to be
a power of two number.


dsavineau:
jenkins test playbook syntax

dsavineau:

Issue #5060, opened 2020-02-14
Support allow_standby_replay for CephFS


secret104278:
When we have multiple MDSs deployed, it's better to have hot standby MDS so that MDS failover will become much more faster.

we should add option to run `ceph fs set <FSNAME> allow_standby_replay true` to enable this function

PR #5082, opened 2020-02-19
ceph-mds: Allow standby replay mds to be configured

secret104278:
Allow standby replay mds to be configured.
Resolve https://github.com/ceph/ceph-ansible/issues/5060

https://docs.ceph.com/docs/master/cephfs/standby/#configuring-standby-replay

Signed-off-by: Yu Chih secret104278@gmail.com

### DIFF
--- a/roles/ceph-mds/tasks/create_mds_filesystems.yml
+++ b/roles/ceph-mds/tasks/create_mds_filesystems.yml
@@ -96,3 +96,8 @@
   when:
     - mds_max_mds > 1
     - not rolling_update
+
+- name: set allow_standby_replay
+  command: "{{ container_exec_cmd | default('') }} ceph --cluster {{ cluster }} fs set {{ cephfs }} allow_standby_replay {{ mds_allow_standby_replay }}"
+  changed_when: false
+  delegate_to: "{{ groups[mon_group_name][0] }}"
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -408,6 +408,13 @@ dummy:
 ## MDS options
 #
 #mds_max_mds: 1
+# Once an MDS has entered the standby-replay state, it will only be 
+# used as a standby for the rank that it is following. If another
+# rank fails, this standby-replay daemon will not be used as a replacement,
+# even if no other standbys are available. For this reason, it is advised 
+# that if standby-replay is used then every active MDS should have a 
+# standby-replay daemon.
+#mds_allow_standby_replay: false
 
 ## Rados Gateway options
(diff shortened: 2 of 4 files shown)


dsavineau:
and run the generate_group_vars_sample.sh script to populate the group_vars samples.

It would be also great to rebase the PR

dsavineau:
@batrick Is there any scenario where we don't want this setting set to true ? Should we ensure that we have at least 1 standby mds node available ? (ie: what happens if we are 3 mds and all are active)

roles/ceph-mds/tasks/create_mds_filesystems.yml
```diff
@@ -77,3 +77,8 @@
   when:
     - mds_max_mds > 1
     - not rolling_update
+
+- name: set allow_standby_replay
+  command: "{{ container_exec_cmd | default('') }} ceph --cluster {{ cluster }} fs set {{ cephfs }} allow_standby_replay {{ mds_allow_standby_replay | bool | lower }}"
```
dsavineau: You don't need the `| lower` filter

batrick:
> @batrick Is there any scenario where we don't want this setting set to true ? Should we ensure that we have at least 1 standby mds node available ? (ie: what happens if we are 3 mds and all are active)

Ya, probably this should be off by default. User needs to be sure they have enough standbys to turn this on. I don't think ceph-ansible wants to get in the business of making that judgement.

secret104278:
> > @batrick Is there any scenario where we don't want this setting set to true ? Should we ensure that we have at least 1 standby mds node available ? (ie: what happens if we are 3 mds and all are active)
> 
> Ya, probably this should be off by default. User needs to be sure they have enough standbys to turn this on. I don't think ceph-ansible wants to get in the business of making that judgement.

I have change the default value to false and add some warning that user should make sure they have enough mds to take over.

roles/ceph-mds/tasks/create_mds_filesystems.yml
```diff
@@ -96,3 +96,8 @@
   when:
     - mds_max_mds > 1
     - not rolling_update
+
+- name: set allow_standby_replay
+  command: "{{ container_exec_cmd | default('') }} ceph --cluster {{ cluster }} fs set {{ cephfs }} allow_standby_replay {{ mds_allow_standby_replay | bool }}"
```
dsavineau: you shouldn't use the bool filter here

guits:

guits:

guits:
@secret104278  UP 🙂

PR #5086, opened 2020-02-20
ceph-rgw: increase connection timeout to 10

dsavineau:
5s as a connection timeout could be low in some setup. Let's increase
it to 10s.


dsavineau:

PR #5101, opened 2020-02-27
ceph-rgw-loadbalancer: Add keepalived non-distro install option

slam1987:
Currently keepalived installs the distro version which is 2 years behind.
Add feature which allows the user the option to specify the version number of keepalived to install.

Signed-off-by: Stanley Lam <stanleylam_604@hotmail.com>


slam1987:
bump

Issue #5104, opened 2020-02-28
key validation


fultonj:
When users supply ceph keys they sometimes make mistakes in the format. This will cause the deployment to fail and an error will be displayed that a ceph command failed but it may not be clear to the user that their key wasn't valid. 

Because keys must match the following regex could there be an additional validation in roles/ceph-validate/ which fails early if the keys don't match the regex? It would make for a nicer user experience and perhaps reduce bug reports or requests for help.

 "^[a-zA-Z0-9+/]{38}==$"

PR #5105, opened 2020-02-28
ceph-validate: add key format validation

dsavineau:
If the user provides manually the key value for a specific keyring then
there's not valation on the content which could lead to unexpected
failures in the ceph_key module.

Closes: #5104


dsavineau:

PR #5109, opened 2020-03-02
Check if containerized_deployment is 'true' before validating contain…

ashishkumsingh:
Check if containerized_deployment is 'true' before validating container registry credentials.
In Baremetal deployment, we may need to pass ceph_docker_registry_username and ceph_docker_registry_password

Fixes BZ : https://bugzilla.redhat.com/show_bug.cgi?id=1808982

Signed-off-by: Ashish Singh <assingh@redhat.com>


dsavineau:
This is wrong.
The container registry authentication is also used for non containerized deployment due to the ceph dashboard (100% containerized).
So we only need to check the `ceph_docker_registry_auth` variable.

PR #5126, opened 2020-03-05
Fix HAProxy httpchk parameter for Ceph RGW

nikosmeds:
Fix the `httpchk` parameter to use valid HTTP verb.

Currently traffic will be forwarded to offline `radosgw` services. These changes resolve the issue.

Signed-off-by: Niko Smeds nikosmeds@gmail.com

### DIFF
--- a/roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
+++ b/roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
@@ -43,9 +43,9 @@ frontend rgw-frontend
 backend rgw-backend
     option forwardfor
     balance static-rr
-    option httpchk Get /
+    option httpchk HEAD /
 {% for host in groups[rgw_group_name] %}
 {% for instance in hostvars[host]['rgw_instances'] %}
-	server {{ 'server-' + hostvars[host]['ansible_hostname'] + '-' + instance['instance_name'] }} {{ instance['radosgw_address'] }}:{{ instance['radosgw_frontend_port'] }} weight 100
+	server {{ 'server-' + hostvars[host]['ansible_hostname'] + '-' + instance['instance_name'] }} {{ instance['radosgw_address'] }}:{{ instance['radosgw_frontend_port'] }} weight 100 check
 {% endfor %}
 {% endfor %}


slam1987:

slam1987:
bump

nikosmeds:
@dsavineau could I bother you for a review?

Without these changes HAProxy will continue to send traffic to down backends.

nikosmeds:
Hello admins? Without this change traffic is sent to unhealthy backends :(

nikosmeds:
:(

nikosmeds:
Bump

roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
```diff
@@ -43,9 +43,8 @@ frontend rgw-frontend
 backend rgw-backend
     option forwardfor
     balance static-rr
-    option httpchk Get /
```
BenoitKnecht, line 46: Why remove the `httpchk` option here? It seems better to do a full HTTP check rather than just a TCP check, doesn't it?

BenoitKnecht: To (maybe) answer my own question, the `httpchk` above fails when checks are enabled, because `Get` isn't a valid HTTP verb. It works with this though:

```
    option httpchk GET /
```

nikosmeds: Ah, that makes more sense - thanks @BenoitKnecht. That would be a more appropriate fix. Will give it a shot and update this PR afterwards.

nikosmeds:
@BenoitKnecht appreciate the feedback. I've tested locally and confirmed your suggestion fixes the issue 🙏  PR has been updated.

roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
```diff
@@ -43,7 +43,7 @@ frontend rgw-frontend
...
+    option httpchk GET /
 {% for host in groups[rgw_group_name] %}
 {% for instance in hostvars[host]['rgw_instances'] %}
 	server {{ 'server-' + hostvars[host]['ansible_hostname'] + '-' + instance['instance_name'] }} {{ instance['radosgw_address'] }}:{{ instance['radosgw_frontend_port'] }} weight 100
```
BenoitKnecht, line 49: You should still be adding `check` at the end of this line in order for HAproxy to actually do the HTTP GET requests.

guits:
jenkins test all

guits:
jenkins test all

guits:
@nikosmeds what's the status on this one? Is it still valid?

nikosmeds:
> @nikosmeds what's the status on this one? Is it still valid?

Sorry about the hold up - updated. Rested on my local Ceph deployment and confirmed I can perform rolling stop/starts of RADOS gateway service without any outage.

nikosmeds:
Please hold up on merging this PR. After applying it, I've discovered a few issues:

1. The following error occurs each time a HAProxy service performs the `httpchk`
   ```
   $ sudo tail -f /var/log/ceph/ceph-rgw-server002.rgw0.log
   2020-08-23T18:55:27.301+0200 7fe4e171a700  0 ERROR: client_io->complete_request() returned Broken pipe
   ```
2. One of the backends started flapping. I logged into HAProxy server and manually cURL'd the backend and received the following response
   ```
   <?xml version="1.0" encoding="UTF-8"?><Error><Code>SlowDown</Code><RequestId>tx000000000000000014a91-005f429b10-8b9cb6-default</RequestId><HostId>8b9cb6-default-default</HostId></Error>
   ```

Will look into these issues first.

BenoitKnecht:
> Please hold up on merging this PR. After applying it, I've discovered a few issues:
> 
>     1. The following error occurs each time a HAProxy service performs the `httpchk`
>        ```
>        $ sudo tail -f /var/log/ceph/ceph-rgw-server002.rgw0.log
>        2020-08-23T18:55:27.301+0200 7fe4e171a700  0 ERROR: client_io->complete_request() returned Broken pipe
>        ```

I think you can get rid of this one by replacing

```
option httpchk GET /
```

with

```
option httpchk HEAD /
```

As for the second issue, I haven't experienced it myself, so I'm not sure it's directly related to this change.

nikosmeds:
@BenoitKnecht thanks for the suggestion - sadly that didn't work either.

So far I've tested...

1. `option httpchk` results in **down** HAProxy backends
2. `option httpchk GET /` results in **up** HAProxy backends
   ... but `ERROR: client_io->complete_request() returned Broken pipe` in Ceph logs
3. `option httpchk HEAD /` results in **up** HAProxy backends
   ... but `ERROR: client_io->complete_request() returned Broken pipe` in Ceph logs

Just FYI, options 2+3 _do_ work, however it doesn't seem ideal to use a setting which results in errors in the logs. Will have to look into this further still.

BenoitKnecht:
> @BenoitKnecht thanks for the suggestion - sadly that didn't work either.

I'm very surprised it didn't work for you. I tested it on my (Nautilus) cluster, and the `Broken pipe` errors are completely gone.

Are you sure the change was taken into account by all HAproxy instances that use this particular RadosGW instance as a backend? Can you confirm with `tcpdump` that nothing else is doing `GET /` requests on that backend?

In any case, since these errors are pretty harmless and checking the backend state is definitely better than assuming it's always up, I would suggest merging this PR (preferably the `HEAD /` version); it can always be refined at a later stage if we can find a better implementation.

nikosmeds:
Hmm, okay I retested this and have narrowed it down a bit. First it's worth mentioning that the `Broken pipe` errors are quite infrequent (a couple within a minute) when using the `HEAD /` option.

We have three servers, and both HAProxy and RADOS gateway services are installed on all three (i.e. each server has both HAProxy and RGW installed). I've realized that when `HEAD /` httpchk is added, the pipe errors _only_ occur in HAProxy logs on the same host.

I tested this by applying the change to only one of the servers and reviewing the logs on all three - only that first server had occasional `Broken pipe` messages. Then I rotated through the three servers and confirmed they all experience same issue. Ain't that interesting.

tcpdump shows `HEAD /` requests. I've updated PR btw, as it's not breaking anything... I'll keep looking into this (sadly have some other priorities, so not spending much time digging further at the moment).

BenoitKnecht:
@guits How do you feel about merging this now?

BenoitKnecht:
Shouldn't this be backported to `stable-4.0` and `stable-5.0`?

PR #5131, opened 2020-03-07
openstack_keys: use openstack_cinder_pool.name

berendt:
Instead of volumes as a static string the openstack_cinder_pool.name
variable should be used as with the other keys.


guits:

PR #5141, opened 2020-03-12
osd: support changing default rule even without osd_crush_location

guits:
Creating crush rules even with no crush hierarchy configuration is a
valid scenario so we shouldn't be bound to the first task result (which
configure crush hierarchy) to be able to add new crush rules.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1816989

### DIFF
--- a/roles/ceph-osd/tasks/crush_rules.yml
+++ b/roles/ceph-osd/tasks/crush_rules.yml
@@ -43,7 +43,6 @@
   with_items: "{{ groups[mon_group_name] }}"
   run_once: true
   when:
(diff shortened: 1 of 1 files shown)


guits:

guits:

fultonj:
how difficult is it to put this into ceph-ansible 3.y ?

guits:
> how difficult is it to put this into ceph-ansible 3.y ?

Considering the change, I don't think it's challenging from code/backport point of view, I guess the question is more about a potential 3.3z5 where it would need to land in?

PR #5055, opened 2020-02-12
ceph-grafana: fix grafana_{crt,key} condition

dsavineau:
The grafana_{crt,key} aren't boolean variables but strings. The default
value is an empty string so we should do the conditional on the string
length instead of the bool filter

Closes: #5053

Signed-off-by: Dimitri Savineau <dsavinea@redhat.com>

Issue #5053, opened 2020-02-12
Grafana setup ignoring defined certificates which cause errors in dashboard


styleart:
**Bug Report**

What happened:
ceph-ansible is skipping the copy of defined certificates which causes errors in performance dashboards. The `http_addr` setting is also an issue, since it picks the IP of the server, but certificates are issued for a domain. So if I head to performance dashboards with generated Grafana certs, I get the error: `Error code: MOZILLA_PKIX_ERROR_SELF_SIGNED_CERT`. If i replace the certificates manually, most of dashboards are working, but 

- Dashboard continuously pops the red error "500 - Internal Server Error"
- Pools performance dashboard is broken and I additionally get "TypeError: l.c[t.type] is undefined" error
- ceph-mgr log still reports an error:
```
SSLError: HTTPSConnectionPool(host='10.1.40.10', port=3000): Max retries exceeded with url: /api/dashboards/uid/z99hzWtmk (Caused by SSLError(CertificateError("hostname '10.1.40.10' doesn't match '*.dc1.redacted.com'",),))
```
- Setting the `http_addr` manually to domain also doesn't help =/

How to reproduce it (minimal and precise):
Just run the rolling update playbook with dashboard settings and certificates defined.

**inventory**
```
[mons]
ceph1-40-10.dc1.redacted.com
ceph1-40-11.dc1.redacted.com
ceph1-40-12.dc1.redacted.com

[osds]
ceph1-40-10.dc1.redacted.com
ceph1-40-11.dc1.redacted.com
ceph1-40-12.dc1.redacted.com

[mgrs]
ceph1-40-10.dc1.redacted.com
ceph1-40-11.dc1.redacted.com
ceph1-40-12.dc1.redacted.com

[rgws]
ceph1-40-10.dc1.redacted.com
ceph1-40-11.dc1.redacted.com
ceph1-40-12.dc1.redacted.com

[nfss]
ceph1-40-10.dc1.redacted.com
ceph1-40-11.dc1.redacted.com
ceph1-40-12.dc1.redacted.com

[grafana-server]
ceph1-40-10.dc1.redacted.com
ceph1-40-11.dc1.redacted.com
ceph1-40-12.dc1.redacted.com

[ceph:children]
mons
osds
mgrs
rgws
nfss
grafana-server
```
**snippet from dashboard config in group vars** (everything else is default)
```
dashboard_enabled: True
dashboard_protocol: https
dashboard_port: 8443
[...]

PR #5155, opened 2020-03-17
Fix for TLS dashboard

thotypous:
Setting up a TLS dashboard currently causes some issues, as initially reported in #5053. Problems were partially fixed by #5055, but some issues remain:

- Setting `dashboard_protocol` to `https` causes the dashboard to try accessing Prometheus API though https too. However, Prometheus [does not](https://prometheus.io/docs/guides/tls-encryption/) support TLS natively. Setting up a reverse proxy such as nginx would be needed for this. For now, we just fix the protocol as plain http, which is what Prometheus supports.

- People are currently using `ceph dashboard set-grafana-api-ssl-verify False` as a workaround to get Grafana to work with TLS. In order to get the certificate to validate correctly, we can allow the user to override `grafana_server_addr` with a FQDN listed in the certificate's CN or altsubjects.

Just for reference, I'm deploying with the following config:

```
dashboard_protocol: https
dashboard_admin_user: admin
dashboard_admin_password: ommited
dashboard_crt: 'ceph.crt'
dashboard_key: 'ceph.key'
grafana_admin_user: admin
grafana_admin_password: ommited
grafana_crt: 'ceph.crt'
grafana_key: 'ceph.key'
grafana_server_addr: somedomain.ufscar.br
```

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -768,6 +768,8 @@ dummy:
 # We only need this for SSL (https) connections
 #grafana_crt: ''
 #grafana_key: ''
+# When using https, please fill with a hostname for which grafana_crt is valid.
+#grafana_server_fqdn: ''
 #grafana_container_image: "docker.io/grafana/grafana:5.4.3"
 #grafana_container_cpu_period: 100000
 #grafana_container_cpu_cores: 2
--- a/group_vars/rhcs.yml.sample
+++ b/group_vars/rhcs.yml.sample
@@ -768,6 +768,8 @@ node_exporter_container_image: registry.redhat.io/openshift4/ose-prometheus-node
 # We only need this for SSL (https) connections
 #grafana_crt: ''
 #grafana_key: ''
+# When using https, please fill with a hostname for which grafana_crt is valid.
+#grafana_server_fqdn: ''
 grafana_container_image: registry.redhat.io/rhceph/rhceph-4-dashboard-rhel8:4
 #grafana_container_cpu_period: 100000
 #grafana_container_cpu_cores: 2
--- a/roles/ceph-dashboard/tasks/configure_dashboard.yml
+++ b/roles/ceph-dashboard/tasks/configure_dashboard.yml
@@ -107,13 +107,13 @@
   changed_when: false
 
 - name: set alertmanager host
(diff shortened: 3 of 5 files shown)


group_vars/all.yml.sample
```diff
@@ -768,6 +768,8 @@ dummy:
 # We only need this for SSL (https) connections
 #grafana_crt: ''
 #grafana_key: ''
+# Server hostname for which grafana_crt is valid
+#grafana_server_addr: ''
```
guits, line 772: 2 things:

1/ you shouldn't edit this file directly.
please, edit `roles/ceph-defaults/defaults/main.yml` and run `./generate_group_vars_sample.sh`

2/ That will work as it is, but I wonder whether we would better setting a default value here, then we would have a condition like this on the tasks you've modified below:  

```yaml
  when:
    - grafana_server_addr != "whatever-default-value"
```

roles/ceph-facts/tasks/grafana.yml
```diff
@@ -21,6 +23,7 @@
     grafana_server_addrs: "{{ (grafana_server_addrs | default([]) + [hostvars[item]['ansible_all_ipv4_addresses'] | ips_in_ranges(public_network.split(',')) | first]) | unique }}"
   with_items: "{{ groups.get(grafana_server_group_name, []) }}"
   when:
+    - grafana_server_addr is not defined
```
guits, line 26: if you skip this task when `grafana_server_addr` is defined by the user, how `grafana_server_addrs` is supposed to be set ? It is consumed

guits:
@fmount @fultonj @gfidente FYI

fmount:
@thotypous how can you deal with multiple grafana instances? 
One of the main supported scenarios which is the reason to properly compute the `grafana_server_addr` fact is to allow users defining multiple grafana servers, configuring the ini file using the correct fact for the current node in which grafana is deployed.

roles/ceph-facts/tasks/grafana.yml
thotypous: Sorry, the original patch was over cae24dd85a979ed40efc7de384c158a04bd3558d and I messed it up when rebasing before sending this PR. The new patch I'm force pushing will employ a new approach.

thotypous:
Sorry for the delay. I had several issues with the cluster I was using for tests and only got to test the patch against master a few minutes ago.

The previous approach was completely incorrect, sorry. I was originally testing against cae24dd85a979ed40efc7de384c158a04bd3558d, then took too long to send the PR and messed up when rebasing. I didn't pay attention to the new `grafana_server_addrs` fact.

@fmount Please take a look at the new approach.

PR #5158, opened 2020-03-17
ceph-defaults: add registry name on dashboard vars

dsavineau:
We don't use the registry name when using the community dashboard
container images (grafana, prometheus, alertmanager & node exporter).
This commit adds the docker.io registry explicitly in the default
dashboard container image name values.


dsavineau:

PR #5163, opened 2020-03-18
Switch to epel-8 mock config for make rpm

bshewale:
Since the Makefile is updated to build rpm on el8 but
mock config was pointing to epel-7. It needs to be fixed
to use epel-8 config.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ NAME = ceph-ansible
 #  A "git describe" value of "v2.2.0" creates an NVR
 #  "ceph-ansible-2.2.0-1.el8"
 
+DIST ?= "el8"
+MOCK_CONFIG ?= "epel-8-x86_64"
 TAG := $(shell git describe --tags --abbrev=0 --match 'v*')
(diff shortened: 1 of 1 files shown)


bshewale:
@gfidente

ktdreyer:
This will alter the stable-4.0 builds without an easy way to go back to el7.

Instead of hard-coding this further, would you please copy the entire Makefile from `master` to `stable-4.0`? That will include the `$(DIST)` and `$(MOCK_CONFIG)` variables. That is far easier to maintain if we need to ever run `make rpm` for el7 again.

ktdreyer:
Thanks. With this change, I confirmed that `make rpm` builds an RPM for CentOS 8, and `make rpm DIST=el7 MOCK_CONFIG=epel-7-x86_64` builds an RPM for CentOS 7.

ktdreyer:
This change now cherry-picks the whole `Makefile` from master.

guits:
> This change now cherry-picks the whole `Makefile` from master.

I don't think it's a proper cherry-pick here

guits:
Closing this in favor of #5194

PR #5184, opened 2020-03-24
ceph-facts: fix rgw_instances_all fact

dsavineau:
The rgw_instances_all fact is supposed to be the list of all radosgw
instances from all rgw nodes.
But the fact is always using the local rgw_instances variable so this
won't work on multiple nodes.


dsavineau:

dsavineau:

dsavineau:

dsavineau:

PR #5238, opened 2020-04-02
tox-docker2podman: update container image tag

dsavineau:
The current docker to podman scenario is using the nautilus container
image tag instead of master.


dsavineau:

PR #5249, opened 2020-04-03
ceph-mgr: add saml python lib for dashboard SSO

dsavineau:
The dashboard SSO mgr module requires the saml python library to be
installed. This is only a valid scenario for RHCS deployment because
the saml python library isn't available in other classic repositories.
This package is present in RHCS Tools repository so we also need to
enable it on the mgr nodes.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1820233


dsavineau:
This probable needs more work since sso isn't a mgr module but a mgr dashboard sub component

https://docs.ceph.com/docs/master/mgr/dashboard/#dashboard-sso-support

dsavineau:

dsavineau:

PR #5250, opened 2020-04-05
rgw: Add default quota configs

clwluvw:
Add user max size and user max objects and bucket max size quota config to ceph conf

Signed-off-by: Seena Fallah <seeenafallah@gmail.com>

### DIFF
--- a/group_vars/rgws.yml.sample
+++ b/group_vars/rgws.yml.sample
@@ -31,6 +31,12 @@ dummy:
 # Consider setting a quota on buckets so that exceeding this
 # limit will require admin intervention.
 #rgw_bucket_default_quota_max_objects: 1638400 # i.e., 100K * 16
+#rgw_bucket_default_quota_max_size: 107374182400 # i.e., 100GB
+
+# Consider setting a quota on users so that exceeding this
+# limit will require admin intervention.
+#rgw_user_default_quota_max_objects: 1638400 # i.e., 100K * 16
+#rgw_user_default_quota_max_size: 107374182400 # i.e., 100GB
 
 # Declaring rgw_create_pools will create pools with the given number of pgs,
 # size, and type. The following are some important notes on this automatic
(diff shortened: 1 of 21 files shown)


guits:
Hi @clwluvw 

please, take a look at the documentation [1].

> We will no longer accept pull requests that modify the ceph.conf template unless it helps the deployment. For simple configuration tweaks please use the ceph_conf_overrides variable.

[1] https://github.com/ceph/ceph-ansible/blob/master/docs/source/index.rst#cephconf-configuration-file

clwluvw:
@guits But why the `max_objects` is defined and `max_size` isn't defined? Isn't that inconsistency?

guits:
this was added before we decided to stop modifying ceph.conf.j2 and was probably kept for backward compatibility concerns.

clwluvw:
@guits Makes sense but if we could add these too the quota default section for ansible would be complete :)

clwluvw:
@guits Makes sense but if we could add these too the quota default section for ansible would be complete :)
Now the ansible config gets dirty because one of is in `ceph_conf_overrides` the other isn't.

guits:
> @guits Makes sense but if we could add these too the quota default section for ansible would be complete :)
> Now the ansible config gets dirty because one of is in `ceph_conf_overrides` the other isn't.

In fact, I would do the opposite: I would deprecate and remove what we've left in the ceph.conf.j2 so we keep consistency.

clwluvw:
> > @guits Makes sense but if we could add these too the quota default section for ansible would be complete :)
> > Now the ansible config gets dirty because one of is in `ceph_conf_overrides` the other isn't.
> 
> In fact, I would do the opposite: I would deprecate and remove what we've left in the ceph.conf.j2 so we keep consistency.

Yes it fully depends on you. I think if we keep consistency here it would be so better in our ansible config.

PR #5262, opened 2020-04-09
library/ceph_volume: look for error messages in stderr

rishabh-d-dave:
Error message were moved to from stdout in stderr here -
https://github.com/ceph/ceph/commit/b8d6dcbe9f803c96c0af68da54f1262e9b6a9e77#diff-20f7c578a4e69ec61a5869d706567a24R137.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1793542

### DIFF
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -649,10 +649,17 @@ def run_module():
         try:
             report_result = json.loads(out)
         except ValueError:
-            strategy_change = "strategy changed" in out
-            if strategy_change:
-                out = json.dumps(
-                    {"changed": False, "stdout": out.rstrip("\r\n")})
+            strategy_changed_in_out = "strategy changed" in out
+            strategy_changed_in_err = "strategy changed" in err
+            strategy_changed = strategy_changed_in_out or \
+                               strategy_changed_in_err
+            if strategy_changed:
+                if strategy_changed_in_out:
+                    out = json.dumps({"changed": False,
+                                      "stdout": out.rstrip("\r\n")})
+                elif strategy_changed_in_err:
+                    out = json.dumps({"changed": False,
+                                      "stderr": err.rstrip("\r\n")})
                 rc = 0
                 changed = False
             else:
@@ -664,7 +671,7 @@ def run_module():
                 rc=rc,
                 changed=changed,
             )
-            if strategy_change:
+            if strategy_changed:
                 module.exit_json(**result)
             module.fail_json(msg='non-zero return code', **result)
 


guits:
@rishabh-d-dave is the fix https://github.com/ceph/ceph/commit/b8d6dcbe9f803c96c0af68da54f1262e9b6a9e77#diff-20f7c578a4e69ec61a5869d706567a24R137 present in nautilus?

dsavineau:
How to manage backward compatibility on this?

andrewschoen:
> @rishabh-d-dave is the fix [ceph/ceph@b8d6dcb#diff-20f7c578a4e69ec61a5869d706567a24R137](https://github.com/ceph/ceph/commit/b8d6dcbe9f803c96c0af68da54f1262e9b6a9e77#diff-20f7c578a4e69ec61a5869d706567a24R137) present in nautilus?
> 
> the question is do we need to backport this in nautilus (and luminous perhaps?)

It looks like this has been backported to both nautilus and luminous already.

nautilus backport commit here: https://github.com/ceph/ceph/commit/033d7ddf6699079aca0dcf43dbf2f7d72df4604a

luminous backport commit here: https://github.com/ceph/ceph/commit/d3a2d41ca5232aa39cf1653e6a6b8d865729a42c

guits:
@rishabh-d-dave Dimitri's concern about backward compatibility makes sense, how are we supposed to deal with this?

andrewschoen:
> @rishabh-d-dave Dimitri's concern about backward compatibility makes sense, how are we supposed to deal with this?

I think we could check both ``out`` and ``err`` and that would take care of any backward compatibility issues. Thoughts?

dsavineau:
> > @rishabh-d-dave Dimitri's concern about backward compatibility makes sense, how are we supposed to deal with this?
> 
> I think we could check both `out` and `err` and that would take care of any backward compatibility issues. Thoughts?

:+1:  I think that's the best solution

library/ceph_volume.py
```diff
@@ -649,7 +649,8 @@ def run_module():
         try:
             report_result = json.loads(out)
         except ValueError:
-            strategy_change = "strategy changed" in out
+            strategy_change = ("strategy changed" in out or
+                               "strategy changed" in err)
```
dsavineau, line 653: `"strategy changed" in err + out` wdyt ?

andrewschoen: I prefer it this way, if we have any sort of error keeping ``out`` and ``err`` separately might make debugging easier.

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

library/ceph_volume.py
```diff
@@ -649,7 +649,8 @@ def run_module():
...
             if strategy_change:
                 out = json.dumps(
                     {"changed": False, "stdout": out.rstrip("\r\n")})
```
dsavineau, line 656: I didn't saw this the first time but we should also use err to print the output instead of out (`out.rstrip`) when the strategy changes ?

dsavineau:

dsavineau:

library/ceph_volume.py
rishabh-d-dave: Nice catch. I've added `err` too. However, it's redundant to make the user look at both stdout and stderr in case of a failure. So, I am printing those conditionally.

Thanks!

library/ceph_volume.py
```diff
@@ -649,10 +649,17 @@ def run_module():
         try:
             report_result = json.loads(out)
         except ValueError:
-            strategy_change = "strategy changed" in out
-            if strategy_change:
-                out = json.dumps(
-                    {"changed": False, "stdout": out.rstrip("\r\n")})
+            strategy_changed_in_out = "strategy changed" in out
+            strategy_changed_in_err "strategy changed" in err
```
dsavineau, line 653: missing `=` character

rishabh-d-dave:
@guits @dsavineau  @andrewschoen Thanks!

PR #5269, opened 2020-04-09
[skip ci] Add playbook for converting cluster to cephadm

dsavineau:
The commit adds a new playbook for converting an existing ceph cluster
deployed by ceph-ansible to the cephadm orchestrator.

### DIFF
--- a/infrastructure-playbooks/cephadm-adopt.yml
+++ b/infrastructure-playbooks/cephadm-adopt.yml
@@ -1,0 +1,847 @@
+---
+#
+# This playbook does a cephadm adopt for all the Ceph services
+#
+
+- name: confirm whether user really meant to adopt the cluster by cephadm
+  hosts: localhost
+  connection: local
+  become: false
+  gather_facts: false
+  vars_prompt:
+    - name: ireallymeanit
+      prompt: Are you sure you want to adopt the cluster by cephadm ?
+      default: 'no'
+      private: no
+  tasks:
+    - name: exit playbook, if user did not mean to adopt the cluster by cephadm
+      fail:
+        msg: >
+          Exiting cephadm-adopt playbook, cluster was NOT adopted.
+           To adopt the cluster, either say 'yes' on the prompt or
+           use `-e ireallymeanit=yes` on the command line when
+           invoking the playbook
+      when: ireallymeanit != 'yes'
+
+- name: gather facts and prepare system for cephadm
+  hosts:
+    - "{{ mon_group_name|default('mons') }}"
+    - "{{ osd_group_name|default('osds') }}"
+    - "{{ mds_group_name|default('mdss') }}"
+    - "{{ rgw_group_name|default('rgws') }}"
+    - "{{ mgr_group_name|default('mgrs') }}"
+    - "{{ rbdmirror_group_name|default('rbdmirrors') }}"
+    - "{{ nfs_group_name|default('nfss') }}"
+    - "{{ iscsi_gw_group_name|default('iscsigws') }}"
+    - "{{ grafana_server_group_name|default('grafana-server') }}"
+  become: true
+  gather_facts: false
+  vars:
+    delegate_facts_host: true
+  tasks:
+    - import_role:
+        name: ceph-defaults
+
+    - name: gather facts
+      setup:
+      when: not delegate_facts_host | bool or inventory_hostname in groups.get(client_group_name, [])
+
+    - name: gather and delegate facts
+      setup:
+      delegate_to: "{{ item }}"
+      delegate_facts: true
+      with_items: "{{ groups['all'] | difference(groups.get('clients', [])) }}"
+      run_once: true
+      when: delegate_facts_host | bool
+
+    - name: fail if one osd node is using filestore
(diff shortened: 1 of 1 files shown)


infrastructure-playbooks/cephadm-adopt.yml
```diff
@@ -0,0 +1,689 @@
...
+  become: true
+  gather_facts: false
+  vars:
+    delegate_facts_host: true
+  tasks:
+    - import_role:
+        name: ceph-defaults
+
+    - name: fail if one osd node is using filestore
+      fail:
```
jmolmo, line 46: Maybe it could be good to restrict the source ceph versions, and only allow to do the "adoption" from "Nautilus"

infrastructure-playbooks/cephadm-adopt.yml
```diff
@@ -0,0 +1,689 @@
...
+        name: ceph-container-engine
+      when: not containerized_deployment | bool
+
+    - import_role:
+        name: ceph-container-common
+      when: not containerized_deployment | bool
+
+    - name: install cephadm requirements
+      package:
+        name: ['python3', 'lvm2']
```
jmolmo, line 76: Also "docker" or "podman"

infrastructure-playbooks/cephadm-adopt.yml
jmolmo, line 46: Also it would be nice to be sure that the cluster is in a good shape.  Checking if we have a "HEALTH_OK" cluster , and inform if there is any health alarm could be useful. The user can abort the adoption or continue with it at his own risk.

jmolmo:
Maybe i do not have somethign into account but in my view the order installing things is not correct:
You are enabling the cephadm orchestrator module, generating and copying the ssh keys, adding hosts to the cephadm cluster, assimilate ceph configuration and making the adoption of the ceph mon and mgr daemons. Then you continue with OSD's and rest of the services

This is going to fail because if you try to enable the cephadm  orchestrator module in a monitor (Nautilus version or previous), it is going to fail because in Nautilus the cephadm orchestrator didn't exists.

I think that the right order is: assimilate ceph configuration, adopt monitor, adopt mgr, enable the cephadm orchestrator, generating and copying the ssh keys, adding hosts to the cephadm cluster. Then continue with OSD's and rest of the services

Take a look to:
https://ceph.readthedocs.io/en/latest/cephadm/adoption/

infrastructure-playbooks/cephadm-adopt.yml
dsavineau: In theory this playbook will only be available in ceph-ansible releases supporting >= octopus but we can probably another validation.

infrastructure-playbooks/cephadm-adopt.yml
dsavineau: this is already handled by ceph-ansible via the ceph-container-engine role.

infrastructure-playbooks/cephadm-adopt.yml
dsavineau: in fact we're already doing this for the upgrade playbook so I think we can do the same here

https://github.com/ceph/ceph-ansible/blob/master/infrastructure-playbooks/rolling_update.yml#L127-L142

dsavineau:
> Maybe i do not have somethign into account but in my view the order installing things is not correct:
> You are enabling the cephadm orchestrator module, generating and copying the ssh keys, adding hosts to the cephadm cluster, assimilate ceph configuration and making the adoption of the ceph mon and mgr daemons. Then you continue with OSD's and rest of the services
> 
> This is going to fail because if you try to enable the cephadm orchestrator module in a monitor (Nautilus version or previous), it is going to fail because in Nautilus the cephadm orchestrator didn't exists.

If you run this playbook it assumes that you're cluster is already upgraded to octopus.

So users managing ceph-ansible clusters from nautilus or mimic will need to upgrade to octopus first, and then convert it to cephadm. The workflow is basically
  - ansible-playbook infrastructure-playbooks/rolling_update.yml
  - ansible-playbook infrastructure-playbooks/cephadm-adopt.yml

> 
> I think that the right order is: assimilate ceph configuration, adopt monitor, adopt mgr, enable the cephadm orchestrator, generating and copying the ssh keys, adding hosts to the cephadm cluster. Then continue with OSD's and rest of the services
> 
> Take a look to:
> https://ceph.readthedocs.io/en/latest/cephadm/adoption/

I've already take a look to the documentation and honnestly I don't see any reasons why the mon/mgr should have to be adopted before enabling the cephadm orchestrator.

jmolmo:
> I've already taken a look to the documentation and honnestly I don't see any reasons why the mon/mgr should be adopted before enabling the cephadm orchestrator.

If you are in Octopus , then the order is not important because your are going to have the cephadm orchestrator available since the first moment.

jmolmo:
LGTM

jmolmo:
ping @dsavineau: It is needed something additional to merge this?

guits:
there's no testing in the CI anyway for this one at the moment, let's merge this as it is and follow up if needed

PR #5237, opened 2020-04-02
Readd CentOS 7 with conditions

dsavineau:
The CentOS 7 distribution could still be used be deploying ceph if
  - it's a containerized deployment
  - it's a non containerized deployment without the dashboard (due to
missing python3 libraries).

The ceph_stable_redhat_distro variable has been remove because we can
rely on the ansible_distribution_major_version fact instead.

The copr el8 repository configuration is only applied for CentOS 8.

The ceph-mgr-dashboard package is only installed when the
dashboard_enabled variable is set to true.

Signed-off-by: Dimitri Savineau <dsavinea@redhat.com>

PR #5280, opened 2020-04-16
docker2podman: pull images from docker daemon

dsavineau:
The docker2podman playbook only installs the podman package and updates
the systemd units with the right container_binary value.

We never pull the container image so if one service is restarted then
the container image will be pulled first before the service can start
which could cause longer downstream.

To avoid to download the container image from internet again we can just
pull it from the local docker daemon.

The container_{binding,package,service}_name variables are removed
because they are only used in the ceph-container-engine role which
isn't call in this playbook.

### DIFF
--- a/infrastructure-playbooks/docker-to-podman.yml
+++ b/infrastructure-playbooks/docker-to-podman.yml
@@ -46,8 +46,9 @@
     - "{{ mgr_group_name | default('mgrs') }}"
     - "{{ iscsi_gw_group_name | default('iscsigws') }}"
     - "{{ rbdmirror_group_name | default('rbdmirrors') }}"
+  gather_facts: false
   become: true
-  pre_tasks:
+  tasks:
     - import_role:
         name: ceph-defaults
     - import_role:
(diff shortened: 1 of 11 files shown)


dsavineau:
we need https://github.com/ceph/ceph-ansible/pull/5237 first to test this

dsavineau:

dsavineau:

dsavineau:

chkumar246:
check-github

guits:
@chkumar246 your job is failing for unrelated reason, any chance you can take a look and fix it? We can't merge this patch because of that

dsavineau:

dsavineau:
@guits do you want to split this PR into two dedicated PR ?

The second commit is associated to a BZ but the first commit is an enhancement.

WDYT ?

dsavineau:

guits:
> @guits do you want to split this PR into two dedicated PR ?
> 
> The second commit is associated to a BZ but the first commit is an enhancement.
> 
> WDYT ?

I think it's fine, I would backport this anyway

PR #5283, opened 2020-04-17
Updated use of deprecated filter in check_system task

abaird-rh:
In the check_system task as part of the ceph-validate role there is the use of a deprecated filter "version_compare" which has been removed from Ansible 2.9, the required version of Ansible for this version of Ceph.

This has been update to use the 'version' function instead.


dsavineau:
Please create the PR against the master branch first.

We will then backport the change in the right stable branches.

You also need to add a sign off to your commit.

Finally, if you change version_compare by version then you also need to do this on the entire code.

dsavineau:
I've opened https://github.com/ceph/ceph-ansible/pull/5284

PR #5335, opened 2020-04-30
tox-update: fix update container tag value

dsavineau:
The update container tag value should be set to latest-octopus and not
latest-master.
Also we don't need the `UPDATE_CEPH_DEV_*` environment variables.


dsavineau:

PR #5341, opened 2020-05-06
ceph-rgw: use match instead of equalto from jinja2

dsavineau:
The '==' jinja2 operator (or 'equalto') has been introduced in jinja2
2.8.
On EL7, jinja2 version is 2.7 so the operator isn't present creating
templating error like:

The error was: TemplateRuntimeError: no test named '=='

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1747206


dsavineau:

dsavineau:

dsavineau:
@alimaredia I've updated the PR because with the `match` operator we need to use ^$ to avoid matching a superset zone/zonegroup/realm

dsavineau:

dsavineau:

PR #5344, opened 2020-05-08
ceph-validate: Fix "fail on unsupported CentOS release"

BenoitKnecht:
The `dashboard_enabled` condition used a `true` filter (which doesn't exist)
instead of the `bool` filter.


dsavineau:

PR #5354, opened 2020-05-12
docker-to-podman: conditional docker commands

dsavineau:
The docker commands should be based on the container_binary variable
otherwise running the playbook on a host without docker (like podman
only) will failed.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1829985


dsavineau:

PR #5366, opened 2020-05-14
test: set sitepackages=false in tox

guits:
Otherwise it might try to use the system installed version of ansible
when there's one available.


guits:
unrelated failure

PR #5370, opened 2020-05-14
tests: force using the more recent build

guits:
We should use  `latest-master-devel` for switch_to_containers job.
Otherwise it might happen we actually downgrade the ceph version when
the image used is older than the rpm initially used for installing ceph.


guits:

PR #5376, opened 2020-05-15
facts: always set ceph_run_cmd and ceph_admin_command

guits:
always set these facts on monitor nodes whatever we run with `--limit`.
Otherwise, playbook will fail when using `--limit` on nodes where these
facts are used on a delegated task to monitor.

(cherry picked from commit e5e81843e918ed9aa57a5675af2888499700eac2)


guits:

PR #5378, opened 2020-05-15
ceph-osd/start_osds.yml: ensure wal and db partitions have proper owner

siavashsardari:
OSD services fails to start due to permission denied error to access wal and db partitions (destination of block.wal and bock.db symlink).
I checked with ceph-deploy so ceph_volume works properly. This should be an ansible issue which this commit solves.
Also the problem might be from ceph_volume ansible module, but I could not find any problem on ceph_volume ansible module.
Note: I think it is a good idea to backport this to stable branch.

Signed-off-by: Siavash Sardari <siavash.sardari@gmail.com>

### DIFF
--- a/roles/ceph-osd/tasks/start_osds.yml
+++ b/roles/ceph-osd/tasks/start_osds.yml
@@ -33,6 +33,15 @@
   include_tasks: systemd.yml
   when: containerized_deployment | bool
 
+- name: ensure files have correct owner
+  file:
+    path: /var/lib/ceph/osd/
+    state: directory
+    recurse: yes
+    owner: "{{ ceph_uid if containerized_deployment else 'ceph' }}"
+    group: "{{ ceph_uid if containerized_deployment else 'ceph' }}"
+  when: cephx | bool
+
 - name: systemd start osd
   systemd:
     name: ceph-osd@{{ item }}


dsavineau:
I don't think changing the owner/group on the whole /var/lib/ceph/osd directory is a good idea.

This will be applied even on filestore configuration whereas the issue you mentioned is bluestore only.

Could you give more context on your issue ?

What was the ceph version ?
Is it a containerized deployment ? If yes, what was the container image ?
What's the OS ?

siavashsardari:
@dsavineau

siavashsardari:
you are right. I didn't think about filestore.
my problem is when I use wal and db partition for OSDs, it cause the OSD services to not start. I looked at the logs and found permission denied error. while the ownership of /var/lib/ceph/osd/osd-0/ files were correct, the ownership of destination block.wal and block.db symlink which are /dev/sdd1 and /dev/sde1 were still owned by root user and group.

siavashsardari:
@dsavineau you are right. I didn't think about filestore. BTW I didn't change the ownership of var/lib/ceph/osd directory, with both bluestore and filestore, the ownership of mentioned directory is the ceph user and group (unless we didn't use cephx). so I thought double checking the ownership of whole directory, while solves my issue, is harmless to other scenarios. but I think there might be a better solution.
my problem is when I use wal and db partition for OSDs, it cause the OSD services to not start. I looked at the logs and found permission denied error. while the ownership of /var/lib/ceph/osd/osd-0/ files were correct, the ownership of destination block.wal and block.db symlink which are /dev/sdd1 and /dev/sde1 were still owned by root user and group.
I used stable-5.0 branch and installed Nautilus version (I didn't use stable-4.0 because I needed to use crush_device_class which can not be used with stable-4.0 branch). I also tested this with stable-5.0 branch and Octopus version.
I installed on baremetal and clean Ubuntu 18.04 installation.

my OSD options is this:
vm_volumes:
  - data: /dev/sdb
    wal: /dev/sdd1
    db: /dev/sde1
    crush_device_class: HDD_WAL_DB
  - data: /dev/sdc
    wal: /dev/sdd2
    db: /dev/sde2
    crush_device_class: HDD_WAL_DB
  - data: /dev/sdf
    crush_device_class: HDD
  - data: /dev/sdg
    crush_device_class: HDD

dsavineau:
> @dsavineau you are right. I didn't think about filestore. BTW I didn't change the ownership of var/lib/ceph/osd directory, with both bluestore and filestore, the ownership of mentioned directory is the ceph user and group (unless we didn't use cephx). so I thought double checking the ownership of whole directory, while solves my issue, is harmless to other scenarios. but I think there might be a better solution.

The ansible task you provided changes the owner/group of this directory recursively. This is unrelated to cephx, cephx is about keyring authentication not system user/group.

> my problem is when I use wal and db partition for OSDs, it cause the OSD services to not start. I looked at the logs and found permission denied error. while the ownership of /var/lib/ceph/osd/osd-0/ files were correct, the ownership of destination block.wal and block.db symlink which are /dev/sdd1 and /dev/sde1 were still owned by root user and group.

So why changing the owner/group on `/var/lib/ceph/osd` directory is the issue is on `/dev/` ?

> I used stable-5.0 branch and installed Nautilus version

First, use the right ceph-ansible branch for deploying Nautilus (stable-4.0).

https://docs.ceph.com/ceph-ansible/master/#releases

> I didn't use stable-4.0 because I needed to use crush_device_class which can not be used with stable-4.0 branch

Why ? The crush_device_class parameter for lvm_volumes is the same since stable-3.0 branch.

Finally this doesn't seem to be a ceph-ansible issue as the owner/group are resetted on the bluestore db/wal partitions.
Only ceph-volume (set ceph:ceph) [1] and probably udev can do this kind of operation.

There's a private bugzilla [2] about this and this is definitely not related to ceph-ansible.

[1] https://github.com/ceph/ceph/blob/nautilus/src/ceph-volume/ceph_volume/devices/lvm/activate.py#L179-L188
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1822134

siavashsardari:
Thanks for the response. this is completely related to ceph-volume.

PR #5387, opened 2020-05-22
common: quick fix in raw_install_python.yml

guits:
Typical error:

```
  msg: |-
    The conditional check 'item.stat.exists | bool' failed. The error was: error while evaluating conditional (item.stat.exists | bool): 'dict object' has no attribute 'stat'

```

This can happen when none of the binaries was found.


guits:
note: this will just make the playbook failing later, it would be probably nice to add an other check to fail earlier in `raw_install_python.yml`

PR #5391, opened 2020-05-22
Fix deamon socket checks

rossmartyn04:
Where there are multiple socket paths, i.e OSDs, the paths are aggregated into one long string due to the --printf option for the stat tool.

This replaces the option with --format which correctly formats the output into multiple lines. For completeness this change has been applied to every check.


rossmartyn04:
It looks like this is failing on "[05] Use shell only when shell functionality is required".

Through my testing, it seems I need to use shell here, but I can rework if anyone believes this isn't true.

rossmartyn04:
Hello,

Any reviews would be apprecaited. It sounded like this was ready to merge on the last PR.

Thanks

PR #5401, opened 2020-06-01
ceph-rgw-loadbalancer: Fix keepalived master selection

BenoitKnecht:
While 2ca33641 fixed a bug in the way the `keepalived.conf.j2` template matched
hostnames to set the VRRP `MASTER`/`BACKUP` states, it also introduced a
regression in the case where `virtual_ips` is a list of more than one IP
address.

The previous behavior would result in each host in the `rgwloadbalancers` group
to be `MASTER` for one of the `virtual_ips`, but the new behavior caused the
first host to be `MASTER` for all the IP address in `virtual_ips`.

This commit restores the original behavior.

### DIFF
--- a/roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
+++ b/roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
@@ -19,7 +19,7 @@
 
 - name: set_fact vip to vrrp_instance
   set_fact:
-    vrrp_instances: "{{ vrrp_instances | default([]) | union([{ 'name': 'VI_' + index|string , 'vip': item }]) }}"
+    vrrp_instances: "{{ vrrp_instances | default([]) | union([{ 'name': 'VI_' + index|string , 'vip': item, 'master': groups[rgwloadbalancer_group_name][index] }]) }}"
   loop: "{{ virtual_ips | flatten(levels=1) }}"
   loop_control:
     index_var: index
--- a/roles/ceph-rgw-loadbalancer/templates/keepalived.conf.j2
+++ b/roles/ceph-rgw-loadbalancer/templates/keepalived.conf.j2
@@ -15,8 +15,8 @@ vrrp_script check_haproxy {
 
 {% for instance in vrrp_instances %}
 vrrp_instance {{ instance['name'] }} {
-    state {{ 'MASTER' if inventory_hostname == groups[rgwloadbalancer_group_name][0] else 'BACKUP' }}
-    priority {{ '100' if inventory_hostname == groups[rgwloadbalancer_group_name][0] else '90' }}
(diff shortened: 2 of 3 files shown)


roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
```diff
@@ -19,7 +19,7 @@
 
 - name: set_fact vip to vrrp_instance
   set_fact:
-    vrrp_instances: "{{ vrrp_instances | default([]) | union([{ 'name': 'VI_' + index|string , 'vip': item }]) }}"
+    vrrp_instances: "{{ vrrp_instances | default([]) | union([{ 'name': 'VI_' + index|string , 'vip': item, 'master': groups[rgwloadbalancer_group_name][index] }]) }}"
```
dsavineau, line 22: The problem with this is that it doesn't work when you have more VIP than loadbalancer nodes because the iteration is made on the VIP list but you're using the index on the loadbalancer group list.

BenoitKnecht:
I don't think this should be marked as stale, it still doesn't work as expected.

guits:
> I don't think this should be marked as stale, it still doesn't work as expected.

so what's the next step? It didn't get any update for a while. @dsavineau @BenoitKnecht what's the status then?

BenoitKnecht:
@guits I still believe this should be merged. It fixes a regression for anyone who has a setup where each load-balancer is supposed to be `MASTER` for a separate VIP (it worked fine before 2ca33641091366a9f69150a331b0fdf364833bf4, it doesn't work now).

I can add validation to make sure `len(virtual_ips) <= len(groups[rgwloadbalancer_group_name])`. As for the use-case with multiple networks, it wasn't supported before, and as mentioned above, it should be handled by adding multiple `virtual_ipaddress` in a single `vrrp_instance`. If you want to support that use-case, I think it should be addressed in a separate PR.

BenoitKnecht:
I really don't think this should have been closed, the issue is still not fixed. If there's any changes that need to be made before merging this please let me know, but in the meantime, can this be reopened?

BenoitKnecht:
Bump.

guits:
Stuck for 6 months now...  It looks like we can't have a consensus here on the implementation but let's settle this.

@BenoitKnecht can you add the check you suggested in `ceph-validate`? (`len(virtual_ips) <= len(groups[rgwloadbalancer_group_name])`)

guits:

BenoitKnecht:
@guits Thanks for revisiting this and adding validation. I've been pretty busy these past few days and haven't had a chance to do it myself, so I appreciate the help!

guits (approved):

BenoitKnecht:
@guits I just realized that this hasn't been backported to the `stable-4.0` and `stable-5.0` branches. Should I create PRs for these, or can it be done automatically?

PR #5404, opened 2020-06-04
rgw: allow rgws can be run with or without multisite at the same

alimaredia:
Allows rgws in a ceph cluster to be run with multisite and without multisite at the same time.

### DIFF
--- a/roles/ceph-facts/tasks/set_radosgw_address.yml
+++ b/roles/ceph-facts/tasks/set_radosgw_address.yml
@@ -73,4 +73,4 @@
   with_items: "{{ groups.get(rgw_group_name, []) }}"
   when:
     - inventory_hostname in groups.get(rgw_group_name, [])
-    - rgw_multisite | bool
+    - hostvars[item]["rgw_multisite"] | default(False) | bool


roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -126,7 +126,7 @@ rgw frontends = {{ frontend_line(radosgw_frontend_type) }} {{ radosgw_frontend_o
 {% if 'num_threads' not in radosgw_frontend_options %}
 rgw thread pool size = {{ radosgw_thread_pool_size }}
 {% endif %}
-{% if rgw_multisite | bool %}
+{% if hostvars[host]['rgw_multisite'] | bool %}
```
dsavineau, line 129: This won't work because when using the default value (ie: not configuring the variable in group_vars) so you probably need to add the `default` filter with the `rgw_multisite` value.

roles/ceph-facts/tasks/set_radosgw_address.yml
```diff
@@ -74,3 +74,4 @@
   when:
     - inventory_hostname in groups.get(rgw_group_name, [])
     - rgw_multisite | bool
+    - hostvars[item]["rgw_multisite"] | bool
```
dsavineau, line 77: Why are you keeping both `rgw_multisite` and `hostvars[item]["rgw_multisite"]` ? the last one should be enough right ?

alimaredia:

roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -126,7 +126,7 @@ rgw frontends = {{ frontend_line(radosgw_frontend_type) }} {{ radosgw_frontend_o
 {% if 'num_threads' not in radosgw_frontend_options %}
 rgw thread pool size = {{ radosgw_thread_pool_size }}
 {% endif %}
-{% if rgw_multisite | bool %}
+{% if hostvars[host]['rgw_multisite'] | default(False) %}
```
dsavineau, line 129: you still need the bool filter at the end

roles/ceph-facts/tasks/set_radosgw_address.yml
```diff
@@ -73,4 +73,4 @@
   with_items: "{{ groups.get(rgw_group_name, []) }}"
   when:
     - inventory_hostname in groups.get(rgw_group_name, [])
-    - rgw_multisite | bool
+    - hostvars[item]["rgw_multisite"] | bool
```
dsavineau, line 76: you should do the same change here than above

roles/ceph-config/templates/ceph.conf.j2
guits: @alimaredia

roles/ceph-facts/tasks/set_radosgw_address.yml
guits: @alimaredia

alimaredia:

alimaredia:

dsavineau:
Could you rebase your PR ?

alimaredia:
@guits: Yes, it's to be able to have some rgws using multisite on a cluster and others not.

guits:
@alimaredia isn't setting `rgw_multisite` at inventory and/or host_vars level enough?

alimaredia:
@guits I thought it would be, but the reason for this PR is that these changes need to be there to be able to just set it at the host_vars level.

guits:
@alimaredia, there must be a misunderstanding somewhere, if you set `rgw_multisite` at host_vars level like the following : 

```shell
$ grep -rn --color rgw_multisite *
group_vars/all.yml:2:rgw_multisite: foobar
host_vars/mon1:2:rgw_multisite: false
host_vars/mon2:2:rgw_multisite: Foo
host_vars/mon0:2:rgw_multisite: true
```

here is what I get if I run a debug task:

```yaml
       - name: debug                                                                                                                                        
         debug:                                                                                                                                             
           msg: "{{ rgw_multisite }}"                                                                                                                       

```

```
ok: [mon0] => 
  msg: true
ok: [mon1] => 
  msg: false
ok: [mon2] => 
  msg: Foo
ok: [mon3] => 
  msg: foobar
```

there's no need to call `hostvars[host]['rgw_multisite']`, which is expected behavior

guits:

guits:

PR #5413, opened 2020-06-12
lvm_setup: lookup device from inventory, default to /dev/sd* names

jan--f:
This fixes a long standing fail in ceph-volumes lvm test suite.
Otherwise the default behaviour should not change.

Signed-off-by: Jan Fajerski <jfajerski@suse.com>


jan--f:
Test failures are fixed by https://github.com/ceph/ceph/pull/35552 I think.

guits:

PR #5424, opened 2020-06-17
switch-to-containers: set and unset osd flags

guits:
The workflow in this playbook should be the same than in rolling_update,
we should first set noout and nodeep-scrub flags before migrating the
first osd and unset osd flags after the last osd is migrated.

(cherry picked from commit 2cfaa056e020615bb99eb9db1520a977e5ac3ef4)


guits:

PR #5426, opened 2020-06-17
ceph-common: remove copr and sepia repositories

dsavineau:
All EL8 dependencies are now present on EPEL 8 so we don't need the
additional repositories that were only a temporary solution.


dsavineau:
@guits ping maggle

PR #5427, opened 2020-06-17
Adds missing dependency for restful mgr module

evindunn:


dsavineau:
Please open this on master first and we will backport it if needed to stable-5.0 branches.

Also add a signoff and commit body.

Finally you probably need to handle python3 instead of python2. If I'm right, since octopus it's python3 only and python2 as been dropped.

dsavineau:
BTW the python3-pecan package is automatically installed as a dependency when installing ceph-mgr package.

So if you propose this change that's because you're probably running on CentOS 7, but:

```Note that the dashboard, prometheus, and restful manager modules will not work on the CentOS 7 build due to Python 3 module dependencies that are missing in CentOS 7.```

https://docs.ceph.com/docs/master/releases/octopus/#general

PR #5428, opened 2020-06-18
Fixes for ceph-mgr install and incorrect ansible test usage

jrosser:


dsavineau:
Could you elaborate on the reason for this ? because there's nothing explained here.

jrosser:
There is a bug only revealed on ubuntu focal related to the use of tests as filters as follows:

fatal: [aio1_ceph-nfs_container-b8cf862c]: FAILED! => {"msg": "The conditional check 'add_ganesha_apt_repo | changed' failed. The error was: template error while templating string: no filter named 'changed'. String: {% if add_ganesha_apt_repo | changed %} True {% else %} False {% endif %}\n\nThe error appears to be in '/etc/ansible/roles/ceph-ansible/roles/ceph-nfs/tasks/pre_requisite_non_container_debian.yml': line 17, column 11, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n        - name: update apt cache\n

jrosser:
ceph-mgr cannot be deployed at all on python3 only Debian/Ubuntu OS becasue the python-routes package is not available, nor would there be a python2 interpreter present. There is no hook to allow the deployer to specify the python3 version of that package.

Rather than introduce even more complexity in the conditionals to account for this I have simplified the approach to make it more robust, and the differences in package installs needed for different operating systems are now cleanly described in the data, rather than complex conditionals.

dsavineau:
> ceph-mgr cannot be deployed at all on python3 only Debian/Ubuntu OS becasue the python-routes package is not available, nor would there be a python2 interpreter present. There is no hook to allow the deployer to specify the python3 version of that package.

What is the ceph version you're trying to installed ?

As far as I remenber this task was added for Nautilus which was still based on python2 but for Octopus and master we don't need this task anymore especially since python3-routes is a depencendy for ceph-mgr.

So instead doing such refact, should we just remove that python-routes task and fix the filter ?

jrosser:
Sure, if the python-routes package is no longer required I can just do that. We're installing Octopus using 0.5.0 currently. I'll push a new set of changes.

Do you have a suggestion about what to do with python3-saml, as that appears to break installs on Centos-8.

dsavineau:
> Do you have a suggestion about what to do with python3-saml, as that appears to break installs on Centos-8.

It shouldn't since the condition only impacts RedHat distribution but not CentOS because the saml package is only available via RHCS repository and not upstream.

jrosser:
> > Do you have a suggestion about what to do with python3-saml, as that appears to break installs on Centos-8.
> 
> It shouldn't since the condition only impacts RedHat distribution but not CentOS because the saml package is only available via RHCS repository and not upstream.

Of course you are right - I am confusing ansible_distribution with ansible_os_family.

dsavineau:
@jrosser Could you amend both commits and add a signoff ?

https://docs.ceph.com/ceph-ansible/master/dev/index.html#submit-a-patch

jrosser:
> @jrosser Could you amend both commits and add a signoff ?
> 
> https://docs.ceph.com/ceph-ansible/master/dev/index.html#submit-a-patch

Done.

guits:
@jrosser could you rebase on master please? thanks!

dsavineau:

PR #5461, opened 2020-06-26
ceph-volume.py: add support for batch refactored code

jan--f:
See https://github.com/ceph/ceph/pull/34740 for the batch changes.

Signed-off-by: Jan Fajerski <jfajerski@suse.com>

Note that this is backwards compatible. Intention is to backport this to stable-4.0 and newer.

### DIFF
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -714,10 +714,17 @@ def run_module():
             module.fail_json(msg='non-zero return code', **result)
 
         if not report:
-            # if not asking for a report, let's just run the batch command
-            changed = report_result['changed']
-            if changed:
-                # Batch prepare the OSD
+            if 'changed' in report_result:
+                # we have the old batch implementation
+                # if not asking for a report, let's just run the batch command
+                changed = report_result['changed']
+                if changed:
+                    # Batch prepare the OSD
+                    rc, cmd, out, err = exec_command(
+                        module, batch(module, container_image))
+            else:
+                # we have the refactored batch, its idempotent so lets just
+                # run it
                 rc, cmd, out, err = exec_command(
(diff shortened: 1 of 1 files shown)


guits:
@dsavineau i'm basically -1 on the reason that has led to this PR but I'm afraid we will have to merge this given ceph-volume leaves us no choice.

jan--f:
The sooner we get this fix out to users the lower the chance for friction with this update. I don't think its unreasonable to assume that users will update ceph with and up-to-date ceph-ansible release.

guits:
> The sooner we get this fix out to users the lower the chance for friction with this update. I don't think its unreasonable to assume that users will update ceph with and up-to-date ceph-ansible release.

as you said, this is assumption only with no guarantee, this is exactly my concern.

leseb:
> > The sooner we get this fix out to users the lower the chance for friction with this update. I don't think its unreasonable to assume that users will update ceph with and up-to-date ceph-ansible release.
> 
> as you said, this is assumption only with no guarantee, this is exactly my concern.

Perhaps a warning in the changelog could help users and give this more visibility? What do you think big?

guits:
> > > The sooner we get this fix out to users the lower the chance for friction with this update. I don't think its unreasonable to assume that users will update ceph with and up-to-date ceph-ansible release.
> > 
> > 
> > as you said, this is assumption only with no guarantee, this is exactly my concern.
> 
> Perhaps a warning in the changelog could help users and give this more visibility? What do you think big?

adding a note in the changelog won't hurt for sure, but definitely not addressing this concern.

In any case, there's nothing more we can do at the ceph-ansible level so I don't see any reason to block this PR.

Issue #5463, opened 2020-06-26
Useless dependency on python-routes


amarao:
**Bug Report**
There is a hard-coded dependency on python-routes in roles/ceph-mgr/tasks/pre_requisite.yml:
```
- name: install routes python library for dashboard module
  apt:
    name: python-routes
  register: result
  until: result is succeeded
  when:
    - ansible_os_family == 'Debian'
    - "'ceph-mgr-dashboard' in ceph_mgr_packages"
```

The modern ceph does not need python-routes, as it uses python3-routes:

```
apt-cache show ceph-mgr-dashboard
Version: 15.2.3-1focal
Depends: ceph-mgr (= 15.2.3-1focal), python3-cherrypy3, python3-jwt, python3-bcrypt, python3-werkzeug, python3-routes
```

Focal distribution does not provide python-routes anymore.  I think this dependency should be removed.

Ubuntu/focal, community distribution of Ceph Octopus, ceph-ansible stable-5.0

dsavineau:
https://github.com/ceph/ceph-ansible/pull/5428

PR #5464, opened 2020-06-26
Exclude Ubuntu focal from python-routes installation.

amarao:
python-routes is not available in Focal and it's not needed for
Octopus ceph-mgr-dashboards module.

Fixes: #5463

I'm not sure if it need to be more inclusive, but it fixes Ubuntu Focal issue here.


dsavineau:
duplicate https://github.com/ceph/ceph-ansible/pull/5428

PR #5466, opened 2020-06-26
dashboard: configure mgr backend before restart

dsavineau:
We need to set the mgr dashboard server ip address before restarting the
dashboard module otherwise we can try to bind the dashboard module on an
already used address.
We already do this configuration for the dashboard port value and ssl
setup so we should do the same for server address too.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1851455


dsavineau:
@fmount FYI

dsavineau:

PR #5485, opened 2020-06-30
ceph-common: remove copr and sepia repositories (bp #5426)

dsavineau:
All EL8 dependencies are now present on EPEL 8 so we don't need the
additional repositories that were only a temporary solution.

Backport: #5426

(cherry picked from commit 3592ba1d61c967de3810850d178a1febf189d309)


guits:

PR #5489, opened 2020-07-01
tests: enforce pytest-rerunfailures version

guits:
This commit enforces the pytest-rerunfailures installed so it's <9.0

This is to avoid the following error:

```
ERROR: pytest-rerunfailures 9.0 has requirement pytest>=5.0, but you'll have pytest 4.6.11 which is incompatible.
```

latest version of pytest-rerunfailures isn't compatible with the version
of pytest we are using.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -8,4 +8,4 @@ Jinja2>=2.10
 netaddr
 mock
 jmespath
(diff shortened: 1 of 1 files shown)


guits:

guits:

PR #5506, opened 2020-07-03
core: unify site-container.yml and site.yml

guits:
This is an attempt to unify main playbooks.
There are pros and cons, it needs to be discussed.

With this commit, we only have 1 main playbook to manage.


site.yml.sample
```diff
@@ -11,21 +11,23 @@
   - clients
   - mgrs
   - iscsigws
-  - iscsi-gws # for backward compatibility only!
   - grafana-server
-  - rgwloadbalancers
+  - "{{ 'rgwloadbalancers' if not containerized_deployment | bool else omit }}"
```
guits, line 15: useless and doesn't even work anyway

site.yml.sample
```diff
@@ -11,21 +11,23 @@
...
+  any_errors_fatal: true
 
   tags: always
 
   vars:
     delegate_facts_host: True
 
   pre_tasks:
-    # If we can't get python2 installed before any module is used we will fail
+    - import_role:
```
guits, line 27: ditto

Issue #5392, opened 2020-05-23
No information in documentation about containerized vs non-containerized


modir:
The current documentation lists in some chapters that it is possible to install ceph containerized or normal. But the information is rather sparse. What I am missing is:

1. What is the recommended way to install ceph these days when using ceph-ansible? Containerized or not?
2. Is there something to consider when going for containerized? E.g are there limitations?
3. What technology is used in the background when installing ceph containerized? (e.g. docker or LXD)

I came up with this because I stumbled upon this commit and realized there is not enough information in the official documentation of this project: https://github.com/ceph/ceph-ansible/commit/2547ab601a5187251f9dd41246d3dcedb653030f

dsavineau:
> * What is the recommended way to install ceph these days when using ceph-ansible? Containerized or not?

There's no recommended way. It's up to you to choose.

> * Is there something to consider when going for containerized? E.g are there limitations?

You have the same features between non containerized and containerized deployment. The only thing that is different is the name of some playbook (like site, purge)

> * What technology is used in the background when installing ceph containerized? (e.g. docker or LXD)

The container runtime depends on the OS you're using but at the moment ceph-ansible is only supporting docker and podman.

modir:
> You have the same features between non containerized and containerized deployment. The only thing that is different is the name of some playbook (like site, purge)

This means that for the "clients" (e.g. nodes in a KVM cloud cluster) that they have no differences how they mount data disks to a VM? 

> The container runtime depends on the OS you're using but at the moment ceph-ansible is only supporting docker and podman.

It would be great if right after this title https://docs.ceph.com/ceph-ansible/master/#installation would be a sub-title "Requirements for containerized deployment". And there you would have this sentence. (and maybe more information.)
I will try to make a commit with this change.

dsavineau:
> This means that for the "clients" (e.g. nodes in a KVM cloud cluster) that they have no differences how they mount data disks to a VM?

I don't understand what you mean.

The "clients" nodes only import the keyrings and create the pools.

modir:
> The "clients" nodes only import the keyrings and create the pools.

Well, I am not that good with ceph yet, but according to https://docs.ceph.com/docs/master/start/quick-rbd/ the clients create the map of a device image to a local block device as well. Or is this outdated? And if I understood it correctly then I could use those device images to create VMs and mount the images with librados to the VM.
[...]

PR #5538, opened 2020-07-09
[skip ci] Update methods.rst

modir:
Based on the discussion in issue https://github.com/ceph/ceph-ansible/issues/5392 I added now this paragraph to this page.

Signed-off-by: Raffael Luthiger <r.luthiger@huanga.com>

### DIFF
--- a/docs/source/installation/methods.rst
+++ b/docs/source/installation/methods.rst
@@ -8,3 +8,5 @@ ceph-ansible can deploy Ceph either in a non-containerized context (via packages
 
    non-containerized
    containerized
+
+The difference here is that you don't have the rbd command on the host when using the containerized deployment so everything related to ceph needs to be executed within a container. So in the case there is software like e.g. Open Nebula which requires that the rbd command is accessible directly on the host (non-containerized) then you have to install the rbd command by yourself on those servers outside of containers. (Or make sure that this software somehow runs within containers as well and that it can access rbd.


modir:
It would be nice if either this text or something similar would be added to the documentation.

docs/source/installation/methods.rst
```diff
@@ -8,3 +8,5 @@ ceph-ansible can deploy Ceph either in a non-containerized context (via packages
 
    non-containerized
    containerized
+
+The difference here is that you don't have the rbd command on the host when using the containerized deployment so everything related to ceph needs to be executed within a container. So in the case there is software like e.g. Open Nebula which requires that the rbd command is accessible directly on the host (non-containerized) then you need to have to install the rbd command by yourself on those servers outside of containers. (Or make sure that this software somehow runs within containers as well and that it can access rbd.
```
guits, line 12:
```suggestion
The difference here is that you don't have the rbd command on the host when using the containerized deployment so everything related to ceph needs to be executed within a container. So in the case there is software like e.g. Open Nebula which requires that the rbd command is accessible directly on the host (non-containerized) then you have to install the rbd command by yourself on those servers outside of containers. (Or make sure that this software somehow runs within containers as well and that it can access rbd.
```

guits: @modir

guits (approved):

PR #5540, opened 2020-07-10
[skip ci] rhcs: default to containerized deployment

dsavineau:
Starting RHCS 5, only containerized deployment is available.


dsavineau:
I guess this will maybe break the RHCS upgrade workflow

@guits WDYT ?

PR #5541, opened 2020-07-10
[skip ci] cephadm-adopt: use placement parameter

dsavineau:
It's better to use the --placement parameter when using ceph orch apply
commands to avoid confusion in the parameters.


dsavineau:

PR #5554, opened 2020-07-15
[skip ci] cephadm-adopt: delegate task for orch apply

dsavineau:
This is a partial revert of b38019e because we don't want to execute
the whole play on the monitor otherwise if we have some empty group
like rgws or mdss then the orchestrator commands will still be
executed.
Instead we should keep the real target group name at play level and
delegate the orchestator commands to the monitor. The whole play
will be skipped is the group is empty.


dsavineau:

PR #5564, opened 2020-07-20
[skip ci] cephadm: set the command as a fact

dsavineau:
Set the cephadm cmd as a fact instead of rewriting the same command
over and over.
This also fix an issue when using docker as container engine because
the --docker cephadm parameter should be use before the subcommand
not after.


dsavineau:
jenkins test centos-container-cephadm

dsavineau:

dsavineau:

dsavineau:

PR #5544, opened 2020-07-13
facts: fix broken facts when using --limit

guits:
This commit fixes these tasks when --limit is used.

It makes sure the fact is set on right nodes even when the playbook is
run with `--limit`

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #5566, opened 2020-07-20
facts: fix broken facts when using --limit (bp #5544)

guits:
This commit fixes these tasks when --limit is used.

It makes sure the fact is set on right nodes even when the playbook is
run with `--limit`

(cherry picked from commit f8a951f50c6a64ab3d60a1bf66ca9d2db2f6bc35)


dsavineau:

dsavineau:

PR #5570, opened 2020-07-21
podman: always remove container on start

dsavineau:
In case of failure, the systemd ExecStop isn't executed so the container
isn't removed. After a reboot of a failed node, the container doesn't
start because the old container is still present in created state.
We should always try to remove the container in ExecStartPre for this
situation.
A normal reboot doesn't trigger this issue and this also doesn't affect
nodes running containers via docker.
This behaviour was introduced by d43769d.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1858865


dsavineau:

dsavineau:

dsavineau:

dsavineau:
@guits I guess you have to apply to same fix to the ceph-crash PR

dsavineau:

PR #5573, opened 2020-07-21
ceph-handler: add missing condition on ceph-crash

dsavineau:
The ceph-crash tasks present in the ceph-handler role don't need to be
executed on all nodes.


dsavineau:

dsavineau:

PR #5574, opened 2020-07-21
ceph-handler: remove iscsigws restart scripts

dsavineau:
The iscsigws restart scripts for tcmu-runner and rbd-target-{api,gw}
services only call the systemctl restart command.
We don't really need to copy a shell script to do it when we can use
the ansible service module instead.


dsavineau:

dsavineau:

guits:
I'm not sure we should backport this

dsavineau:
> I'm not sure we should backport this

Is there any reason for not backporting this ?

PR #5583, opened 2020-07-23
Commit on behalf of Daniel Parkes

rmahique:
Change the radosgw_frontend_port to take in account more than 1 RGW instance, in it’s original form 'radosgw_frontend_port': radosgw_frontend_port | int , it configured the 8080 port to all instances, with the following modification 'radosgw_frontend_port': radosgw_frontend_port | int + item|int we increase in 1 the port count.

Removed the conditional check rgw_instances is undefined, this has to be checked in detail but this variable was never arriving undefined at this point of the playbook.

Signed-off-by: Raul <rmahique@redhat.com>

### DIFF
--- a/roles/ceph-container-common/tasks/fetch_image.yml
+++ b/roles/ceph-container-common/tasks/fetch_image.yml
@@ -82,13 +82,12 @@
     - ceph_mon_inspect.get('rc') == 0
 
 - name: "inspecting ceph osd container image before pulling"
-  command: "{{ container_binary }} inspect {{ (ceph_osd_inspect.stdout | from_json)[0].Image }}"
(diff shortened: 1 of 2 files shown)


rmahique:
This error has nothing to do with the change.

fatal: [rgw0]: FAILED! => changed=false 
  attempts: 3
  msg: 'Failed to download metadata for repo ''extras'': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried'
  rc: 1
  results: []

dsavineau:
Why are you closing this ?

https://bugzilla.redhat.com/show_bug.cgi?id=1859872 doesn't need to be fixed anymore ?

rmahique:
it does, i think i closed by mistake

rmahique:
I created another pull request:
https://github.com/ceph/ceph-ansible/pull/5613

PR #5591, opened 2020-07-24
ceph-handler: remove iscsigws restart scripts (bp #5574)

dsavineau:
The iscsigws restart scripts for tcmu-runner and rbd-target-{api,gw}
services only call the systemctl restart command.
We don't really need to copy a shell script to do it when we can use
the ansible service module instead.

Backport: #5574

(cherry picked from commit cbe79428e687e383f9764668a56171e5582451be)


dsavineau:

dsavineau:

PR #5600, opened 2020-07-29
wait for all osd to be up

quangln94:


dsavineau:
What is the purpose of this PR ?

PR #5603, opened 2020-07-29
tests: install pyyaml on osd nodes

dsavineau:
Due to [1], ceph-volume has now a dependency on pyyaml but it's not
installed by default via the package dependency.
This patch only add the required package on non containerized
deployment and as temporary workaround for the CI.

[1] https://tracker.ceph.com/issues/46759


dsavineau:

dsavineau:

PR #5608, opened 2020-07-30
docker2podman: set disk_list for non lvm scenario

dsavineau:
When using non lvm scenarios (collocated or non-collocated) then the
disk_list variable isn't set because this is done during the ceph-osd
role (start_osds.yml) which isn't executed in the docker2podman
playbook.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1862046


dsavineau:
jenkins test centos-container-docker2podman

dsavineau:

Issue #5619, opened 2020-08-03
{{ ceph_conf_key_directory }}/{{ cluster }}.conf ownership discrepancy (ceph-config role & switch-from-non-containerized-to-containerized-ceph-daemons.yml)


kevincoakley:
**Bug Report**

What happened:

There is a file ownership discrepancy for {{ ceph_conf_key_directory }}/{{ cluster }}.conf (/etc/ceph/ceph.conf) between the ceph-config role & and the switch-from-non-containerized-to-containerized-ceph-daemons.yml infrastructure-playbook. 

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/tasks/main.yml#L163 sets the owner to **root:root**

https://github.com/ceph/ceph-ansible/blob/master/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml#L98 sets the owner to **ceph:ceph** for everything in /etc/ceph/.  

I have to convert ~1000 OSDs to containerized ceph daemons and as the code is currently written, all of the OSDs need to be restarted when switch-from-non-containerized-to-containerized-ceph-daemons.yml is ran (obviously). Then the first time I run ceph-ansible after the switch, the ownership of /etc/ceph/ceph.conf changes. This ownership change requires all of the OSDs to be restarted a second time with no real effect. Since I am dealing with so many OSDs it takes hours to restart all of the OSDs and I want to prevent it. 

Should the ownership of {{ ceph_conf_key_directory }}/{{ cluster }}.conf be **root:root** or **ceph:ceph**? I would be happy to submit a pull request for the fix, but I am not 100% sure what ownership you want it to be. ceph:ceph seems correct, but changing the ownership in the ceph-config role will cause every existing cluster to restart all daemons when they update the role, which doesn't sound desirable. 

If you prefer neither file be changed, I can make the fix locally, but I would like to push my fix to the community so others don't have to duplicate my effort. 

**Environment**:
* OS (e.g. from /etc/os-release): CentOS Linux release 7.8.2003 (Core)
* Kernel (e.g. `uname -a`):  3.10.0-1127.el7.x86_64
* Docker version if applicable (e.g. `docker version`):
* Ansible version (e.g. `ansible-playbook --version`): 2.9.10
* ceph-ansible version (e.g. `git head or tag or stable branch`): 4.0.26 (this issue is still present in master)
* Ceph version (e.g. `ceph -v`): 14.2.10

PR #5642, opened 2020-08-06
switch_to_containers: Fix cluster config file permissions to match the ceph-config role

kevincoakley:
switch_to_containers changes the ownership of /etc/ceph/* to ceph:ceph using
the find command multiple times. The "generate {{ cluster }}.conf configuration
file" task under the "config file operations for containerized scenarios" block
in the ceph-config role sets the ownership of
"{{ ceph_conf_key_directory }}/{{ cluster }}.conf" to root:root. After running
the switch_to_containers playbook, the ownership of {{ cluster }}.conf is
ceph:ceph. The next time the ceph-ansible playbook (site.yml) is ran, the
ownership is changed to root:root and the handler restarts all ceph services.
For large clusters, this is a time consuming and unnecessary process. This fix
runs the ceph-config role prior to running the service role in order to sync
the ownership of {{ cluster }}.conf so the service only needs to be restarted
once when running switch_to_containers. Also, by importing the ceph-config role
the ownership of {{ cluster }}.conf will remain synced with the ceph-config
role should the "generate {{ cluster }}.conf configuration file" task change.
Alternatively, a task can be added to switch_to_containers after the find
command to change the ownership of
"{{ ceph_conf_key_directory }}/{{ cluster }}.conf" to root:root. This would
results in fewer service restarts when running the switch_to_containers,
however it could result in the ownership of
"{{ ceph_conf_key_directory }}/{{ cluster }}.conf" to not be in sync with the
ceph-config role, should the role be changed.

### DIFF
--- a/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
+++ b/infrastructure-playbooks/switch-from-non-containerized-to-containerized-ceph-daemons.yml
@@ -129,6 +129,9 @@
         name: ceph-container-common
 
     - import_role:
+        name: ceph-config
+
+    - import_role:
         name: ceph-mon
 
   post_tasks:
@@ -191,6 +194,9 @@
 
     - import_role:
         name: ceph-container-common
+
+    - import_role:
+        name: ceph-config
 
     - import_role:
         name: ceph-mgr
@@ -308,6 +314,9 @@
         name: ceph-container-common
 
     - import_role:
+        name: ceph-config
+
+    - import_role:
         name: ceph-osd
 
   post_tasks:
@@ -390,6 +399,9 @@
         name: ceph-container-common
 
     - import_role:
+        name: ceph-config
+
+    - import_role:
          name: ceph-mds
 
 
@@ -447,6 +459,9 @@
         name: ceph-container-common
 
     - import_role:
+        name: ceph-config
+
+    - import_role:
         name: ceph-rgw
 
 
@@ -499,6 +514,9 @@
         name: ceph-container-common
 
     - import_role:
+        name: ceph-config
+
+    - import_role:
         name: ceph-rbd-mirror
 
 
@@ -546,6 +564,9 @@
 
     - import_role:
         name: ceph-container-common
+
+    - import_role:
+        name: ceph-config
 
     - import_role:
         name: ceph-nfs


kevincoakley:
See issue https://github.com/ceph/ceph-ansible/issues/5619

guits:
@kevincoakley why not simply excluding `ceph.conf` from the find command instead of rerunning the whole `ceph-config` role?

```bash
find /var/lib/ceph /etc/ceph -not -( -user {{ ceph_uid }} -or -group {{ ceph_uid }} -or -name ceph.conf -) -execdir chown {{ ceph_uid }}:{{ ceph_uid }} {} +
```

kevincoakley:
In the ceph-config role the ownership of {{ cluster }}.conf (https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/tasks/main.yml#L104) is ceph:ceph for non-containerized daemons, excluding {{ cluster }}.conf from the find statement would mean the ownership was still ceph:ceph. 

If it is a non-containerized daemon install yum/apt would create a ceph user, if it is a containerized daemon install then yum/apt doesn't install ceph and there is no ceph user, therefore the ceph-config role sets the ownership to root.

guits:
> In the ceph-config role the ownership of {{ cluster }}.conf (https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/tasks/main.yml#L104) is ceph:ceph for non-containerized daemons, excluding {{ cluster }}.conf from the find statement would mean the ownership was still ceph:ceph.
> 
> If it is a non-containerized daemon install yum/apt would create a ceph user, if it is a containerized daemon install then yum/apt doesn't install ceph and there is no ceph user, therefore the ceph-config role sets the ownership to root.

Ok, I didn't have all details off the top of my head, but yes, what you say makes sense.

nice catch @kevincoakley  👍

guits:
> I guess it would be better to fix the owner/group on the ceph.conf file for containerized deployment by modifying [1] with the `ceph_uid` fact instead of calling the `ceph-config` role everywhere.
> 
> [1] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/tasks/main.yml#L168-L169

+1 I agree this is a more elegant solution (even though it means those who are already running with `root:root`  will hit the issue reported here).
rerunning the whole `ceph-config` role just for setting `root:root` on 1 file seems too much and would keep this discrepancy between RPM/container deployments

kevincoakley:
Another solution for containerized daemon installs is to update the ceph-config role to set the ownership to match the ceph user uid/gid in the ceph container. 

I don't know how consistent the ceph user uid/gid is in every container build, but if it is installed in the container using the rpm then it should be consistent between container builds. Having everything in /etc/ceph and /var/lib/ceph with uid 167 and gid 167, instead of using the username to set the ownership, would make it look like the files are owned by the ceph user inside the ceph container. However, there are potential security issues if the systems already have a non-privileged user with uid or gid 167 (a rare case, but possible). Also, like mentioned any change to ownership in the ceph-config role will cause everyones services to restart, unless the ceph-config role is updated to not have the handlers run on just a permission/ownership change (this may not be a good idea, maybe ownership change should always cause a service restart).

kevincoakley:
I have another version of switch_to_containers that I can commit that just changes the ownership of "{{ ceph_conf_key_directory }}/{{ cluster }}.conf" to root:root. I pushed this version instead because it would be in sync with the ceph-config role always by having the ceph-config role set the permissions. I can push the other version with just the ownership change if you like.

guits:
> Another solution for containerized daemon installs is to update the ceph-config role to set the ownership to match the ceph user uid/gid in the ceph container.

Isn't what @dsavineau suggested?




> I have another version of switch_to_containers that I can commit that just changes the ownership of `"{{ ceph_conf_key_directory }}/{{ cluster }}.conf"` to `root:root`. I pushed this version instead because it would be in sync with the ceph-config role always by having the ceph-config role set the permissions. I can push the other version with just the ownership change if you like.

That would keep the current discrepancy

kevincoakley:
> Isn't what @dsavineau suggested?

Sorry, yes, it is. I wasn't aware of the existence of the ceph_uid fact. 

> That would keep the current discrepancy

Yes, it would. Fixing the ceph-config role to use the ceph_uid fact is easy to do, however the change will cause every current deployment to restart all ceph services if they run ceph-ansible again. I am not sure if that desired side-effect, but the fix is probably the best practice.  

If that is the path you want to take to fix this issue then I can close this pull request since no changes will need to be made to switch-from-non-containerized-to-containerized-ceph-daemons.yml. 

I would be happy to submit a pull request for the changes to ceph-config. It could just set the owner/group on ""generate {{ cluster }}.conf configuration file" [1] and everyone will have their ceph services restarted. Alternatively, I can use the file module to set the owner/group before [1] without a handler, which should change the owner/group on ceph.conf without restarting the services but is a ugly hack. Otherwise, I can just submit an issue and someone else can solve the issue.


[1] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/tasks/main.yml#L163

PR #5643, opened 2020-08-07
nfs: do not copy rgw keyring when `nfs_obj_gw` is true

guits:
This keyring shouldn't be copied when `nfs_obj_gw` is `True` if the
cluster doesn't contain a rgw node, which can be the case given we are
using `nfs_obj_gw` instead of `nfs_file_gw` (cephfs vs. object), the
deployment will fail trying to copy a key that doesn't exist.


guits:

PR #5659, opened 2020-08-13
[skip ci] infra: only install logrotate on right nodes

guits:
For intsance, there is no need to install logrotate on clients nodes.

### DIFF
--- a/roles/ceph-infra/tasks/main.yml
+++ b/roles/ceph-infra/tasks/main.yml
@@ -24,8 +24,16 @@
     state: present
   register: result
   until: result is succeeded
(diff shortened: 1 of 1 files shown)


roles/ceph-infra/tasks/main.yml
```diff
@@ -24,8 +24,16 @@
     state: present
   register: result
   until: result is succeeded
-  when: not is_atomic | bool
   tags: with_pkg
+  when:
+    - not is_atomic | bool
+    - containerized_deployment
```
dsavineau, line 30: `| bool` maggle

Issue #5677, opened 2020-08-18
alertmanager container fails to start with ansible-2.9.12


karelyatin:
**Bug Report**

Tried deploying TripleO(which uses ceph-ansible to deploy ceph services) with ansible-2.9.12 https://review.rdoproject.org/r/#/c/28996/ but alertmanager container is not able to start. It's due to ansible-2.9.12 made the file permissions strict[1] and now "mode" needs to be specified explicitly, alertmanger.yml create task[2] doesn't have "mode" defined so permissions are set to 600 instead of 666 as in earlier ansible versions.
 
Warning is also reported for /etc/prometheus/prometheus.yml but issue not seen in it as "owner" is set[3] and it has read permission.

[1] https://github.com/ansible/ansible/commit/0199b1cf05e1838bc65ac26d99f2c5505c24ad0f
[2] https://github.com/ceph/ceph-ansible/blob/622d9feae924b216d02fc10a90c5a2089ab98794/roles/ceph-prometheus/tasks/main.yml#L41-L45
[3] https://github.com/ceph/ceph-ansible/blob/622d9feae924b216d02fc10a90c5a2089ab98794/roles/ceph-prometheus/tasks/main.yml#L15

What you expected to happen:

Deployment should succeed with ansible-2.9.12.

How to reproduce it (minimal and precise):

A deployment of alertmanager with ansible-2.9.12 should reproduce it.

Share your group_vars files, inventory and **full** ceph-ansibe log
Log link:- https://logserver.rdoproject.org/96/28996/1/check/rdoinfo-tripleo-master-testing-centos-8-scenario004-standalone/e13825e/logs/undercloud/home/zuul/standalone-ansible-0ofy65b3/ceph-ansible/ceph_ansible_command.log.txt.gz
Pasting relevent warnings from log here:-
[WARNING]: File '/etc/prometheus/prometheus.yml' created with default permissions '600'. The previous default was '666'. Specify 'mode' to avoid this warning.
[WARNING]: File '/etc/alertmanager/alertmanager.yml' created with default permissions '600'. The previous default was '666'. Specify 'mode' to avoid this warning.

**Environment**:
* OS (e.g. from /etc/os-release): CentOS-8
[...]

PR #5680, opened 2020-08-18
[skip ci] Set default permission for prometheus config files

fultonj:
Regardless of the outcome of Ansible 2.9.12 issue 71200
we can set a default permission for these files.

Closes: https://github.com/ceph/ceph-ansible/issues/5677

### DIFF
--- a/roles/ceph-prometheus/tasks/main.yml
+++ b/roles/ceph-prometheus/tasks/main.yml
@@ -13,6 +13,7 @@
     src: prometheus.yml.j2
     dest: "{{ prometheus_conf_dir }}/prometheus.yml"
     owner: "{{ prometheus_user_id }}"
+    mode: 0640
   notify: service handler
 
 - name: make sure the alerting rules directory exists
@@ -42,7 +43,8 @@
   template:
     src: alertmanager.yml.j2
     dest: "{{ alertmanager_conf_dir }}/alertmanager.yml"
-    owner: "root"
+    owner: "{{ prometheus_user_id }}"
+    mode: 0640
   notify: service handler
 
 - name: include setup_container.yml


roles/ceph-prometheus/tasks/main.yml
```diff
@@ -13,6 +13,7 @@
     src: prometheus.yml.j2
     dest: "{{ prometheus_conf_dir }}/prometheus.yml"
     owner: "{{ prometheus_user_id }}"
+    owner: 0644
```
dsavineau, line 16: s/owner/mode/

dsavineau:
IMHO we should check this on all roles and not only prometheus/alertmanager

dsavineau:
> Signed-off-by: John Fulton [fulton@redhat.com](mailto:fulton@redhat.com)

Technically you email should have <> before and after.

dsavineau:
If you want to fix the alertmanager issue only then I think we should set the mode as `0640` instead of `0644` and add the `prometheus_user_id` variable for the owner

roles/ceph-prometheus/tasks/main.yml
fultonj: whops, thanks for spotting that. fixed

fultonj:
With this PR all uses the the template module have a mode (config_template doesn't use mode and is a diff module so that doesn't apply here).

```
[fultonj@runcible ceph-ansible{issues/5677}]$ grep -R template -A 5 | curl -F 'f:1=<-' ix.io
http://ix.io/2uxy
[fultonj@runcible ceph-ansible{issues/5677}]$ 
```

fultonj:
> If you want to fix the alertmanager issue only then I think we should set the mode as `0640` instead of `0644` and add the `prometheus_user_id` variable for the owner

OK, I've done that. thanks for the review

fultonj:
> > Signed-off-by: John Fulton [fulton@redhat.com](mailto:fulton@redhat.com)
> 
> Technically you email should have <> before and after.

fixed in commit message

dsavineau:

dsavineau:

dsavineau:

PR #5710, opened 2020-08-25
client: add nodelete pool flag option

administratorius:
Add an option to change nodelete flag on a pool

Signed-off-by: Vytenis Sabaliauskas <vytenis.sabaliauskas@gmail.com>


administratorius:
Ok, makes sense. Closing PR.

PR #5724, opened 2020-09-03
[skip ci] ceph_pool: add idempotency to absent state

dsavineau:
When using the "absent" state on a non existing pool then the ceph_pool
module will fail and return a python traceback.

Instead we should check if the pool exit or not and execute the pool
deletion according to the result.
The state changed is now set when the pool is actually deleted.

This also disable add_file_common_args because we don't manipulate
files with this module.

### DIFF
--- a/library/ceph_pool.py
+++ b/library/ceph_pool.py
@@ -475,7 +475,6 @@ def run_module():
     module = AnsibleModule(
         argument_spec=module_args,
         supports_check_mode=True,
-        add_file_common_args=True,
     )
 
     # Gather module parameters in variables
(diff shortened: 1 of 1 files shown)


library/ceph_pool.py
```diff
@@ -597,7 +596,13 @@ def run_module():
...
 
     elif state == "absent":
-        rc, cmd, out, err = exec_commands(module, remove_pool(cluster, name, user, user_key, container_image=container_image))
+        rc, cmd, out, err = exec_commands(module, check_pool_exist(cluster, name, user, user_key, container_image=container_image))
+        if rc == 0:
+            rc, cmd, out, err = exec_commands(module, remove_pool(cluster, name, user, user_key, container_image=container_image))
+            changed = True
+        else:
+            rc = 0
+            out = "Pool {} doesn't exist".format(name)
```
guits, line 605:
```suggestion
            out = "Skipped, since pool {} doesn't exist".format(name)
```

guits (approved):

library/ceph_pool.py
guits: @dsavineau do you mind? 🙂

dsavineau: Yes I'll change that

guits (approved):

PR #5751, opened 2020-09-10
[skip ci] ceph_pool: set state as optional

dsavineau:
Most ansible module using a state parameter default to the present
value (when available) instead of using it as a mandatory option.


guits:
even though I like when things are explicit.... why not... 🙂

guits (approved):

PR #5752, opened 2020-09-10
Fix delegation when facts are computed in rolling_update context

fmount:
This change is trying to fix an issue that occurs when rolling_update
playbook is run: in particular the following error is reported:

 FAILED! => {"msg": "list object has no element 0"}

and this patch is just an attempt to fix it.
The change also make the playbook and the delegation consistent with
the rest of the playbook.

Closes: https://bugzilla.redhat.com/1877426


fmount:
@gfidente FYI

dsavineau:
Closing in favor of  #5755

PR #5758, opened 2020-09-10
tests: add quay registry for collocation baremetal

dsavineau:
Even if the non containerized collocation scenario deploys ceph with
RPMs then we also deploy the dashboard/monitoring but with containers.
This requires to set the registry variable to ceph's quay.


dsavineau:

PR #5763, opened 2020-09-11
dashboard: use entity with ceph config get mgr

dsavineau:
That commit [1] introduced a regression in the dashboard configuration
because the `ceph config get mgr xxxx` command doesn't work with
nautilus.
In that release the get operation needs an entity.

[1] f607857f2a58b2ed14faf49f2b10d056a7f96b30

### DIFF
--- a/roles/ceph-dashboard/tasks/configure_dashboard.yml
+++ b/roles/ceph-dashboard/tasks/configure_dashboard.yml
@@ -10,7 +10,7 @@
   block:
    - name: get SSL status for dashboard
      run_once: true
-     command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} config get mgr mgr/dashboard/ssl"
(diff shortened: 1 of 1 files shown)


dsavineau:

dsavineau:

dsavineau:

dsavineau:

roles/ceph-dashboard/tasks/configure_dashboard.yml
```diff
@@ -10,7 +10,7 @@
   block:
    - name: get SSL status for dashboard
      run_once: true
-     command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} config get mgr mgr/dashboard/ssl"
+     command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} config get mgr.{{ hostvars[groups[mgr_group_name][0]]['ansible_hostname'] }} mgr/dashboard/ssl"
```
dsavineau, line 13: We probably need to do some conditioning here if the mgr group isn't defined and fallback to mons[0] (implicite mgr on mon)

guits:
I would say it's probably better to revert this patch in stable-4.0 instead of striving to fix it (given there was no BZ)

PR #5780, opened 2020-09-15
[skip ci] infra: support log rotation for tcmu-runner

guits:
This commit adds the log rotation support for tcmu-runner.

ceph-container related PR: ceph/ceph-container#1726

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1873915

### DIFF
--- a/roles/ceph-infra/templates/logrotate.conf.j2
+++ b/roles/ceph-infra/templates/logrotate.conf.j2
@@ -10,3 +10,16 @@
     notifempty
     su root root
 }
+
+/var/log/tcmu-runner/*.log {
+    rotate 7
+    daily
+    compress
+    sharedscripts
+    postrotate
+        killall -q -1 tcmu-runner || pkill -1 -x "tcmu-runner"
+    endscript
+    missingok
+    notifempty
+    su root root
+}
--- a/roles/ceph-iscsi-gw/templates/tcmu-runner.service.j2
+++ b/roles/ceph-iscsi-gw/templates/tcmu-runner.service.j2
@@ -28,7 +28,7 @@ ExecStart=/usr/bin/{{ container_binary }} run --rm \
   -v /dev:/dev \
   -v /lib/modules:/lib/modules \
   -v /etc/ceph:/etc/ceph \
-  -v /var/log/ceph:/var/log/ceph:z \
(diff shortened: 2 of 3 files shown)


roles/ceph-iscsi-gw/templates/tcmu-runner.service.j2
```diff
@@ -29,6 +29,7 @@ ExecStart=/usr/bin/{{ container_binary }} run --rm \
   -v /lib/modules:/lib/modules \
   -v /etc/ceph:/etc/ceph \
   -v /var/log/ceph:/var/log/ceph:z \
+  -v /var/log/tcmu-runner:/var/log/tcmu-runner:z \
```
dsavineau, line 32: Could we just bind mount the default log file ?

And I think wa can remove the `/var/log/ceph` bind mount

dsavineau (changes requested):

roles/ceph-iscsi-gw/templates/tcmu-runner.service.j2
guits: > Could we just bind mount the default log file ?

so we logrotate a file which is bind-mounted by podman?

roles/ceph-infra/templates/logrotate.conf.j2
```diff
@@ -10,3 +10,16 @@
...
     su root root
 }
+
+/var/log/tcmu-runner/*.log {
+    rotate 7
+    daily
+    compress
+    sharedscripts
+        killall -q -1 tcmu-runner || pkill -1 -x "tcmu-runner"
+    postrotate
```
dsavineau, line 20: postrotate should be set before the killall/pkill command

roles/ceph-iscsi-gw/templates/tcmu-runner.service.j2
dsavineau: looks like it's not possible so let's continue with the directory

dsavineau (approved):

dsavineau:

dsavineau:

dsavineau:

dsavineau:
The failure is unrelated with the current change and we can see that the tcmu-runner container is started correctly with the right bindmount for the log file.

PR #5782, opened 2020-09-15
[skip ci] Include plugin grafana-image-renderer

ismaelpuerto:
BZ1879001


guits (changes requested):
please, remove the last commit `remove last line`, add a real commit message for the first one explaining why you are adding this plugin and finally, add the `Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1879001` line in your commit message.

Thanks!

guits:
@ismaelpuerto I'm not sure this is supposed to work out of the box like this.

see https://grafana.com/grafana/plugins/grafana-image-renderer 

> This plugin is not compatible with the current Grafana Docker image without installing further system-level dependencies. We recommend setting up another Docker container for rendering and using remote rendering, see Remote Rendering Using Docker for reference.

> If you still want to install the plugin in the Grafana docker image we provide instructions for how to build a custom Grafana image, see Grafana Docker documentation for further instructions.

I'm not a grafana expert, not sure whether the image we use has everything needed in order to run with this plugin? @dsavineau any thoughts?

ismaelpuerto:
In this moment, ceph-ansible use [5.4.3](https://github.com/ceph/ceph-ansible/blob/55cd6e83e475ab9ad8d684b88da5325d869e9d1c/group_vars/all.yml.sample#L775) I suppose that is not "the current Grafana Docker", but you are right, in the future, this could be a problem if ceph use other version of grafana more updated.

epuertat:
Hey, @ismaelpuerto. Nice to see you here!

Based on the priority of this issue, we were not planning this for 4.*, but for the next major release (5.0), which is planned to work with cephadm orchestrator and a [pre-built Grafana image](https://github.com/ceph/ceph/blob/ebdb8e56e55488bf2280b4da6c370936940ee554/monitoring/grafana/build/Makefile#L16-L30).

guits:
The linked BZ has been closed, so I'm closing this PR as well. Feel free to reopen if needed.

PR #5789, opened 2020-09-15
switch2container: fixes for running containers on Debian based host (bp #5785)

dsavineau:
The switch-from-non-containerized-to-containerized-ceph-daemons.yml playbook is currently only used for switching from a RPM based deployment to containerized deployment.
When using a DEB based system then there's some differences in the packaging (systemd unit path) and other thing to fix.

Backport: #5785


dsavineau:

PR #5786, opened 2020-09-15
container: add optional http(s) proxy option

dsavineau:
When using a http(s) proxy with either docker or podman we can rely on
the `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` environment variables.
But with ansible, even if those variables are defined in a source file
then they aren't loaded during the container pull/login tasks.
This implements the http(s) proxy support with docker/podman.
Both implementations are different:
  1/ docker doesn't rely en the environment variables with the CLI.
Thos are needed by the docker daemon via systemd.
  2/ podman uses the environment variables so we need to add them to
the login/pull tasks.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1876692

Signed-off-by: Dimitri Savineau <dsavinea@redhat.com>

PR #5794, opened 2020-09-16
container: add optional http(s) proxy option (bp #5786)

dsavineau:
When using a http(s) proxy with either docker or podman we can rely on
the HTTP_PROXY, HTTPS_PROXY and NO_PROXY environment variables.
But with ansible, even if those variables are defined in a source file
then they aren't loaded during the container pull/login tasks.
This implements the http(s) proxy support with docker/podman.
Both implementations are different:
  1/ docker doesn't rely en the environment variables with the CLI.
Thos are needed by the docker daemon via systemd.
  2/ podman uses the environment variables so we need to add them to
the login/pull tasks.

Backport: #5786
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1876692

(cherry picked from commit bda3581294c8f29eda598522c331a4c009243884)

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -583,6 +583,9 @@ dummy:
 #ceph_docker_registry_auth: false
 #ceph_docker_registry_username:
 #ceph_docker_registry_password:
+#ceph_docker_http_proxy:
+#ceph_docker_https_proxy:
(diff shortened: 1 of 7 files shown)


dsavineau:

dsavineau:

dsavineau:

PR #5799, opened 2020-09-17
ceph-dashboard: increase radosgw-admin timeout

dsavineau:
20s could be too short in some case so raising this to 60s.

### DIFF
--- a/roles/ceph-dashboard/tasks/configure_dashboard.yml
+++ b/roles/ceph-dashboard/tasks/configure_dashboard.yml
@@ -183,7 +183,7 @@
   run_once: true
   block:
     - name: get radosgw system user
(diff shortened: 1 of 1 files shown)


dsavineau:

dsavineau:

dsavineau:

dsavineau:
so it could be related to nfs-ganesha is segfauting so the radosgw-admin command timeouts to notify the watchers (rgw and ganesha processes) which is set to 10s. Because there's 2 watches for ganesha and the command takes only 2s to be executed then we reach the command timeout set to 20s (10+10+2)

```console
# rados listwatchers -p default.rgw.control 'notify.0'
watcher=192.168.42.17:0/620183115 client.10868 cookie=93845376345680
watcher=192.168.42.16:0/3241210843 client.5015 cookie=94664958685568
watcher=192.168.42.17:0/2673994322 client.10892 cookie=94858711346784
```

.17 is ganesha
.16 is rgw daemon

PR #5803, opened 2020-09-17
tests: disable container nfs testing

dsavineau:
Looks like nfs-ganesha 3.3 and 4.-dev doesn't work with recent changes
in librgw 16.0.0.
The nfs-ganesha daemon is segfaulting and restart in a loop.

See https://tracker.ceph.com/issues/47520

### DIFF
--- a/tests/functional/all_daemons/container/hosts
+++ b/tests/functional/all_daemons/container/hosts
@@ -19,8 +19,8 @@ mds2
 [rgws]
 rgw0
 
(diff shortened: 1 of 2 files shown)


dsavineau:

dsavineau:
if this fixes the dashboard radosgw user creation issue then we don't need https://github.com/ceph/ceph-ansible/pull/5799 anymore

PR #5593, opened 2020-07-24
defaults: change default grafana-server name

guits:
This change default value of grafana-server group name.
Adding some tasks in ceph-defaults in order to keep backward
compatibility.

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #5810, opened 2020-09-18
grafana-server - Invalid characters were found in group names

rh-dluong:
In the upcoming release of ansible 2.10, grafana-server group will no longer
be viable to use by default, it will fail instead of give a warning.

To fix this we'll need to convert all the references to the group from '-' to '_'.

I _believe_ I got all the exceptions like grafana_services vars to stay grafana-server instead of grafana_server, and I believe I changed the tests/vagrantfiles to be correct too, but someone definitely needs to double check that.


guits:
This patch is a breaking change, by the way this is already handled by #5593

PR #5812, opened 2020-09-21
Bluelvm

jjict:
I have updated my changes to the LVM Creation Playbook. These Changes make them support Bluestore and multiple NVMe devices as well as all devices may now be specified at the inventory.


jjict:
can anyone tell me what I am doing wrong? to me it looks like there is no link between my changes and the failing test.

PR #5820, opened 2020-09-23
[skip ci] Remove libjemalloc1 installation task

noonedeadpunk:
libjemalloc1 package is not required neither for ganesha dependency nor
for the package build process. So this task can be simply dropped.

Signed-Off-By: Dmitriy Rabotyagov <noonedeadpunk@ya.ru>

### DIFF
--- a/roles/ceph-nfs/tasks/pre_requisite_non_container_debian.yml
+++ b/roles/ceph-nfs/tasks/pre_requisite_non_container_debian.yml
@@ -49,13 +49,6 @@
         - (ceph_origin == 'repository' or ceph_origin == 'distro')
         - ceph_repository != 'rhcs'
       block:
-        - name: install jemalloc for debian
-          apt:
-            name: libjemalloc1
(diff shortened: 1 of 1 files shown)


dsavineau:
I'm pretty sure that nfs-ganesha isn't built at all with jemalloc so I don't think we need this task at all.

I'll double check this.

noonedeadpunk:
I think you might be right. Even if it's needed, it will be pulled as a dependency, since community repos don't contain nfs-ganesha for modern Ubuntu, and it's installed from native repos

dsavineau:
I've checked both rpm and deb spec files and there's nothing related to jemalloc in the dependency nor the build process (you can use jemalloc with ganesha but the option is off by default).
I've tested on Debian Buster with ganesha 3.3 and 2.8 and there's no dynamic jemalloc library linked with the ganesha binary or library whereas ceph is using tcmalloc.
I think you can update your PR and remove the task, it will be fine.

dsavineau:

dsavineau:

dsavineau (approved):

guits (approved):

PR #5852, opened 2020-09-25
ceph-handler: set handler on xxx_stat result

dsavineau:
In non containerized deployment we check if the service is running
via the socket file presence.
This is done via the xxx_socket_stat variable that check the file
socket in the /var/run/ceph/ directory.
In some scenarios, we could have the socket file still present in
that directory but not used by any process.
That's why we have the xxx_stat variable which clean those leftovers.

The problem here is that we're set the variable for the handlers status
(like handler_mon_status) based on xxx_socket_stat instead of xxx_stat.
That means we will trigger the handlers if there's an old socket file
present on the system without any process associated.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1866834

### DIFF
--- a/roles/ceph-handler/tasks/main.yml
+++ b/roles/ceph-handler/tasks/main.yml
@@ -5,37 +5,37 @@
 # We do not want to run these checks on initial deployment (`socket.rc == 0`)
 - name: set_fact handler_mon_status
   set_fact:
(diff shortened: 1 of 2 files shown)


dsavineau:

dsavineau:

dsavineau:

dsavineau:

guits (approved):

PR #5854, opened 2020-09-25
library/ceph_pool: use ceph config show command

dsavineau:
The `ceph config get` command doesn't return the running configuration
but only the configuration option(s) with default ceph values.
That means that even if we set osd_pool_default_size to 1 or 2 in the
ceph configuration file then the ceph config get command will always
return 3 because that's the ceph default value but not the running
value.
Instead we need to use the `ceph config show` command which uses the
existing mon entity (wildcard isn't allowed).

```console
-----
config get <who> [<key>] :  Show configuration option(s) for an entity
config show <who> [<key>] :  Show running configuration
-----
```

```console
$ grep 'osd pool default size' /etc/ceph/ceph.conf
osd pool default size = 1
$ ceph config get mon.* osd_pool_default_size
3
$ ceph config show mon.mon0 osd_pool_default_size
1
```

### DIFF
--- a/library/ceph_pool.py
+++ b/library/ceph_pool.py
@@ -137,6 +137,7 @@ from ansible.module_utils.basic import AnsibleModule  # noqa E402
 import datetime  # noqa E402
 import json  # noqa E402
 import os  # noqa E402
+import platform  # noqa E402
 import stat  # noqa E402
 import time  # noqa E402
 
@@ -237,8 +238,8 @@ def generate_get_config_cmd(param, cluster, user, user_key, container_image=None
         '--cluster',
         cluster,
         'config',
-        'get',
-        'mon.*',
+        'show',
+        'mon.{}'.format(platform.node().split('.')[0]),
         param
     ]
     cmd = _cmd + args
--- a/tests/library/test_ceph_pool.py
+++ b/tests/library/test_ceph_pool.py
@@ -1,5 +1,6 @@
 import json
 import os
+import platform
 import sys
 sys.path.append('./library')
 import ceph_pool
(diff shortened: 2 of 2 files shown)


dsavineau:
of course this doesn't work for external setup because we don't delegate the task on a monitor...

dsavineau:

guits:
> of course this doesn't work for external setup because we don't delegate the task on a monitor...

from the `ceph-client` role perspective, the `ceph_pool` module call implies you have the required keyring (admin), whatever you are running in an external setup or not.

see: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-client/tasks/create_users_keys.yml#L68

are you concerned because of these other following calls?

```py
roles/ceph-iscsi-gw/tasks/common.yml:56:                   ceph_pool:
roles/ceph-mds/tasks/create_mds_filesystems.yml:13:        ceph_pool:
roles/ceph-osd/tasks/openstack_config.yml:5:               ceph_pool:
roles/ceph-rgw/tasks/rgw_create_pools.yml:31:              ceph_pool:
roles/ceph-rgw/tasks/rgw_create_pools.yml:50:              ceph_pool:
```

do we support the standalone deployment of these daemons?

dsavineau:
> from the `ceph-client` role perspective, the `ceph_pool` module call implies you have the required keyring (admin), whatever you are running in an external setup or not.
> 
> see: https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-client/tasks/create_users_keys.yml#L68
> 
> are you concerned because of these other following calls?

No I'm concerned about the client role because we have the external scenario. With this scenario we don't delegate the ceph_pool task on the monitor but from one of the client node.

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-client/tasks/create_users_keys.yml#L25-L27

That means this patch doesn't work in this situation because we will try to get the running cluster configuration from `mon.<client node>` which doesn't exist.

> do we support the standalone deployment of these daemons?

Nope

guits:
I think we are good to close it in favor of #5865
we can still re-open if needed

PR #5883, opened 2020-10-01
requirements: don't exclude 2.9.10 explicitly

guits:
Since 2.9.10 isn't the latest 2.9 release anymore, we don't need to
explicitly exclude this version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 # These are Python requirements needed to run ceph-ansible master
-ansible>=2.9,<2.10,!=2.9.10
+ansible>=2.9,<2.10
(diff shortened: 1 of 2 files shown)


dsavineau (commented):
even if new installation should get newer 2.9 release then I think we should keep the 2.9.10 exclude just in case someone explicitely use 2.9.10 which is completly broken with ceph-ansible

guits:
> even if new installation should get newer 2.9 release then I think we should keep the 2.9.10 exclude just in case someone explicitely use 2.9.10 which is completly broken with ceph-ansible

sounds good to me maggle

PR #5894, opened 2020-10-02
Disable redirect mechanism in a ssl+haproxy context

fmount:
When ssl is enabled and haproxy is configured (because the vip is
provided only when we have haproxy entries), according to the doc
[1][2] we want to disable the redirection behaviour to prevent
situations when internal (unresolvable) URL’s are published to the
frontend client.
By doing this haproxy is able to respond with a HTTP error (503),
sending the user to the active instance.

[1] https://docs.ceph.com/en/latest/mgr/dashboard/#disable-the-redirection
[2] https://docs.ceph.com/en/latest/mgr/dashboard/#haproxy-example-configuration

### DIFF
--- a/roles/ceph-dashboard/tasks/configure_dashboard.yml
+++ b/roles/ceph-dashboard/tasks/configure_dashboard.yml
@@ -91,6 +91,12 @@
       delegate_to: "{{ groups[mon_group_name][0] }}"
       run_once: true
 
+    - name: disable redirect mechanism of the dashboard
+      command: "{{ container_exec_cmd }} ceph --cluster {{ cluster }} config set mgr mgr/dashboard/standby_behaviour error"
+      delegate_to: "{{ groups[mon_group_name][0] }}"
+      run_once: true
+      when: dashboard_frontend_vip is defined and dashboard_frontend_vip |length > 0
+
 - name: "set the dashboard port ({{ dashboard_port }})"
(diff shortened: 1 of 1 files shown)


guits:
@fmount does it need to be backported? is there a BZ related to this? 
Thanks!

guits:

fmount:
@guits I have the bz and I can easily update the PR, just waiting for a feedback and still figuring out if we can avoid this task in the ssl context

dsavineau:
@fmount according to the documentation you need to do this change when running the dashboard behing a proxy so should we do this for both http and https dashboard_protocol values ? Currently you're only doing this for https scenario.

fmount:
@guits @dsavineau after some investigation,  we ended up with this upstream change [1] on OSP side and sounds like disabling the ssl check on haproxy is enough to process the redirects generated by the inactive backends.
With this configuration we are able to reach the active dashboard instance and, since with [2] we can control the status generated from the backends, there's no need of having [3] and [4]: this means the behaviour can still be "redirect".

Closing this PR as we have a solution that can avoid a new ceph-ansible change.

[1] https://review.opendev.org/#/c/755730/
[2] https://review.opendev.org/#/c/755730/12/manifests/haproxy.pp@985
[3] https://github.com/ceph/ceph-ansible/pull/5894
[4] https://bugzilla.redhat.com/show_bug.cgi?id=1884543
[5] https://docs.ceph.com/en/latest/mgr/dashboard/#haproxy-example-configuration

PR #5896, opened 2020-10-02
[skip ci] flake8: run the workflow conditionally

dsavineau:
We don't need to run flake8 on ansible modules and their tests if we
don't have any modifitions.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -1,5 +1,11 @@
 name: flake8
-on: [pull_request]
+on:
+  pull_request:
+    paths:
+      - 'library/**.py'
+      - 'tests/conftest.py'
+      - 'tests/library/**.py'
+      - 'tests/functional/tests/**.py'
 jobs:
   build:
     runs-on: ubuntu-latest


.github/workflows/flake8.yml
```diff
@@ -1,5 +1,9 @@
 name: flake8
-on: [pull_request]
+on:
+  pull_request:
+    paths:
+      - 'library/**.py'
+      - 'tests/library/**.py'
```
guits, line 6: we also have `./tests/conftest.py` and everything in `./tests/functional/tests/`

guits (changes requested):

.github/workflows/flake8.yml
dsavineau: but we're not testing those files at the moment.

Should we wait for #5897 to be merged first ?

guits: > but we're not testing those files at the moment.

I wouldn't be concerned because of that anyway :man_shrugging:

guits: > Should we wait for #5897 to be merged first ?

Why would we wait for #5897 ? You can still edit this patch anyway

guits (approved):

PR #5900, opened 2020-10-05
ceph_key: remove backward compatibility

dsavineau:
It's time to remove this backward compatibility. Users had enough time
to convert their openstack_keys and key values.
We now fail in ceph-validate if the caps key isn't set.

### DIFF
--- a/roles/ceph-client/tasks/create_users_keys.yml
+++ b/roles/ceph-client/tasks/create_users_keys.yml
@@ -1,15 +1,4 @@
 ---
-- name: set_fact keys_tmp - preserve backward compatibility after the introduction of the ceph_keys module
-  set_fact:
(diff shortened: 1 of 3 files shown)


guits (approved):

guits:
is there any reason we shouldn't backport this in stable-5.0 ? (OSP?)

dsavineau:
> is there any reason we shouldn't backport this in stable-5.0 ? (OSP?)

OSP will never use stable-5.0 so I guess we can backport this safely.

As far as I know OSP was already using the right structure in stable-4.0.

dsavineau:
> > is there any reason we shouldn't backport this in stable-5.0 ? (OSP?)
> 
> OSP will never use stable-5.0 so I guess we can backport this safely.
> 
> As far as I know OSP was already using the right structure in stable-4.0.

For reference https://github.com/openstack/tripleo-ansible/blob/stable/train/tripleo_ansible/roles/tripleo-ceph-work-dir/tasks/build_keys.yml (train == OSP16 == RHCS 4)

guits:
> > is there any reason we shouldn't backport this in stable-5.0 ? (OSP?)
> 
> OSP will never use stable-5.0 so I guess we can backport this safely.
> 
> As far as I know OSP was already using the right structure in stable-4.0.

from downstream point of view, we could get rid of this in stable-4.0 as well, but for upstream it's probably safer to stick to backporting in stable-5.0 only I guess

Issue #5910, opened 2020-10-06
ceph-rgw: Pool get the wrong crush rule when rule_name isn't specified


BenoitKnecht:
**Bug Report**

What happened:

With `group_vars/rgws.yml` set to

```yaml
rgw_create_pools:
  "{{ rgw_zone }}.rgw.buckets.data":
    type: replicated
    pg_num: 2048
```

and `group_vars/osds.yml` set to

```yaml
crush_rule_config: true

crush_rules:
- name: replicated-default-hdd
  root: default
  type: host
  class: hdd
  default: true
```

`ceph-rgw` changed the crush rule on `default.rgw.buckets.data` from `replicated-default-hdd` to `replicated_rule`, leading the cluster to try and move a lot of objects, and several OSDs to crash in the process.

What you expected to happen:

`ceph-rgw` shouldn't have tried to change the crush rule, or it should have set it to the correct default rule, which is `replicated-default-hdd` in this case.

How to reproduce it (minimal and precise):

Run `ansible-playbook site.yml.sample` with the configuration above.

**Environment**:

* OS (e.g. from /etc/os-release): CentOS 7
* Ansible version (e.g. `ansible-playbook --version`): 2.9.13
* ceph-ansible version (e.g. `git head or tag or stable branch`): stable-4.0
* Ceph version (e.g. `ceph -v`): 14.2.11

The issue comes from the fact that this task:

https://github.com/ceph/ceph-ansible/blob/765db7ceecf1b75b18c113495673c45c3ae2375e/roles/ceph-rgw/tasks/rgw_create_pools.yml#L63-L74

sets the crush rule to `ceph_osd_pool_default_crush_rule_name` if `rule_name` is unspecified, but that fact is only being set in

https://github.com/ceph/ceph-ansible/blob/765db7ceecf1b75b18c113495673c45c3ae2375e/roles/ceph-facts/tasks/get_def_crush_rule_name.yml#L2-L14

which is not called on the `rgws` hosts. So instead, the default value

https://github.com/ceph/ceph-ansible/blob/765db7ceecf1b75b18c113495673c45c3ae2375e/roles/ceph-defaults/vars/main.yml#L6

ends up being used.

[...]

PR #5921, opened 2020-10-07
ceph-facts: Fix osd_pool_default_crush_rule fact

### LINKED ISSUE #5912: ceph-facts: Fact osd_pool_default_crush_rule is not set correctly
**Bug Report**

What happened:

After deploying a cluster with `group_vars/osds.yml` set to

```yaml
crush_rule_config: true

crush_rules:
- name: replicated-default-hdd
  root: default
  type: host
  class: hdd
  default: true
```

using `ansible-playbook site.yml.sample`, I noticed that `/etc/ceph/ceph.conf` contained

```ini
[global]
osd pool default crush rule = -1
```

What you expected to happen:

I expected to see

```ini
[global]
osd pool default crush rule = 1
```

since the ID of the `replicated-default-hdd` crush rule is `1`.

How to reproduce it (minimal and precise):

Run `ansible-playbook site.yml.sample` with the configuration above.

**Environment**:
* OS (e.g. from /etc/os-release): CentOS 7
* Ansible version (e.g. `ansible-playbook --version`): 2.9.13
* ceph-ansible version (e.g. `git head or tag or stable branch`): stable-4.0
* Ceph version (e.g. `ceph -v`): 14.2.11

The issue comes from

https://github.com/ceph/ceph-ansible/blob/765db7ceecf1b75b18c113495673c45c3ae2375e/roles/ceph-facts/tasks/facts.yml#L293-L315

The `get default crush rule value from ceph configuration` task gets the correct value from `/etc/ceph/ceph.conf`, and sets `crush_rule_variable` accordingly. But then `get default crush rule value from running monitor ceph configuration` is skipped, but it still overwrites `crush_rule_variable` with

```yaml
crush_rule_variable:
  changed: false
  skip_reason: Conditional result was False
  skipped: true
```

After that, `set_fact osd_pool_default_crush_rule` sets `osd_pool_default_crush_rule` to `ceph_osd_pool_default_crush_rule` (which is equal to `-1`), since `crush_rule_variable.get('rc', 1)` returns `1`.

And finally, `ceph-config` generates `/etc/ceph/ceph.conf` with that value:

https://github.com/ceph/ceph-ansible/blob/765db7ceecf1b75b18c113495673c45c3ae2375e/roles/ceph-config/templates/ceph.conf.j2#L35

BenoitKnecht:
The `osd_pool_default_crush_rule` is set based on `crush_rule_variable`, which
is the output of a `grep` command.

However, two consecutive tasks can set that variable, and if the second task is
skipped, it still overwrites the `crush_rule_variable`, leading the
`osd_pool_default_crush_rule` to be set to `ceph_osd_pool_default_crush_rule`
instead of the output of the first task.

This commit ensures that the fact is set right after the `crush_rule_variable`
is assigned, before it can be overwritten.

Closes #5912

### DIFF
--- a/roles/ceph-facts/tasks/facts.yml
+++ b/roles/ceph-facts/tasks/facts.yml
@@ -308,29 +308,35 @@
     path: '/etc/ceph/{{ cluster }}.conf'
   register: ceph_conf
 
+- name: set default osd_pool_default_crush_rule fact
+  set_fact:
+    osd_pool_default_crush_rule: "{{ ceph_osd_pool_default_crush_rule }}"
+
 - name: get default crush rule value from ceph configuration
-  command: grep 'osd pool default crush rule' /etc/ceph/{{ cluster }}.conf
-  register: crush_rule_variable
-  changed_when: false
-  check_mode: no
-  failed_when: false
+  block:
+    - &read-osd-pool-default-crush-rule
+      name: read osd pool default crush rule
+      command: grep 'osd pool default crush rule' /etc/ceph/{{ cluster }}.conf
+      register: crush_rule_variable
+      changed_when: false
+      check_mode: no
+      failed_when: crush_rule_variable.rc not in (0, 1)
+    - &set-osd-pool-default-crush-rule-fact
+      name: set osd_pool_default_crush_rule fact
+      set_fact:
+        osd_pool_default_crush_rule: "{{ crush_rule_variable.stdout.split(' = ')[1] }}"
+      when: crush_rule_variable.rc == 0
   when: ceph_conf.stat.exists | bool
 
 - name: get default crush rule value from running monitor ceph configuration
-  command: grep 'osd pool default crush rule' /etc/ceph/{{ cluster }}.conf
-  register: crush_rule_variable
-  changed_when: false
-  check_mode: no
-  failed_when: false
-  run_once: true
-  delegate_to: "{{ running_mon }}"
+  block:
+    - <<: *read-osd-pool-default-crush-rule
+      run_once: true
+      delegate_to: "{{ running_mon }}"
+    - *set-osd-pool-default-crush-rule-fact
   when:
     - running_mon is defined
     - not ceph_conf.stat.exists | bool
-
-- name: set_fact osd_pool_default_crush_rule
-  set_fact:
-    osd_pool_default_crush_rule: "{{ crush_rule_variable.stdout.split(' = ')[1] if crush_rule_variable.get('rc', 1) | int == 0 else ceph_osd_pool_default_crush_rule }}"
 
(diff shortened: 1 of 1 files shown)


roles/ceph-facts/tasks/facts.yml
```diff
@@ -308,30 +308,36 @@
...
 - name: get default crush rule value from running monitor ceph configuration
-  command: grep 'osd pool default crush rule' /etc/ceph/{{ cluster }}.conf
-  register: crush_rule_variable
-  changed_when: false
-  check_mode: no
-  failed_when: false
-  run_once: true
-  delegate_to: "{{ running_mon }}"
+  block:
+    - <<: *read-osd-pool-default-crush-rule
```
guits, line 333: @BenoitKnecht I'm a bit confused, I think I'm missing something.
Why would we run the task `read osd pool default crush rule` if we have a condition `- not ceph_conf.stat.exists | bool` in that block?

BenoitKnecht: That's the way it was implemented before, I didn't change the logic there.

It makes sense because the task is delegated to `running_mon` in this case (line 335 below).

Maybe it's the YAML override syntax that isn't clear here. The

```yaml
<<: *anchor
option: value
```

syntax simply takes all the keys referred to by `*anchor`, and adds (or overrides if it was already present in `*anchor`) the `option` key.

So here, we run the same command as in the `read osd pool default crush rule` task from the block above (line 317), but with

```yaml
run_once: true
delegate_to: "{{ running_mon }}"
```

added.

BenoitKnecht: @guits Have you had a chance to take another look at this? Is it clearer now, or did I misunderstand your question?

BenoitKnecht:
I just want to add a quick note about the severity of the issue that this PR fixes.

In conjunction with #5910, setting `osd_pool_default_crush_rule` to the wrong value results in existing pools having their crush rule changed. On pools containing a large amount of data, it's pretty catastrophic, as it remaps a huge amount of PGs and tries to reshuffle massive amounts of data. In our test cluster, this caused several OSDs to become overwhelmed and crash.

And even if we work around #5910, the fact that the `osd pool default crush rule` setting in `/etc/ceph/ceph.conf` changes back and forth during a `site.yml.sample` run causes `ceph-handler` to restart all Ceph daemons (including `ceph-osd`), which is a big drawback on large clusters when the playbook was not expected to make significant changes.

So while it may seem like a fairly minor fix on the surface, it has a pretty significant operational impact, and I would love to see it merged and backported ASAP. Let me know if there's anything I can do to help in that regard.

guits (approved):

PR #5922, opened 2020-10-07
Update lvm-batch.yml

quangln94:
Deploy OSD on each node, instead of all OSDs at the same time:
Example:
Node 1: OSD.1 OSD.2 OSD.3
Node 2: OSD.4 OSD.5 OSD.6
Node 3: OSD.7 OSD.8 OSD.9


guits:
a couple of details:

1/ in a cloud mindset, you don't care about where a specific osd id is deployed
2/ in a large cluster deployment with hundreds of nodes, deploying serially would be really time-consuming.

Thanks for your interest in ceph-ansible and your contribution but unfortunately this patch can't be merged @quangln94

PR #5932, opened 2020-10-14
ceph-osd: don't start the OSD services twice

dsavineau:
Using the + operation on two lists doesn't filter out the duplicate
keys.
Currently each OSDs is started (via systemd) twice.
Instead we could use the union filter.


guits:

guits (approved):
union for the win maggle

PR #5945, opened 2020-10-19
config: Always use osd_memory_target if set

gaudenz:
The osd_memory_target variable was only used if it was higher than the
calculated value based on the number of OSDs. This is changed to always
use the value if it is set in the configuration. This allows this value
to be intentionally set lower so that it does not have to be changed
when more OSDs are added later.

### DIFF
--- a/roles/ceph-config/templates/ceph.conf.j2
+++ b/roles/ceph-config/templates/ceph.conf.j2
@@ -82,18 +82,16 @@ filestore xattr use omap = true
 {% if osd_objectstore == 'bluestore' %}
 {% set _num_osds = num_osds | default(0) | int %}
 [osd]
-{% if is_hci | bool and _num_osds > 0 %}
+{% if is_hci | bool and _num_osds > 0 and not osd_memory_target is defined %}
 {# hci_safety_factor is the safety factor for HCI deployments #}
-{% if ansible_memtotal_mb * 1048576 * hci_safety_factor / _num_osds > osd_memory_target %}
 {% set _osd_memory_target = (ansible_memtotal_mb * 1048576 * hci_safety_factor / _num_osds) | int %}
-{% endif %}
-{% elif _num_osds > 0 %}
+{% elif _num_osds > 0 and not osd_memory_target is defined %}
 {# non_hci_safety_factor is the safety factor for dedicated nodes #}
-{% if ansible_memtotal_mb * 1048576 * non_hci_safety_factor / _num_osds > osd_memory_target %}
 {% set _osd_memory_target = (ansible_memtotal_mb * 1048576 * non_hci_safety_factor / _num_osds) | int %}
 {% endif %}
-{% endif %}
+{% if _osd_memory_target is defined or osd_memory_target is defined %}
 osd memory target = {{ _osd_memory_target | default(osd_memory_target) }}
+{% endif %}
 {% endif %}
 {% endif %}
 
--- a/roles/ceph-defaults/defaults/main.yml
+++ b/roles/ceph-defaults/defaults/main.yml
@@ -338,10 +338,12 @@ cephfs_pools:
 
 ## OSD options
 #
+# Variables to calculate a suitable osd_memory_target based on the number of OSDs
 is_hci: false
 hci_safety_factor: 0.2
 non_hci_safety_factor: 0.7
-osd_memory_target: 4294967296
+# You can also set an explicit value which will override the calculation if set.
+# osd_memory_target: 4294967296
 journal_size: 5120 # OSD journal size in MB
 block_db_size: -1 # block db size in bytes for the ceph-volume lvm batch. -1 means use the default of 'as big as possible'.
 public_network: 0.0.0.0/0
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -346,10 +346,12 @@ dummy:
 
 ## OSD options
 #
+# Variables to calculate a suitable osd_memory_target based on the number of OSDs
 #is_hci: false
 #hci_safety_factor: 0.2
 #non_hci_safety_factor: 0.7
-#osd_memory_target: 4294967296
+# You can also set an explicit value which will override the calculation if set.
+# osd_memory_target: 4294967296
 #journal_size: 5120 # OSD journal size in MB
 #block_db_size: -1 # block db size in bytes for the ceph-volume lvm batch. -1 means use the default of 'as big as possible'.
 #public_network: 0.0.0.0/0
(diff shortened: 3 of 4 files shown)


guits (approved):

guits:

guits:

guits:

roles/ceph-defaults/defaults/main.yml
```diff
@@ -338,10 +338,12 @@ cephfs_pools:
 
 ## OSD options
 #
+# Variables to calculate a suitable osd_memory_target based on the number of OSDs
 is_hci: false
 hci_safety_factor: 0.2
 non_hci_safety_factor: 0.7
-osd_memory_target: 4294967296
+# You can also set an explicit value which will override the calculation if set.
+# osd_memory_target: 4294967296
```
guits, line 346: This should be left uncommented out
```suggestion
osd_memory_target: 4294967296
```

otherwise the `osd memory target = {{ _osd_memory_target | default(osd_memory_target) }}` in ceph.conf.j2 doesn't make sense anymore since it won't work

guits (changes requested):

BenoitKnecht:
I think it would be nice to still support the current behavior (i.e. compute the "ideal" memory target, but only if above a minimum threshold), while also account for @gaudenz use-case.

In essence, I would introduce a new variable called `min_osd_memory_target=4294967296`. Then if `osd_memory_target` is undefined, set the memory target to `max(ansible_memtotal_mb * 1048576 * non_hci_safety_factor / _num_osds, min_osd_memory_target)`, and if `osd_memory_target` is defined, simply set the memory target to that value.

Seems to me like the least surprising behavior, while still allowing for the same functionality as before (albeit with some configuration changes required).

What do you think?

roles/ceph-defaults/defaults/main.yml
gaudenz: Your proposed fix won't work. If the `osd_memory_target` is defined in the defaults, it will always take precedence after my change. AFAIK you can't unset a variable that's set in role defaults and the logic is based on the assumption that the calculated value should be used if `osd_memory_target` is not defined.

But there is indeed an issue if `_num_osds == 0` because then `_osd_memory_target` will be undefined even if `osd_memory_target` is not defined. I fixed this by adding another default value directly to the template to cover this edge case .

gaudenz:
@BenoitKnecht I don't think your logic is useful. It will lead to overcommiting host memory if the number of OSDs is too high. What might be useful is to introduce a `max_osd_memory_target` variable and set `min(calculated_value, max_osd_memory_target)`. But this would also make the use case where one just want's to set a fixed value hard. The calculated value would then serve as a safety measure to avoid overcommiting memory.

The main problem I see with the autocalculation is that the `ceph.conf` file changes if OSDs are added and the handler restarts all OSDs which can take very long and is quite unexpected. It can also be disruptive on a busy cluster because of the additional peering required.

BenoitKnecht:
@gaudenz My issue with your proposal is that it would not only change the current default behavior, but it wouldn't even let users keep it with a configuration change if they wanted to. Their only option to keep the same `osd memory target` after upgrading Ceph Ansible would be to hard-code the value where needed.

I'm only suggesting a way for users who upgrade to keep the current behavior. And if you don't think it makes sense to have a minimum value for `osd memory target`, users can set `min_osd_memory_target=0` to obtain the behavior you're describing.

gaudenz:
@BenoitKnecht I don't understand your comment. The current default behavior is unchanged by my proposal. If `osd_memory_target` is not set (the default) then the calculated value is used and nothing changes about the calculation. So as far as I understand the default behavior does not change. And the current behavior if `osd_memory_target` is set is IMHO a bug. I don't see how anyone would want to set a value that is higher than the result of the automatic calculation. This is IMHO a clear recipe to shoot yourself in the foot. The automatic calculation is meant to be an upper limit of what's acceptable to not overcommit memory given the current number of OSDs and the current RAM installed on the host. So I don't see how a static value which is only applied if it's greater than the upper limit of what would be a sensible value is an option anyone would want.

What's your use case for the current behavior?

@guits What's do you think about this. If there is a general consensus that this is something useful, I won't oppose and can rewrite my patch. As long as there is a way to set a value that is always consistently applied, my use case is covered.

roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -82,18 +82,14 @@ filestore xattr use omap = true
 {% if osd_objectstore == 'bluestore' %}
 {% set _num_osds = num_osds | default(0) | int %}
 [osd]
-{% if is_hci | bool and _num_osds > 0 %}
+{% if is_hci and _num_osds > 0 and not osd_memory_target is defined %}
```
dsavineau, line 85: You have to put back the `bool` filter there

roles/ceph-config/templates/ceph.conf.j2
```diff
@@ -82,18 +82,14 @@ filestore xattr use omap = true
...
-{% endif %}
-{% elif _num_osds > 0 %}
+{% elif _num_osds > 0 and not osd_memory_target is defined %}
 {# non_hci_safety_factor is the safety factor for dedicated nodes #}
-{% if ansible_memtotal_mb * 1048576 * non_hci_safety_factor / _num_osds > osd_memory_target %}
 {% set _osd_memory_target = (ansible_memtotal_mb * 1048576 * non_hci_safety_factor / _num_osds) | int %}
 {% endif %}
-{% endif %}
-osd memory target = {{ _osd_memory_target | default(osd_memory_target) }}
+osd memory target = {{ _osd_memory_target | default(osd_memory_target | default(0)) }}
```
dsavineau, line 92: If you use default(0) then you will have `osd memory target = 0` when there's no OSDs which breaks the backward compatibility.

dsavineau (changes requested):
> The current default behavior is unchanged by my proposal

That doesn't seem right because you removed the condition on the safety_fact value > osd_memory_target.

Running both code return different values.

Maybe I've missed something in the discussion but why not just set a custom `osd_memory_target` value via the {group,host}_vars ?
`
osd_memory_target: 4294967296` in ceph-defaults role in only present to avoid undefined variable but that doesn't mean this can't be overrided by users.

roles/ceph-config/templates/ceph.conf.j2
gaudenz: Right, will fix this.

roles/ceph-config/templates/ceph.conf.j2
gaudenz: Yes I can see how this is bad in this edge case. I fixed this by not including any `osd memory target` value into the file if neither `_osd_memory_target` and `osd_memory_target` are set.

gaudenz:
I tried to address both of @dsavineau s comments. Please review again.

guits (approved):

guits:

PR #5969, opened 2020-10-22
ceph-handler: fix curl ipv6 command with rgw

dsavineau:
When using the curl command with ipv6 address and brackets then we need
to use the -g option otherwise the command fails.

```console
$ curl http://[fdc2:328:750b:6983::6]:8080
curl: (3) [globbing] error: bad range specification after pos 9
```


dsavineau:

dsavineau:

dsavineau:

guits (approved):

PR #5975, opened 2020-10-24
Improve the checked way of the OSD created checking process.

wangxiaotong2020:
messages: The original method use "ceph -s" to check the OSD's status . In this way, it contains a lot of useless data . In order to exclude these invalid data , I replace the original method with "ceph osd stat" .

Signed-off-by: wangxiaotong <wangxiaotong@fiberhome.com>

### DIFF
--- a/roles/ceph-osd/tasks/main.yml
+++ b/roles/ceph-osd/tasks/main.yml
@@ -76,15 +76,15 @@
     - inventory_hostname == ansible_play_hosts_all | last
 
 - name: wait for all osd to be up
-  command: "{{ hostvars[groups[mon_group_name][0]]['container_exec_cmd'] | default('') }} ceph --cluster {{ cluster }} -s -f json"
+  command: "{{ hostvars[groups[mon_group_name][0]]['container_exec_cmd'] | default('') }} ceph --cluster {{ cluster }} osd stat -f json"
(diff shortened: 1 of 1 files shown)


dsavineau (changes requested):
Even if I approve this, I don't see the relation with https://tracker.ceph.com/issues/43430.

This issue (in fact we don't use the ceph tracker for tracking our issues) is related to users not using the correct ceph-ansible version against a ceph release due to the osdmap structure that changed in the the ceph status command after the nautilus release.

I think you should drop it from the commit message.

wangxiaotong2020:
> Even if I approve this, I don't see the relation with https://tracker.ceph.com/issues/43430.
> 
> This issue (in fact we don't use the ceph tracker for tracking our issues) is related to users not using the correct ceph-ansible version against a ceph release due to the osdmap structure that changed in the the ceph status command after the nautilus release.
> 
> I think you should drop it from the commit message.

Thanks for your advice and answering my questions (in my heart) about this issue.

I have drop it from the commit message according to your suggestion. But this is my first time to participate in community contirbutions and I don't know whether your requirements could be met. If not , please point out the problem and I will try my best to fix them.

Finally , thank you again very much.

dsavineau:
closing in flavor of #5978

PR #6012, opened 2020-11-09
Add ceph_client tag to execute or skip the playbook

fmount:
There are some use cases where there's a need to skip the execution
of the ceph-ansible client role even though the client section of the
inventory isn't empty.
This can happen in contexts where the services are colocated or when
a all-in-one deployment is performed.
The purpose of this change is adding a 'ceph_client' tag to avoid
altering the ceph-ansible execution flow but at the same time be able
to include or exclude a set of tasks using this tag.

### DIFF
--- a/site-container.yml.sample
+++ b/site-container.yml.sample
@@ -282,18 +282,19 @@
 
     - import_role:
         name: ceph-defaults
-      tags: ['ceph_update_config']
+      tags: ['ceph_update_config', 'ceph_client']
     - import_role:
         name: ceph-facts
-      tags: ['ceph_update_config']
+      tags: ['ceph_update_config', 'ceph_client']
     - import_role:
         name: ceph-handler
-      tags: ['ceph_update_config']
+      tags: ['ceph_update_config', 'ceph_client']
     - import_role:
         name: ceph-config
-      tags: ['ceph_update_config']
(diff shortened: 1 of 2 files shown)


fmount:
@fultonj @gfidente FYI

fmount:
@guits do you have concerns adding this tag?

guits:
@fmount I don't really get why you want to do this

fmount:
> @fmount I don't really get why you want to do this

My idea is using this new tag (and minimize the changes on OpenStack side) [1] to skip the client role execution.
I think it doesn't hurt and can avoid adding a lot of logic just to skip this playbook.
I see that this could be achieved providing an empty [clients] in the inventory, but for the Ceph integration customizing
the inventory is not an easy task: we're just able to create a mapping [2] to make the inventory ceph-ansible compatible.

[1] https://review.opendev.org/#/c/760915/8/deployment/ceph-ansible/ceph-base.yaml 
[2] https://github.com/openstack/tripleo-common/blob/master/tripleo_common/inventory.py#L320-L330

fmount:
In addition, during stack update (when there's no need to reconfigure clients), in a scale lab we want to add the ability (exposing a variable) to disable this role instead of blacklisting several nodes.

PR #6023, opened 2020-11-12
Add additional check for radosgw network validation

xelexin:
When user doesn't want to have a radosgw, this check with fail without
this option. We should check if user defined hosts in rgw group.

Signed-off-by: Rafał Wądołowski <rwadolowski@cloudferro.com>


guits:
as @dsavineau said, this is already covered by the condition `rgw_group_name in group_names`.

I'm closing this issue since this isn't needed. Feel free to re-open with more details if needed.

xelexin:
sure, yesterday I had problem with that, after i deleted group from inventory, I didn't check again... My mistake, sorry guys for bothering you!

PR #6041, opened 2020-11-17
containers: modify bindmount option (mon/mgr)

guits:
This commit changes the bind mount option for the mount point
/var/lib/ceph in the systemd template for mon and mgr containers. This
is needed in case of collocating mon/mgr with osds using dmcrypt
scenario.
Once mon/mgr got converted to containers, the dmcrypt layer sub mount is
still seen in /var/lib/ceph. For some reason it makes the
corresponding devices busy so any other container can't open/close it.
As a result, it prevents osds from starting properly.

Since it only happens on the nodes converted before the OSD play, the idea is
to bind mount /var/lib/ceph on mon and mgr with the rshared option
so once the sub mount is unmounted, it is propagated inside the
container so it doesn't see that mount point.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1896392


guits:
jenkins test playbook syntax

PR #6042, opened 2020-11-17
osd: ensure /var/lib/ceph/osd/{cluster}-{id} is present

guits:
This commit ensures that the `/var/lib/ceph/osd/{{ cluster }}-{{ osd_id }}` is
present before starting OSDs.

This is needed specificly when redeploying an OSD in case of OS upgrade
failure.
Since ceph data are still present on its devices then the node can be
redeployed, however those directories aren't present since they are
initially created by ceph-volume. We could recreate them manually but
for better user experience we can ask ceph-ansible to recreate them.

NOTE:
this only works for OSDs that were deployed with ceph-volume.
ceph-disk deployed OSDs would have to get those directories recreated manually.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1898486

### DIFF
--- a/roles/ceph-osd/tasks/start_osds.yml
+++ b/roles/ceph-osd/tasks/start_osds.yml
@@ -41,6 +41,15 @@
     - ceph_osd_systemd_overrides is defined
     - ansible_service_mgr == 'systemd'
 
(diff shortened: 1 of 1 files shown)


dsavineau (approved):

dsavineau:

Issue #6073, opened 2020-11-25
wait for all osd to be up task stuck on not healthy cluster


clwluvw:
**Bug Report**

What happened:
If there is an OSD that is down/destroyed because of failure and we are using ceph-ansible to add nodes `wait for all osd to be up` task will be stuck because `num_osds` won't equal to `num_up_osds` anytime!

What you expected to happen:
I think the task should only check the OSDs that is related to the target host by ansible.

How to reproduce it (minimal and precise):
1. Mark one OSD down
1. Try to add a new node to your cluster

guits:
why would you want to add a node with an OSD marked down? can't you fix this before adding your new node?
I would say most of the time, a day 2 operation requires your cluster is in a good state.

clwluvw:
@guits I think it has lots of scenarios for this situation, the ones I get into was I wanted to add new nodes to my SSD tier in an urgent but a minutes before one of my disks from HDD tier fails and I can't replace/recover the disk too fast!

guits:
@clwluvw we (@dsavineau and I) had a discussion, he has a good point:

since your HDD disk is failed anyway, why don't you remove it completely before adding your new nodes?

The proposed patch would prevent from doing multiple operations in one single run if you want to add new nodes + adding a new crush rule for instance.

clwluvw:
@guits @dsavineau I don't think making decisions for clients is good behavior here, I mean we can't preserve the client situation that for example, the client should replace the failed disk first and then add its new node! Maybe the client doesn't have extra disks and should buy a new one and it takes time (the new node disks are SSD it can't be replaced with HDD tier disks!) or the disks aren't hot-swappable and the client can't replace it in a meantime.
I was in urgent to add a new node to my SSD tier and there was no time to replace my failed disk! Also having multiple operations in two different tiers isn't very strange!
Also, I mention here again that it can have **too many other scenarios** that this task stuck the operations and I think the correct behavior for this task is to check only the OSDs in the target host!

guits:
@clwluvw that wouldn't require that you replace the failed disk by a new one.
It just requires that you remove it "logically" from the cluster, not "physically".
You can then add your new nodes this way.

> Also, I mention here again that it can have **too many other scenarios** that this task stuck the operations and I think the correct behavior for this task is to check only the OSDs in the target host!

That's incorrect, this task has to be run against *all* osd nodes, see original commit [1]

[1] 9d5265fe11fb5c1d0058525e8508aba80a396a6b

[...]

PR #6077, opened 2020-11-26
[skip ci] osd: add tag on 'wait for all osd to be up' task

guits:
This allows skipping this task if really desired.
Use it carefully. Use it at your own risk.

Fixes: #6073

### DIFF
--- a/roles/ceph-osd/tasks/main.yml
+++ b/roles/ceph-osd/tasks/main.yml
@@ -88,10 +88,12 @@
   when:
     - not ansible_check_mode
     - inventory_hostname == ansible_play_hosts_all | last
+  tags: wait_all_osds_up
 
 - name: include crush_rules.yml
   include_tasks: crush_rules.yml
   when: hostvars[groups[mon_group_name][0]]['crush_rule_config'] | default(crush_rule_config) | bool
+  tags: wait_all_osds_up
 
 # Create the pools listed in openstack_pools
 - name: include openstack_config.yml
@@ -101,3 +103,4 @@
     - not rolling_update | default(False) | bool
     - openstack_config | bool
     - inventory_hostname == groups[osd_group_name] | last
+  tags: wait_all_osds_up


guits:

clwluvw:
@guits Can't we have this in `stable-4.0`?

guits:
@dsavineau / @clwluvw couldn't it be a middle ground addressing both your issue/concern?

clwluvw:
> @dsavineau / @clwluvw couldn't it be a middle ground addressing both your issue/concern?

@guits What was the mistake from the previous one?

guits:
> > @dsavineau / @clwluvw couldn't it be a middle ground addressing both your issue/concern?
> 
> @guits What was the mistake from the previous one?

Using `--skip-tags` allows you to skip this task for your specific use case without enforcing a new default behavior.

clwluvw:
> > > @dsavineau / @clwluvw couldn't it be a middle ground addressing both your issue/concern?
> > 
> > 
> > @guits What was the mistake from the previous one?
> 
> Using `--skip-tags` allows you to skip this task for your specific use case without enforcing a new default behavior.

@guits I know but what was the mistake to skip this task when `add_osd`? For the crush_rule creation are you sure it is necessary that all the OSDs get up and then create it? I have tested creating the crush rule without any OSDs and it was successful.
For OpenStack pool creation, can't it just check the `num_up_osds` is upper than zero or move this task to ceph-client like other pool creation or move it to a new role?
I think making them together is just making complexity.

guits:
> > > > @dsavineau / @clwluvw couldn't it be a middle ground addressing both your issue/concern?
> > > 
> > > 
> > > @guits What was the mistake from the previous one?
> > 
> > 
> > Using `--skip-tags` allows you to skip this task for your specific use case without enforcing a new default behavior.
> 
> @guits I know but what was the mistake to skip this task when `add_osd`? 

what if you want to add a new OSD node and create and/or modify crush rules? That would require 2 runs when you could do it in only 1 run.

I'm unsure why you keep asking us to enforce this when we provide you an easy solution for your use case.

guits:
> For OpenStack pool creation, can't it just check the `num_up_osds` is upper than zero or move this task to ceph-client like other pool creation or move it to a new role?

ceph-client isn't the right place for this, that would require admin keyring which isn't something we want in ceph-client by default.
I like the idea of a new role. It could make sense but it won't be backported, neither in stable-5.0 nor stable-4.0.

clwluvw:
> > > > > @dsavineau / @clwluvw couldn't it be a middle ground addressing both your issue/concern?
> > > > 
> > > > 
> > > > @guits What was the mistake from the previous one?
> > > 
> > > 
> > > Using `--skip-tags` allows you to skip this task for your specific use case without enforcing a new default behavior.
> > 
> > 
> > @guits I know but what was the mistake to skip this task when `add_osd`?
> 
> what if you want to add a new OSD node and create and/or modify crush rules? That would require 2 runs when you could do it in only 1 run.
> 
> I'm unsure why you keep asking us to enforce this when we provide you an easy solution for your use case.

I thought it's not just my problem and anyone else can get in this trouble so I just try to find a solution that keeps it general. I think being smarter and less complex can make more value than the runs count :man_shrugging:

Issue #6102, opened 2020-12-02
Assumptions about consistent container runtime in a Ceph cluster running mixed OS versions


oneswig:
**Bug Report**

I am working with a containerised deployment that is migrating from CentOS 7 (and Docker) to CentOS 8 (and Podman).  It does not appear possible to deploy MGRs or MDSs on the new CentOS 8 image due to an assumption that all nodes have the same container runtime.  This is causing problems when forming delegated commands to run on other cluster hosts.

In the role `ceph-mgr`, task `get keys from monitors`, I see errors of the form:

```
failed: [host6 -> host5] (item={u'path': u'/var/lib/ceph/mgr/ceph-host6/keyring', u'name': u'mgr.host6', u'copy_key': True}) => changed=false  
  ansible_loop_var: item 
  cmd: podman exec ceph-mon-host5 ceph --cluster ceph auth get mgr.host6 
  item: 
    copy_key: true 
    name: mgr.host6 
    path: /var/lib/ceph/mgr/ceph-host6/keyring 
  msg: '[Errno 2] No such file or directory' 
  rc: 2 
```

The issue being that `host5` is running Docker and `host6` is running Podman.

How to reproduce it (minimal and precise):

It should be relatively easy to reproduce by adding a new host running CentOS8/Podman to a cluster running CentOS7/Docker.

**Environment**:
* OS is being upgraded from `CentOS 7.8.2003` to `CentOS 8.2.2004`
* Docker RPM is `docker-1.13.1-203`, Podman RPM is `podman-1.6.4-10`
* Ansible version is `2.9.15`
* ceph-ansible version is stable-4.0
* Ceph version is Nautilus 14.2.15

dsavineau:
Hi Stig,

Unfortunatly that's a known issue as we assume the entire cluster is running the same OS.

The main issue is that we are delegating some commands by using the `container_binary` global variable instead of the scoped one to the delegated node (`container_binary` vs `hostvars[xxx]['container_binary']`). I think there's a lot of places in the roles where this need to be changed.

As a temp workaround, did you try to add a symlink podman->docker on host5 ?

PR #6103, opened 2020-12-02
Bugfix #6102: Assumptions about consistent container runtime in a Ceph cluster running mixed OS versions

oneswig:
Use hostvars from the delegated host to determine container binary.

### DIFF
--- a/roles/ceph-mds/tasks/common.yml
+++ b/roles/ceph-mds/tasks/common.yml
@@ -11,7 +11,7 @@
     - /var/lib/ceph/mds/{{ cluster }}-{{ mds_name }}
 
 - name: get keys from monitors
-  command: "{{ container_exec_cmd | default('') }} ceph --cluster {{ cluster }} auth get {{ item.name }}"
+  command: "{{ hostvars[groups.get(mon_group_name)[0]]['container_exec_cmd'] | default('') }} ceph --cluster {{ cluster }} auth get {{ item.name }}"
   register: _mds_keys
   with_items:
     - { name: "client.bootstrap-mds", path: "/var/lib/ceph/bootstrap-mds/{{ cluster }}.keyring", copy_key: true }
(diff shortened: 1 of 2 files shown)


dsavineau:
Could you open the PR agains the master branch and add a signoff ?

We don't allow pushing changes on stable branches directly (unless it's the only solution) so you should create the PR on master and we will create the backport to the stable branches.

oneswig:
Thanks Dimitri!  I looked at master branch and the approach there has been refactored so applying the patch is not trivial.  The issue does appear to be still there (unless it is resolved at a different level).  Unfortunately I cannot test against master branch.

guits:
If the issue reported here boils down to an issue because of different OS versions, I would say we have never supported deployments with a mix of different host OS versions.
That being said, from what I'm understanding here, the underlying question is about supporting to deploy with a mix of different container runtimes, regardless of the OS version.

@dsavineau why should we support this kind of deployment? I'm not 100% sure so I might be wrong but off the top of my head, this has never really been supported nor tested in the CI. (On the other hand, I'm thinking about deployments which would be collocated with nodes that already come with a container runtime installed or something like that)

A couple of questions:
- Is this patch enough to fully support this? Is it broken as well in stable-5.0 and master?
- If we decide to support this, it means we should add functional testings

oneswig:
The situation where it arises for me is in rolling upgrades from 7 to 8.  During that process, the cluster will be partly running 7+Docker, and partly 8+Podman

dsavineau:
> @dsavineau why should we support this kind of deployment? I'm not 100% sure so I might be wrong but off the top of my head, this has never really been supported nor tested in the CI. (On the other hand, I'm thinking about deployments which would be collocated with nodes that already come with a container runtime installed or something like that)

That's true, we never supported this kind of setup but in the same time we've never mentioned anywhere that it wasn't supported.

> * Is this patch enough to fully support this?

It's only a partial fix to me (especially for Stig's environment) because there's a lot of other part of the code where we use `delegate_to` + unscoped `container_binary/container_exec_cmd/etc variables`.

> Is it broken as well in stable-5.0 and master?

Yes. Note that it should be easier to fix this in master/stable-5.0 as we use more standalone container commands (docker/podman run) via the ansible module vs the docker/podman exec workflow in stable-4.0.

guits:
> That's true, we never supported this kind of setup but in the same time we've never mentioned anywhere that it wasn't supported.

That doesn't mean that if we don't explicitly say something isn't supported one can state it is.
To be honest, Although I don't mind introducing patches for that I'm not sure we should start supporting this given the future of ceph-ansible, it all depends on how much effort it represents for adding this kind of support (thinking about ci testing, etc.)

PR #6111, opened 2020-12-07
ceph-facts: fix grafana group conversion

dsavineau:
The conversion fact task was only executed when the grafana_server_group_name
variable was explicitly set in the user configuration. If an user was using
the default value then the conversion wasn't executed.

This also adds back the default `grafana_server_group_name` value in case user
was using the default value and to avoid undefined variable error.

Instead of hardcoding the "monitoring" group name then we can reuse the
monitoring_group_name variable.

There's no need to override the `monitoring_group_name` variable, it's either
using the default value or the one defined by the user.

Finally removing the delegate_to statement on the add_host task since it's
always executed on the ansible controller.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1903732

### DIFF
--- a/roles/ceph-facts/tasks/convert_grafana_server_group_name.yml
+++ b/roles/ceph-facts/tasks/convert_grafana_server_group_name.yml
@@ -2,15 +2,8 @@
 - name: convert grafana-server group name if exist
   add_host:
     name: "{{ item }}"
-    groups: "monitoring"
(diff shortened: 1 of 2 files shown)


dsavineau:

dsavineau:

dsavineau:

dsavineau:

guits (approved):

PR #6112, opened 2020-12-07
tests: update openssl package for containerized

dsavineau:
This is a temporary workaround since CentOS 8.3 has been released (the
content of the repository have been rebuild to 8.3) but we're still using
CentOS 8.2 because there's no newer image available.

At the moment, this causes the podman command to fails

```console
stderr: '/usr/libexec/platform-python: error while loading shared libraries:
libcrypto.so.1.1: cannot change memory protections'
```

### DIFF
--- a/tests/functional/setup.yml
+++ b/tests/functional/setup.yml
@@ -75,6 +75,16 @@
         resizefs: yes
       when: is_atomic | bool
 
+    - name: update openssl for CentOS 8.3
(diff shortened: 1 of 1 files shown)


dsavineau:

dsavineau:
Closing as it doesn't seems to come from openssl but the CentOS release coming from the Vagrant image

PR #6114, opened 2020-12-08
ceph-config: Fix RadosGW remote address with HAproxy

BenoitKnecht:
When RadosGW is behind a reverse proxy, it should get the client address from
the `X-Forwarded-For` header, so that S3 policies that refer to `aws:SourceIp`
work as expected.

This commit sets `rgw remote addr param = HTTP_X_FORWARDED_FOR` when the
`rgwloadbalancers` group is not empty.

### DIFF
--- a/roles/ceph-config/templates/ceph.conf.j2
+++ b/roles/ceph-config/templates/ceph.conf.j2
@@ -121,6 +121,9 @@ rgw_realm = {{ instance['rgw_realm'] }}
 rgw_zonegroup = {{ instance['rgw_zonegroup'] }}
 rgw_zone = {{ instance['rgw_zone'] }}
 {% endif %}
+{% if groups.get(rgwloadbalancer_group_name) %}
+rgw remote addr param = HTTP_X_FORWARDED_FOR
+{% endif %}
 {% endfor %}
 {% endif %}
 {% endif %}


guits:
hi @BenoitKnecht see https://github.com/ceph/ceph-ansible/blob/master/docs/source/index.rst#cephconf-configuration-file

> We will no longer accept pull requests that modify the ceph.conf template unless it helps the deployment. For simple configuration tweaks please use the ceph_conf_overrides variable.

BenoitKnecht:
I would argue that it does "help the deployment" in this case. It's not really a configuration tweak, the RadosGW setup ends up not working as expected if `rgw remote addr param` isn't set in this case.

BenoitKnecht:
@guits Any thoughts on my comment above?

I understand the reluctance to add options to `ceph.conf.j2`, as it would lead to an endless list of parameters that could just as well be set with `ceph_conf_overrides`.

But to reiterate, in this case the `rgw remote addr param` option is _required_ in case you deploy RadosGW with HAproxy in front of it, otherwise you get bucket policies that don't work as intended (with potential security implications). So to me it doesn't make sense to let users deploy a broken configuration by default, and rely on them to manually set `ceph_conf_overrides` where needed.

This PR adds very little complexity to `ceph.conf.j2` and results in a properly configured RadosGW whether it's behind HAproxy or not, so I would urge you to reconsider merging it.

PR #6117, opened 2020-12-09
[skip ci] Revert "config: Always use osd_memory_target if set"

dsavineau:
This reverts commit 4d1fdd2b05d55f8028fb5593d41fa61dbddd7095.

This breaks the backward compatibility with previous osd_memory_target
calculation and we could have a value lower than the minimum value allowed
(896M) which causes some ceph commands to fail (like ceph assimilate-conf).

### DIFF
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -342,12 +342,10 @@ dummy:
 
 ## OSD options
 #
-# Variables to calculate a suitable osd_memory_target based on the number of OSDs
 #is_hci: false
 #hci_safety_factor: 0.2
 #non_hci_safety_factor: 0.7
-# You can also set an explicit value which will override the calculation if set.
-# osd_memory_target: 4294967296
+#osd_memory_target: 4294967296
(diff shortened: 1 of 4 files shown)


guits:
@gaudenz FYI

why not using `ceph_conf_override` instead?

guits (approved):

gaudenz:
Using `ceph_conf_override` works, but you have to do it like this:
```
ceph_conf_override:
  osd:
    'osd memory target': '{{ osd_memory_target }}'
```
Writing the key in quotes and with spaces is important to really override the value.  If using `osd_memory_target` (with underscores) you will end up with the same setting twice in the config file with possibly different values.

dsavineau:
> Writing the key in quotes and with spaces is important to really override the value. If using `osd_memory_target` (with underscores) you will end up with the same setting twice in the config file with possibly different values.

You don't need quotes on the key, it works without.

```yaml
osd_memory_target: 12345
ceph_conf_override:
  osd:
    osd memory target: '{{ osd_memory_target }}'
```

Issue #6137, opened 2020-12-15
[ceph-dashboard : add iscsi gateways - ipv4 fails when `api_secure: true` due to ip being used instead of fqdn


Venator-Fox:
**Bug Report**

What happened: `TASK [ceph-dashboard : add iscsi gateways - ipv4 fails` when `api_secure: true` due to the task running `podman exec <someHost> ceph --cluster ceph dashboard iscsi-gateway-add https://<ip>` instead of `https://<fqdn>`. I probably am not understanding the task correctly, but expecting the IP to be in the subject alternative names in the certificate seems a bit strange? A PR is submitted here to change it to use the fqdn from ip, which allowed the play to run successfully in my environment but I'm not sure if it will break other things (or if I'm configuring incorrectly in the first place).

I'm not sure if #5165 is related

What you expected to happen: The task to complete successfully

How to reproduce it (minimal and precise):

Set `api_secure: true`, add tls certificates, and add hosts to `[iscsigws]` in inventory

Share your group_vars files, inventory and **full** ceph-ansible log
~~~
TASK [ceph-dashboard : add iscsi gateways - ipv4] **********************************************************************Monday 14 December 2020  12:20:42 -0700 (0:00:00.128)       0:42:37.637 *******
failed: [redacted-ceph-node-000 -> redacted-ceph-node-000] (item=redacted-ceph-node-000) => changed=false
  ansible_loop_var: item
  cmd:
  - podman
  - exec
  - ceph-mon-redacted-ceph-node-000
  - ceph
  - --cluster
  - ceph
  - dashboard
  - iscsi-gateway-add
  - https://admin:<redacted>@10.224.71.11:5000
  delta: '0:00:01.093222'
  end: '2020-12-14 12:20:44.063241'
  item: redacted-ceph-node-000
  msg: non-zero return code
  rc: 22
  start: '2020-12-14 12:20:42.970019'
  stderr: |-
    Error EINVAL: Traceback (most recent call last):
      File "/lib/python3.6/site-packages/urllib3/connectionpool.py", line 600, in urlopen
        chunked=chunked)
      File "/lib/python3.6/site-packages/urllib3/connectionpool.py", line 343, in _make_request
        self._validate_conn(conn)
      File "/lib/python3.6/site-packages/urllib3/connectionpool.py", line 839, in _validate_conn
        conn.connect()
      File "/lib/python3.6/site-packages/urllib3/connection.py", line 364, in connect
[...]

PR #6138, opened 2020-12-15
dashboard: connect with fqdn instead of first ip

Venator-Fox:
Regarding ssl failures due to the api being accessed by an ip and not the fqdn, fixes https://github.com/ceph/ceph-ansible/issues/6137

Signed-off-by: Adam W Zheng <adam.zheng@colorado.edu>

### DIFF
--- a/roles/ceph-dashboard/tasks/configure_dashboard.yml
+++ b/roles/ceph-dashboard/tasks/configure_dashboard.yml
@@ -266,14 +266,14 @@
         - generate_crt | default(false) | bool
 
     - name: add iscsi gateways - ipv4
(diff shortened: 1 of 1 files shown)


dsavineau (changes requested):
if you change `ansible_all_ipv{4,6}_addresses` by `ansible_fqdn`  then both `add iscsi gateways` tasks are identical so you can remove one of them (and the associated `when` condition)

The problem is that there's no guarantee that the `ansible_fqdn` is configured with an IP address on the public_network since the mgr dashboard is binding on that network.

Venator-Fox:
Hi @dsavineau, may I have a bit more time on this?
Had a question on #6137 to see if that would work.

PR #6187, opened 2021-01-12
purge-cluster: use parted ansible module

dsavineau:
Instead of doing some scripting via the shell module, we can use the
parted ansible module to check the boot flag on partitions.


dsavineau:

guits (approved):

PR #6205, opened 2021-01-21
cephadm-adopt: use ceph_osd_flag module

dsavineau:
There's no reason to not use the ceph_osd_flag module to set/unset osd
flags.
Also if there's no OSD nodes in the inventory then we don't need to
execute the set/unset play.


dsavineau:

guits (approved):

PR #6212, opened 2021-01-25
rgw/multisite: add endpoints key to zones list

dsavineau:
This removes the zone_endpoints_list fact since we can define it as the
zones fact to avoid duplicate facts and two calls to radosgw_zone module
(one with access/secret key and another one with the endpoints).
Now the endpoints key is merged into the zones definition so we can remove
the dedicated task which adds the endpoints to the zone.

### DIFF
--- a/roles/ceph-rgw/tasks/multisite/main.yml
+++ b/roles/ceph-rgw/tasks/multisite/main.yml
@@ -13,21 +13,15 @@
   loop: "{{ rgw_instances_all }}"
   when: item.rgw_zonegroupmaster | default(hostvars[item.host]['rgw_zonemaster']) | bool
 
-- name: create list zones
-  set_fact:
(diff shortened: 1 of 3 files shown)


dsavineau:

dsavineau:

dsavineau:
@alimaredia it is unclear to me why we are adding the zone endpoints as the zonegroup endpoints to.

If we have a mapping 1:1 between the zonegroup and zone then it seems to be fine, but if we have multiple zones in the same zonegroup then we might have an issue.

What's the meaning of the zonegroup endpoints ?

dsavineau:

dsavineau:

alimaredia:
@dsavineau Per Casey: buckets are owned by a zonegroup. if you send a request on a bucket that's owned by a different zonegroup, we forward the request to that zonegroups endpoints. 

Also it is possible to have multiple zones in the same zone group (even on the same cluster). The requirements for zonegroup endpoints is that endpoints of the master zone should be the endpoints for a zone group.

alimaredia:
@guits should I close this and make sure the upstream documentation is updated.

guits:
@alimaredia I didn't follow to be honest, I would trust your judgement on this one.

guits:
@alimaredia  :slightly_smiling_face:

PR #6219, opened 2021-01-27
[skip ci] docs: rgw multisite docs with new rgw_instances config

alimaredia:
Docs reflect that each instance of `rgw_instances`
can now take rgw_zonemaster, rgw_zonesecondary,
rgw_zonegroupmaster, rgw_multisite_proto.

### DIFF
--- a/README-MULTISITE.md
+++ b/README-MULTISITE.md
@@ -449,11 +449,15 @@ Here is an example:
 ```yaml
 rgw_instances:
   - instance_name: rgw1
+    rgw_zonemaster: true
+    rgw_zonesecondary: false
+    rgw_zonegroupmaster: true
     rgw_realm: usa
     rgw_zonegroup: alaska
     rgw_zone: juneau
     radosgw_address: "{{ _radosgw_address }}"
     radosgw_frontend_port: 8080
+    rgw_multisite_proto: http
     rgw_zone_user: edward.lewis
     rgw_zone_user_display_name: "Edward Lewis"
     system_access_key: yu17wkvAx3B8Wyn08XoF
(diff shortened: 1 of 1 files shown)


README-MULTISITE.md
```diff
@@ -494,7 +502,7 @@ rgw_instances:
 
 This example starts up 2 rgws on host rgw-001. `rgw1` is configured to be in realm usa and `rgw2` is configured to be in realm france.
 
-The variables `rgw_zonemaster`, `rgw_zonesecondary`, `rgw_zonegroupmaster`, `rgw_multisite_proto` cannot be set in an item of rgw_instances. All of these variables must be set in group_vars/ or host_vars/.
+**Note:** The old format of declaring `rgw_zonemaster`, `rgw_zonesecondary`, `rgw_zonegroupmaster`, `rgw_multisite_proto` outside of rgw_instances still works but declaring the values at the instance level (as seen above) is preferred.
```
guits, line 505:
```suggestion
**Note:** The old format of declaring `rgw_zonemaster`, `rgw_zonesecondary`, `rgw_zonegroupmaster`, `rgw_multisite_proto` outside of `rgw_instances` still works but declaring the values at the instance level (as seen above) is preferred.
```

README-MULTISITE.md
```diff
@@ -505,28 +513,32 @@ The value of `endpoint` should be the endpoint of an RGW in the master zone of t
...
 ```yaml
-rgw_zonemaster: false
-rgw_zonesecondary: true
-rgw_zonegroupmaster: true
-rgw_multisite_proto: "http"
 rgw_instances:
   - instance_name: rgw3
+    rgw_zonemaster: false
+    rgw_zonesecondary: true
+    rgw_zonegroupmaster: true
```
guits, line 520: @alimaredia I'm still very confused with the concept of `zonegroupmaster`.
In the CI, we have `rgw_zonegroumaster: false` when it's a secondary zone see https://github.com/ceph/ceph-ansible/blob/master/tests/functional/rgw-multisite/host_vars/osd0#L17-L19

README-MULTISITE.md
```diff
@@ -536,8 +548,9 @@ rgw_instances:
 
 This example starts up 2 rgws on the host that will pull the realm from the rgws on rgw-001 above. `rgw3` is pulling from the rgw endpoint in realm usa in the master zone example above (instance name rgw1). `rgw4` is pulling from the rgw endpoint in realm france in the master zone example above (instance name rgw2).
 
-Just like the example on the master zone, the variables `rgw_zonemaster`, `rgw_zonesecondary`, `rgw_zonegroupmaster`, `rgw_multisite_proto` cannot be set in an item of rgw_instances. All of these variables must be set in group_vars/ or host_vars/.
+**Note:** The old format of declaring `rgw_zonemaster`, `rgw_zonesecondary`, `rgw_zonegroupmaster`, `rgw_multisite_proto` outside of rgw_instances still works but declaring the values at the instance level (as seen above) is preferred.
```
guits, line 551:
```suggestion
**Note:** The old format of declaring `rgw_zonemaster`, `rgw_zonesecondary`, `rgw_zonegroupmaster`, `rgw_multisite_proto` outside of `rgw_instances` still works but declaring the values at the instance level (as seen above) is preferred.
```

guits (changes requested):

README-MULTISITE.md
alimaredia: @guits the value of `rgw_zonegroupmaster` doesn't matter for secondary zones because when secondary zones are created the zonegroup they are in already exists, thus we don't need to specify whether or not that zonegroup is master. I think you can leave that as it is.

guits: so this parameter shouldn't be set at all for secondary zones it means

alimaredia: It could be removed or it could be kept, there's really no way the value could be configured that is `wrong`. Maybe we should add that into the docs?

guits (approved):

PR #6247, opened 2021-02-05
Add option to configure container --memory-swappiness for OSDs

fultonj:
Add new variable ceph_osd_docker_memory_swappiness which is
undefined by default. If ceph_osd_docker_memory_swappiness
is defined, OSD service is started with --memory-swappiness
set to the value in this variable.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1925555


dsavineau:
No need to add a new variable to achieve this since we can do it via the existing `ceph_osd_docker_extra_env` variable.

```yaml
ceph_osd_docker_extra_env: "--memory-swappiness=20"
```

fultonj:
ack

PR #6248, opened 2021-02-05
Set idmap config for Ceph-NFS

tombarron:
Currently NFS Ganesha (ceph-nfs) consumes /etc/idmapd.conf, which
controls mapping of user/owner identities under NFSv4+.  With
containerized service deployment, this file is an immutable part of the
container image and cannot be modified.

Here we provide group variables, and a taskk and templates for the
ceph-nfs role, to set the path of the idmap configuration file and
to make the most common adjustment to the contents of that file --
namely to set the 'Domain'.  We default the path to /etc/ganesha/idmap.conf
so that we will not conflict with /etc/idmapd.conf on the controller nodes
where ganesha runs.  NFSv4 clients, as used for example by the Cinder NFS
driver, consume /etc/idmapd.conf and may require different settings than
what is wanted for NFS Ganesha.  Additionally, because we already bind
/etc/ganesha from the host into the ceph-nfs container, the file NFS
Ganesha consumes will no longer be an immutable part of the container.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1925646

Signed-off-by: Tom Barron <tpb@dyncloud.net>

### DIFF
--- a/group_vars/nfss.yml.sample
+++ b/group_vars/nfss.yml.sample
@@ -57,6 +57,11 @@ dummy:
 # when setting 'nfs_file_gw' option as true.
 #ceph_nfs_disable_caching: false
 
+# This is the file ganesha will use to control NFSv4 ID mapping
+#ceph_nfs_idmap_conf: "/etc/ganesha/idmap.conf"
+# ``domain`` setting in the idmap.conf file.  Default is host's DNS domain.
+#ceph_nfs_idmap_domain: ""
+
 ####################
 # FSAL Ceph Config #
 ####################
--- a/roles/ceph-nfs/defaults/main.yml
+++ b/roles/ceph-nfs/defaults/main.yml
@@ -49,6 +49,11 @@ ceph_nfs_rados_export_index: "ganesha-export-index"
 # when setting 'nfs_file_gw' option as true.
 ceph_nfs_disable_caching: false
 
+# This is the file ganesha will use to control NFSv4 ID mapping
+ceph_nfs_idmap_conf: "/etc/ganesha/idmap.conf"
+# ``domain`` setting in the idmap.conf file.  Default is host's DNS domain.
+ceph_nfs_idmap_domain: ""
+
 ####################
 # FSAL Ceph Config #
 ####################
--- a/roles/ceph-nfs/tasks/start_nfs.yml
+++ b/roles/ceph-nfs/tasks/start_nfs.yml
@@ -47,6 +47,15 @@
     mode: "0644"
   notify: restart ceph nfss
 
+- name: generate ganesha idmap.conf file
+  template:
+    src: "idmap.conf.j2"
+    dest: "{{ ceph_nfs_idmap_conf }}"
+    owner: "root"
+    group: "root"
+    mode: "0644"
+  notify: restart ceph nfss
+
 - name: create exports directory
   file:
     path: /etc/ganesha/export.d
--- a/roles/ceph-nfs/templates/ganesha.conf.j2
+++ b/roles/ceph-nfs/templates/ganesha.conf.j2
@@ -34,7 +34,9 @@ RADOS_URLS {
 
 NFSv4 {
 	RecoveryBackend = 'rados_kv';
+        IdmapConf = "{{ ceph_nfs_idmap_conf }}";
 }
+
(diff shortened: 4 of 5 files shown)


dsavineau:
As ceph-ansible is slowly disappearing then we will need a bugzilla in order to priorize and triage this.

Any downstream target for this ? AFAIK this could be only usefull for RHCS 4 (nautilus) because RHCS 5 uses cephadm so this would be targeted as an RFE for RHCS 4.3

I'm not a ganesha expert but instead of using a dedicated idmap file for ganesha (with only the domain part set) why don't you use the `DomainName` [1] option in the NFSv4 section ?

[1] https://github.com/nfs-ganesha/nfs-ganesha/wiki/NFS4Params

tombarron:
> As ceph-ansible is slowly disappearing then we will need a bugzilla in order to priorize and triage this.
> 
> Any downstream target for this ? AFAIK this could be only usefull for RHCS 4 (nautilus) because RHCS 5 uses cephadm so this would be targeted as an RFE for RHCS 4.3
> 
> I'm not a ganesha expert but instead of using a dedicated idmap file for ganesha (with only the domain part set) why don't you use the `DomainName` [1] option in the NFSv4 section ?
> 
> [1] https://github.com/nfs-ganesha/nfs-ganesha/wiki/NFS4Params

Downstream we'd target OSP16.z, so corresponding to  RHCS4.  (We'd want to integrate it with the rest of the CephNFS functionality provided currently by ceph-ansible in the solution for CephNFS  with cephadm for OSP17 and RHCS5.)

The OSP bug driving this patch is https://bugzilla.redhat.com/show_bug.cgi?id=1917356 Should I clone it for ceph-ansible as well?

With regard to using the NFSv4 section of ganesha.conf instead, we may well have other parameters from idmap.conf than Domainname that need to be set in the future that are not also exposed in that file.

dsavineau:
> The OSP bug driving this patch is https://bugzilla.redhat.com/show_bug.cgi?id=1917356 Should I clone it for ceph-ansible as well?

Yes please

> With regard to using the NFSv4 section of ganesha.conf instead, we may well have other parameters from idmap.conf than Domainname that need to be set in the future that are not also exposed in that file.

Ack.
In that case we will probably need to have a smarter solution instead of adding new variable each time. So one solution would be to use the `config_template` plugin [1] and an override variable like we're doing with the ceph config [2] and `ceph_conf_override` [3] (as it looks like the idmap config file is in INI format).
With this kind of solution, one would be able to change anything in the idmap configuration without modifying the ceph-ansible code.
We're also doing this in the ganesha configuration file (`ganesha_conf_override` [4]) but differently since the format isn't standard.

Maybe something to consider is the backward compatibility (if any) if one was relying on the default idmap configuration file (/etc/idmapd.conf) with something custom settings as the PR force to use the new file.

[1] https://github.com/ceph/ceph-ansible/blob/master/plugins/actions/config_template.py
[2] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/tasks/main.yml#L107-L115
[3] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/defaults/main.yml#L504-L526
[4] https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-nfs/defaults/main.yml#L79-L100

tombarron:
@dsavineau Thanks for the quick helpful review!
1) I cloned ceph-ansible bug https://bugzilla.redhat.com/show_bug.cgi?id=1925646
2) Thanks for the pointers to the config_template stuff, which does sound like a more general solution.
3) We're OK for OSP consumers of ganesha because the version of /etc/idmapd.conf they see is baked into a container so they could not have changed its parameter values from the default.  If we need to worry about other consumers (RGW?) then I can set the default value of the idmap file to /etc/idmapd.conf and override it from THT when OSP is using ceph-nfs.  Let me know what seems best, please.

guits:
hey @tombarron :slightly_smiling_face: 
do you mind adding the `Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1925646` in your commit message please?
also, could you rebase on master?
Thanks!

guits:
@tombarron any news?

guits:
Closing in favor of #6338

PR #6272, opened 2021-02-11
[skip ci] switch2container: do not serialize the ceph-crash migration

guits:
There's no need to slow down the playbook execution time by migrating
all the `ceph-crash` instances in a serial way. Let's remove the
`serial: 1` so the migration is achieved in a parallel way.


guits:

PR #6295, opened 2021-02-19
library: ceph_crush: set containerized param to false to respect docstrings

k0ste:
doc:
```yaml
    containerized:
        description:
            - Weither or not this is a containerized cluster. The value is
            assigned or not depending on how the playbook runs.
        required: false
        default: None
```
fixes:
```python
failed: [mon1.example.com] (item=Set bucket placement for: osd1.example.com) => {"ansible_loop_var": "item", "changed": false, "item": {"host": "osd1.example.com", "rack": "rack1"}, "msg": "missing required arguments: containerized"}
```

Also delete couple of rudiments in `main()`

### DIFF
--- a/library/ceph_crush.py
+++ b/library/ceph_crush.py
@@ -139,16 +139,14 @@ def exec_commands(module, cmd_list):
     return rc, cmd, out, err
 
 
-def run_module():
-    module_args = dict(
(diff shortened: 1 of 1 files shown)


guits:
in fact, it could be nice to rework this module so we get rid of the `containerized` parameter (like it has been done for almost all other module)

PR #6064, opened 2020-11-24
mon: refact initial keyring generation

guits:
adding monitor is no longer possible because we generate a new mon
keyring each time the playbook is run.
    
Fixes: #5864

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #5946, opened 2020-10-19
Fix Ansible check mode in ceph-mon and ceph-osd

BenoitKnecht:
Skip tasks that depend on the output of commands that are not being run in check mode.

PR #6301, opened 2021-03-01
ceph-mon: Fix check mode for deploy monitor tasks

BenoitKnecht:
Skip the `get initial keyring when it already exists` task when both commands
whose `stdout` output it requires have been skipped (e.g. when running in check
mode).

This was previously fixed in #5946, but got lost in the refactoring from #6064.

The fix is also slightly different from the original, because the `ceph_key` module is now being used instead of `command`, and that task is not skipped in check mode, it just returns an empty `stdout`.


guits:

guits:

guits:

guits (approved):

BenoitKnecht:
@guits It would be great if this could be backported to the stable branches.

BenoitKnecht:
@guits How about backporting this? I assume you'd just need to add the proper labels to this PR, but if it's not possible anymore after the PR is closed, I can submit new PRs for the stable branches. Let me know.

PR #6319, opened 2021-03-09
[skip ci] rgw-loadbalancer: update haproxy.cfg.j2

kalebskeithley:
haproxy gets an AVC when configured to connect to port 8081

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1923890

Signed-off-by: Kaleb S. KEITHLEY <kkeithle@redhat.com>


roles/ceph-rgw-loadbalancer/templates/haproxy.cfg.j2
```diff
@@ -40,6 +40,14 @@ frontend rgw-frontend
...
     default_backend rgw-backend
 
+# when running in an selinux environment, selinux restricts the ports that haproxy can
+# connect to to:
+#  * 80, 81, 443, 488, 8008, 8009, 8443, 9000 (http_port_t) and,
+#  * 8080, 8118, 8123, 10001-10010 (http_cache_port_t)
+#
+# Practically speaking, it would be preferable (and perhaps easier) to configure the
+# rgw daemons to listen on ports 10001-10010 and configure haproxy here to match.
+#
```
mattbenjamin, line 50: could we add a snippet of selinux policy to hypothetically do the thing for an imaginary port  number?

mattbenjamin (dismissed):
that seems ideal

kalebskeithley:
jenkins test playbook syntax

dsavineau (dismissed):
> haproxy gets an AVC when configured to connect to port 8081

IMO we should either
  - set `haproxy_connect_any` to true via the `seboolean` ansible module [1]
  - define the selinux network port type for all RGW instances via the `seport` ansible module [2]

But this has to be done via ansible in the ceph-rgw-loadbalancer role.
Putting this as a pseudo documentation in a template doesn't really make any sense.

[1] https://docs.ansible.com/ansible/2.9/modules/seboolean_module.html
[2] https://docs.ansible.com/ansible/2.9/modules/seport_module.html

> jenkins test playbook syntax

You should amend your commit and add a signoff

mattbenjamin:
> 
> 
> > haproxy gets an AVC when configured to connect to port 8081
> 
> IMO we should either
> 
>     * set `haproxy_connect_any` to true via the `seboolean` ansible module [1]

this is a security problem--that can't be valid

> 
>     * define the selinux network port type for all RGW instances via the `seport` ansible module [2]

what effect does this have?

> 
> 
> But this has to be done via ansible in the ceph-rgw-loadbalancer role.
> Putting this as a pseudo documentation in a template doesn't really make any sense.

I don't agree, but the then, I suggested it.  We are trying to educate the user writing the playbook--if they had chosen 1 rgw daemon per host, it there would be no avc

Matt

> 
> [1] https://docs.ansible.com/ansible/2.9/modules/seboolean_module.html
> [2] https://docs.ansible.com/ansible/2.9/modules/seport_module.html
> 
> > jenkins test playbook syntax
> 
> You should amend your commit and add a signoff

mattbenjamin:
@dsavineau Dimitri, could you please reach out to us (me, @kalebskeithley ) to help us understand how your proposed approaches help customers solve this issue, while preserving the utility of selinux enforcement?  We need to make a concrete change to address the issue, and we don't know what it is.

guits:
> > IMO we should either
> > ```
> > * set `haproxy_connect_any` to true via the `seboolean` ansible module [1]
> > ```
> 
> this is a security problem--that can't be valid

@mattbenjamin if I understand correctly, for security reasons, this option is a no-go.

> > ```
> > * define the selinux network port type for all RGW instances via the `seport` ansible module [2]
> > ```
> 
> what effect does this have?


@mattbenjamin I think this is the ansible way to apply what @kalebskeithley suggests [here](https://github.com/ceph/ceph-ansible/pull/6319/files#diff-1ea21f2851186f2a01ff25e715ed670b9d96629c6b7bc385aefd9e4154204bdeR51-R53) @dsavineau am I correct?
If yes, I understand the issue with that is that the range of ports `8081-8084` is already reserved and can't be used by haproxy.

To sum it up, from an SElinux point of view, having `8080` as the default port for [`radosgw_frontend_port`](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-defaults/defaults/main.yml#L376) is not a good choice.

Ideally, it would be nice to switch the default value to 10001 or whatever isn't already reserved but there's an issue with existing deployments, it's a breaking change.

@dsavineau any thoughts?

kalebskeithley:
> I understand the issue with that is that the range of ports 8081-8084 is already reserved and can't be used by haproxy.

They are reserved in the sense that selinux has policy that allows other applications to use those ports without throwing an AVC.

My attempts to add/mod policy to also allow rgw to use those ports too met with errors. Perhaps the selinux devs can suggest a way around that?

guits:
> > I understand the issue with that is that the range of ports 8081-8084 is already reserved and can't be used by haproxy.
> 
> They are reserved in the sense that selinux has policy that allows other applications to use those ports without throwing an AVC.
> 
> My attempts to add/mod policy to also allow rgw to use those ports too met with errors. Perhaps the selinux devs can suggest a way around that?

I'm not a selinux expert but I don't think you can have a port assigned to more than one selinux type, right?
That being said, yes, I agree it could be nice to have input from selinux devs.

guits:
```console
# semanage port -m -t http_port_t -p tcp 8081
# semanage port -l | grep 8081
http_port_t                    tcp      8081, 80, 81, 443, 488, 8008, 8009, 8443, 9000
transproxy_port_t              tcp      8081
# 
```

guits:
> I'm not a selinux expert but I don't think you can have a port assigned to more than one selinux type, right?

looks like it's possible @kalebskeithley

guits:

guits:

guits:

guits:

guits:

guits:

guits:

guits:

guits:
check-rdo

PR #6326, opened 2021-03-11
Disable facts by default in ansible.cfg

mwhahaha:
As a continuation of a7f2fa73e63e69dba2e41aaac9732397eec437c9, this
change switches fact injection to off by default in the provided
ansible.cfg.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -22,6 +22,7 @@ nocows = 1
 callback_whitelist = profile_tasks
 stdout_callback = yaml
 force_valid_group_names = ignore
+inject_facts_as_vars = False
 
 # Disable them in the context of https://review.openstack.org/#/c/469644
 retry_files_enabled = False
--- a/roles/ceph-facts/tasks/facts.yml
+++ b/roles/ceph-facts/tasks/facts.yml
@@ -19,6 +19,16 @@
     discovered_interpreter_python: "{{ ansible_python_interpreter }}"
   when: ansible_python_interpreter is defined
 
+# If ansible_python_interpreter is not defined, this can result in the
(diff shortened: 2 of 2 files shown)


dsavineau:
@mwhahaha I think there's still some `discovered_interpreter_python` occurence in https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-handler/templates/

mwhahaha:
Yea so the latest revision ensures that it is set.  This will allow ansible_python_interpreter to continue to function while setting the value to the fact value if not previously set

mwhahaha:
This looks like it breaks the centos_package_dependency var definition which assumes that distribution_major_version is set when those vars are loaded.

mwhahaha:
Actually no it seems to no changes related to that var, so I don't know why the lvm job fails

guits:

guits:

mwhahaha:
looks like we missed ansible_architecture

   The conditional check '(hostvars[item]['ansible_architecture'] == 'x86_64') or (not containerized_deployment | bool)' failed. The error was: error while evaluating conditional ((hostvars[item]['ansible_architecture'] == 'x86_64') or (not containerized_deployment | bool)): 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'ansible_architecture'

mwhahaha:
It looks like it's changed in code, not certain where that's coming from

guits:

guits:

guits:

guits:

guits:
> It looks like it's changed in code, not certain where that's coming from

Yep, I'm unsure how but I've reverted one of your changes in a recent patch... 😞

guits (approved):

PR #6368, opened 2021-03-20
Make sure the repo url contains the correct arch

badone:
We can end up with an arm only repo unless we are specific about the
architecture we require. Brings the deb code in line with the rpm
equivalent.

### DIFF
--- a/roles/ceph-common/tasks/installs/debian_dev_repository.yml
+++ b/roles/ceph-common/tasks/installs/debian_dev_repository.yml
@@ -1,7 +1,14 @@
 ---
+- name: get latest available build
+  uri:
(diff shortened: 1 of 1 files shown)


badone:
An example, arm repo listed first so gets consumed.

https://shaman.ceph.com/api/repos/ceph/octopus/23148618fcb587daa6bac0ac170c39ee83f3175d/ubuntu/bionic/

guits (approved):

dsavineau:
@guits I think that one need to be backported into stable branches

guits:
backports need to be created manually i think

badone:
> backports need to be created manually i think

I've had a go at doing this since we definitely need this in stable-4.0 for nautilus. Let me know if they need work as I'm not familiar with the process for ca.

Issue #6284, opened 2021-02-12
ceph_purge-container-cluster.yml cannot be run multiple times. 


hmeAtos:
Hello,

In my configuration I have set up Logical Volume.
When I deploy this it works fine.
the problem is during the purge, I cannot run the purge role (ceph_purge-container-cluster.yml) more than once.
I have the impression that this role is not idempotent.
I would like to know if I am the only one having this problem?

Thank you for your help.
regards

**Bug Report**

```
TASK [zap and destroy osds created by ceph-volume with lvm_volumes] ***************************************************************************************************************
Friday 12 February 2021  14:54:08 +0100 (0:00:00.136)       0:10:26.168 *******
The full traceback is:
Traceback (most recent call last):
  File "<stdin>", line 102, in <module>
  File "<stdin>", line 94, in _ansiballz_main
  File "<stdin>", line 40, in invoke_module
  File "/usr/lib64/python3.6/runpy.py", line 205, in run_module
    return _run_module_code(code, init_globals, run_name, mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmp/ansible_ceph_volume_payload_462rwwao/ansible_ceph_volume_payload.zip/ansible/modules/ceph_volume.py", line 759, in <module>
  File "/tmp/ansible_ceph_volume_payload_462rwwao/ansible_ceph_volume_payload.zip/ansible/modules/ceph_volume.py", line 755, in main
  File "/tmp/ansible_ceph_volume_payload_462rwwao/ansible_ceph_volume_payload.zip/ansible/modules/ceph_volume.py", line 633, in run_module
  File "/tmp/ansible_ceph_volume_payload_462rwwao/ansible_ceph_volume_payload.zip/ansible/modules/ceph_volume.py", line 475, in is_lv
  File "/usr/lib64/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
[...]

PR #6371, opened 2021-03-22
ceph_volume: fix bug in `is_lv()`

guits:
This function makes the `ceph_volume` module be not idempotent in
containerized context because it tries to run a container and bindmount
directories that no longer exist.

In that case, the `lvs` command being executed returns something
different than `0` so we can't call `json.loads(out)['report'][0]['lv']`
since it might throw a python error.

The idea is to return `True` only if `rc` is equal to `0` and
`len(result)` is greater than `0`, which means the command matched an
LV.

Fixes: #6284

### DIFF
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -446,11 +446,12 @@ def is_lv(module, vg, lv, container_image):
 
     rc, cmd, out, err = exec_command(module, cmd)
 
-    result = json.loads(out)['report'][0]['lv']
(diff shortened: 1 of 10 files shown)


dsavineau (approved):

guits:

guits:

guits:

guits:

guits:

hmeAtos:
Thanks @guits for this patch.
good work.

PR #6449, opened 2021-04-13
[skip ci] ceph-rgw-loadbalancer: Fix rgw_ports fact

BenoitKnecht:
The `set_fact rgw_ports` task was failing due to a templating error, because
`hostvars[item].rgw_instances` is a list, but it was treated as if it was a
dictionary.

Another issue was the fact that the `unique` filter only applied to the list
being appended to `rgw_ports` instead of the entire list, which means it was
possible to have duplicate items.

Lastly, `rgw_ports` would have been a list of integers, but the `seport` module
expects a list of strings.

This commit fixes all of the issues above, allowing the `ceph-rgw-loadbalancer`
role to work on systems with SELinux enabled.

This should be backported to the stable branches as well.

### DIFF
--- a/roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
+++ b/roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
@@ -41,7 +41,7 @@
   block:
     - name: set_fact rgw_ports
       set_fact:
-        rgw_ports: "{{ rgw_ports | default([]) + [hostvars[item]['rgw_instances']['radosgw_frontend_port']] | unique }}"
+        rgw_ports: "{{ rgw_ports | default([]) | union(hostvars[item]['rgw_instances'] | map(attribute='radosgw_frontend_port') | map('string') | list) }}"
       with_items: "{{ groups.get(rgw_group_name, []) }}"
 
     - name: add selinux rules


roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
```diff
@@ -41,12 +41,12 @@
...
     - name: set_fact rgw_ports
       set_fact:
-        rgw_ports: "{{ rgw_ports | default([]) + [hostvars[item]['rgw_instances']['radosgw_frontend_port']] | unique }}"
+        rgw_ports: "{{ rgw_ports | default([]) + hostvars[item].rgw_instances|map(attribute='radosgw_frontend_port')|list }}"
       with_items: "{{ groups.get(rgw_group_name, []) }}"
 
     - name: add selinux rules
       seport:
-        ports: "{{ rgw_ports }}"
+        ports: "{{ rgw_ports|map('string')|unique|list }}"
```
guits, line 49:
```suggestion
        ports: "{{ rgw_ports }}"
```

roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
```diff
@@ -41,12 +41,12 @@
   block:
     - name: set_fact rgw_ports
       set_fact:
-        rgw_ports: "{{ rgw_ports | default([]) + [hostvars[item]['rgw_instances']['radosgw_frontend_port']] | unique }}"
+        rgw_ports: "{{ rgw_ports | default([]) + hostvars[item].rgw_instances|map(attribute='radosgw_frontend_port')|list }}"
```
guits, line 44:
```suggestion
        rgw_ports: "{{ rgw_ports | default([]) | union(hostvars[item]['rgw_instances'] | map(attribute='radosgw_frontend_port') | map('string') | list) }}"
```

guits (changes requested):

roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
BenoitKnecht: Thanks for the suggestion, I've amended the commit accordingly!

roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
```diff
@@ -41,7 +41,7 @@
   block:
     - name: set_fact rgw_ports
       set_fact:
-        rgw_ports: "{{ rgw_ports | default([]) + [hostvars[item]['rgw_instances']['radosgw_frontend_port']] | unique }}"
+        rgw_ports: "{{ rgw_ports | default([]) | union(hostvars[item].rgw_instances | map(attribute='radosgw_frontend_port') | map('string') | list) }}"
```
guits, line 44:
```suggestion
        rgw_ports: "{{ rgw_ports | default([]) | union(hostvars[item]['rgw_instances'] | map(attribute='radosgw_frontend_port') | map('string') | list) }}"
```

guits (changes requested):
sorry, one last detail you missed in the previous review but IMO, it's better for readability and consistency 🙂

roles/ceph-rgw-loadbalancer/tasks/pre_requisite.yml
BenoitKnecht: I'm not sure I would call it more readable, but alright, done. :slightly_smiling_face:

guits (approved):

PR #6486, opened 2021-04-26
ceph-mon: Fix check mode for deploy monitor tasks (backport #6301)

BenoitKnecht:
This is a backport of #6301.


guits:
it looks like we are missing this in stable-6.0 and stable-5.0 too

BenoitKnecht:
Indeed. I can create backport PRs for those too if you want.

guits:
let's see if mergify can take care of it 🙂

PR #6507, opened 2021-05-04
[skip ci] update: fix ceph-crash stop task

guits:
This is a workaround for an issue in ansible.
When trying to stop/mask/disable this service in one task, the stop
didn't actually happen, the task doesn't fail but for some reason the
container is still present and running.
Then the task starting the service in the role ceph-crash fails because
it can't start the container since it's already running with the same
name.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1955393


guits:

guits:

guits:
rhcs note: it needs to be cherry-picked into 4.2z1 branch if we ship an async for 4.2z1

Issue #6501, opened 2021-04-29
ganesha dbus support broken for OSP deployments post stable-4.0


tombarron:
**Bug Report**

What happened:

We deployed OpenStack with stable 6.0 and discovered that 'manila access-allow <share> nfs <ipaddr>' always went to error state.  The manila share logs contained tracebacks indicating: "u'Error org.freedesktop.DBus.Error.ServiceUnknown: The name org.ganesha.nfsd was not provided by any .service files\n'.   In consequence, the manila share cannot be mounted from a client with  <ipaddr>.

What you expected to happen:

manila access-allow command will complete without error and mounts from <ipaddr> work.  No traceback in manila-share log about missing service file for dbus.

How to reproduce it (minimal and precise):
   1. Deploy OpenStack with Manila backed by CephFS via NFS using v6.0.0-stable-6.0-pacific-centos-8-x86_64
   2. Create a manila share.
   3. Run 'manila access-allow <share> nfs 0.0.0.0 # any address will do
   4. Run 'manila access-list <share>' and observe that the access rule is in Error state.
   5. Inspect maniila share log and observe GaneshaCommandFailure with error message mentioned above in traceback at the time of the 'manila access-allow ...' command

 
Should be reproducible with stable 5.0 or later which has commit 33bfb10af993faf97a976972f47344ab7ba51edf [1] or backport of same.  The commit is a good idea for deployments where it is known that the nfs-ganesha package will be installed exactly where the dbus daemon is running -- e.g. where message source, dbus daemon, and ganesha run on the same host or in the same container.

In OpenStack, however, this is not the case.
   - message source is maniila-share process, running in a podman container.
   - dbus daemon is on a bare metal host where nfs-ganesha (and the ceph services) *are not installed*
   - message destination is nfs-ganesha running in another podman container
Both message source and message destination containers run on the same host as the dbus daemon and share a common dbus rendevous socket via bind mounts from the host into the containers.

Before commit [1], ceph-ansible installed the ganesha dbus service file on the host machine and it all worked.
After that commit, the dbus service file is not installed on the host machine, only in the ganesha container where nfs-ganesha package is installed, so we get the failure citead above.

[1] 

        [tbarron@p50 ceph-ansible]$ git show 33bfb10af993faf97a976972f47344ab7ba51edf
        commit 33bfb10af993faf97a976972f47344ab7ba51edf
        Author: Guillaume Abrioux <gabrioux@redhat.com>
[...]

PR #6519, opened 2021-05-06
[skip ci] nfs: get org.ganesha.nfsd.conf from container

guits:
Since we need to revert 33bfb10, this is an alternative to initial approach.
We can avoid maintaining this file since it is present in container
image. The idea is to simply get it from the image container and write
it to the host.

Fixes: #6501

### DIFF
--- a/roles/ceph-nfs/tasks/pre_requisite_container.yml
+++ b/roles/ceph-nfs/tasks/pre_requisite_container.yml
@@ -41,3 +41,22 @@
         - item.item.copy_key | bool
   when: groups.get(mon_group_name, []) | length > 0
 
+- name: dbus related tasks
+  block:
+    - name: get file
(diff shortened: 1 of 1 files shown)


guits:

EmilienM:
:+1: need this :)

dsavineau:
> 👍 need this :)

I'm curious about why you need this since TripleO starting octopus/pacific should have switched from ceph-ansible to tripleo-ansible managing the ceph deployment via cephadm.

Am I missing something ?

EmilienM:
> > +1 need this :)
> 
> I'm curious about why you need this since TripleO starting octopus/pacific should have switched from ceph-ansible to tripleo-ansible managing the ceph deployment via cephadm.
> 
> Am I missing something ?

In my team we were still using ceph-ansible because the cephadm integration was WIP but now we'll switch to the new thing.
Thanks for checkin-in

tombarron:
FWIW tripleo in upstream RDO currently runs two test tracks -- one with ceph-ansible and one with cephadm -- and ganesha based scenarios (scenario004) are still only deployed and tested using the former.  There is WIP to resolve differences in the way OpenStack uses ganesha and what cephadm actually supports today, and when that is resolved we can cut over entirely to cephadm when targeting octopus/pacific.

PR #6526, opened 2021-05-07
[skip ci] nfs: replace a command task

guits:
Let's use the ansible service module instead of using command module.
by the way, killall isn't POSIX standard.

### DIFF
--- a/roles/ceph-nfs/tasks/pre_requisite_container.yml
+++ b/roles/ceph-nfs/tasks/pre_requisite_container.yml
@@ -56,7 +56,11 @@
         owner: "root"
         group: "root"
         mode: "0644"
+      register: dbus_svc_file
 
     - name: reload dbus configuration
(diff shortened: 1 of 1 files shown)


roles/ceph-nfs/tasks/pre_requisite_container.yml
```diff
@@ -56,7 +56,11 @@
...
         group: "root"
         mode: "0644"
+      register: dbus_svc_file
 
     - name: reload dbus configuration
-      command: "killall -SIGHUP dbus-daemon"
+      service:
+        name: dbus
+        state: reloaded
+      when: dbus_svc_file.changed | bool
```
dsavineau, line 65: are you afraid to create a new handler ? :p

guits: > are you afraid to create a new handler ? :p

😅  😏

PR #6533, opened 2021-05-17
Make sure the repo url contains the correct arch

badone:
We can end up with an arm only repo unless we are specific about the
architecture we require. Brings the deb code in line with the rpm
equivalent.

(cherry picked from commit 267cce9e8360fc8cb9c192fde2406e5dca724610)


badone:
Note that this is an attempt to backport https://github.com/ceph/ceph-ansible/pull/6368 to stable-4.0

PR #6545, opened 2021-05-24
cephadm-adopt: purge grafana db

guits:
ceph-ansible leaves /var/lib/grafana content so it overrides the
datasource that cephadm writes. It causes a 502 error "bad gateway".

The idea here is basically removing the grafana content in
/var/lib/grafana which should fix this issue.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1963849

### DIFF
--- a/infrastructure-playbooks/cephadm-adopt.yml
+++ b/infrastructure-playbooks/cephadm-adopt.yml
@@ -1155,6 +1155,11 @@
             enabled: false
           failed_when: false
 
+        - name: purge old grafana db in /var/lib/grafana
+          file:
+            path: /var/lib/grafana/grafana.db
+            state: absent
+
         - name: adopt grafana daemon
           cephadm_adopt:
             name: "grafana.{{ ansible_facts['hostname'] }}"


guits:

guits:

guits:

sebastian-philipp (approved):

sebastian-philipp:
So, there is a side effect: Custom changes made done manually in ceph-ansible are gone afterwards.

epuertat:
@pcuzner this is a fix for an issue detected during migration of ceph-ansible to cephadm while persisting `/var/lib/grafana`: basically there're duplicated datasources (the ceph-ansible one and the cephadm one), and the default is the ceph-ansible one, which breaks Ceph-dashboard. Is it ok to delete the Grafana db?

epuertat (approved):

sebastian-philipp:
relates to https://github.com/ceph/ceph/pull/40537

guits:

pcuzner (commented):
if we delete the grafana.db then any local dashboards or changes to users/passwords within grafana wlll be lost right? Are there warnings/prompts in the playbook to make the admin aware of this?

Alternatively, rhcs4 used a datasource name of Dashboard, and rhcs5 is using a datasource name of Dashboard1. if the grafana.db is intact, any user added dataources will remain in place - so the only issue is the provisioned datasource for the dashboard  - which could just be overridden, discarding the old entry altogether. This would remove the bogus entry and keep any customisations intact.

epuertat:
@pcuzner instead of printing a warning that might sound alarming and, perhaps, intimidate users from migrating, what about doing a preventive back-up in case the file is found? I remember Ansible had such an option when overwriting files, right?

The main issue with the datasource is that regardless the name, it seems that Grafana is trying to use the default one (`Dashboard` instead of `Dashboard1`) and there's no way to manually disable or delete that datasource (you get an error saying that the datasource was added via configuration and that you'll need to delete it in the same way).

To make things worse, there's no datasource managament with `grafana-cli`, you forcibly have to rely on the REST API to do so...

guits:
> @pcuzner instead of printing a warning that might sound alarming and, perhaps, intimidate users from migrating

Yeh, I agree this might intimidate, especially when we are talking about data loss in a storage cluster.

> what about doing a preventive back-up in case the file is found? I remember Ansible had such an option when overwriting files, right?

Yes, ansible can do that.

> To make things worse, there's no datasource managament with `grafana-cli`, you forcibly have to rely on the REST API to do so...

I'm not a grafana expert at all, but I think we can easily change the default datasource with a basic call the the REST API though? right?

In any case, thanks all for getting into that.

pcuzner:
@epuertat not sure I follow. Are you saying that even when the provisioning file has a deletedatasources for "Dashboard" it ignores it?

epuertat:
> @epuertat not sure I follow. Are you saying that even when the provisioning file has a deletedatasources for "Dashboard" it ignores it?

@pcuzner I asked the QE team to repeat the migration process to double check what's going on there, but what we saw is that Grafana was still trying to fetch data from the default datasource. I just read [the PR where you introduced this `$datasource` variable](https://github.com/ceph/ceph/pull/24314), but I don't see in the code where this variable is interpolated...

epuertat:
@pcuzner Fixing this in https://github.com/ceph/ceph/pull/41598. Let's see if this PR is eventually not needed.

guits:
closing this in favor of ceph/ceph#41598

PR #6551, opened 2021-05-25
[skip ci] fs2bs: fix wrong filter when setting osd_ids

guits:
using 'match' filter in that task will lead to bad behavior if I have
the following node names for instance:

- node1
- node11
- node111

with `selectattr('name', 'match', inventory_hostname)` it will match
'node1' along with 'node11' and 'node111'.

using 'equalto' filter will make sure we only match the target node.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1963066

### DIFF
--- a/infrastructure-playbooks/filestore-to-bluestore.yml
+++ b/infrastructure-playbooks/filestore-to-bluestore.yml
@@ -48,7 +48,7 @@
       set_fact:
         osd_ids: "{{ osd_ids | default([]) | union(item) }}"
       with_items:
-        - "{{ ((osd_tree.stdout | default('{}') | trim | from_json).nodes | selectattr('name', 'match', '^' + inventory_hostname + '$') | map(attribute='children') | list) }}"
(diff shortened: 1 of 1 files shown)


guits:
jenkins test centos-non_container-filestore_to_bluestore

guits:
jenkins test centos-container-filestore_to_bluestore

guits:
jenkins test centos-non_container-filestore_to_bluestore

dsavineau:
@guits does that mean the ^$ doesn't work at all ? because we have several occurences of this in the ceph-ansible (like rgw multisite).

Also one of the reason of using `match` instead of `equalto` was the incompatibility with the jinja2 version for python 2.7 on EL 7 [1]. So with that change, the fs2bs playbook won't work on EL7 ansible node.

[1] https://github.com/ceph/ceph-ansible/pull/5341

guits:
damn... I was suspecting there was something that led us to use match instead of equalto... I couldn't really remember... :confused:

PR #6576, opened 2021-06-10
rolling_update: fix mon+rgw/multisite collocation

guits:
When monitors and rgw are collocated with multisite enabled, the
rolling_update playbook fails because during the workflow, we run some
radosgw-admin commands very early on the first mon even though this is
the monitor being upgraded, it means the container doesn't exist since
it was stopped.
With this commit we now run an ephemeral container instead for running
those radosgw-admin commands so we don't need to care if the mon container
is running.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1970232

### DIFF
--- a/roles/ceph-rgw/tasks/main.yml
+++ b/roles/ceph-rgw/tasks/main.yml
@@ -1,4 +1,8 @@
 ---
+- name: set_fact radosgw_adm_cmd
+  set_fact:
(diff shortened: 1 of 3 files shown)


guits:
backport note:

the task `fetch the realm(s)` in stable-6.0, stable-5.0 and stable-4.0 must be modified given it doesn't use the `radosgw_realm` module

guits:
@dsavineau maybe we should skip all of this https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-handler/tasks/main.yml#L52-L78 when `rolling_update` is `True` ?

guits:
jenkins test centos-non_container-external_clients

dsavineau:
> @dsavineau maybe we should skip all of this https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-handler/tasks/main.yml#L52-L78 when `rolling_update` is `True` ?

I think that's a good idea. I don't see any reason to play those tasks during the upgrade workflow.

IMHO those tasks are only needed during the main deployment or other day 2 operations like scale up and others

guits:

guits:

guits:

guits:
closing in favor of #6577

PR #6604, opened 2021-06-16
Don't create RGW pools on upgrade

gaudenz:
Skip the RGW pool creation tasks on upgrade. The same is already
done for OpenStack pools.

When upgrading from an older version which may have created a different
set of pools or pools with different parameters, creating the pools
could create unneeded new pools or cause pool parameters to change.

### DIFF
--- a/roles/ceph-rgw/tasks/main.yml
+++ b/roles/ceph-rgw/tasks/main.yml
@@ -9,7 +9,9 @@
 - name: rgw pool creation tasks
   include_tasks: rgw_create_pools.yml
   run_once: true
-  when: rgw_create_pools is defined
+  when:
+    - rgw_create_pools is defined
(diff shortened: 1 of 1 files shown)


roles/ceph-rgw/tasks/main.yml
```diff
@@ -9,7 +9,9 @@
 - name: rgw pool creation tasks
   include_tasks: rgw_create_pools.yml
   run_once: true
-  when: rgw_create_pools is defined
+  when:
+    - rgw_create_pools is defined
+    - not rolling_update | default(False) | bool
```
dsavineau, line 14:
```suggestion
    - not rolling_update | bool
```

It's just a minor thing since the rolling_update variable value is already set at false by default.

PR #6606, opened 2021-06-16
Create radosgw key on mon host

gaudenz:
Instead of copying the bootstrap key to the rgw host, create the key on
the first mon host and just copy over the resulting key. The bootstrap
key basically gives full permissions to create keys on the cluster and
should be tightly secured and not be on a rgw host which might be
directly connected to the Internet.

If the pools are created by the Ansible playbook, the key permissions are
restricted to just allow the operations needed for radosgw and do not
allow pool manipulation.

### DIFF
--- a/roles/ceph-rgw/tasks/common.yml
+++ b/roles/ceph-rgw/tasks/common.yml
@@ -8,37 +8,32 @@
     mode: "{{ ceph_directories_mode }}"
   with_items: "{{ rbd_client_admin_socket_path }}"
 
-- name: get keys from monitors
+- name: get admin key from monitors
   ceph_key:
-    name: "{{ item.name }}"
+    name: "client.admin"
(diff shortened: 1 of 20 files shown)


gaudenz:
@guits Is there a problem with using the CI for draft PRs? I actually made this a draft because I wanted to run the CI tests. I now made some changes to fix the tests which were failing and would like to recheck. Otherwise I would have to see how I can run the checks locally.

guits:
> @guits Is there a problem with using the CI for draft PRs? I actually made this a draft because I wanted to run the CI tests. I now made some changes to fix the tests which were failing and would like to recheck. Otherwise I would have to see how I can run the checks locally.

Hi @gaudenz sorry for the late response,

yes, there was a couple of issues with the CI (like almost all job failing because of the infra), even though I still see some failures here, that should be fixed now

PR #6624, opened 2021-06-27
dashboard: split monitoring deployment tools from ceph dashboard deployment

clwluvw:
Add monitoring_enabled variable for deploying prometheus, grafana, alertmanager and use dashboard_enabled variable just for deploying and configuring ceph dashboard


guits:
jenkins test playbook syntax

guits:

guits:
@clwluvw unfortunately ceph-ansible is going to be deprecated. As you may know already, cephadm is going to replace ceph-ansible after stable-6.0. By the way, ceph-ansible in stable-6.0 is supported only for upgrading and migrating to cephadm.
Although I think it's a valid and a nice patch (I do value the work you've done here), I feel sorry because it means it doesn't really make sense to merge it given the project is going to be in "maintenance mode" very soon, those changes are too significant to be backported, even though we only merge it in the master branch, it might give us hard times when we will have to push dashboard fixes related in stable branches.

clwluvw:
@guits Makes sense! but for example, if we can have it just for pacific (if backporting is easy) it would be so great!

PR #6704, opened 2021-07-09
[skip ci] rolling_update: check quorum state before upgrade

dsavineau:
If one a the monitor is out of the quorum then nothing prevents the upgrade
playbook to run.
We only check if we have at least three monitor nodes but we should also
check if those monitor nodes are correctly present in the quorum.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1952571


dsavineau:

dsavineau:

guits (approved):

PR #6705, opened 2021-07-09
[skip ci] multisite: use node fqdn for endpoints when https

dsavineau:
When the `rgw_multisite_proto` variable is set to https then we shoudn't use
the IP address in the zone endpoints list but the node FQDN to match the
TLS certificate CN.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1965504

### DIFF
--- a/roles/ceph-rgw/tasks/multisite/main.yml
+++ b/roles/ceph-rgw/tasks/multisite/main.yml
@@ -21,7 +21,7 @@
 
 - name: create a list of dicts with each rgw endpoint and it's zone
   set_fact:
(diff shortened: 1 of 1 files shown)


dsavineau:
jenkins test centos-non_container-multisite

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:

dsavineau:
@guits looks like without this change we can't deploy multisite with https as I always get an error when https + IP addresses but I'm still wondering if we need to allow passing another value instead the ansible fqdn fact. WDYT ?

@alimaredia any thoughts on this ?

dsavineau:

dsavineau:

dsavineau:

dsavineau:

alimaredia (approved):
> Looks good to me. @dsavineau are you able to test this by hand?

Yes I did and it worked well.

The only limitation I can see is that the ansible fqdn of the node should resolves the same IP address than the one used for the radosgw instances.

dsavineau:
> Looks good to me. @dsavineau are you able to test this by hand?

Yes I did and it worked well.

The only limitation I can see is that the ansible fqdn of the node should resolves the same IP address than the one used for the radosgw instances.

guits (approved):

PR #6733, opened 2021-07-20
[skip ci] common: remove unnecessary run_once statements

dsavineau:
1303611 introduced tasks for disabling the pg_autoscaler on pools and
the balancer but thoses tasks are already executed on the first monitor
node so we don't need to add the `run_once` statement.


dsavineau:

dsavineau:

guits (approved):

PR #6734, opened 2021-07-20
common: fix py2 pool_list from_json when skipped

dsavineau:
When using python 2 and the task with a loop is skipped then it generates
an error.

Unexpected templating type error occurred on
({{ (pool_list.stdout | from_json)['pools'] }}): expected string or buffer


dsavineau:

guits (approved):

PR #6765, opened 2021-07-27
cephadm-adopt: check for version only once in containerized_deployment

clwluvw:
Based on running container on the same image for all hosts it can be done with one host

### DIFF
--- a/infrastructure-playbooks/cephadm-adopt.yml
+++ b/infrastructure-playbooks/cephadm-adopt.yml
@@ -93,14 +93,17 @@
 
     - name: get the ceph version
       command: "{{ container_binary + ' run --rm --entrypoint=ceph ' + ceph_docker_registry + '/' + ceph_docker_image + ':' + ceph_docker_image_tag if containerized_deployment | bool else 'ceph' }} --version"
+      run_once: "{{ containerized_deployment }}"
       changed_when: false
       register: ceph_version_out
 
     - name: set_fact ceph_version
+      run_once: "{{ containerized_deployment }}"
       set_fact:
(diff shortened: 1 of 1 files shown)


dsavineau:
> Based on running container on the same image for all hosts it can be done with one host

The task is done for both containerized and non containerized deployment so you're breaking the current workflow.

Also using `run_once: true` doesn't make sense since this task aims to check all nodes and not just one.

There no guarantee that you have the same image on all nodes. You can have an environment with nodes which were not upgrade so using `run_once: true` will hides this behaviour

clwluvw:
@dsavineau You are not checking each container in hosts and you just running the same container on all hosts and get the same result as you can see the command in task:
```
ceph_docker_registry + '/' + ceph_docker_image + ':' + ceph_docker_image_tag
```

This is redundant and with this PR this can be efficient!

dsavineau:
> @dsavineau You are not checking each container in hosts and you just running the same container on all hosts and get the same result as you can see the command in task:
> 
> ```
> ceph_docker_registry + '/' + ceph_docker_image + ':' + ceph_docker_image_tag
> ```
> 
> This is redundant and with this PR this can be efficient!

```console
$ cat group_vars/*.yml
---
ceph_docker_registry: quay.io
ceph_docker_image: ceph/daemon
ceph_docker_image_tag: latest-master
...
---
ceph_docker_registry: quay.io
ceph_docker_image: ceph/daemon
ceph_docker_image_tag: latest-master-fix_issue_42
...
```

How do you deal with different container images then ?

clwluvw:
@dsavineau Yes, I didn't remember this situation. You are right!

PR #6766, opened 2021-07-27
cephadm-adopt: pull image when containerized_deployment

clwluvw:


dsavineau:
This doesn't make sense at all.

You need to pull the image when you're using package deployment.

If you're already using containers then you already have the container image.

Issue #6197, opened 2021-01-19
Error when installing : TASK [ceph-dashboard : create dashboard admin user] stderr: 'Error ENOENT: User ''admin'' does not exist'


BojungKim:
**Bug Report**

What happened:
ceph-ansible fatal failed. TASK [ceph-dashboard : create dashboard admin user] stderr: 'Error ENOENT: User ''admin'' does not exist'.

What you expected to happen:
Creating admin does not working. 

How to reproduce it (minimal and precise):

Share your group_vars files, inventory and **full** ceph-ansibe log

**Environment**:
* OS (e.g. from /etc/os-release): Ubuntu 18.04
* Kernel (e.g. `uname -a`):  4.15.0-128-generic
* Docker version if applicable (e.g. `docker version`): No use
* Ansible version (e.g. `ansible-playbook --version`): ansible-playbook 2.9.16
* ceph-ansible version (e.g. `git head or tag or stable branch`):
I don't know git well. Just latest commit id : 13427eddac03f9f62e98f6689948e98e7baeef29

* Ceph version (e.g. `ceph -v`): ceph version 15.2.8 (bdf3eebcd22d7d0b3dd4d5501bee5bac354d5b55) octopus (stable)

### hosts (file)
[mgrs]
ceph-mon01 ansible_host=10.254.3.91
ceph-mon02 ansible_host=10.254.3.92
ceph-mon03 ansible_host=10.254.3.93

[rgws]
ceph-mon01 ansible_host=10.254.3.91
ceph-mon02 ansible_host=10.254.3.92
ceph-mon03 ansible_host=10.254.3.93

[osds]
ceph-osd01 ansible_host=10.254.3.94
ceph-osd02 ansible_host=10.254.3.95
ceph-osd03 ansible_host=10.254.3.96

[mons]
ceph-mon01 ansible_host=10.254.3.91
ceph-mon02 ansible_host=10.254.3.92
ceph-mon03 ansible_host=10.254.3.93

[mdss]
ceph-mon01 ansible_host=10.254.3.91
ceph-mon02 ansible_host=10.254.3.92
ceph-mon03 ansible_host=10.254.3.93

[rgwloadbalancers]
ceph-mon01 ansible_host=10.254.3.91
ceph-mon02 ansible_host=10.254.3.92
ceph-mon03 ansible_host=10.254.3.93

[monitoring]
ceph-mon01 ansible_host=10.254.3.91
[...]

PR #6773, opened 2021-07-28
[skip ci] library: exit on user creation failure

dsavineau:
When the ceph dashboard user creation fails then the issue is hidden
as we don't check the return code and don't print the error message
in the module output.

This ends up with a failure on the ceph dashboard set roles command saying
that the user doesn't exist.

By failing on the user creation, we will have an explicit explaination of
the issue (like weak password).

Closes: #6197


dsavineau:

dsavineau:

guits (approved):

PR #6776, opened 2021-07-28
[skip ci] cephadm-adopt: use --realm and --zone for RGW

dsavineau:
Historically, ceph orch apply rgw was using:
 - 1st arg as service id
 - 2nd arg as realm value
 - 3rd arg as zone value

There's now dedicated parameters for that (even if the current implementation
is still working) but we have now the same parameters used for both multisite
and non multisite setup.


sebastian-philipp (approved):

dsavineau:
closing in flavor of #6776

PR #6779, opened 2021-07-29
[skip ci] rolling_update: get ceph version when mons exist

dsavineau:
eec3878 introduced a regression for upgrade scenarios where there's no
monitor nodes at all (like ganesha standalone, external clients, etc..)

```console
TASK [get the ceph release being deployed] ************************************
task path: infrastructure-playbooks/rolling_update.yml:121
Thursday 29 July 2021  15:55:29 +0000 (0:00:00.484)       0:00:15.802 *********
fatal: [client0]: FAILED! =>
  msg: '''dict object'' has no attribute ''mons'''
```

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -114,7 +114,9 @@
         - ceph_docker_registry_auth | bool
 
     - name: check ceph release in container image
-      when: containerized_deployment | bool
+      when:
+        - groups.get(mon_group_name, []) | length > 0
+        - containerized_deployment | bool
       delegate_to: "{{ groups[mon_group_name][0] }}"
       run_once: true
       block:


dsavineau:
jenkins test centos-container-external_clients

dsavineau:

dsavineau:
@guits I don't know what's the best solution for this.

At the moment, this patch ignores the ceph version check when there's no monitor nodes but maybe we could do something else (like delegate on the first client node if there's no monitor).

What do you think ?

guits:
> @guits I don't know what's the best solution for this.
> 
> At the moment, this patch ignores the ceph version check when there's no monitor nodes but maybe we could do something else (like delegate on the first client node if there's no monitor).
> 
> What do you think ?

followup RHCS 9.3 maggle 😏

guits (approved):

PR #6804, opened 2021-08-06
add-osd: use container_exec_cmd fact from mon host

dsavineau:
Because we're delegating the task to the first monitor node, we need to be
sure that the container_exec_cmd fact is the one from that node too otherwise
we could have a mismatch on the ceph-mon container name.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1990772

### DIFF
--- a/infrastructure-playbooks/add-osd.yml
+++ b/infrastructure-playbooks/add-osd.yml
@@ -113,7 +113,7 @@
         add_osd: True
 
     - name: set noup flag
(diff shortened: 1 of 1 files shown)


dsavineau:
jenkins test centos-container-add_osd

dsavineau:
jenkins test centos-container-add_osds

dsavineau:
jenkins test centos-non_container-add_osds

guits (approved):

PR #6811, opened 2021-08-09
tests: add pytest-timeout

dsavineau:
There's no global timeout on pytest meaning if one test hang then the
CI job will only timeout after 3h instead of failing earlier.
Adding a global timeout to 300s.


dsavineau:

guits (approved):

PR #6829, opened 2021-08-12
tests: use old build of ceph@master

guits:
for unlocking the ci.
this is intended to be reverted.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -297,6 +297,7 @@ setenv=
   ANSIBLE_KEEP_REMOTE_FILES = 1
   ANSIBLE_CACHE_PLUGIN = memory
   ANSIBLE_GATHERING = implicit
+  CEPH_DEV_SHA1 = 6436cc5e13174b8b206301b0a073b8a776eea490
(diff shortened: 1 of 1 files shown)


guits:

guits:

dsavineau (commented):
Do we need to do the same for other tox file ?

I know we only have reboot in all_daemons and collocation scenarios but we won't test the same ceph@master version on every scenarios

guits:
> Do we need to do the same for other tox file ?
> 
> I know we only have reboot in all_daemons and collocation scenarios but we won't test the same ceph@master version on every scenarios

this is a temporary workaround for unlocking the ci.
as you said, we only reboot in all_daemons and collocation, having ceph@master being tested against two differents commits doesn't really matter IMO.

dsavineau (approved):

guits:
the only downside with this is that I don't really know how long this build of ceph@master will be available

guits:
let's give it a shot, will see how it goes :man_shrugging:

PR #6846, opened 2021-08-18
[skip ci] iscsi: don't set default value for trusted_ip_list

guits:
It restricts access to the iSCSI API.
It can be left empty if the API isn't going to be access from outside the
gateway node

Even though this seems to be a limited use case, it's better to leave it
empty by default than having a meaningless default value.

We could make this variable mandatory but that would be a breaking
change. Let's just add a logic in the template in order to set this
variable in the configuration file only if it was specified by users.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1994930

### DIFF
--- a/roles/ceph-iscsi-gw/tasks/common.yml
+++ b/roles/ceph-iscsi-gw/tasks/common.yml
@@ -30,7 +30,7 @@
 
 - name: add mgr ip address to trusted list with dashboard - ipv4
   set_fact:
-    trusted_ip_list: '{{ trusted_ip_list }},{{ hostvars[item]["ansible_facts"]["all_ipv4_addresses"] | ips_in_ranges(public_network.split(",")) | first }}'
+    trusted_ip_list: '{{ trusted_ip_list | default("") }}{{ "," if trusted_ip_list is defined else "" }}{{ hostvars[item]["ansible_facts"]["all_ipv4_addresses"] | ips_in_ranges(public_network.split(",")) | first }}'
   with_items: '{{ groups[mgr_group_name] | default(groups[mon_group_name]) }}'
   when:
     - dashboard_enabled | bool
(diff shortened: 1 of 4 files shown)


dsavineau (commented):
I guess this will break the iscsi integration within the ceph dashboard 

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-iscsi-gw/tasks/common.yml#L31-L45

dsavineau:
> I guess this will break the iscsi integration within the ceph dashboard
> 
> https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-iscsi-gw/tasks/common.yml#L31-L45

Not the best solution but at least it works

```diff
    - name: add mgr ip address to trusted list with dashboard - ipv4
      set_fact:
-        trusted_ip_list: '{{ trusted_ip_list }},{{ hostvars[item]["ansible_facts"]["all_ipv4_addresses"] | ips_in_ranges(public_network.split(",")) | first }}'
+        trusted_ip_list: '{{ trusted_ip_list | default("") }}{{ "," if trusted_ip_list is defined else "" }}{{ hostvars[item]["ansible_facts"]["all_ipv4_addresses"] | ips_in_ranges(public_network.split(",")) | first }}'
      with_items: '{{ groups[mgr_group_name] | default(groups[mon_group_name]) }}'
      when:
        - dashboard_enabled | bool
        - ip_version == 'ipv4'
```

roles/ceph-iscsi-gw/tasks/common.yml
```diff
@@ -30,15 +30,15 @@
...
+    trusted_ip_list: '{{ trusted_ip_list | default('') }}{{ "," if trusted_ip_list is defined else "" }}{{ hostvars[item]["ansible_facts"]["all_ipv4_addresses"] | ips_in_ranges(public_network.split(",")) | first }}'
   with_items: '{{ groups[mgr_group_name] | default(groups[mon_group_name]) }}'
   when:
     - dashboard_enabled | bool
     - ip_version == 'ipv4'
 
 - name: add mgr ip address to trusted list with dashboard - ipv6
   set_fact:
-    trusted_ip_list: '{{ trusted_ip_list }},{{ hostvars[item]["ansible_facts"]["all_ipv6_addresses"] | ips_in_ranges(public_network.split(",")) | last }}'
+    trusted_ip_list: '{{ trusted_ip_list | default('') }}{{ "," if trusted_ip_list is defined else "" }}{{ hostvars[item]["ansible_facts"]["all_ipv6_addresses"] | ips_in_ranges(public_network.split(",")) | last }}'
```
dsavineau, line 41: I think the single quote in the default filter will conflict with the global quotes. You need to use double quotes here

dsavineau (approved):

PR #3718, opened 2019-03-15
rolling_update: Remove ceph aliases

dsavineau:
ceph aliases have been introduced in stable-3.2 during the ceph
deployment. On master this has been removed but we don't handle
this removal in the upgrade from stable-3.2 to master via the
rolling_update playbook.
Also remove the task from purge-docker-cluster missing from
d9e7835

Signed-off-by: Dimitri Savineau <dsavinea@redhat.com>

PR #6906, opened 2021-09-28
common: ensure ceph-aliases.sh is absent

guits:
For pre-existing cluster where this fail was already deployed we need to
ensure it is removed.
See https://github.com/ceph/ceph-ansible/pull/3447#issue-390617983

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -135,6 +135,12 @@
       set_fact:
         rolling_update: true
 
+    - name: ensure legacy ceph aliases file is absent
+      file:
+        path: /etc/profile.d/ceph-aliases.sh
+        state: absent
+
+
 - name: upgrade ceph mon cluster
   vars:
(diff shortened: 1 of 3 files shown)


infrastructure-playbooks/rolling_update.yml
```diff
@@ -135,6 +135,12 @@
       set_fact:
         rolling_update: true
 
+    - name: ensure legacy ceph aliases file is absent
+      file:
+        path: /etc/profile.d/ceph-aliases.sh
```
dsavineau, line 140: This is already done in the upgrade playbook since the beginning of the stable-4.0 branch and tags (look at first task in the mon play) so we shouldn't have any deployment (from scratch or upgraded) with that file anymore.

https://github.com/ceph/ceph-ansible/commit/57b4e76d11e0e3bb58c299ae5d37617ea06e2c17
https://github.com/ceph/ceph-ansible/pull/3718

guits:
that's correct, i didn't notice this.

PR #7019, opened 2021-12-07
Make cluster network optional for vagrant

rubenk:
Signed-off-by: Ruben Kerkhof <ruben@42on.com>

### DIFF
--- a/roles/ceph-defaults/defaults/main.yml
+++ b/roles/ceph-defaults/defaults/main.yml
@@ -335,7 +335,6 @@ osd_memory_target: 4294967296
 journal_size: 5120 # OSD journal size in MB
 block_db_size: -1 # block db size in bytes for the ceph-volume lvm batch. -1 means use the default of 'as big as possible'.
 public_network: 0.0.0.0/0
-cluster_network: "{{ public_network | regex_replace(' ', '') }}"
 osd_mkfs_type: xfs
 osd_mkfs_options_xfs: -f -i size=2048
 osd_mount_options_xfs: noatime,largeio,inode64,swalloc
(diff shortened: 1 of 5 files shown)


roles/ceph-defaults/defaults/main.yml
```diff
@@ -335,7 +335,6 @@ osd_memory_target: 4294967296
 journal_size: 5120 # OSD journal size in MB
 block_db_size: -1 # block db size in bytes for the ceph-volume lvm batch. -1 means use the default of 'as big as possible'.
 public_network: 0.0.0.0/0
-cluster_network: "{{ public_network | regex_replace(' ', '') }}"
```
guits, line 338: why are you removing this?

guits: this is a breaking change for the ceph-infra role

rubenk: if cluster_network is defined, it ends up in /etc/ceph/ceph.conf, even if it's the same as public_network

guits: What is the issue with that?

This is one of the multiple possible reasons for writing a commit message actually: describing what you are changing and why...
By the way, I don't think having 2 commits is needed here, you might want to squash them.

Thanks!

rubenk: I don't need a cluster network, and prefer having less settings in ceph.conf. I thought that was obvious but I'll amend the commit message and squash.

guits:
you have to run `./generate_group_vars_sample.sh`

PR #7033, opened 2021-12-21
dashboard: fix radosgw system user creation

guits:
The radosgw system user creation will fail when `rgw_instances`
is set at the host_var level because this variable won't bet set
on monitor nodes, given that this is where the tasks is delegated, it fails.

The idea here is to check over all rgw instances that are defined and set a
boolean fact in order to check if at least one instance has `rgw_zonemaster` set
to `True`

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2034595


guits:

PR #7086, opened 2022-02-11
library/ceph_pool: Fix potential null value when creating pools

Papawy:
Before, creating a pool by providing *only* `pg_num` would result in an
error as it would produce `--pgp-num null`.

This commit fix this behavior by defaulting `pgp_num` value to `pg_num`.

As described in Ceph documentation, when creating a pool `pgp_num` and `pg_num` should be equal.


guits:

guits (approved):

PR #7127, opened 2022-03-30
do not update Debian cache when package-install is disabled

insatomcat:
When deploying with --skip-tags=package-install (when there is no access to a repository), the playbook is still trying to update the package cache, which makes the playbook fail.
This change prevents the playbook to try to update the cache when the package-install tag is skipped.

PR #7132, opened 2022-04-04
do not update Debian cache when package-install is disabled (backport #7127)

insatomcat:
When deploying with --skip-tags=package-install (when there is no access to a repository), the playbook is still trying to update the package cache, which makes the playbook fail.
This change prevents the playbook to try to update the cache when the package-install tag is skipped.

Signed-off-by: Florent CARLI <florent.carli@rte-france.com>
(cherry picked from commit 63f20f59418a668261c91e408ad7ae5ac67a63d1)


guits:
jenkins test centos-non_container-subset_update

PR #7135, opened 2022-04-04
facts: always set `docker_exec_cmd`

guits:
This ensures `docker_exec_cmd` is always set on the first monitor.
Otherwise, when running the playbook with `--limit`, it fails on tasks
calling `docker_exec_cmd` with a delegate_to on the first monitor since
it was not set.


guits:

PR #7163, opened 2022-04-25
ceph_pool: set target size ratio on both 'on' and 'warn' mode

clwluvw:
when we set target_size_ratio to warn it means that the administrator wants to get suggestion from the mgr module but apply it manually when he/she wants. So it's in the same approach as 'on' mode just triggered by hand.
So there is no need to set pg_num when target_size_ratio is 'warn' and the mgr module will calculate the correct pg_num and the administrator will adjust it whenever he/she wants.

It is the same approach that was in #6471

### DIFF
--- a/library/ceph_pool.py
+++ b/library/ceph_pool.py
@@ -391,7 +391,7 @@ def create_pool(cluster,
     args = ['create', user_pool_config['pool_name']['value'],
             user_pool_config['type']['value']]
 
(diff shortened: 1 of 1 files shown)


clwluvw:
ping @guits

clwluvw:
ping @guits

guits (approved):

clwluvw:
@Mergifyio backport stable-6.0

clwluvw:
@Mergifyio backport stable-5.0

PR #7185, opened 2022-05-18
Add python default path for debian/ubuntu

savi0r:
Since debian based OSes dropped supporting python version 2 which resided in /usr/bin/python we need to add a softlink or otherwise ansible would fail.

Signed-off-by: Aliakbar Naji dr.nadji@gmail.com


guits:
@savi0r you are missing a sign off to your commit message. thanks!

savi0r:
> @savi0r you are missing a sign off to your commit message. thanks!

PR #7190, opened 2022-05-25
[skip ci] Add ceph_infra tag to rolling_update

fmount:
When the upgrade from Ceph 4 to 5 is performed in the OpenStack context,
ceph-ansible triggers the rolling_update playbook, which is supposed to
rollout new Ceph containers.  The ceph-infra role tries to take care
about firewall, ntp config and logrotate; however, TripleO manages them
through tripleo-heat-templates.  This patch just add an additional tag
to skip the ceph-infra role in the OpenStack context.

Closes: https://bugzilla.redhat.com/2090456

### DIFF
--- a/infrastructure-playbooks/rolling_update.yml
+++ b/infrastructure-playbooks/rolling_update.yml
@@ -151,6 +151,7 @@
 
     - import_role:
         name: ceph-infra
+      tags: ceph_infra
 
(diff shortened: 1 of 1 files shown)


fmount:
@guits does it make sense for you?

guits (approved):

PR #7196, opened 2022-06-07
ansible: use ansible.utils.ipwrap instead of ansible.netcommon.ipwrap

clwluvw:
ansible.netcommon.ipwrap is deprecated and is not being redirected with ansible 2.9.*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ ceph-ansible.spec
 !.travis.yml
 !.mergify.yml
 !raw_install_python.yml
+!requirements.yml
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -63,6 +63,16 @@ You can acquire Ansible on Ubuntu by using the `Ansible PPA <https://launchpad.n
    $ sudo apt update
    $ sudo apt install ansible
 
+Ansible collections
+-------------------
+
+In order to install third-party collections that are required for ceph-ansible,
+please run:
+
+.. code-block:: console
+
+   $ ansible-galaxy install -r requirements.yml
+
 
 Releases
 ========
--- a/group_vars/all.yml.sample
+++ b/group_vars/all.yml.sample
@@ -752,7 +752,7 @@ dummy:
 # For example:
 # If the ceph public network is 2a00:8a60:1:c301::/64 and the iSCSI Gateway resides
 # at a dedicated gateway network (2a00:8a60:1:c300::/64) (With routing between those networks).
-# It means "{{ hostvars[item]['ansible_facts']['all_ipv4_addresses'] | ips_in_ranges(public_network.split(',')) | last | ansible.netcommon.ipwrap }}" will be empty.
+# It means "{{ hostvars[item]['ansible_facts']['all_ipv4_addresses'] | ips_in_ranges(public_network.split(',')) | last | ansible.utils.ipwrap }}" will be empty.
 # As a consequence, this prevent from deploying dashboard with iSCSI node when it reside in a subnet different than `public_network`.
 # Using `igw_network` make it possible, set it with the subnet used by your iSCSI node.
 #igw_network: "{{ public_network }}"
--- a/group_vars/rhcs.yml.sample
+++ b/group_vars/rhcs.yml.sample
@@ -752,7 +752,7 @@ alertmanager_container_image: registry.redhat.io/openshift4/ose-prometheus-alert
 # For example:
 # If the ceph public network is 2a00:8a60:1:c301::/64 and the iSCSI Gateway resides
 # at a dedicated gateway network (2a00:8a60:1:c300::/64) (With routing between those networks).
-# It means "{{ hostvars[item]['ansible_facts']['all_ipv4_addresses'] | ips_in_ranges(public_network.split(',')) | last | ansible.netcommon.ipwrap }}" will be empty.
+# It means "{{ hostvars[item]['ansible_facts']['all_ipv4_addresses'] | ips_in_ranges(public_network.split(',')) | last | ansible.utils.ipwrap }}" will be empty.
 # As a consequence, this prevent from deploying dashboard with iSCSI node when it reside in a subnet different than `public_network`.
 # Using `igw_network` make it possible, set it with the subnet used by your iSCSI node.
 #igw_network: "{{ public_network }}"
--- a/infrastructure-playbooks/cephadm-adopt.yml
+++ b/infrastructure-playbooks/cephadm-adopt.yml
@@ -375,7 +375,7 @@
       when: ip_version == 'ipv4'
 
     - name: manage nodes with cephadm - ipv6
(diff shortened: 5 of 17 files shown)


clwluvw:
@guits I got a problem with using `ansible.netcommon.ipwrap` with ansible 2.9.* even if I installed `ansible.netcommon` galaxy
It is solved when changing to `ansible.utils.ipwrap`

guits:
@clwluvw how do you install ansible? is it ansible 3 ? or ansible-base 2.10 ?

clwluvw:
> @clwluvw how do you install ansible? is it ansible 3 ? or ansible-base 2.10 ?

@guits Tested with ansible 2.9.27

clwluvw:
Also btw it's deprecated and in `main` branch we are using `ansible.utils.ipwrap`
Seems it makes sense to use `ansible.utils.ipwrap` instead of `ansible.common.ipwrap` anyways 🤔

guits:
> Also btw it's deprecated and in the `main` branch we are using `ansible.utils.ipwrap` Seems it makes sense to use `ansible.utils.ipwrap` instead of `ansible.netcommon.ipwrap` anyways 🤔

in fact, it's a bit more complicated. It depends on how you get your Ansible engine.

guits:
> > @clwluvw how do you install ansible? is it ansible 3 ? or ansible-base 2.10 ?
> 
> @guits Tested with ansible 2.9.27

what version of `ansible.netcommon` did you install?

clwluvw:
> > > @clwluvw how do you install ansible? is it ansible 3 ? or ansible-base 2.10 ?
> > 
> > 
> > @guits Tested with ansible 2.9.27
> 
> what version of `ansible.netcommon` did you install?

3.0.1

clwluvw:
> > Also btw it's deprecated and in the `main` branch we are using `ansible.utils.ipwrap` Seems it makes sense to use `ansible.utils.ipwrap` instead of `ansible.netcommon.ipwrap` anyways 🤔
> 
> in fact, it's a bit more complicated. It depends on how you get your Ansible engine.

pip3 install ansible==2.9.27 🤔 
@guits But can you give me more hints about your fact? because I use pipenv for installing ansible so maybe other deps are in higher version or...

guits:
how can you run `stable-6.0` with ansible 2.9 given https://github.com/ceph/ceph-ansible/blob/stable-6.0/roles/ceph-validate/tasks/check_system.yml#L7-L10 ?

clwluvw:
> how can you run `stable-6.0` with ansible 2.9 given https://github.com/ceph/ceph-ansible/blob/stable-6.0/roles/ceph-validate/tasks/check_system.yml#L7-L10 ?

@guits Good point! =)))
Actually I was only testing your last commit here (https://github.com/ceph/ceph-ansible/commit/c9dd9a09d21a25ee67b0dd002461f1197d6f67d0) with a simple ansible ad-hoc

So seems I need to update my ansible as well :))
And I won't face this issue anymore

So it would be optional here to change to ansible.utils maybe just because of deprecation and being sync with the main branch

guits:
> So it would be optional here to change to ansible.utils maybe just because of deprecation and being sync with the main branch

It might make sense if we change the requirement from `ansible>=3,<4` to  `ansible>=2.10,<2.11,!=2.9.10`

clwluvw:
> It might make sense if we change the requirement from ansible>=3,<4 to ansible>=2.10,<2.11,!=2.9.10

But seems ansible 3 is based on ansible-base 2.10 (https://docs.ansible.com/ansible/latest/roadmap/COLLECTIONS_3_0.html) so this condition should work

guits:
> > It might make sense if we change the requirement from ansible>=3,<4 to ansible>=2.10,<2.11,!=2.9.10
> 
> But seems ansible 3 is based on ansible-base 2.10 (https://docs.ansible.com/ansible/latest/roadmap/COLLECTIONS_3_0.html) so this condition should work

ansible-base 2.10 alone doesn't include the collection `ansible.utils`, hence the need of `requirements.yml`
ansible-3.x comes with ansible-base 2.10 + collections. But the `ansible.utils` collection version shipped doesn't include the filter `ipwrap`, it's still present in `ansible.netcommon`. Hence the patch I've sent yesterday.

clwluvw:
> > > It might make sense if we change the requirement from ansible>=3,<4 to ansible>=2.10,<2.11,!=2.9.10
> > 
> > 
> > But seems ansible 3 is based on ansible-base 2.10 (https://docs.ansible.com/ansible/latest/roadmap/COLLECTIONS_3_0.html) so this condition should work
> 
> ansible-base 2.10 alone doesn't include the collection `ansible.utils`, hence the need of `requirements.yml` ansible-3.x comes with ansible-base 2.10 + collections. But the `ansible.utils` collection version (2.1.0) shipped with ansible-3.x doesn't include the filter `ipwrap`, it's still present in `ansible.netcommon`. Hence the patch I've sent yesterday.

Now I get what did you mean about how I installed the ansible 👍 😃

guits:
yeh these recent changes from Ansible are quite confusing and prone to error 😅

mitch000001:
> how can you run `stable-6.0` with ansible 2.9 given https://github.com/ceph/ceph-ansible/blob/stable-6.0/roles/ceph-validate/tasks/check_system.yml#L7-L10 ?

Actually, the documentation at https://docs.ceph.com/projects/ceph-ansible/en/latest/index.html#releases tells that the version to be used should be ansible 2.9. So this should definitely be corrected.

guits:
> > how can you run `stable-6.0` with ansible 2.9 given https://github.com/ceph/ceph-ansible/blob/stable-6.0/roles/ceph-validate/tasks/check_system.yml#L7-L10 ?
> 
> Actually, the documentation at https://docs.ceph.com/projects/ceph-ansible/en/latest/index.html#releases tells that the version to be used should be ansible 2.9. So this should definitely be corrected.

yup, correct. Thanks for noticing this. The documentation  has been updated

guits:
it looks like we can't support both `ansible 3.x` and `ansible-base 2.10`, we have to pick one.
I feel we should go with `ansible-base 2.10` to get closer to newer ansible releases, on the other hand, it might be a breaking change for users with `ansible 3.x`

any thoughts?

clwluvw:
> it looks like we can't support both `ansible 3.x` and `ansible-base 2.10`, we have to pick one. I feel we should go with `ansible-base 2.10` to get closer to newer ansible releases, on the other hand, it might be a breaking change for users with `ansible 3.x`
> 
> any thoughts?

I'm thinking like you too. but the users with ansible 3.x won't face any breaking changes because ansible 3.x is based on 2.10, so the new users only need to take care of the collections that are missing with ansible-base (we might need to list them in requirements.yml)

guits:
@clwluvw remember what I said:

> > But the `ansible.utils` collection version (2.1.0) shipped with ansible-3.x doesn't include the filter `ipwrap`, it's still present in `ansible.netcommon`.

clwluvw:
@guits So we might need to include the minimum version (2.5.0) to the requirements.yml as well 🤔 (and users should take care of the changelogs btw 😃)

guits (approved):

PR #7291, opened 2022-08-11
facts: set is_rgw_instances_defined from configure_dashboard

guits:
When we come from configure_dashboard.yml, this fact should be set if
`rgw_instances` is defined in group_vars/host_vars. Otherwise, the next
task that set the fact `rgw_instances` will be run as it will assume it
wasn't user defined.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2117294


guits:

guits:

PR #7293, opened 2022-08-16
ceph-config: allow overriding osd_memory_target

guits:
Typical failure:

```
"    The conditional check 'item > osd_memory_target' failed. The error was: Unexpected templating type error occurred on ({% if item > osd_memory_target %} True {% else %} False {% endif %}): '>' not supported between instances of 'str' and 'int'"
```

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2118544


guits:

PR #7318, opened 2022-09-24
OSD memory target improvements/fixes

clwluvw:
- ceph-osd: remove unused ceph config set for osd_memory_target
    - As the conf is always being set in the config file there is no need to set it in with `ceph config`.
    - Also this will make it hard to run the playbook with the `ceph_update_config` tag as it won't run and will create an inconsistency between the config managements of the cluster
- ceph-config: fix overriding osd_memory_target
    - When the value is overriding in `ceph_conf_overrides`, there is no need to calculate and set `osd_memory_target` again as we wanted to override the conf by our desired value.
- ceph-config: don't check for devices on existing osds
    - When osd_auto_discovery is true the `devices` var will be empty (as the disks have holders).
    - Also in general there is no need to check for devices to list the devices with ceph-volume as we have `default({})` on the stdout in `num_osds` set fact in the next task
- ceph-config: always set _osd_memory_target
    - this should be set when rolling_update is true as well, otherwise, it will reset to default on the upgrade

### DIFF
--- a/roles/ceph-config/tasks/main.yml
+++ b/roles/ceph-config/tasks/main.yml
@@ -97,9 +97,9 @@
 - name: set osd related config facts
   when: inventory_hostname in groups.get(osd_group_name, [])
   block:
-    - name: set_fact osd_memory_target, override from ceph_conf_overrides
+    - name: set_fact _osd_memory_target, override from ceph_conf_overrides
       set_fact:
-        osd_memory_target: "{{ item }}"
+        _osd_memory_target: "{{ item }}"
       loop:
         - "{{ ceph_conf_overrides.get('osd', {}).get('osd memory target', '') }}"
         - "{{ ceph_conf_overrides.get('osd', {}).get('osd_memory_target', '') }}"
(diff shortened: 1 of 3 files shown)


clwluvw:
@guits Will you take a look, please?

guits:
@clwluvw at first glance, there are some statements that are false here. I need to take a better look. I'll get back to you asap...

guits:
@clwluvw by any chance, are you available on IRC?

clwluvw:
> @clwluvw by any chance, are you available on IRC?

Yes you can find me with `seena` nickname

guits:
> seena

which IRC network?

clwluvw:
> which IRC network?

irc.oftc.net
Available on #ceph and #ceph-devel channel as well

guits (approved):

guits:

guits:

guits:
jenkins test centos-non_container-subset_update

guits:
jenkins test centos-container-subset_update

PR #7418, opened 2023-04-19
Replace ip_version check with ansible test

xbezdick:
Instead of checking ip_version variable we should check the input address for ip version and select code path based on that.

This solves ceph adoption with mixed ipv6 and ipv4 networks.

Resolves: rhbz#2186226

### DIFF
--- a/infrastructure-playbooks/cephadm-adopt.yml
+++ b/infrastructure-playbooks/cephadm-adopt.yml
@@ -364,13 +364,13 @@
       command: "{{ ceph_cmd }} orch host add {{ ansible_facts['nodename'] }} {{ ansible_facts['all_ipv4_addresses'] | ips_in_ranges(cephadm_mgmt_network.split(',')) | first }} {{ group_names | intersect(adopt_label_group_names) | join(' ') }}"
       changed_when: false
       delegate_to: '{{ groups[mon_group_name][0] }}'
-      when: ip_version == 'ipv4'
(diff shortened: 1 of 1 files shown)


fultonj:
LGTM

fmount (approved):
Thanks

fmount:
@guits can you review this patch when you have the chance?

asm0deuz:
@xbezdick why not set ip_version to ivp4? cephadm_mgmt_network shouldn't be used in the condition to chose between ipv4 or ipv6

asm0deuz:

asm0deuz:

asm0deuz:

guits (dismissed):
that looks good to me, I don't see any issue with this patch.
I let @asm0deuz take a final look

asm0deuz:
@xbezdick LGTM as well but could you sign-off your commit? Thx

asm0deuz (approved):

ktdreyer:
It's actually even more complicated. `ceph-ansible` still uses Ansible 2.9 downstream. Did you develop this and test it with Ansible 2.14?

ktdreyer:
> It's actually even more complicated. `ceph-ansible` still uses Ansible 2.9 downstream. Did you develop this and test it with Ansible 2.14?

I see now, when I install the `ansible-collection-ansible-utils` RPM, it works. So we need to add this to `ceph-ansible.spec.in`.

PR #7423, opened 2023-05-22
check ip_version with built-in filters for ansible 2.9

ktdreyer:
Ansible 2.9's IP address filters are simply available as `ipv4` or `ipv6`.

https://docs.ansible.com/ansible/2.9/user_guide/playbooks_filters.html#ip-address-filter

In newer Ansible versions, these are `ansible.utils.ipv4` and `ansible.utils.ipv6`.


ktdreyer:
With `ansible-2.9.27-1.el8ae`, I tested the following playbook:

```yaml
- name: testing rhbz#2207872
  hosts: localhost
  gather_facts: false
  vars:
    cephadm_mgmt_network: "192.168.1.0/24" 
    #cephadm_mgmt_network: "fe80::1/64" 
  tasks:
    - name: run date cmd on network success
      command: "date"
      when: cephadm_mgmt_network | ipv4
      #when: cephadm_mgmt_network | ipv6
```

It works when you comment or un-comment the various lines to test it.

ktdreyer:
I discussed this with @tserlin and @asm0deuz on Slack. We don't need this now. I found that installing `ansible-collection-ansible-utils` allows ceph-ansible `v6.0.28.4` to work on RHEL 8 with `ansible-2.9.27-1.el8ae`.

Instead, we'll simply add `ansible-collection-ansible-utils` to `ceph-ansible.spec.in`.

PR #7427, opened 2023-05-31
ceph_key: drop unused result parameter

guits:
This parameter is not used anywhere in this function, so we can drop it.


asm0deuz (approved):
LGTM

PR #7431, opened 2023-06-02
tests: updates functional tests with new image

guits:
let's use quay.io/ceph/daemon-base in every tests instead of `ceph/daemon` since it's not supposed to be built anymore soon.


asm0deuz (approved):
LGTM

PR #7451, opened 2023-08-16
do not use update debian cache or try to install packages when package-install is disabled

insatomcat:
When deploying with --skip-tags=package-install (when there is no access to a repository), the playbook is still trying to update the package cache, or to install ceph-mgr packages, which makes the playbook fail.
This change prevents the playbook to try to update the cache or install ceph-mgr packages when the package-install tag is skipped.


asm0deuz (approved):
LGTM

PR #7466, opened 2023-10-25
Do not pass NoneType as argument to ceph_crush_rule

noonedeadpunk:
With ansible-core 2.15 it is not possible to pass argument of unexpected
type, as otherwise module will fail with:
`'None' is not a string and conversion is not allowed`

With that we want to only get all existing crush rules, so we can simply
supply an empty string as a name argument, which would satisfy
requirements and have same behaviour for previous ansible versions.

Alternative approach would be to stop making `name` as a required
argument to the module and use empty string as default value
when info state is used.

### DIFF
--- a/library/ceph_crush_rule.py
+++ b/library/ceph_crush_rule.py
@@ -46,7 +46,8 @@ description:
 options:
     name:
         description:
-            - name of the Ceph Crush rule.
+            - name of the Ceph Crush rule. If state is 'info' - empty string
+              can be provided as a value to get all crush rules
         required: true
     cluster:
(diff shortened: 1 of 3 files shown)


jrosser:
@guits would it be possible to review this PR? It's blocking some progress over in the OpenStack world.

jrosser:
Reviews of this patch would be greatly appreciated, we are fast approaching a release deadline for OpenStack and this error is blocking ceph support in the OpenStack-Ansible project.

asm0deuz (approved):
LGTM

PR #7491, opened 2024-03-07
ceph-mon: add missing no_log for admin keyring tasks

clwluvw:


clwluvw:
jenkins test this please

PR #7492, opened 2024-03-08
core: bump to ansible 2.15

guits:
2.12 is EOL since May 2023.

Let's bump to 2.15.


berendt:
Wouldn't it make sense to go straight to 2.16?

guits:
> Wouldn't it make sense to go straight to 2.16?

it might make sense, yes.

berendt:
> > Wouldn't it make sense to go straight to 2.16?
> 
> it might make sense, yes. (keep in mind this is for stable-7.0)

Ah, I thought that was for the stable-8.0. For stable-7.0 I think Ansible 2.15 is fully sufficient. We have been using stable-7.0 with Ansible 2.14 for a long time without any problems.

clwluvw:

PR #7535, opened 2024-04-03
ceph_default: stop using ceph/daemon-base

jnemeiksis:
let's use reef tag quay.io/ceph/ceph instead of ceph/daemon-base

### DIFF
--- a/roles/ceph-defaults/defaults/main.yml
+++ b/roles/ceph-defaults/defaults/main.yml
@@ -463,7 +463,7 @@ ceph_tcmalloc_max_total_thread_cache: 134217728
 ##########
 # DOCKER #
 ##########
-ceph_docker_image: "ceph/daemon-base"
+ceph_docker_image: "ceph/ceph"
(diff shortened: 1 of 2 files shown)


roles/ceph-defaults/defaults/main.yml
```diff
@@ -463,8 +463,8 @@ ceph_tcmalloc_max_total_thread_cache: 134217728
 ##########
 # DOCKER #
 ##########
-ceph_docker_image: "ceph/daemon-base"
-ceph_docker_image_tag: latest-main
+ceph_docker_image: "ceph/ceph"
+ceph_docker_image_tag: v18
```
guits, line 467: the branch `main` is intended to test against the cutting edge of ceph devel which is well `latest-main`

these changes would have to be introduced directly to stable branches.

jnemeiksis: Agree, but I don't see `latest-main` tag here: https://quay.io/repository/ceph/ceph?tab=tags

PR #7540, opened 2024-04-04
ceph_default: Use vars cluster, setuser and setgroup for all daemons

jnemeiksis:
Allow all daemons drop privilege themselves, except ceph-exporter


jnemeiksis:
@guits I've tried this PR, but tests totally failed:) These variables will not be used in the future?

PR #7551, opened 2024-04-24
radosgw_user: handle non-existence user on check mode

clwluvw:
Set return code to 0 on check mode when the user doesn't exists and it will fail the module.


clwluvw:
jenkins test centos-container-subset_update

PR #7554, opened 2024-05-07
ceph-rgw: introduce rgw zone to the name schema

clwluvw:
This is needed by ceph-exporter as it is parsing the socket by the number of dots. Although the rgw_zone variable is only used for constructing the client name and has nothing to do with multisiting.
Reference: https://github.com/ceph/ceph/pull/52698

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ ceph-ansible.spec
 !.mergify.yml
 !raw_install_python.yml
 !requirements.yml
+.vscode/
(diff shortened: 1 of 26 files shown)


Eli-Tarrago:
I have a multi-site rgw configured by ceph-ansible and I'm concerned this change doesn't address previously configured radosgw zones.

I agree this change is the proper way of configuring radosgw

guits (approved):

PR #7560, opened 2024-05-16
ceph-defaults: use ceph as user and group

clwluvw:
To avoid mixup permissions on host and ceph-crash runs as `ceph` which can't send crash reports generated by others with root user.


clwluvw:

clwluvw:

guits (approved):

PR #7586, opened 2024-07-30
devices: test devices before collecting on auto discovery

clwluvw:
In some scenarios with NVMe, a device might be identified by Ansible but could actually be a multipath device rather than an actual device. We need to exclude these as Ceph cannot create OSDs on them.


clwluvw:

clwluvw:
jenkins test centos-container-external_clients

clwluvw:
jenkins retest this please

guits (approved):

PR #7607, opened 2024-09-06
Dashboard: replace hardcoded groupname with groupname variable

gael-o0o:
A little PR for replacing hard-coded group names with the group name variables defined.

Without this modification, we can't deploy dashboard with our global dynamic inventory.

PR #7609, opened 2024-09-12
playbooks: use group name variables consistently

stuartgrace-bbc:
In many places, playbooks allow host group names such as 'mons' to be overriden by defining a variable such as 'mon_group_name'. This is very useful when ceph-ansible is sharing the ansible inventory with other playbooks and other hosts which expect different group names. But this was unfinished, with the variables used in some places but hard-coded names used in others places, sometimes within the same playbook.

This commit uses the group name variables everywhere so they can be reliably used to overrride the host group names.

Signed-off-by: Stuart Grace <stuart.grace@bbc.co.uk>

### DIFF
--- a/dashboard.yml
+++ b/dashboard.yml
@@ -105,7 +105,7 @@
 # using groups[] here otherwise it can't fallback to the mon if there's no mgr group.
 # adding an additional | default(omit) in case where no monitors are present (external ceph cluster)
 - name: Deploy dashboard
-  hosts: "{{ groups['mgrs'] | default(groups['mons']) | default(omit) }}"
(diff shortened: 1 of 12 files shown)


gael-api:
I have a[ PR](https://github.com/ceph/ceph-ansible/pull/7607/files) with the same goal. (But with unknown commits attached to it) 
Some group_names are also present in prometheus jinja file : 
- https://github.com/ceph/ceph-ansible/blob/main/roles/ceph-prometheus/templates/alertmanager.yml.j2
- https://github.com/ceph/ceph-ansible/blob/main/roles/ceph-prometheus/templates/prometheus.yml.j2

I think an **`all_group_name`** that replace **`all`** generic group must be implemented when inventory used is larger than the simple cluster inventory (as in a cloud dynamic inventory).

PR #7645, opened 2025-01-30
Pass cluster in rolling_update call to ceph_config

fultonj:
This patch is necessary to migrate older clusters which were deployed when custom names were supported.

Follow up to 3cda8e31c4da81efeade9009b2fc9a1e379f2c66


fmount:
@asm0deuz FYI ^

asm0deuz (approved):
LGTM

PR #7658, opened 2025-02-11
common/container: drop CEPH_USE_RANDOM_NONCE env var

clwluvw:
It is not necessary anymore as this behavior became the default one.

Ceph Ref. https://github.com/ceph/ceph/pull/50344


clwluvw:
jenkins retest this please

PR #7660, opened 2025-02-24
Allow to control dashboard standby_behaviour

noonedeadpunk:
In case of placing Ceph Dashboard behind of the LoadBalancer
user needs to ensure setting `dashboard_standby_behavior: error`
to notify the LoadBalancer which backend is currently active one
as suggested by the documentation [1]

[1] https://docs.ceph.com/en/reef/mgr/dashboard/#disable-the-redirection


guits:
jenkins retest this please

PR #7664, opened 2025-03-07
fix for issue #11863. Optimize Image Removal

pemsith:
**Summary**
This PR optimizes the removal of Ceph dashboard container images by replacing repetitive command module calls with a single shell script execution.

**Changes Made**
Replaced multiple ansible.builtin.command calls in a loop with a dynamically generated script.
The script is templated using Jinja2 (remove_images.sh.j2) to ensure variable substitution for image names.
The script is copied to the target host and executed in one step, reducing SSH overhead and improving efficiency.
Ensures compatibility with both Docker and Podman by passing container_binary dynamically.

**Why This Change?**
Reduces multiple SSH calls by executing a single script instead of looping through docker rmi commands in Ansible.
Enhances performance.

### DIFF
--- a/infrastructure-playbooks/purge-dashboard.yml
+++ b/infrastructure-playbooks/purge-dashboard.yml
@@ -132,12 +132,20 @@
         - /run/prometheus.service-cid
         - /run/grafana-server.service-cid
 
-    - name: Remove ceph dashboard container images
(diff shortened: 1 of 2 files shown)


guits:
@pemsith why not simply calling `podman rmi <image1> <image2>  <...>` ?

PR #7682, opened 2025-12-05
Fix: osd_memory_target is now set correctly for hosts

### LINKED ISSUE #7681: osd_memory_target not set at all/correctly
**Bug Report**

What happened: `osd_memory_target` is not set

What you expected to happen: `osd_memory_target` to be set to calculated value

How to reproduce it (minimal and precise): 

Any time `Set osd_memory_target to cluster host config` runs it sets `osd_memory_target` for `"osd.*/{{ ansible_facts['hostname'] }}:host"`. But this doesn't set anything (its set in config dump - even tho it does nothing in ceph osd).
(you can check by `ceph config show-with-defaults osd.x | grep osd_memory_target`)

* Ceph version (e.g. `ceph -v`): tested on versions > pacific (expect tentacle)

You can also check this in [docs](https://docs.ceph.com/en/latest/rados/configuration/ceph-conf/#sections-and-masks).

tlapik123:
Based on [docs](https://docs.ceph.com/en/squid/rados/configuration/ceph-conf/#sections-and-masks) and testing `osd_memory_target` is not being set at all after the memory calculation.

Yes you can see it in `ceph config dump` but its not propagated onto the host because `ansible_facts['hostname']` is taken as CRUSH bucket.

Correct way is other way around:  `host:{{ ansible_facts['hostname'] }}`.

Also while at it the `osd.*` is IMO not required when `osd` does the same job but I'm definitely open to critique :).

This fixes #7681

### DIFF
--- a/roles/ceph-config/tasks/main.yml
+++ b/roles/ceph-config/tasks/main.yml
@@ -109,7 +109,7 @@
     - name: Set osd_memory_target to cluster host config
       ceph_config:
         action: set
-        who: "osd.*/{{ ansible_facts['hostname'] }}:host"
+        who: "osd/host:{{ ansible_facts['hostname'] }}"
         option: "osd_memory_target"
         value: "{{ _osd_memory_target }}"
       when:


tlapik123:
Hopefully I didn't do some kind of PR mistake :sweat_smile:. Feel free to correct me if anything is wrong.

tlapik123:
reopen - still active

anfimovir:
Hello,

Please reopen.

anfimovir:
@tlapik123 or may be you can create new PR?

tlapik123:
@anfimovir reopened as #7707 🙏

PR #7685, opened 2026-02-13
Increase stop timeout in ceph-osd.service

dswarbrick:
Wait a bit longer for clean BlueFS umount to prevent unclean shutdowns resulting in full recovery upon next boot.


clwluvw (approved):
Aligned with cephadm

PR #7692, opened 2026-04-08
support upgrade to tentacle:

badfiles:
- mds upgrade according to docs
 - fix missing status/versions on upgrade success


noonedeadpunk (commented):
This patch looks good to me, but I'd wish that changes to ‎library/ceph_fs.py were made in a separate PR, as this is kinda hard to review.

I don't think that adding new functionality to the module should really be incorporated to the upgrade support...

noonedeadpunk:
bump

PR #7698, opened 2026-05-12
Make sure CEPH_CONTAINER_IMAGE is fixed everywhere

damiandabrowski5:
In [1], `CEPH_CONTAINER_IMAGE` was fixed to work properly with ansible-core 2.19, where the way ansible handles environment variables with "None" value has changed.

Unfortunately, `CEPH_CONTAINER_IMAGE` was left unchanged in a few places.
This patch fixes that by making sure that the value of `CEPH_CONTAINER_IMAGE` is correct everywhere.

[1] https://github.com/ceph/ceph-ansible/commit/e70b2e4a1113d323da50b9ec9d16c5a7d7df4bf4

### DIFF
--- a/roles/ceph-nfs/tasks/pre_requisite_container.yml
+++ b/roles/ceph-nfs/tasks/pre_requisite_container.yml
@@ -43,7 +43,7 @@
         mode: "0600"
       no_log: "{{ no_log_on_ceph_key_tasks }}"
       environment:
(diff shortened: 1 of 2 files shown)


damiandabrowski5:
@clwluvw I'm super sorry, but I need to ask you for one more review :/ It looks like I somehow missed to update `CEPH_CONTAINER_IMAGE` in two places.

With this PR applied, openstack-ansible CI(where we use ceph-ansible) works fine with ansible-core 2.19 & ceph-ansible 9.0 - https://review.opendev.org/c/openstack/openstack-ansible/+/988043

^ @noonedeadpunk FYI

clwluvw (approved):

PR #7703, opened 2026-05-20
Switch to ganesha5

noonedeadpunk:
PPAs for Ubuntu does not contain modern distros for ganesha4.

Ganesha 5 is available both on EL and Ubuntu systems, and should be working just fine with ceph.

Signed-off-by: Dmitriy Rabotyagov <noonedeadpunk@gmail.com>


guits (approved):

noonedeadpunk:
So functional CI is never triggered, so is not going to provide any reply... I have no means to fix it, or at least can't find anything related to it...
Have submitted #7706 with a hope of somebody could spend some time to recover CI execution for the repo...

PR #7708, opened 2026-07-09
rgw: add support to create/delete accounts and assign user to accounts

BobBinaryBuilder:
ceph-ansible doesn't support the account feature introduced with ceph-squid. 

This pr adds a radosgw_account module focusing on account creation and deletion and modifies radosgw_user.py to assign users to an account. This can be useful when creating users for testing which should be able to interact with the rgw IAM API.

### DIFF
--- a/library/radosgw_account.py
+++ b/library/radosgw_account.py
@@ -1,0 +1,274 @@
+# Copyright 2026, Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


library/radosgw_account.py
```diff
@@ -0,0 +1,275 @@
+# Copyright 2020, Red Hat, Inc.
```
noonedeadpunk: it's not 2020 now

library/radosgw_account.py
```diff
@@ -0,0 +1,275 @@
...
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = '''
+---
+module: radosgw_account
+
+short_description: Manage RADOS Gateway Account
+
+version_added: "2.8"
```
noonedeadpunk: versioning looks off here

noonedeadpunk (commented):
@guits can you kindly allow CI execution for this PR, as at least flake8 should be running.

library/radosgw_account.py
BobBinaryBuilder: Sorry, looks like a dumb c/p error. I've updated both to the recent year and version.

noonedeadpunk (commented):
LGTM.

Don't have approve privileges, but +1 from me :)

anfimovir:
@BobBinaryBuilder Hello, please look flake8 results, all another good, thank you.

clwluvw (approved):

PR #7719, opened 2026-07-21
Use deb822_repository for Ubuntu/Debian

noonedeadpunk:
Since Ubuntu 24.04 [1] and Debian 12 [2] deb822 [3] format is
a preffered method of adding and managing repositories
for distros.

This also solves the upgrade issue, when a new repository is being adeed
each time you change the release name or ceph_repository value,
as apt_repository module currently constructs list filename from the URI

Using deb822_repository allows to replace the content of repo
instead of keep adding new repos, which might
require their manual clean-up in case of incorrect values.

[1] https://discourse.ubuntu.com/t/spec-apt-deb822-sources-by-default/29333
[2] https://discourse.ubuntu.com/t/spec-apt-deb822-sources-by-default/29333
[3] https://repolib.readthedocs.io/en/latest/deb822-format.html


anfimovir:
@clwluvw Hello, may be we can restore this PR for merge good changes?

PR #7729, opened 2026-08-28
chore/cleanup: ceph-ec-profile: ISA-L is default EC plugin

k0ste:
Since Tentalce release the default was changed. We should switch there default plugin, because Jerasure is unmaintained

From T release notes [1]:

```
The default plugin for erasure coded pools has been changed from Jerasure to ISA-L. Clusters created on Tentacle or later releases will use ISA-L as the default plugin when creating a new pool. Clusters that upgrade to the T release will continue to use their existing default values. The default values can be overridden by creating a new erasure code profile and selecting it when creating a new pool. ISA-L is recommended for new pools because the Jerasure library is no longer maintained.
```

[1] https://ceph.io/en/news/blog/2025/v20-2-0-tentacle-released/

### DIFF
--- a/library/ceph_ec_profile.py
+++ b/library/ceph_ec_profile.py
@@ -173,7 +173,7 @@ def run_module():
         state=dict(type='str', required=False,
                    choices=['present', 'absent'], default='present'),
         stripe_unit=dict(type='str', required=False),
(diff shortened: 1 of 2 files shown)


anfimovir:
LGTM

@clwluvw

clwluvw (approved):

PR #7733, opened 2026-09-04
Update ceph_key.py

dc0809:
create_key() always generated the entity's secret locally via
generate_secret(), which hardcodes cephx key type 1 (aes). Keys
created this way (e.g. ceph-mgr) could never be aes256k, causing
mgr auth failures on 20.2.4+ clusters that no longer accept aes.

Signed-off-by: Dmitriy Chubinidze <dcu995@gmail.com>


anfimovir:
@clwluvw Hello, this change include in latest squid release too:
https://ceph.io/en/news/blog/2026/v20-2-4-v19-2-6-combo-released/

I think, we can backport it in stable-9.0 too.

clwluvw (approved):

PR #7734, opened 2026-09-04
ceph-mon: handle rejected mon. keyring on rerun

dc0809:
On rerun, if the mon. keyring's cipher is no longer accepted, the initial keyring check fails with rc=13 and aborts before it can be fixed. Accept rc=13, skip re-creating an already-valid key, and add a step to rotate when needed.

Signed-off-by: Dmitriy Chubinidze <dcu995@gmail.com>


anfimovir:
LGTM from my side.

noonedeadpunk:
I am slightly confused now about the overall key generation flow.

Apparently, https://github.com/ceph/ceph-ansible/commit/29c82b0ea1c1c72d9d5714e55598978aead3833c was supposed to fix behavior and produce a proper aes256k instead of legacy ones, correct?

And I assume rc13 happens when monitor still has an old key? So the following patch is affecting upgrades only, and is not needed for freshly spawned clusters, after the `29c82b0ea1c1c72d9d5714e55598978aead3833c`?

If that is all true, what does `On rerun, if the mon. keyring's cipher is no longer accepted` mean? As that is probably not re-run, but upgrade from prior versions?
As if it is re-run, then smth is completely wrong with the flow, and I'd assume some other fix is required for `ceph_key` module.
Or is it just commit message is confusing?

dc0809:
> I am slightly confused now about the overall key generation flow.
> 
> Apparently, [29c82b0](https://github.com/ceph/ceph-ansible/commit/29c82b0ea1c1c72d9d5714e55598978aead3833c) was supposed to fix behavior and produce a proper aes256k instead of legacy ones, correct?
> 
> And I assume rc13 happens when monitor still has an old key? So the following patch is affecting upgrades only, and is not needed for freshly spawned clusters, after the `29c82b0ea1c1c72d9d5714e55598978aead3833c`?
> 
> If that is all true, what does `On rerun, if the mon. keyring's cipher is no longer accepted` mean? As that is probably not re-run, but upgrade from prior versions? As if it is re-run, then smth is completely wrong with the flow, and I'd assume some other fix is required for `ceph_key` module. Or is it just commit message is confusing?

Thanks for reply, i think the better approach overall is really to use ceph-authtool to generate the key correctly
from the start, so rerunning won't be needed anymore.

